PR #5, opened 2021-04-20
Adapt QFR to new DD Package

burgholzer:
This PR refactors the QFR library to work with the new DD Package version released recently. As a consequence the performance of practically all DD related routines should be improved.

✨ separation of matrix and vector DDs
✨ dynamic qubit numbers for DD package
✨ introduce `Definitions` file containing the most used definitions/constants
🔥 removed `line` parameter
🎨 `unsigned short` -> `dd::Qubit` for qubits, `unsigned short` to `dd::QubitCount` for numbers of qubits, `std::size_t` mostly everywhere else for consistency
♻️ controls are no longer a vector but a set
♻️ Measurement operations now hold separate vectors for qubits and classics
🐛 fix bug where permutation was not correctly accounted for in recursive functionality construction
🐛 fix bug where reference counting was not done correctly when recursively building the functionality of Grover's algorithm
🐛 fix reference counting bug in `reduceAncillary` and `reduceGarbage` methods
🔥 removed non-implemented `dumpReal`
🔥 remove DEBUG_MODE_QC
♻️ moved DD generation routines (e.g. for SWAP) to DD Package
🐛 Enable LTO for all targets
🐧 disable LTO in case clang is used under linux
✅ allow some up and down in project coverage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,6 @@ jobs:
         run:   |
                if [ "$RUNNER_OS" == "Windows" ]; then
                  cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=$BUILD_TYPE -T "ClangCl" -DBUILD_QFR_TESTS=ON
-               elif [ "$RUNNER_OS" == "macOS" ]; then
-                 cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_QFR_TESTS=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
                else
                  cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_QFR_TESTS=ON
                fi
@@ -61,8 +59,6 @@ jobs:
         run:   |
                if [ "$RUNNER_OS" == "Windows" ]; then
                  cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=$BUILD_TYPE -T "ClangCl" -DBUILD_QFR_TESTS=ON
-               elif [ "$RUNNER_OS" == "macOS" ]; then
-                 cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_QFR_TESTS=ON -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
                else
                  cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_QFR_TESTS=ON
                fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,11 +92,11 @@ jobs:
           python-version: '3.8'
       - name:  Configure CMake
         shell: bash
-        run:   cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=Release -DCOVERAGE=1 -DBUILD_QFR_TESTS=ON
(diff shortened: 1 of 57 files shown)


.github/workflows/ci.yml
```diff
@@ -16,106 +16,106 @@ jobs:
...
+      - name:  Configure CMake
+        shell: bash
+        run:   cmake -S "${{github.workspace}}" -B "${{github.workspace}}/build" -DCMAKE_BUILD_TYPE=Release -DCOVERAGE=1 -DBUILD_QFR_TESTS=ON
 
-    - name: Build
-      shell: bash
-      run: cmake --build "${{github.workspace}}/build" --config Debug --target qfr_test
+      - name:  Build
+        shell: bash
+        run:   cmake --build "${{github.workspace}}/build" --config Release --target qfr_test
```
hillmich: Shouldn't the coverage run in the debug build?

include/QuantumComputation.hpp
```diff
@@ -106,55 +170,95 @@ namespace qc {
...
+        void                      setLogicalQubitGarbage(dd::Qubit logical_qubit_index) { garbage[logical_qubit_index] = true; }
+        MatrixDD                  createInitialMatrix(std::unique_ptr<dd::Package>& dd) const; // creates identity matrix, which is reduced with respect to the ancillary qubits
 
         /// strip away qubits with no operations applied to them and which do not pop up in the output permutation
         /// \param force if true, also strip away idle qubits occurring in the output permutation
         void stripIdleQubits(bool force = false, bool reduceIOpermutations = true);
         // apply swaps 'on' DD in order to change 'from' to 'to'
         // where |from| >= |to|
-        static void changePermutation(dd::Edge& on, qc::permutationMap& from, const qc::permutationMap& to, std::array<short, qc::MAX_QUBITS>& line, std::unique_ptr<dd::Package>& dd, bool regular = true);
+        template<class DD>
```
hillmich: The name `DD` is a bit unlucky, as this this refers to vector edges or matrix edges. How about `template <class EdgeType>`?

include/operations/StandardOperation.hpp
```diff
@@ -6,81 +6,88 @@
...
-    //constexpr long double PARAMETER_TOLERANCE = dd::ComplexNumbers::TOLERANCE * 10e-2;
-    constexpr fp PARAMETER_TOLERANCE = 10e-6;
-    inline bool  fp_equals(const fp a, const fp b) { return (std::abs(a - b) < PARAMETER_TOLERANCE); }
-
-    class StandardOperation: public Operation {
+    class StandardOperation final: public Operation {
     protected:
-        static void checkInteger(fp& ld) {
+        static void checkInteger(dd::fp& ld) {
             auto nearest = std::nearbyint(ld);
```
hillmich: Maybe spell out `dd::fp nearest = ...`? `nearbyint` looks as it returns an integer at first glance.

include/operations/StandardOperation.hpp
```diff
@@ -6,81 +6,88 @@
...
             if (std::abs(ld - nearest) < PARAMETER_TOLERANCE) {
                 ld = nearest;
             }
         }
 
-        static void checkFractionPi(fp& ld) {
-            auto div     = qc::PI / ld;
+        static void checkFractionPi(dd::fp& ld) {
+            auto div     = dd::PI / ld;
```
hillmich: Maybe spell out `dd::fp div = ...`? `dd::PI` is in another package and its type is not that obvious.

hillmich (changes requested):

.github/workflows/ci.yml
burgholzer: Yeah, definitely. This has been changed since the build times in debug mode were horrendous. But since this has been resolved, it should be safe to go back to debug mode.

include/QuantumComputation.hpp
burgholzer: What about `DDType`? I feel that the concept of matrix and vector decision diagrams is better in the context of the QFR than "edges" (which are fine in the DD package itself).

include/operations/StandardOperation.hpp
burgholzer: agreed 👍🏼

include/operations/StandardOperation.hpp
burgholzer: agreed 👍🏼

include/QuantumComputation.hpp
hillmich: Works for me.

hillmich (approved):

PR #8, opened 2021-06-08
Dynamic Quantum Circuits

burgholzer:
This PR integrates dynamic quantum circuits into the JKQ framework based on the ideas described in https://arxiv.org/abs/2106.01099. To this end,

- [x] the Quantum Phase Estimation algorithm is implemented.
- [x] the Iterative Quantum Phase Estimation algorithm is implemented.
- [x] a simulation method (along with an RNG) is added that can handle dynamic circuit primitives
- [x] the unitary transformation scheme is realized.
- [x] the measurement outcome extraction scheme is realized.

🚸 increased usability of `ClassicControlledOperation`
🐛 fix `clone` method

### DIFF
--- a/include/CircuitOptimizer.hpp
+++ b/include/CircuitOptimizer.hpp
@@ -55,6 +55,8 @@ namespace qc {
         static void changeControls(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
 
         static void deferMeasurements(QuantumComputation& qc);
+
+        static bool isDynamicCircuit(QuantumComputation& qc);
     };
 } // namespace qc
 #endif //QFR_CIRCUITOPTIMIZER_HPP
--- a/include/CircuitOptimizer.hpp
+++ b/include/CircuitOptimizer.hpp
@@ -50,9 +50,9 @@ namespace qc {
 
         static void eliminateResets(QuantumComputation& qc);
 
-        static void changeTargets(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
+        static void changeTargets(Targets& targets, const std::map<dd::Qubit, dd::Qubit>& replacementMap);
 
-        static void changeControls(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
+        static void changeControls(dd::Controls& controls, const std::map<dd::Qubit, dd::Qubit>& replacementMap);
 
         static void deferMeasurements(QuantumComputation& qc);
 
--- a/include/CircuitOptimizer.hpp
+++ b/include/CircuitOptimizer.hpp
@@ -50,6 +50,10 @@ namespace qc {
 
         static void eliminateResets(QuantumComputation& qc);
 
+        static void changeTargets(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
+
+        static void changeControls(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
+
         static void deferMeasurements(QuantumComputation& qc);
     };
 } // namespace qc
--- a/include/QuantumComputation.hpp
+++ b/include/QuantumComputation.hpp
@@ -542,6 +542,8 @@ namespace qc {
 
         template<class T>
         std::vector<std::unique_ptr<Operation>>::iterator insert(std::vector<std::unique_ptr<Operation>>::const_iterator pos, T&& op) { return ops.insert(pos, std::forward<T>(op)); }
+
+        [[nodiscard]] const auto& at(std::size_t i) const { return ops.at(i); }
     };
 } // namespace qc
 #endif //QFR_QUANTUMCOMPUTATION_H
--- a/include/QuantumComputation.hpp
+++ b/include/QuantumComputation.hpp
@@ -165,7 +165,7 @@ namespace qc {
             seed(seed) {
             addQubitRegister(nqubits);
             addClassicalRegister(nqubits);
-            if (seed == 0) {
+            if (seed != 0) {
                 mt.seed(seed);
             } else {
                 // create and properly seed rng
@@ -179,7 +179,7 @@ namespace qc {
         explicit QuantumComputation(const std::string& filename, std::size_t seed = 0):
             seed(seed) {
             import(filename);
-            if (seed == 0) {
+            if (seed != 0) {
                 mt.seed(seed);
             } else {
                 // create and properly seed rng
--- a/include/QuantumComputation.hpp
+++ b/include/QuantumComputation.hpp
@@ -474,6 +474,7 @@ namespace qc {
         }
         virtual void dump(std::ostream&& of, Format format);
         virtual void dumpOpenQASM(std::ostream& of);
+        virtual void dumpTensorNetwork(std::ostream& of) const;
 
         virtual void reset() {
             ops.clear();
--- a/include/QuantumComputation.hpp
+++ b/include/QuantumComputation.hpp
@@ -51,7 +51,7 @@ namespace qc {
         QuantumRegisterMap   ancregs{};
 
         std::mt19937_64 mt;
-        std::size_t seed = 0;
+        std::size_t     seed = 0;
 
         void importOpenQASM(std::istream& is);
         void importReal(std::istream& is);
@@ -161,7 +161,8 @@ namespace qc {
 
     public:
         QuantumComputation() = default;
-        explicit QuantumComputation(std::size_t nqubits, std::size_t seed = 0): seed(seed) {
+        explicit QuantumComputation(std::size_t nqubits, std::size_t seed = 0):
+            seed(seed) {
             addQubitRegister(nqubits);
             addClassicalRegister(nqubits);
             if (seed == 0) {
@@ -171,11 +172,12 @@ namespace qc {
                 std::array<std::mt19937_64::result_type, std::mt19937_64::state_size> random_data{};
                 std::random_device                                                    rd;
                 std::generate(std::begin(random_data), std::end(random_data), [&rd]() { return rd(); });
-                std::seed_seq   seeds(std::begin(random_data), std::end(random_data));
+                std::seed_seq seeds(std::begin(random_data), std::end(random_data));
                 mt.seed(seeds);
             }
         }
-        explicit QuantumComputation(const std::string& filename, std::size_t seed = 0): seed(seed) {
+        explicit QuantumComputation(const std::string& filename, std::size_t seed = 0):
+            seed(seed) {
             import(filename);
             if (seed == 0) {
                 mt.seed(seed);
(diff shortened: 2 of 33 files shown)


include/CircuitOptimizer.hpp
```diff
@@ -47,6 +47,14 @@ namespace qc {
         static void decomposeSWAP(QuantumComputation& qc, bool isDirectedArchitecture);
 
         static void decomposeTeleport(QuantumComputation& qc);
+
+        static void eliminateResets(QuantumComputation& qc);
+
+        static void changeTargets(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
+
+        static void changeControls(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap);
```
burgholzer:
```suggestion
        static void changeTargets(Targets& targets, const std::map<dd::Qubit, dd::Qubit>& replacementMap);

        static void changeControls(dd::Controls& controls, const std::map<dd::Qubit, dd::Qubit>& replacementMap);
```
It is (almost) always best to pass STL data structures such as map, vector, etc. as reference (i.e., `&`) to avoid explicit copies. Furthermore, `const` should be used wherever applicable to denote that the parameter is not going to be affected by the function call.

Furthermore, these functions should probably just receive a reference to the corresponding targets vector or controls set and modify it accordingly.

src/CircuitOptimizer.cpp
```diff
@@ -667,4 +667,79 @@ namespace qc {
...
+                    changeTargets(reinterpret_cast<Operation&>(*it), replacementMap);
+                }
+                if ((*it)->isCompoundOperation()) {
+                    changeTargets(reinterpret_cast<Operation&>(*it), replacementMap);
+                    changeControls(reinterpret_cast<Operation&>(*it), replacementMap);
+                }
+                if ((*it)->isClassicControlledOperation()) {
+                    changeTargets(reinterpret_cast<Operation&>(*it), replacementMap);
+                    changeControls(reinterpret_cast<Operation&>(*it), replacementMap);
+                }
```
burgholzer: These `reinterpret_cast`s will all fail and result in `nullptr`. You are trying to reinterpret something as an `Operation` that is actually a pointer type, specifically a unique pointer to some type that might be derived from `Operation`.
This is the main reason why calling the method currently results in a segfault.

Depending on the type of operation, you might need to correctly (dynamic) cast to a pointer type of the underlying operation, e.g., in case of compound operations.

Furthermore, treatment of the different operations can be unified, see the following suggestion:
```suggestion
                if ((*it)->isStandardOperation() || (*it)->isClassicControlledOperation()) {
                    auto& targets = (*it)->getTargets();
                    auto& controls = (*it)->getControls();
                    changeTargets(targets, replacementMap);
                    changeControls(controls, replacementMap);
                } else if ((*it)->isNonUnitaryOperation()) {
                    auto& targets = (*it)->getTargets();
                    changeTargets(targets, replacementMap);
                }
                if ((*it)->isCompoundOperation()) {
                    auto* compOp = dynamic_cast<qc::CompoundOperation*>((*it)->get());
                    /// TODO: do the appropriate replacements on every operation of the compound operation
                }
```

src/CircuitOptimizer.cpp
```diff
@@ -667,4 +667,79 @@ namespace qc {
...
+            }
+            if ((*it)->getType() == qc::Reset) {
+                auto indexAddQubit = static_cast<dd::Qubit>(qc.getNqubits());
+                qc.addQubit(indexAddQubit, indexAddQubit, indexAddQubit);
+                for (const auto& target: (*it)->getTargets()) {
+                    replacementMap.insert(std::pair(static_cast<dd::Qubit>(target), static_cast<dd::Qubit>(indexAddQubit)));
+                }
+                it = qc.erase(it);
```
burgholzer: if a reset has multiple targets, this will only add a single additional qubit. However, a new qubit should be added for every qubit that is reset.

src/CircuitOptimizer.cpp
```diff
@@ -667,4 +667,79 @@ namespace qc {
...
+    void CircuitOptimizer::changeTargets(Operation& op, std::map<dd::Qubit, dd::Qubit> replacementMap) {
+        auto targetIt = op.getTargets().begin();
+        while (targetIt != op.getTargets().end()) {
+            auto newTarget = replacementMap.find((static_cast<dd::Qubit>(*targetIt)));
+            if (newTarget != replacementMap.end()) {
+                op.setTargets(static_cast<Targets>(newTarget->second));
+            }
+            targetIt++;
+        }
+    }
```
burgholzer: As described above, this method should just receive a reference to a vector of targets (and a `const` reference to the replacement map) and modify the vector accordingly. See below for a suggestion:
```suggestion
    void CircuitOptimizer::changeTargets(Targets& targets, const std::map<dd::Qubit, dd::Qubit>& replacementMap) {
        for(auto& target: targets) {
            auto newTargetIt = replacementMap.find(target);
            if (newTargetIt != replacementMap.end()) {
                target = newTarget->second;
            }
        }
    }
```

src/CircuitOptimizer.cpp
```diff
@@ -667,4 +667,79 @@ namespace qc {
...
+        auto controlIt = op.getControls().begin();
+        while (controlIt != op.getControls().end()) {
+            auto newControl = replacementMap.find((*controlIt).qubit);
+            if (newControl != replacementMap.end()) {
+                op.getControls().erase(controlIt);
+                op.getControls().insert(dd::Control{newControl->second});
+            }
+            controlIt++;
+        }
+    }
```
burgholzer: Same as above. You should be able to figure it out based on the `changeTargets` suggestion above.
Small hint: A control consists of a `Qubit control` and a `Type type`. You need to copy the type of the old control to the newly emplaced one or it won't work properly.

burgholzer (commented):
I had a look at the code you wrote:
 - Fixed the merge conflicts (please do not change/commit anything in the submodules).
 - The project wasn't even compiling so I fixed the compilation error
 - Also fixed some of the warnings that popped up in CLion and ran clang-format so that code formatting checks pass
 - Added a basic set of tests for the `eliminateResets` method
 - Did a quick review of the `eliminateResets` method. Lots of stuff to address there, see the respective comments.

include/QuantumComputation.hpp
```diff
@@ -49,6 +50,9 @@ namespace qc {
         ClassicalRegisterMap cregs{};
         QuantumRegisterMap   ancregs{};
 
+        std::mt19937_64 mt;
+        std::size_t     seed = 0;
```
hillmich, line 54: Make this const?

include/QuantumComputation.hpp
```diff
@@ -157,12 +161,34 @@ namespace qc {
 
     public:
         QuantumComputation() = default;
-        explicit QuantumComputation(std::size_t nqubits) {
+        explicit QuantumComputation(std::size_t nqubits, std::size_t seed = 0):
+            seed(seed) {
             addQubitRegister(nqubits);
             addClassicalRegister(nqubits);
+            if (seed == 0) {
```
hillmich: Pretty sure this should be `!=`.

include/QuantumComputation.hpp
```diff
@@ -157,12 +161,34 @@ namespace qc {
...
+                std::generate(std::begin(random_data), std::end(random_data), [&rd]() { return rd(); });
+                std::seed_seq seeds(std::begin(random_data), std::end(random_data));
+                mt.seed(seeds);
+            }
         }
-        explicit QuantumComputation(const std::string& filename) {
+        explicit QuantumComputation(const std::string& filename, std::size_t seed = 0):
+            seed(seed) {
             import(filename);
+            if (seed == 0) {
```
hillmich: Pretty sure this should be `!=`.

include/QuantumComputation.hpp
```diff
@@ -418,11 +446,14 @@ namespace qc {
             max_controls = std::max(ncontrols, max_controls);
         }
 
-        virtual VectorDD simulate(const VectorDD& in, std::unique_ptr<dd::Package>& dd) const;
-        virtual MatrixDD buildFunctionality(std::unique_ptr<dd::Package>& dd) const;
-        virtual MatrixDD buildFunctionalityRecursive(std::unique_ptr<dd::Package>& dd) const;
-        virtual bool     buildFunctionalityRecursive(std::size_t depth, std::size_t opIdx, std::stack<MatrixDD>& s, Permutation& permutation, std::unique_ptr<dd::Package>& dd) const;
+        virtual VectorDD                           simulate(const VectorDD& in, std::unique_ptr<dd::Package>& dd) const;
+        virtual std::map<std::string, std::size_t> simulate(const VectorDD& in, std::unique_ptr<dd::Package>& dd, std::size_t shots);
```
hillmich, line 450: Why is this method not const?

src/QuantumComputation.cpp
```diff
@@ -581,6 +581,339 @@ namespace qc {
...
+
+                const auto& targets = op->getTargets();
+                if (targets.size() != 1) {
+                    throw qc::QFRException("Resets on multiple qubits are currently not supported. Please split them into multiple single resets.");
+                }
+
+                const auto& [pzero, pone] = dd->determineMeasurementProbabilities(state, targets[0], true);
+
+                //                if (dd::ComplexTable<>::Entry::approximatelyOne(pone)) {
+                if (pone > 0.1) {
```
hillmich: Comment on why `0.1`?

src/QuantumComputation.cpp
```diff
@@ -581,6 +581,339 @@ namespace qc {
...
+                    qc::MatrixDD xGate      = dd->makeGateDD(dd::Xmat, state.p->v + 1, targets[0]);
+                    qc::VectorDD resetState = dd->multiply(xGate, state);
+                    dd->incRef(resetState);
+                    dd->decRef(state);
+                    state = resetState;
+                    continue;
+                }
+
+                //                if (!dd::ComplexTable<>::Entry::approximatelyOne(pzero)) {
+                if (pzero < 0.9) {
```
hillmich: Comment on why `0.9`?

hillmich (commented):
See comments

include/QuantumComputation.hpp
burgholzer: I'd rather keep it non-const so that the defaulted move-assignment still works as it is.

include/QuantumComputation.hpp
burgholzer: Definitely! Good catch.

include/QuantumComputation.hpp
burgholzer: This can't be `const` due to the function using the RNG and, consequently, modifying its state.

src/QuantumComputation.cpp
burgholzer: I fixed this to be more in line with what is described to happen in the code comment above.

src/QuantumComputation.cpp
burgholzer: Same as above

PR #10, opened 2021-06-11
Refactor Python Bindings

burgholzer:
The JKQ QFR Python bindings unnecessarily used JSON as an intermediate layer between C++ and Python.
This PR refactors the bindings to directly use Python objects for that purpose, which also fixes the LGTM build
Overall, this change reduces boilerplate code and should also increase performance marginally.

Furthermore, this PR modifies the CI pipeline so that
 - C++ bindings code is now also built during the `build` step
 - Python bindings code (sdist and wheels) are now also tested on every commit to the master branch or every pull request


hillmich (approved):
LGTM

PR #12, opened 2021-07-06
Fix iSWAP & Peres OpenQASM export

ollpu:
The `iSWAP`, `Peres` and `Peresdag` gates were exported with invalid OpenQASM caused by missing spaces when there are no control qubits.

A dump of an `iSWAP(0, 1)` gate previously looked as follows:

```qasm
swap q[0], q[1];
s q[0];
s q[1];
czq[0], q[1];
```

### DIFF
--- a/src/operations/StandardOperation.cpp
+++ b/src/operations/StandardOperation.cpp
@@ -401,7 +401,7 @@ namespace qc {
                 of << op.str() << "cz";
                 for (const auto& c: controls)
                     of << " " << qreg[c.qubit].second << ",";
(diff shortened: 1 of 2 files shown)


burgholzer:
Good catch. Thanks for this PR. This already looks great!
Would you mind adding some small tests in the `qfr/test/unittests/test_io.cpp` file that cover these cases?

ollpu:
Done. The tests assert that an OpenQASM dump with these gates is parsed successfully when importing. Should they be more comprehensive?

burgholzer:
This is great! Thanks for your contribution 👍🏼

PR #16, opened 2021-09-02
Dump Tensor Network File

burgholzer:
This PR adds the possibility to dump a tensor network representation of a given `QuantumComputation`.
This should allow to feed circuits exported from our framework into TN-based tools, such as Cotengra, quimb, etc.

In addition, the Qiskit `QuantumCircuit` import now natively supports ancillary registers and `c3x`/`c4x` gates.


hillmich (approved):
LGTM!

PR #17, opened 2021-09-27
Feature/ecc

pichristoph:
add Quantum Error Correcting Code features to QFR


include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "QuantumComputation.hpp"
+#include "EccStatistics.hpp"
+
+#ifndef QFR_Ecc_HPP
+#define QFR_Ecc_HPP
+
+class Ecc {
+public:
+
+    enum EccID {
```
burgholzer:
```suggestion
    enum ID {
```
This could just be named `ID` as accessing it involves calling `Ecc::ID`. A possible alternative would be to call this enum `Code` or `Type`.

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+		Id, Q3Shor, Q9Shor, Q7Steane, QxCustom
+	};
+    struct EccInfo {
```
burgholzer:
```suggestion
    struct Info {
```
Same logic as above. Naming this `Info` is sufficient as calling it requires `Ecc::Info`. Helps to avoid redundancies in naming.

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+        EccID enumID;
```
burgholzer:
```suggestion
        EccID id;
```

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+        int nRedundantQubits;
+        int nClassicalBits;
+        std::string name;
+    };
+
+    const struct EccInfo ecc;
```
burgholzer:
```suggestion
    const EccInfo ecc;
```
The struct keyword is not needed in C++.

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+	Ecc(struct EccInfo ecc_type, qc::QuantumComputation& qc);
```
burgholzer:
```suggestion
	Ecc(EccInfo ecc, qc::QuantumComputation& qc);
```

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+	virtual ~Ecc() = default;
+
+	qc::QuantumComputation& applyEcc();
```
burgholzer:
```suggestion
	qc::QuantumComputation& apply();
```
This is a method of class `Ecc`. I guess naming this method `apply` is sufficient.

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+
+protected:
+
+    qc::QuantumComputation& qc;
+	qc::QuantumComputation qcMapped;
+	EccStatistics statistics{};
+
+	virtual void writeEccEncoding()=0;
```
burgholzer:
```suggestion
	virtual void writeEncoding()=0;
```

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+	virtual void measureAndCorrect()=0;
+
+	virtual void writeEccDecoding()=0;
```
burgholzer:
```suggestion
	virtual void writeDecoding()=0;
```

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+	virtual void mapGate(std::unique_ptr<qc::Operation> &gate)=0;
+
+    void writeToffoli(unsigned short c1, unsigned short c2, unsigned short target);
+    void writeCnot(unsigned short control, unsigned short target);
+
+    dd::Control createControl(unsigned short qubit, bool pos);
```
burgholzer: This function should not be necessary as a control can always be constructed using `dd::Control{qubit, pos}` or possibly `dd::Control{static_cast<dd::Qubit>(qubit), pos}`.

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "QuantumComputation.hpp"
+#include "EccStatistics.hpp"
+
+#ifndef QFR_Ecc_HPP
+#define QFR_Ecc_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/EccStatistics.hpp
```diff
@@ -0,0 +1,28 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include <stdlib.h>
+
+#ifndef QFR_EccStatistics_HPP
+#define QFR_EccStatistics_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/IdEcc.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "QuantumComputation.hpp"
+#include "Ecc.hpp"
+
+
+#ifndef QFR_IdEcc_HPP
+#define QFR_IdEcc_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/Q3ShorEcc.hpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "QuantumComputation.hpp"
+#include "Ecc.hpp"
+
+#ifndef QFR_Q3ShorEcc_HPP
+#define QFR_Q3ShorEcc_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/Q7SteaneEcc.hpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "QuantumComputation.hpp"
+#include "Ecc.hpp"
+
+#ifndef QFR_Q7SteaneEcc_HPP
+#define QFR_Q7SteaneEcc_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/Q9ShorEcc.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "QuantumComputation.hpp"
+#include "Ecc.hpp"
+
+
+#ifndef QFR_Q9ShorEcc_HPP
+#define QFR_Q9ShorEcc_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/QxCustomEcc.hpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "QuantumComputation.hpp"
+#include "Ecc.hpp"
+
+#ifndef QFR_QxCustomEcc_HPP
+#define QFR_QxCustomEcc_HPP
```
burgholzer, line 7: Header guards should come before anything in the file, especially before includes.

include/eccs/IdEcc.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+class IdEcc: public Ecc {
+public:
+    IdEcc(qc::QuantumComputation& qc);
+
+    static const std::string getEccName() {
+        return "Id";
+    }
```
burgholzer: why is this not a (purely) virtual function in the base class?
```suggestion
    static const std::string getName() {
        return "Id";
    }
```

src/eccs/Ecc.cpp
```diff
@@ -0,0 +1,89 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "eccs/Ecc.hpp"
+
+//#include <stdlib.h>
```
burgholzer: can be removed

src/eccs/Ecc.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+Ecc::Ecc(struct EccInfo eccInfo, qc::QuantumComputation& quantumcomputation): ecc(eccInfo), qc(quantumcomputation) {
+}
+
+qc::QuantumComputation& Ecc::applyEcc() {
+    qc.stripIdleQubits(true, false);
+    statistics.nInputQubits = qc.getNqubits();
+	statistics.nOutputQubits = qc.getNqubits()*ecc.nRedundantQubits;
+	statistics.nOutputClassicalBits = qc.getNqubits()*ecc.nClassicalBits;
```
burgholzer: Is it intended that the number of classical bits is a multiple of the number of qubits instead of the number of classical bits?

src/eccs/Ecc.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    measureAndCorrect();
+
+    writeEccDecoding();
+
+	long nOutputGates = 0;
+	for(auto& gate: qcMapped) {
+        nOutputGates++;
+    }
+    statistics.nOutputGates = nOutputGates;
```
burgholzer:
```suggestion
    statistics.nOutputGates = qcMapped.getNindividualOps();
```

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+	qc::QuantumComputation qcMapped;
+	EccStatistics statistics{};
+
+	virtual void writeEccEncoding()=0;
+
+	virtual void measureAndCorrect()=0;
+
+	virtual void writeEccDecoding()=0;
+
+	virtual void mapGate(std::unique_ptr<qc::Operation> &gate)=0;
```
burgholzer:
```suggestion
	virtual void mapGate(const std::unique_ptr<qc::Operation> &gate)=0;
```
I guess `const` can be added here.

src/eccs/Ecc.cpp
```diff
@@ -0,0 +1,89 @@
...
+    statistics.nInputQubits = qc.getNqubits();
+	statistics.nOutputQubits = qc.getNqubits()*ecc.nRedundantQubits;
+	statistics.nOutputClassicalBits = qc.getNqubits()*ecc.nClassicalBits;
+	qcMapped.addQubitRegister(statistics.nOutputQubits);
+	qcMapped.addClassicalRegister(statistics.nOutputClassicalBits);
+
+	writeEccEncoding();
+
+	long nInputGates = 0;
+    for(auto& gate: qc) {
```
burgholzer:
```suggestion
    for(const auto& gate: qc) {
```
I guess `const` can also be added here.

src/eccs/Ecc.cpp
```diff
@@ -0,0 +1,89 @@
...
+		size_t dot = outputFilename.find_last_of('.');
+		std::string extension = outputFilename.substr(dot + 1);
+		std::transform(extension.begin(), extension.end(), extension.begin(), [](unsigned char c) { return ::tolower(c); });
+		if (extension == "real") {
+			dumpResult(outputFilename, qc::Real);
+		} else if (extension == "qasm") {
+			dumpResult(outputFilename, qc::OpenQASM);
+		} else {
+			throw qc::QFRException("[dump] Extension " + extension + " not recognized/supported for dumping.");
+		}
```
burgholzer: at the moment, only the `OpenQASM` dump is actually implemented in the QFR. So you could simplify this statement.

src/eccs/IdEcc.cpp
```diff
@@ -0,0 +1,66 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "eccs/IdEcc.hpp"
+
+#include <chrono>
+//#include <stdlib.h>
```
burgholzer: can be removed

src/eccs/IdEcc.cpp
```diff
@@ -0,0 +1,66 @@
...
+    case qc::Vdag:
+        for(std::size_t j=0;j<gate.get()->getNtargets();j++) {
+            i = gate.get()->getTargets()[j];
+            if(gate.get()->getNcontrols()) {
+                auto& ctrls = gate.get()->getControls();
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, ctrls, i, gate.get()->getType());
+            } else {
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, i, gate.get()->getType());
+            }
+        }
```
burgholzer:
```suggestion
        for (const auto& target: gate->getTargets()) {
            if(!gate->getControls().empty()) {
                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, gate->getControls(), target, gate->getType());
            } else {
                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, target, gate->getType());
            }
        }
```
This can be refactored in a more C++ like fashion.

src/eccs/IdEcc.cpp
```diff
@@ -0,0 +1,66 @@
...
+    case qc::SWAP:
+    case qc::iSWAP:
+    case qc::Peres:
+    case qc::Peresdag:
+    case qc::Compound:
+    case qc::ClassicControlled:
+    default:
+        statistics.nOutputGates = -1;
+        statistics.nOutputQubits = -1;
+        throw qc::QFRException("Gate not possible to encode in error code!");
```
burgholzer, line 55: Is this exception  caught anywhere? If not, then setting the statistics members is redundant.
The error message could also include the name of the gate that was not possible to encode.

src/eccs/Q3ShorEcc.cpp
```diff
@@ -0,0 +1,118 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "eccs/Q3ShorEcc.hpp"
+
+#include <chrono>
+//#include <stdlib.h>
```
burgholzer: can be removed

src/eccs/Q3ShorEcc.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+void Q3ShorEcc::measureAndCorrect() {
+    const int nQubits = qc.getNqubits();
+    for(int i=0;i<nQubits;i++) {
+
+        qcMapped.h(i+3*nQubits);
+        qcMapped.h(i+4*nQubits);
+        dd::Control c3 = createControl(i+3*nQubits, true);
+        dd::Control c4 = createControl(i+4*nQubits, true);
```
burgholzer:
```suggestion
        auto c3 = createControl(i+3*nQubits, true);
        auto c4 = createControl(i+4*nQubits, true);
```
You can use `auto` in cases like these.

src/eccs/Q3ShorEcc.cpp
```diff
@@ -0,0 +1,118 @@
...
+                    ctrls2.insert(createControl(ct.qubit+nQubits, ct.type==dd::Control::Type::pos));
+                    ctrls3.insert(createControl(ct.qubit+2*nQubits, ct.type==dd::Control::Type::pos));
+                }
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, ctrls2, i+nQubits, gate.get()->getType());
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, ctrls3, i+2*nQubits, gate.get()->getType());
+            } else {
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, i, gate.get()->getType());
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, i+nQubits, gate.get()->getType());
+                qcMapped.emplace_back<qc::StandardOperation>(nQubits*ecc.nRedundantQubits, i+2*nQubits, gate.get()->getType());
+            }
```
burgholzer: see comment above

src/eccs/Q9ShorEcc.cpp
```diff
@@ -0,0 +1,216 @@
+/*
+ * This file is part of JKQ QFR library which is released under the MIT license.
+ * See file README.md or go to http://iic.jku.at/eda/research/quantum/ for more information.
+ */
+
+#include "eccs/Q9ShorEcc.hpp"
+
+#include <chrono>
+//#include <stdlib.h>
```
burgholzer: can be removed

include/eccs/Ecc.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+    void writeToffoli(unsigned short c1, unsigned short c2, unsigned short target);
+    void writeCnot(unsigned short control, unsigned short target);
```
burgholzer: Are these methods really needed? It seems easy enough to call `qcMapped.x(...)` with the appropriate arguments.

burgholzer (changes requested):
Just did a quick review round. Didn't make it quite through to the very end, but you should get a rough feel for some necessary changes in the remaining code by studying the other comments.

As you might have seen by the failing CI test, we use `clang-format` to ensure consistent formatting throughout the project. See for example https://clang.llvm.org/docs/ClangFormat.html. This should also make reviewing this PR easier.

src/eccs/Q3ShorEcc.cpp
```diff
@@ -0,0 +1,118 @@
...
+        qcMapped.measure(i+4*nQubits, i+nQubits);
+
+        dd::Control cn3 = createControl(i+3*nQubits, false);
+        dd::Control cn4 = createControl(i+4*nQubits, false);
+        dd::Controls cp3n4;
+        cp3n4.insert(c3);cp3n4.insert(cn4);
+        dd::Controls cp3p4;
+        cp3p4.insert(c3);cp3p4.insert(c4);
+        dd::Controls cn3p4;
+        cn3p4.insert(cn3);cn3p4.insert(c4);
```
33Gjl1Xe: Consider not putting multiple statements into a single line. (See also below multiple times)

src/eccs/Q7SteaneEcc.cpp
```diff
@@ -0,0 +1,63 @@
...
+Q7SteaneEcc::Q7SteaneEcc(qc::QuantumComputation& qc): Ecc({EccID::Q3Shor, 13, 6, Q7SteaneEcc::getEccName()}, qc) {}
+
+void Q7SteaneEcc::writeEccEncoding() {
+	measureAndCorrect();
+}
+
+void Q7SteaneEcc::measureAndCorrect() {
+    const int nQubits = qc.getNqubits();
+    for(int i=0;i<nQubits;i++) {
+        //
```
33Gjl1Xe: TODO?

33Gjl1Xe (commented):
LGTM at a glance.

include/eccs/Ecc.hpp
pichristoph: True, thanks for the shortcut!

include/eccs/Ecc.hpp
pichristoph: With the simple way of creating `dd::Control` structs, `qcMapped.x(...)` is definitely easier to use.

include/eccs/IdEcc.hpp
pichristoph: The name information is needed before Ecc objects are created (e.g. for reading the ecc kind of the console), thus I chose to use static methods for that. This has the drawback that static and virtual do not work at the same time.

src/eccs/Ecc.cpp
pichristoph: Yes and no - it is currently wrong if the input circuit contains classical bits. Otherwise, for every ECC, measurements are taken in the end to correct each qubit. `ecc.nClassicalBits` denotes the number of classical bits needed by measurement per input circuit qubit. The total number of classical bits is therefore a multiple. 
However, 1) your question shows me that I should change the naming of the variable to make this more precise, and 2) I have to take care of input circuit classical bits - thanks!

include/eccs/IdEcc.hpp
burgholzer: Makes sense! Didn't think that one through 🙃

src/eccs/Ecc.cpp
burgholzer: Thanks for clearing that up! Changing the name might already clear up most of this.
Typically, if the input circuit has `n` qubits, it will also contain `n` classical bits. I am not that familiar with the details of ECCs but I guess you can just extend these to `n*ecc.ClassicalBits` for the mapped circuit.

src/eccs/Ecc.cpp
pichristoph: Good to know, thanks :)

src/eccs/IdEcc.cpp
pichristoph: True, thanks! The error message now contains the kind of code (Q3Shor, ...) and the "failing" gate.

PR #22, opened 2021-11-15
Algorithm and Circuit Improvements

burgholzer:
This PR primarily brings many improvements for pre-defined algorithms and the handling of dynamic circuits. In particular it:

 - ⚡  switches the extraction of the probability amplitudes of a dynamic circuits to sparse vectors for increased performance and scalability
 - ⚡ significantly improves the QPE algorithms:
   - ♻️ unifies QPE and IQPE algorithms, eliminating lots of redundant code
   - ♻️ switches up gate order in QPE algorithm for better performance
   -  🐛 fixes some bugs that prevented the QPE algorithms from working for more than 32 qubits
   - ✅ adapts dynamic circuit tests for new improvements
 - ⚡ significantly improves the BV algorithm
   - 🐛 fixes algorithm implementation
   - ✨ adds dynamic circuit variant
   - ✨ adds routine that automatically generated an instance for a given number of qubits
   - ✅ adds tests for new functions
 - ⚡ Improves Grover's algorithm
   - ⚡ DD-tailoring for diffusion operation
   - ⚡ improves algorithm description so that it works for an arbitrary number of qubits
 - ⚡ Improves QFT algorithm
   - ⚡ DD-tailoring for order of operations
   - ✨ adds dynamic circuit variant
   - ✅ adds tests for new functions
 - ⚡ DD-tailoring for GHZ state preparation
 - 🐛 fixes an endianness and a reference counting bug in the `simulate` routine with measurements
 - 🐛 reduces the parameter tolerance to the level of the default decision diagram package tolerance in order to introduce less approximations when reading in circuits
 - ✨ introduces a new optimization pass that allows to reorder the operations in a `QuantumComputation` to achieve a canonical ordering
 - ⬆️ updates all external dependencies to their latest version.

Depends on iic-jku/dd_package#26.


include/Definitions.hpp
```diff
@@ -42,6 +42,8 @@ namespace qc {
 
     using Targets = std::vector<dd::Qubit>;
 
+    using BitString = std::bitset<std::numeric_limits<dd::QubitCount>::max()>;
```
hillmich: Wouldn't `std::numeric_limits<dd::Qubit>::max()+1` better upper bound?

include/algorithms/QFT.hpp
```diff
@@ -11,9 +11,16 @@
...
     class QFT: public QuantumComputation {
     public:
-        explicit QFT(dd::QubitCount nq);
+        explicit QFT(dd::QubitCount nq, bool includeMeasurements = true, bool dynamic = false);
 
         std::ostream& printStatistics(std::ostream& os) const override;
+
+        dd::QubitCount precision{};
+        bool           includeMeasurements = true;
+        bool           dynamic             = false;
```
hillmich: These are always set via ctor init list and probably should be `const`.

include/algorithms/QPE.hpp
```diff
@@ -8,14 +8,17 @@
...
 #include <QuantumComputation.hpp>
 
+using namespace dd::literals;
+
 namespace qc {
     class QPE: public QuantumComputation {
     public:
         dd::fp         lambda    = 0.;
         dd::QubitCount precision = 0;
+        bool           iterative = false;
```
hillmich: These two may be `const`.

include/operations/OpType.hpp
```diff
@@ -0,0 +1,179 @@
...
+            case Compound: return "compound";
+            case Measure: return "measure";
+            case Reset: return "reset";
+            case Snapshot: return "snapshot";
+            case ShowProbabilities: return "show probabilities";
+            case Barrier: return "barrier";
+            case Teleportation: return "teleportation";
+            case ClassicControlled: return "classic controlled";
+            default:
+                return "";
```
hillmich: Should the default case throw an exception?

src/QuantumComputation.cpp
```diff
@@ -830,12 +828,12 @@ namespace qc {
 
                 // base case -> determine the basis state from the measurement and safe the probability
                 if (measurements.size() == nclassics - 1) {
-                    std::size_t idx0 = 0;
-                    std::size_t idx1 = 0;
+                    std::uint_least64_t idx0 = 0;
+                    std::uint_least64_t idx1 = 0;
```
hillmich: Keep these as `std::size_t`? The probability vector takes a `std::size_t` as key.

hillmich (changes requested):

hillmich (approved):
LGTM

PR #23, opened 2021-11-28
Extract circuit name from Qiskit `QuantumCircuit` object

burgholzer:
Previously, importing a Qiskit `QuantumCircuit` into the QFR would just result in a blank name. 
This PR adds the functionality to correctly extract and set the name.


hillmich (approved):
LGTM

PR #36, opened 2021-12-27
Test for compiler flag -march=native

katringoogoo:
Added a test in the CMakeLists.txt before adding the -march=native flag

The latest clang (clang-1300.0.29.30) on Apple M1 does not support it (yet?)

Depends on https://github.com/iic-jku/dd_package/pull/30 & submodule reference needs to be updated


katringoogoo:
I updated the dd_package version to the latest one that has the same change for compiler flags as this PR

burgholzer (approved):
LGTM. Thanks for this PR!

PR #37, opened 2022-01-06
🍎 Revert to building separate instead of universal binaries and wheels

burgholzer:
This PR reverts the default behaviour of building universal binaries and, in turn, universal wheels for macOS.
The main reason for this is that universal wheels are hard to create once external dependencies are involved (see, e.g., iic-jku/qmap#16).

This PR also adopts a method that allows to infer the architecture to be used for compilation from the environment (thanks @katringoogoo).


hillmich (approved):
LGTM

Issue #35, opened 2021-12-16
Export to Qiskit `QuantumCircuit`


burgholzer:
mqt-core already provides the means to import a Qiskit `QuantumCircuit` into our own C++ `qc::QuantumComputation` (see https://github.com/cda-tum/mqt-core/blob/main/src/mqt/core/plugins/qiskit.py).

In the future, it might be beneficial to provide support for translating a C++ `qc::QuantumComputation` to a Qiskit `QuantumCircuit`. At the moment, this is handled via translating to QASM and then loading the resulting string into Qiskit.
Possible applications could be the compilation of quantum circuits.

Important aspects to think about are how the `initial_layout` and the `output_permutation` are translated to Qiskit `circ.layout` objects.
In addition, the final layout should be indicated via measurements at the end of the circuits (if no measurements are provided already).

PR #96, opened 2022-05-02
✨ Measurements from output permutation

burgholzer:
This PR adds the functionality to add measurements according to the output permutation of a quantum circuit.
This is a first step towards realizing #35, in that it correctly conveys the expected output permutation in a standardized fashion.

Furthermore, this PR adds a convenience function to unify all quantum registers of a quantum circuit into a single register, which is useful for mapped circuits.


hillmich (approved):
LGTM

PR #126, opened 2022-06-17
Integrate changes of noise aware simulation

33Gjl1Xe:
This PR adds the latest changes of the noise aware quantum circuit simulation to the QFR.
More precisely, this includes the noise functionality of the stochastic noise aware quantum circuit simulator ) and the deterministic noise aware quantum circuit simulator.


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "extern/dd_package"]
 	path = extern/dd_package
 	url = https://github.com/cda-tum/dd_package.git
-	branch = main
+	branch = noiseComputeTable
```
burgholzer: Do not forget to change this back, once the DDPackage PR is merged.

include/operations/OpType.hpp
```diff
@@ -48,7 +48,13 @@ namespace qc {
...
         Teleportation,
         // Classically-controlled Operation
-        ClassicControlled
+        ClassicControlled,
+        // Noise operations
+        ATrue,
+        AFalse,
+        multiATrue,
+        multiAFalse,
+        opCount
```
burgholzer: With a small exception, every member of this Enum is capitalized. So should these
```suggestion
        MultiATrue,
        MultiAFalse,
        OpCount
```

src/parsers/RealParser.cpp
```diff
@@ -277,6 +277,9 @@ void qc::QuantumComputation::readRealGateDescriptions(std::istream& is, int line
                 case Teleportation:
                     std::cerr << "Operation with invalid type " << gate << " read from real file. Proceed with caution!" << std::endl;
                     break;
+                default:
+                    std::cerr << "Unknown operation:  " << gate << "!" << std::endl;
+                    break;
```
burgholzer: Remove the `case Compound:` ... `case Teleportation:` cases and simply replace them with `default:`.
Maybe this even fixes the Code Coverage 🤷🏼‍♂️

burgholzer (changes requested):
A few really minor things to consider.

include/operations/OpType.hpp
33Gjl1Xe: done

.gitmodules
33Gjl1Xe: indeed

src/parsers/RealParser.cpp
33Gjl1Xe: Removed the lines as suggested.

include/operations/OpType.hpp
```diff
@@ -48,7 +48,13 @@ namespace qc {
...
         Teleportation,
         // Classically-controlled Operation
-        ClassicControlled
+        ClassicControlled,
+        // Noise operations
+        ATrue,
+        AFalse,
+        MultiATrue,
+        MultiAFalse,
+        OpCount
```
burgholzer, line 58: Maybe separate the `OpCount` from the `// Noise operations` by a suitable comment

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
+/*
+* This file is part of the MQT DD Package which is released under the MIT license.
+* See file README.md or go to https://www.cda.cit.tum.de/research/quantum_dd/ for more information.
+*/
+
+#ifndef DDSIM_NOISEFUNCTIONALITY_HPP
+#define DDSIM_NOISEFUNCTIONALITY_HPP
```
burgholzer: use `#pragma once` instead of include guards.

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+        const std::unique_ptr<DDPackage>&      package;
+        dd::QubitCount                         nQubits;
+        std::uniform_real_distribution<dd::fp> dist;
+
+        double                           noiseProbability;
+        double                           noiseProbabilityMulti;
+        dd::ComplexValue                 sqrtAmplitudeDampingProbability;
+        dd::ComplexValue                 oneMinusSqrtAmplitudeDampingProbability;
+        dd::ComplexValue                 sqrtAmplitudeDampingProbabilityMulti;
+        dd::ComplexValue                 oneMinusSqrtAmplitudeDampingProbabilityMulti;
```
burgholzer: these should probably be `const`

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+        dd::GateMatrix                   ampDampingTrue{};
+        dd::GateMatrix                   ampDampingTrueMulti{};
+        dd::GateMatrix                   ampDampingFalse{};
+        dd::GateMatrix                   ampDampingFalseMulti{};
```
burgholzer: technically, these should be const as well, but then you have to change the constructor slightly.

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            package->incRef(identityDD);
+        }
+
+        ~StochasticNoiseFunctionality() {
+            package->decRef(identityDD);
+        }
+
+    protected:
+        const std::unique_ptr<DDPackage>&      package;
+        dd::QubitCount                         nQubits;
```
burgholzer: can be made `const` as it is not supposed to change

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            identityDD           = package->makeIdent(nQubits);
+            package->incRef(identityDD);
+        }
+
+        ~StochasticNoiseFunctionality() {
+            package->decRef(identityDD);
+        }
+
+    protected:
+        const std::unique_ptr<DDPackage>&      package;
```
burgholzer, line 61: I am really surprised that this `const` here works since you are conducting operations that alter the DD Package (such as the `incRef` call in the constructor.

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+        [[nodiscard]] dd::GateMatrix getAmplitudeDampingOperationMatrix(bool multiQubitNoiseFlag, bool amplitudeDampingFlag) const {
+            if (amplitudeDampingFlag) {
+                return multiQubitNoiseFlag ? ampDampingTrueMulti : ampDampingTrue;
+            } else {
+                return multiQubitNoiseFlag ? ampDampingFalseMulti : ampDampingFalse;
+            }
+        }
+
+    public:
+        dd::mEdge identityDD;
```
burgholzer: why is this member public?

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                package->decRef(state);
+                state = tmp;
+
+                // I only need to apply the operations once
+                operation = identityDD;
+            }
+        }
+
+    protected:
+        dd::mEdge generateNoiseOperation(dd::mEdge        operation,
```
burgholzer, line 136: shouldn't `operation` be defined as `const dd::mEdge&`?

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            qc::OpType effect;
+            for (const auto& noiseType: noiseEffects) {
+                if (noiseType != dd::amplitudeDamping) {
+                    effect = returnNoiseOperation(noiseType, dist(generator), multiQubitOperation);
+                } else {
+                    if (amplitudeDamping) {
+                        effect = qc::ATrue;
+                    } else {
+                        effect = qc::AFalse;
+                    }
```
burgholzer:
```suggestion
                    effect = amplitudeDamping? qc::ATrue: qc::AFalse;
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                }
+                switch (effect) {
+                    case (dd::I): {
```
burgholzer: why is it `dd::I` but `qc::...` for all the other switch cases?

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                        continue;
+                    }
+                    case (qc::ATrue): {
+                        auto tmpOperation = package->stochasticNoiseOperationCache.lookup(getAmplitudeDampingOperationType(multiQubitOperation, true), target);
```
burgholzer: extract `getAmplitudeDampingOperationType(multiQubitOperation, true)` into a constant and reuse it instead of computing it 3 times.
Same goes for the switch case down below.

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                    }
+                    case (qc::Z): {
+                        auto tmpOperation = package->stochasticNoiseOperationCache.lookup(effect, target);
+                        if (tmpOperation.p == nullptr) {
+                            tmpOperation = package->makeGateDD(dd::Zmat, getNumberOfQubits(), target);
+                            package->stochasticNoiseOperationCache.insert(effect, target, tmpOperation);
+                        }
+                        operation = package->multiply(tmpOperation, operation);
+                        break;
+                    }
```
burgholzer: there is lots of code duplication here. all of these cases follow a similar structure. even both amplitude damping cases share lots of the structure here. Consider pulling that out into a separate function.

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                }
+            }
+            return operation;
+        }
+
+        [[nodiscard]] qc::OpType returnNoiseOperation(dd::NoiseOperations noiseOperation, double prob, bool multiQubitNoiseFlag) const {
+            switch (noiseOperation) {
+                case dd::NoiseOperations::depolarization: {
+                    if (prob >= (getNoiseProbability(multiQubitNoiseFlag) * 0.75)) {
```
burgholzer, line 182: extract `getNoiseProbability(multiQubitNoiseFlag)` into a constant and avoid repeatedly calculating it.

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                        // prob > prob apply qc::I, also 25 % of the time when depolarization is applied nothing happens
+                        return qc::I;
+                    } else if (prob < (getNoiseProbability(multiQubitNoiseFlag) * 0.25)) {
+                        // if 0 < prob < 0.25 (25 % of the time when applying depolarization) apply qc::X
+                        return qc::X;
+                    } else if (prob < (getNoiseProbability(multiQubitNoiseFlag) * 0.5)) {
+                        // if 0.25 < prob < 0.5 (25 % of the time when applying depolarization) apply qc::Z
```
burgholzer:
```suggestion
                        // if 0.25 < prob < 0.5 (25 % of the time when applying depolarization) apply qc::Y
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                    } else {
+                        // if 0.5 < prob < 0.75 (25 % of the time when applying depolarization) apply qc::Z
+                        return qc::Z;
+                    }
+                }
+                case dd::NoiseOperations::phaseFlip: {
+                    if (prob > getNoiseProbability(multiQubitNoiseFlag)) {
+                        return dd::I;
+                    } else {
+                        return dd::Z;
```
burgholzer: namespace mixup. carefully check throughout the file

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+        }
+
+    protected:
+        const std::unique_ptr<DDPackage>& package;
+        dd::QubitCount                    nQubits;
+
+        double noiseProbSingleQubit;
+        double noiseProbMultiQubit;
+        double ampDampingProbSingleQubit;
+        double ampDampingProbMultiQubit;
```
burgholzer: can be made `const`

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+
+        std::vector<dd::NoiseOperations> noiseEffects;
+        bool                             useDensityMatrixType;
+        bool                             sequentiallyApplyNoise;
```
burgholzer: can also be made `const`

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+
+        const std::map<dd::NoiseOperations, int> sequentialNoiseMap = {
+                {dd::phaseFlip, 2},        //Phase-flip
+                {dd::amplitudeDamping, 2}, //Amplitude Damping
+                {dd::depolarization, 4},   //Depolarisation
```
burgholzer, line 252: just to double check: is it on purpose that `2` is occurring twice here?

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+        }
+
+    public:
+        void applyNoiseEffects(dEdge& originalEdge, const std::unique_ptr<Operation>& qcOperation) {
+            auto usedQubits = qcOperation->getTargets();
+            for (auto control: qcOperation->getControls()) {
+                usedQubits.push_back(control.qubit);
+            }
+
+            [[maybe_unused]] auto cacheSizeBefore = package->cn.cacheCount();
```
burgholzer:
```suggestion
            [[maybe_unused]] const auto cacheSizeBefore = package->cn.cacheCount();
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+
+                if (useDensityMatrixType) {
+                    dEdge::applyDmChangesToEdge(&originalEdge);
+                    nodeAfterNoise = applyNoiseEffects(originalEdge, usedQubits, false);
+                    dEdge::revertDmChangesToEdge(&originalEdge);
+                } else {
+                    nodeAfterNoise = applyNoiseEffects(originalEdge, usedQubits, true);
+                }
+                if (!nodeAfterNoise.w.exactlyZero() && !nodeAfterNoise.w.exactlyOne()) {
+                    auto tmpComplexValue = package->cn.getTemporary(CTEntry::val(nodeAfterNoise.w.r), CTEntry::val(nodeAfterNoise.w.i));
```
burgholzer:
```suggestion
                    const auto tmpComplexValue = package->cn.getTemporary(CTEntry::val(nodeAfterNoise.w.r), CTEntry::val(nodeAfterNoise.w.i));
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+
+                package->incRef(nodeAfterNoise);
+                dEdge::alignDensityEdge(&originalEdge);
+                package->decRef(originalEdge);
+                originalEdge = nodeAfterNoise;
+                if (useDensityMatrixType) {
+                    dEdge::setDensityMatrixTrue(&originalEdge);
+                }
+            }
+            [[maybe_unused]] auto cacheSizeAfter = package->cn.cacheCount();
```
burgholzer:
```suggestion
            [[maybe_unused]] const auto cacheSizeAfter = package->cn.cacheCount();
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+
+            //    auto noiseLookUpResult = dd->densityNoise.lookup(originalCopy, usedQubits);
+            //
+            //    if (noiseLookUpResult.p != nullptr) {
+            //        auto tmpComplexValue = dd->cn.getCached();
+            //        CN::mul(tmpComplexValue, noiseLookUpResult.w, originalEdge.w);
+            //        noiseLookUpResult.w = dd->cn.lookup(tmpComplexValue);
+            //        dd->cn.returnToCache(tmpComplexValue);
+            //        return noiseLookUpResult;
+            //    }
```
burgholzer, line 323: is this still needed?

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            std::array<dEdge, std::tuple_size_v<decltype(dd::dNode::e)>> new_edges{};
```
burgholzer: camelCase (`new_edges`)

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            for (short i = 0; i < 4; i++) {
```
burgholzer: the hard-coded `4` here triggers me...

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                if (firstPathEdge || i == 1) {
+                    // If I am to the useDensityMatrix I cannot minimize the necessary operations anymore
+                    dEdge::applyDmChangesToEdge(&originalCopy.p->e[i]);
+                    new_edges[i] = applyNoiseEffects(originalCopy.p->e[i], usedQubits, true);
+                    dEdge::revertDmChangesToEdge(&originalCopy.p->e[i]);
+                } else if (i == 2) {
+                    // Size e[1] == e[2] (due to density matrix representation), I can skip calculating e[2]
```
burgholzer:
```suggestion
                    // Since e[1] == e[2] (due to density matrix representation), I can skip calculating e[2]
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            //        auto tmpComplexValue = dd->cn.getCached();
+            //        CN::mul(tmpComplexValue, noiseLookUpResult.w, originalEdge.w);
+            //        noiseLookUpResult.w = dd->cn.lookup(tmpComplexValue);
+            //        dd->cn.returnToCache(tmpComplexValue);
+            //        return noiseLookUpResult;
+            //    }
+            std::array<dEdge, std::tuple_size_v<decltype(dd::dNode::e)>> new_edges{};
+            for (short i = 0; i < 4; i++) {
+                if (firstPathEdge || i == 1) {
+                    // If I am to the useDensityMatrix I cannot minimize the necessary operations anymore
```
burgholzer: that doesn't sound like a proper sentence 🙃

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                        case dd::identity:
+                            continue;
+                    }
+                }
+            }
+
+            e = package->makeDDNode(originalCopy.p->v, new_edges, true, firstPathEdge);
+
+            // Adding the noise operation to the cache, note that e.w is from the complex number table
+            //    package->densityNoise.insert(originalCopy, e, usedQubits);
```
burgholzer, line 363: is this still needed?

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                            applyAmplitudeDampingToEdges(new_edges, (usedQubits.size() == 1) ? ampDampingProbSingleQubit : ampDampingProbMultiQubit);
+                            break;
+                        case dd::phaseFlip:
+                            applyPhaseFlipToEdges(new_edges, (usedQubits.size() == 1) ? noiseProbSingleQubit : noiseProbMultiQubit);
+                            break;
+                        case dd::depolarization:
+                            applyDepolarisationToEdges(new_edges, (usedQubits.size() == 1) ? noiseProbSingleQubit : noiseProbMultiQubit);
+                            break;
+                        case dd::identity:
+                            continue;
```
burgholzer, line 355: it should be easy to cover these lines with tests

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                    e.w = package->cn.getCached(CTEntry::val(originalEdge.w.r), CTEntry::val(originalEdge.w.i));
+                } else {
+                    CN::mul(e.w, e.w, originalEdge.w);
+                }
+            }
+            return e;
+        }
+
+        void
+        applyPhaseFlipToEdges(std::array<dEdge, std::tuple_size_v<decltype(dd::dNode::e)>>& e, double probability) {
```
burgholzer: please consider a `using` typedef declaration for the rather verbose `std::array<dEdge, std::tuple_size_v<decltype(dd::dNode::e)>>` type

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+            package->cn.returnToCache(complexProb);
+        }
+
+        void applyDetNoiseSequential(dEdge& originalEdge, const qc::Targets& targets) {
+            dd::dEdge tmp = {};
+
+            std::array<mEdge, std::tuple_size_v<decltype(dd::dNode::e)>> idleOperation{};
+
+            // Iterate over qubits and check if the qubit had been used
+            for (auto targetQubit: targets) {
```
burgholzer:
```suggestion
            for (const auto targetQubit: targets) {
```

include/dd/NoiseFunctionality.hpp
```diff
@@ -0,0 +1,714 @@
...
+                    idleNoiseGate[0][0] = idleNoiseGate[0][3] = tmp;
+                    idleNoiseGate[0][1] = idleNoiseGate[0][2] = dd::complex_zero;
+                    tmp.r                                     = std::sqrt(probability) * dd::complex_one.r;
+                    idleNoiseGate[1][0]                       = tmp;
+                    tmp.r *= -1;
+                    idleNoiseGate[1][3] = tmp;
+                    idleNoiseGate[1][1] = idleNoiseGate[1][2] = dd::complex_zero;
+
+                    pointerForMatrices[0] = package->makeGateDD(idleNoiseGate[0], getNumberOfQubits(), target);
+                    pointerForMatrices[1] = package->makeGateDD(idleNoiseGate[1], getNumberOfQubits(), target);
```
burgholzer, line 631: this is missing a comment with the matrix description.
please also check the formatting of the other comments as it seems to be quite a bit off.

test/unittests/test_dd_noise_functionality.cpp
```diff
@@ -0,0 +1,351 @@
...
+                                             DensityMatrixSimulatorDDPackageConfig::CT_MAT_MAT_MULT_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::CT_VEC_KRON_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::CT_MAT_KRON_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::CT_VEC_INNER_PROD_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::CT_DM_NOISE_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::UT_DM_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::UT_DM_INITIAL_ALLOCATION_SIZE,
+                                             DensityMatrixSimulatorDDPackageConfig::CT_DM_DM_MULT_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::CT_DM_ADD_NBUCKET,
+                                             DensityMatrixSimulatorDDPackageConfig::STOCHASTIC_CACHE_OPS>;
```
burgholzer, line 83: see also the comments in the DD Package PR.
Are these the same packages that are actually being used in the simulator?

if yes, then avoid duplicating them and define them once in the DD Package include files.
if no, then this just contributes to the compilation time of the whole project.

test/unittests/test_dd_noise_functionality.cpp
```diff
@@ -0,0 +1,351 @@
...
+    EXPECT_NEAR(m.find("0110")->second, 0.0244576087400, tolerance);
+    EXPECT_NEAR(m.find("0111")->second, 0.0116282811276, tolerance);
+    EXPECT_NEAR(m.find("1000")->second, 0.1731941264570, tolerance);
+    EXPECT_NEAR(m.find("1001")->second, 0.4145855071998, tolerance);
+    EXPECT_NEAR(m.find("1010")->second, 0.0138062113213, tolerance);
+    EXPECT_NEAR(m.find("1011")->second, 0.0184033482066, tolerance);
+    EXPECT_NEAR(m.find("1100")->second, 0.0242454336917, tolerance);
+    EXPECT_NEAR(m.find("1101")->second, 0.0262779844799, tolerance);
+    EXPECT_NEAR(m.find("1110")->second, 0.0239296920989, tolerance);
+    EXPECT_NEAR(m.find("1111")->second, 0.0110373166627, tolerance);
```
burgholzer, line 165: can it really be expected, that the amplitudes are that close to these cryptic numbers all the time?
phrased differently: how high is the chance of these tests becoming flaky?

test/unittests/test_dd_noise_functionality.cpp
```diff
@@ -0,0 +1,351 @@
...
+    for (size_t i = 0U; i < stochRuns; i++) {
+        auto rootEdge = dd->makeZeroState(qc.getNqubits());
+        dd->incRef(rootEdge);
+
+        for (auto const& op: qc) {
+            auto        operation  = dd::getDD(op.get(), dd);
+            std::vector usedQubits = op->getTargets();
+            for (auto control: op->getControls()) {
+                usedQubits.push_back(control.qubit);
+            }
```
burgholzer: code like this is used in several places. please extract this into a method of the `Operation` class and suitably override it for the individual operation types.

burgholzer (changes requested):
Also here: quite close to the finish line. A couple more remarks and requests, but nothing major.

include/operations/OpType.hpp
33Gjl1Xe: This is the case.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: Changed it to `#pragma once`

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: To me it seems like the pointer is constant and not the object the pointer points to.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: Because I need it in the Simulator.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: The noise `operation` are "stacked" onto the operation. Thereby the operation is modified and hence no const qualifier is used.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: I am not sure, what you mean.

include/operations/OpType.hpp
burgholzer: In the current code, there is no comment between the noise operations and the `OpCount`

include/dd/NoiseFunctionality.hpp
burgholzer: True, still surprised.

include/dd/NoiseFunctionality.hpp
burgholzer: Then make it protected and provide a getter for it.

include/dd/NoiseFunctionality.hpp
burgholzer: I see. Would it make sense that this function actually takes a reference as a parameter and returns void instead?
If not, then feel free to close leave it as is and resolve this conversation

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
burgholzer: In each of the switch cases the value `getAmplitudeDampingOperationType(multiQubitOperation, true)` (or similar) is computed three times. Just extract that into a constant
```c++
const auto effect = getAmplitudeDampingOperationType(multiQubitOperation, true)
```

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: Changed that

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: I am not calculating anything here. Here only the correct probability is returned depending if it is a multi qubit operation or not.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: Removed it

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: fixed

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: Indeed, that sentence broke sometime during refactoring. I fixed it.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

test/unittests/test_dd_noise_functionality.cpp
33Gjl1Xe: These are the expected probabilities for the provided parameters (as calculated with another quantum circuit simulator). I have never had any problems with the expected accuracy. However, if problems should occur the tolerance can 
be increased.

test/unittests/test_dd_noise_functionality.cpp
33Gjl1Xe: Done

test/unittests/test_dd_noise_functionality.cpp
33Gjl1Xe: see discussion in the DD Package PR.

test/unittests/test_dd_noise_functionality.cpp
burgholzer: Sound's reasonable 👍🏼

include/dd/NoiseFunctionality.hpp
burgholzer: Yeah, but that is done like three times over the course of the `if` and `else if` branches 🙃

include/operations/OpType.hpp
33Gjl1Xe: My bad. I misread your comment. I am gonna add such a comment.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: I refactored this code block and extracted duplicate code.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: yes!

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: yes. I would like to keep that.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: yes. I would like to keep that.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: done

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: I refactored the code. Now all information is stored within const variables.

include/operations/Operation.hpp
```diff
@@ -96,6 +96,14 @@ namespace qc {
             return startQubit;
         }
 
+        [[nodiscard]] virtual std::vector<dd::Qubit> getUsedQubits() const {
+            auto usedQubits = targets;
+            for (auto control: controls) {
+                usedQubits.push_back(control.qubit);
+            }
+            return usedQubits;
+        }
```
burgholzer: It is not sufficient to just define this in the base class without providing suitable overloads in the derived classes.
You can get around the overload for `ClassicControlledOperation` by using `getTargets()` and `getControls()` instead of `targets` and `controls` here.
For compound operations you have to loop through all operations and accumulate the respective qubits.
For non-unitary operations, it's just calling `getTargets()`.

include/dd/NoiseFunctionality.hpp
```diff
@@ -3,16 +3,18 @@
...
 #include "dd/Export.hpp"
+#include "operations/OpType.hpp"
+#include "operations/Operation.hpp"
 
 #include <random>
 #include <utility>
 
-using CN = dd::ComplexNumbers;
+using CN           = dd::ComplexNumbers;
+using arrayOfEdges = std::array<dd::DensityMatrixDD, std::tuple_size_v<decltype(dd::dNode::e)>>;
```
burgholzer: Types should start with a capital letter.

include/dd/NoiseFunctionality.hpp
```diff
@@ -344,43 +326,43 @@ namespace dd {
             //        dd->cn.returnToCache(tmpComplexValue);
             //        return noiseLookUpResult;
             //    }
-            std::array<dEdge, std::tuple_size_v<decltype(dd::dNode::e)>> new_edges{};
-            for (short i = 0; i < 4; i++) {
+            arrayOfEdges newEdges{};
+            for (size_t i = 0; i < std::tuple_size_v<decltype(dd::dNode::e)>; i++) {
```
burgholzer:
```suggestion
            for (size_t i = 0; i < newEdges.size(); i++) {
```

burgholzer (changes requested):
Also here. One more thing that needs fixing and 2 very little details, then I'll gladly merge this PR.

include/operations/Operation.hpp
33Gjl1Xe: I see. I added the overloads and also included tests for the new functions.

include/dd/NoiseFunctionality.hpp
33Gjl1Xe: I changed that

include/operations/CompoundOperation.hpp
```diff
@@ -158,5 +158,20 @@ namespace qc {
...
+
+        [[nodiscard]] std::vector<dd::Qubit> getUsedQubits() const override {
+            std::vector<dd::Qubit> usedQubits{};
+            for (const auto& op: ops) {
+                for (auto target: op->getTargets()) {
+                    usedQubits.push_back(target);
+                }
+                for (auto control: op->getControls()) {
+                    usedQubits.push_back(control.qubit);
+                }
```
burgholzer: this is not guaranteed to work properly. you have to call `getUsedQubits` on each sub-operation and then add build the union of the current set and the newly generated set. To this end, it is probably easiest and cleanest to just replace the `std::vector<dd::Qubit>` with a `std::set<dd::Qubit>` which handles sorting an uniqueness automatically.
If you really need a `vector` as the result of the call (which I don't think you need), then you can still construct a vector from the set by (pseudocode)
```c++
auto vec = std::vector{set.begin(), set.end()};
```

(the union of two `set`s can be easily be computed in-place by using `set1.insert(set2.begin(), set2.end())`)

include/operations/NonUnitaryOperation.hpp
```diff
@@ -103,5 +103,9 @@ namespace qc {
 
         void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg, const RegisterNames& creg) const override;
         void dumpQiskit(std::ostream& of, const RegisterNames& qreg, const RegisterNames& creg, const char* anc_reg_name) const override;
+
+        [[nodiscard]] std::vector<dd::Qubit> getUsedQubits() const override {
+            return getTargets();
```
burgholzer: See comment above. This would have to be changed to
```c++
const auto& targets = getTargets();
return {targets.begin(), targets.end()};
```

include/operations/Operation.hpp
```diff
@@ -97,8 +97,8 @@ namespace qc {
         }
 
         [[nodiscard]] virtual std::vector<dd::Qubit> getUsedQubits() const {
-            auto usedQubits = targets;
-            for (auto control: controls) {
+            auto usedQubits = getTargets();
+            for (auto control: getControls()) {
                 usedQubits.push_back(control.qubit);
             }
             return usedQubits;
```
burgholzer: As pointed out above, this would have to change to something like
```c++
const auto& targets = getTargets();
const auto& controls = getControls();
auto usedQubits = std::set{targets.begin(), targets.end()};
for (const auto& control: controls) {
  usedQubits.insert(control.qubit);
}
```

test/unittests/test_dd_noise_functionality.cpp
```diff
@@ -343,3 +344,27 @@ TEST_F(DDNoiseFunctionalityTest, StochSimulateAdder4IdentiyError) {
...
+    compoundOp.emplace_back<qc::StandardOperation>(nqubits, 0, qc::X);
+    EXPECT_EQ(compoundOp.getUsedQubits().size(), 2);
+    EXPECT_EQ(compoundOp.getUsedQubits()[0], 0);
+    EXPECT_EQ(compoundOp.getUsedQubits()[1], 1);
+
+    std::unique_ptr<qc::Operation> xOp                   = std::make_unique<qc::StandardOperation>(nqubits, 0, qc::X);
+    auto                           classicalControlledOp = qc::ClassicControlledOperation(xOp, std::pair{0, nqubits}, 1U);
+    EXPECT_EQ(nonUnitaryOp.getUsedQubits().size(), 1);
+    EXPECT_EQ(nonUnitaryOp.getUsedQubits()[0], 0);
+}
```
burgholzer: there is a newline missing at the end of this file

burgholzer (changes requested):
The new function is not fully functional. See the inline comments for further details.

test/unittests/test_dd_noise_functionality.cpp
33Gjl1Xe: added a newline

include/operations/Operation.hpp
33Gjl1Xe: Changed the code as suggested

include/operations/NonUnitaryOperation.hpp
33Gjl1Xe: Changed the code as suggested

include/operations/CompoundOperation.hpp
33Gjl1Xe: Indeed! I changed the code as suggested.

PR #135, opened 2022-06-27
Qubit sanity check

pehamTom:
Added methods that check whether executing a given operation makes sense with respect to the indices of the initial layout.


src/QuantumComputation.cpp
```diff
@@ -998,4 +998,37 @@ namespace qc {
             measure(qubit, clbit);
         }
     }
+
+    void QuantumComputation::checkQubitRange(dd::Qubit qubit) const {
+        if (initialLayout.find(qubit) == initialLayout.end() || initialLayout.at(qubit) >= getNqubits())
```
burgholzer: Very tiny nitpick: `initialLayout.find(qubit)` already returns an iterator to the element, if it was found. The subsequent `at` call searches for the entry again. I think the idiomatic way to do this would be
```suggestion
        if (const auto it = initialLayout.find(qubit); it == initialLayout.end() || *it >= getNqubits())
```

PR #199, opened 2022-10-05
Symbolic

pehamTom:
This PR adds a representation for Quantum Circuits that have symbolic parameters. The symbolic computations are handled within the MQT ZX Package.


.gitmodules
```diff
@@ -1,25 +1,25 @@
...
+        branch = master
+        shallow = true
 [submodule "extern/zx"]
-	path = extern/zx
-	url = https://github.com/cda-tum/zx.git
-	branch = main
-	shallow = true
+        path = extern/zx
+        url = https://github.com/cda-tum/zx.git
+        branch = symbolic
```
burgholzer: just as a reminder: this has to be changed back once the ZX package PR is merged.

include/QuantumComputation.hpp
```diff
@@ -638,9 +671,23 @@ namespace qc {
...
+
+        [[nodiscard]] const std::unordered_set<sym::Variable>& getVariables() const {
+            return occuringVariables;
+        }
+
         /**
-		 * printing
-		 */
+                 * printing
+                 */
```
burgholzer: is this whitespace change intentional?

include/QuantumComputation.hpp
```diff
@@ -675,8 +722,8 @@ namespace qc {
         }
 
         /**
-		 * Pass-Through
-		 */
+                 * Pass-Through
+                 */
```
burgholzer: is this whitespace change intentional?

include/QuantumComputation.hpp
```diff
@@ -638,9 +671,23 @@ namespace qc {
             max_controls = std::max(ncontrols, max_controls);
         }
 
+        void instantiate(const VariableAssignment& assignment);
+
+        void addVariables(const SymbolOrNumber& expr);
+        void addVariables(const SymbolOrNumber& expr1, const SymbolOrNumber& expr2);
+        void addVariables(const SymbolOrNumber& expr1, const SymbolOrNumber& expr2, const SymbolOrNumber& expr3);
+
+        [[nodiscard]] bool isVariableFree() const {
```
burgholzer, line 688: I am wondering whether it be more suitable to provide a `isSymbolic()` function.
I feel like "not symbolic" is more common and straight-forward than "variable-free".

include/QuantumComputation.hpp
```diff
@@ -638,9 +671,23 @@ namespace qc {
             max_controls = std::max(ncontrols, max_controls);
         }
 
+        void instantiate(const VariableAssignment& assignment);
+
+        void addVariables(const SymbolOrNumber& expr);
+        void addVariables(const SymbolOrNumber& expr1, const SymbolOrNumber& expr2);
+        void addVariables(const SymbolOrNumber& expr1, const SymbolOrNumber& expr2, const SymbolOrNumber& expr3);
```
burgholzer: This is really just a suggestion, so feel free to ignore it, but I believe this is a perfect use case for "fold expressions" (see https://en.cppreference.com/w/cpp/language/fold), i.e., 
call the first function (that adds one variable) `addVariable` and then create a function
```c++
template <SymbolOrNumber... Vars> void addVariables(const Vars&... vars) {
  (addVariable(vars), ...);
}
```
This allows to call the function with an arbitrary amount of parameters. So nothing has to be changed if, in the future, a gate with four parameters comes along.

mqt/qfr/bindings.cpp
```diff
@@ -107,6 +107,7 @@ py::dict construct_circuit(const py::object& circ, const ConstructionMethod& met
         ss << "Could not import circuit: " << e.what();
         return py::dict("error"_a = ss.str());
     }
+
```
burgholzer: Even though it's just a whitespace change, I don't believe it belongs here.

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -7,6 +7,11 @@
 #define QFR_QUANTUMCIRCUIT_HPP
 
 #include "pybind11/pybind11.h"
+#include "regex"
```
burgholzer:
```suggestion
```

src/CMakeLists.txt
```diff
@@ -80,35 +82,35 @@ enable_lto(${PROJECT_NAME})
...
+        set_target_properties(${PROJECT_NAME} PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
 endif()
 
 # add coverage compiler and linker flag to the library and all targets that link against it, if COVERAGE is set
 if (COVERAGE)
-	target_compile_options(${PROJECT_NAME} PUBLIC --coverage)
-	target_link_libraries(${PROJECT_NAME} PUBLIC --coverage)
+        target_compile_options(${PROJECT_NAME} PUBLIC --coverage)
+        target_link_libraries(${PROJECT_NAME} PUBLIC --coverage)
 endif()
```
burgholzer, line 114: these whitespace changes probably originate from some difference in tab width or such.

(which reminds me, that we should really set up pre-commit for the QFR as well)

src/QuantumComputation.cpp
```diff
@@ -1032,4 +1032,142 @@ namespace qc {
...
+
+    //         if (symOp) {
+    //             const auto stdOp = symOp->getInstantiatedOperation(assignment);
+    //             qc.emplace_back<StandardOperation>(stdOp.getNqubits(), stdOp.getControls(), stdOp.getTargets(), stdOp.getType(), stdOp.getParameter()[0], stdOp.getParameter()[1], stdOp.getParameter()[2], stdOp.getStartingQubit());
+    //         } else {
+    //             qc.emplace_back<Operation>(std::make_unique<Operation>(op->clone()));
+    //         }
+    //     }
+    //     return qc;
+    // }
```
burgholzer: left-over commented code. can this be removed?

src/QuantumComputation.cpp
```diff
@@ -1032,4 +1032,142 @@ namespace qc {
...
+
+    // Instantiates this computation
+    // Afterwards casting to QuantumComputation can be done if assignment is total
```
burgholzer: This comment does seem like a relic from when there was a `SymbolicQuantumComputation`

src/QuantumComputation.cpp
```diff
@@ -1032,4 +1032,142 @@ namespace qc {
...
+    void QuantumComputation::instantiate(const VariableAssignment& assignment) {
+        for (auto& op: *this) {
```
burgholzer: I think its cleaner to directly iterate over the `ops` member
```suggestion
        for (auto& op: ops) {
```

src/QuantumComputation.cpp
```diff
@@ -1032,4 +1032,142 @@ namespace qc {
...
+            auto* symOp = dynamic_cast<SymbolicOperation*>(op.get());
+            if (symOp) {
```
burgholzer: The scope of this variable can be limited further by defining it in the `if`
```suggestion
            if (auto* symOp = dynamic_cast<SymbolicOperation*>(op.get()); symOp) {
```
Maybe this can even be
```c++
            if (auto* symOp = dynamic_cast<SymbolicOperation*>(op.get())) {
```

src/operations/StandardOperation.cpp
```diff
@@ -11,6 +11,7 @@ namespace qc {
     /***
      * Protected Methods
      ***/
+
```
burgholzer: another whitespace change that technically does not belong to this PR.

test/CMakeLists.txt
```diff
@@ -41,19 +41,21 @@ endmacro()
...
+                 ${CMAKE_CURRENT_SOURCE_DIR}/algorithms/test_random_clifford.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/algorithms/test_qpe.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/unittests/test_io.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/unittests/test_dd_functionality.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/unittests/test_qfr_functionality.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/algorithms/eval_dynamic_circuits.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/unittests/test_zx_functionality.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/unittests/test_symbolic.cpp
+                 ${CMAKE_CURRENT_SOURCE_DIR}/unittests/test_dd_noise_functionality.cpp
+               )
```
burgholzer: unintended whitespace change

test/unittests/test_symbolic.cpp
```diff
@@ -0,0 +1,226 @@
...
+    Symbolic yMonom = Symbolic{Term<fp>{y}};
+    Symbolic zMonom = Symbolic{Term<fp>{z}};
+
+    std::unique_ptr<QuantumComputation> symQc = std::make_unique<QuantumComputation>(3);
+    std::unique_ptr<QuantumComputation> qc    = std::make_unique<QuantumComputation>(3);
+
+protected:
+    // void TearDown override{};
+
+    void SetUp() override{};
```
burgholzer:
```suggestion
```

test/unittests/test_symbolic.cpp
```diff
@@ -0,0 +1,226 @@
...
+    noRealSymQc->rz(0, {1_pc, 2_nc}, xVal);
+    noRealSymQc->rz(0, {1_pc}, xVal);
+    noRealSymQc->rz(0, xVal);
+
+    EXPECT_TRUE(noRealSymQc->isVariableFree());
+
+    for (auto it1 = symQc->begin(), it2 = noRealSymQc->begin(); it1 != symQc->end() && it2 != noRealSymQc->end(); ++it1, ++it2) {
+        // const auto* symOp1 = dynamic_cast<SymbolicOperation*>(it1->get());
+        // const auto* symOp2 = dynamic_cast<SymbolicOperation*>(it2->get());
+        // EXPECT_FALSE(symOp1->equalsSymbolic(*symOp1));
```
burgholzer: could be removed

test/unittests/test_symbolic.cpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    VariableAssignment assignment{{x, xVal}, {y, yVal}, {z, zVal}};
+
+    symQc->instantiate(assignment);
+
+    for (auto it1 = symQc->begin(), it2 = noRealSymQc->begin(); it1 != symQc->end() && it2 != noRealSymQc->end(); ++it1, ++it2) {
+        // const auto* symOp1 = dynamic_cast<SymbolicOperation*>(it1->get());
+        // const auto* symOp2 = dynamic_cast<SymbolicOperation*>(it2->get());
+        // EXPECT_TRUE(symOp1->equalsSymbolic(*symOp2));
```
burgholzer: could be removed

test/unittests/test_symbolic.cpp
```diff
@@ -0,0 +1,226 @@
...
+    Variable x = Variable("x");
+    Variable y = Variable("y");
+    Variable z = Variable("z");
+
+    Symbolic xMonom = Symbolic{Term<fp>{x}};
+    Symbolic yMonom = Symbolic{Term<fp>{y}};
+    Symbolic zMonom = Symbolic{Term<fp>{z}};
+
+    std::unique_ptr<QuantumComputation> symQc = std::make_unique<QuantumComputation>(3);
+    std::unique_ptr<QuantumComputation> qc    = std::make_unique<QuantumComputation>(3);
```
burgholzer: Not that it is per-se wrong, but these could just be regular `QuantumComputation`s (without the smart pointers)

test/unittests/test_zx_functionality.cpp
```diff
@@ -13,6 +13,7 @@
 #include "gtest/gtest.h"
 #include <cstddef>
 #include <iostream>
+#include <memory>
```
burgholzer: maybe I am wrong, but I don't see why this include is necessary. typically it's only included for the smart pointers.

src/zx/FunctionalityConstruction.cpp
```diff
@@ -7,12 +7,14 @@
...
 #include <string>
+#include <variant>
 #include <vector>
 
 namespace zx {
 
-    bool FunctionalityConstruction::checkSwap(const op_it it, const op_it end, const Qubit ctrl,
-                                              const Qubit target, const qc::Permutation& p) {
+    bool FunctionalityConstruction::checkSwap(const op_it& it, const op_it& end, Qubit ctrl,
+                                              Qubit                  target,
```
burgholzer:
```suggestion
    bool FunctionalityConstruction::checkSwap(const op_it& it, const op_it& end, const Qubit ctrl,
                                              const Qubit                  target,
```
The const devil strikes again 😈

src/zx/FunctionalityConstruction.cpp
```diff
@@ -30,23 +32,23 @@ namespace zx {
 
     void FunctionalityConstruction::addZSpider(ZXDiagram& diag, zx::Qubit qubit,
                                                std::vector<Vertex>& qubits,
-                                               const Expression& phase, EdgeType type) {
-        auto newVertex = diag.addVertex(
+                                               const PiExpression& phase, EdgeType type) {
```
burgholzer:
```suggestion
                                               const PiExpression& phase, const EdgeType type) {
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -30,23 +32,23 @@ namespace zx {
 
     void FunctionalityConstruction::addZSpider(ZXDiagram& diag, zx::Qubit qubit,
```
burgholzer:
```suggestion
    void FunctionalityConstruction::addZSpider(ZXDiagram& diag, const zx::Qubit qubit,
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -30,23 +32,23 @@ namespace zx {
...
-        auto newVertex = diag.addVertex(
+                                               const PiExpression& phase, EdgeType type) {
+        auto new_vertex = diag.addVertex(
                 qubit, diag.getVData(qubits[qubit]).value().col + 1, phase,
                 VertexType::Z);
 
-        diag.addEdge(qubits[qubit], newVertex, type);
-        qubits[qubit] = newVertex;
+        diag.addEdge(qubits[qubit], new_vertex, type);
+        qubits[qubit] = new_vertex;
```
burgholzer: why the naming convention change?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -30,23 +32,23 @@ namespace zx {
...
     }
 
     void FunctionalityConstruction::addXSpider(ZXDiagram& diag, Qubit qubit,
```
burgholzer:
```suggestion
    void FunctionalityConstruction::addXSpider(ZXDiagram& diag, const Qubit qubit,
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -30,23 +32,23 @@ namespace zx {
...
                                                std::vector<Vertex>& qubits,
-                                               const Expression& phase, EdgeType type) {
-        const auto newVertex = diag.addVertex(
+                                               const PiExpression& phase, EdgeType type) {
```
burgholzer:
```suggestion
                                               const PiExpression& phase, const EdgeType type) {
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -30,23 +32,23 @@ namespace zx {
...
+        const auto new_vertex = diag.addVertex(
                 qubit, diag.getVData(qubits[qubit]).value().col + 1, phase,
                 VertexType::X);
-        diag.addEdge(qubits[qubit], newVertex, type);
-        qubits[qubit] = newVertex;
+        diag.addEdge(qubits[qubit], new_vertex, type);
+        qubits[qubit] = new_vertex;
```
burgholzer: why the naming convention change?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -56,13 +58,18 @@ namespace zx {
         diag.addEdge(qubits[ctrl], qubits[target]);
     }
 
-    void FunctionalityConstruction::addCphase(ZXDiagram& diag, const PiRational& phase, Qubit ctrl, Qubit target,
-                                              std::vector<Vertex>& qubits) {
-        addZSpider(diag, ctrl, qubits, Expression(phase / 2));
+    void
+    FunctionalityConstruction::addCphase(ZXDiagram& diag, const PiExpression& phase,
+                                         Qubit ctrl, Qubit target,
```
burgholzer:
```suggestion
                                         const Qubit ctrl, const Qubit target,
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -56,13 +58,18 @@ namespace zx {
...
+                                         std::vector<Vertex>& qubits) {
+        auto new_const = phase.getConst() / 2;
+        auto new_phase = phase / 2.0;
```
burgholzer: wrong naming convention

src/zx/FunctionalityConstruction.cpp
```diff
@@ -56,13 +58,18 @@ namespace zx {
...
+        addZSpider(diag, ctrl, qubits, new_phase); //todo maybe should provide a method for int division
         addCnot(diag, ctrl, target, qubits);
-        addZSpider(diag, target, qubits, Expression(-phase / 2));
+        addZSpider(diag, target, qubits, -new_phase);
         addCnot(diag, ctrl, target, qubits);
-        addZSpider(diag, target, qubits, Expression(phase / 2));
+        addZSpider(diag, target, qubits, new_phase);
     }
 
     void FunctionalityConstruction::addSwap(ZXDiagram& diag, Qubit ctrl, Qubit target,
```
burgholzer:
```suggestion
    void FunctionalityConstruction::addSwap(ZXDiagram& diag, const Qubit ctrl, const Qubit target,
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -72,30 +79,28 @@ namespace zx {
...
         const auto t1 = diag.addVertex(ctrl, diag.getVData(qubits[target]).value().col + 1);
-
         diag.addEdge(s0, t1);
         diag.addEdge(s1, t0);
-
         qubits[target] = t0;
         qubits[ctrl]   = t1;
     }
 
     void FunctionalityConstruction::addCcx(ZXDiagram& diag, Qubit ctrl0, Qubit ctrl1, Qubit target,
```
burgholzer:
```suggestion
    void FunctionalityConstruction::addCcx(ZXDiagram& diag, const Qubit ctrl0, const Qubit ctrl1, const Qubit target,
```

src/zx/FunctionalityConstruction.cpp
```diff
@@ -333,13 +350,14 @@ namespace zx {
...
+        return std::all_of(qc->cbegin(), qc->cend(), [&](const auto& op) { return transformableToZX(op.get()); });
     }
 
     bool FunctionalityConstruction::transformableToZX(qc::Operation* op) {
         if (op->getType() == qc::OpType::Compound) {
-            auto const* compOp = dynamic_cast<qc::CompoundOperation*>(op);
-            return std::all_of(compOp->cbegin(), compOp->cend(), [](const auto& subOp) { return transformableToZX(subOp.get()); });
+            auto* compOp = dynamic_cast<qc::CompoundOperation*>(op);
+
+            return std::all_of(compOp->cbegin(), compOp->cend(), [&](const auto& op) { return transformableToZX(op.get()); });
```
burgholzer: over the course of this minor change, `compOp` lost its `const`. Was that on purpose?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -333,13 +350,14 @@ namespace zx {
         return diag;
     }
     bool FunctionalityConstruction::transformableToZX(const qc::QuantumComputation* qc) {
-        return std::all_of(qc->cbegin(), qc->cend(), [](const auto& op) { return transformableToZX(op.get()); });
+        return std::all_of(qc->cbegin(), qc->cend(), [&](const auto& op) { return transformableToZX(op.get()); });
     }
 
     bool FunctionalityConstruction::transformableToZX(qc::Operation* op) {
```
burgholzer: could `op` here technically also be a `const` pointer?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -397,4 +417,21 @@ namespace zx {
         return false;
     }
 
+    PiExpression FunctionalityConstruction::parseParam(const qc::Operation* op,
+                                                       std::size_t          i) {
```
burgholzer:
```suggestion
                                                       const std::size_t          i) {
```

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
             }
         }
 
+        static SymbolOrNumber parseSymbolicExpr(const py::object& pyExpr) {
+            static const std::regex summands("[+|-]?[^+-]+");
+            static const std::regex products("[\\*/]?[^\\*/]+");
+            // static const std::regex products(".*");
+            // static const products("")
```
burgholzer: can these left-over comments be removed?

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+            auto exprStr = pyExpr.attr("__str__")().cast<std::string>();
+            exprStr.erase(std::remove(exprStr.begin(), exprStr.end(), ' '),
+                          exprStr.end()); // strip whitespace
+
+            auto sumIt  = std::sregex_iterator(exprStr.begin(), exprStr.end(), summands);
+            auto sumEnd = std::sregex_iterator();
```
burgholzer:
```suggestion
            const auto sumEnd = std::sregex_iterator();
```

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+
+            qc::Symbolic sym;
+            bool         isConst = true;
+
+            while (sumIt != sumEnd) {
+                auto match    = *sumIt;
+                auto matchStr = match.str();
+                int  sign     = matchStr[0] == '-' ? -1 : 1;
```
burgholzer:
```suggestion
                const int  sign     = matchStr[0] == '-' ? -1 : 1;
```

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+                auto prodIt  = std::sregex_iterator(matchStr.begin(), matchStr.end(), products);
+                auto prodEnd = std::sregex_iterator();
+
+                dd::fp      coeff = 1.0;
+                std::string var   = "";
+                while (prodIt != prodEnd) {
+                    auto prodMatch = *prodIt;
+                    auto prodStr   = prodMatch.str();
+
+                    bool isDiv = prodStr[0] == '/';
```
burgholzer:
```suggestion
                    const bool isDiv = prodStr[0] == '/';
```

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+                    if (prodStr[0] == '*' || prodStr[0] == '/') {
+                        prodStr.erase(0, 1);
+                    }
+
+                    std::istringstream iss(prodStr);
+                    dd::fp             f;
+                    iss >> f;
+                    bool isNum = iss.eof() && !iss.fail();
+
+                    if (isNum) {
```
burgholzer:
```suggestion
                    if (iss.eof() && !iss.fail()) {
```

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+                        coeff *= isDiv ? 1.0 / f : f;
+                    } else {
+                        var = prodStr;
+                    }
+
+                    prodIt++;
```
burgholzer: prefer prefix over postfix increment

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+                }
+                if (var.empty()) {
+                    sym += coeff;
+                } else {
+                    isConst = false;
+                    sym += sym::Term(sign * coeff, sym::Variable{var});
+                }
+                sumIt++;
```
burgholzer: prefer prefix over postfix increment

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -192,6 +197,77 @@ namespace qc::qiskit {
...
+            }
+
+            return isConst ? SymbolOrNumber{sym.getConst()} : SymbolOrNumber{sym};
```
burgholzer: While not necessarily bad, people seem to prefer to avoid ternary if constructs.
I tend to agree and would probably write this as
```c++
if (isConst) {
  return {sim.getConst()};
}
return {sim};
```

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -200,19 +276,26 @@ namespace qc::qiskit {
             }
             auto target = qubits.back().qubit;
             qubits.pop_back();
-            dd::fp theta = 0., phi = 0., lambda = 0.;
+            // dd::fp theta = 0., phi = 0., lambda = 0.;
```
burgholzer: left-over comment that can be removed

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -225,19 +308,25 @@ namespace qc::qiskit {
             qubits.pop_back();
             auto target0 = qubits.back().qubit;
             qubits.pop_back();
-            dd::fp theta = 0., phi = 0., lambda = 0.;
+            // dd::fp             theta = 0., phi = 0., lambda = 0.;
```
burgholzer: left-over comment

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+    template<typename... Ts>
+    struct Overload: Ts... {
+        using Ts::operator()...;
+    };
+    template<class... Ts>
+    Overload(Ts...) -> Overload<Ts...>;
+
+    class SymbolicOperation final: public StandardOperation {
+    protected:
+        std::array<std::optional<Symbolic>, MAX_PARAMETERS> symbolicParameter{{std::nullopt, std::nullopt, std::nullopt}};
```
burgholzer: I might be wrong, but isn't `std::nullopt` the default anyway?
Then this can just be 
```suggestion
        std::array<std::optional<Symbolic>, MAX_PARAMETERS> symbolicParameter{};
```

(which also wouldn't break if `MAX_PARAMETERS` were to change)

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+        static OpType parseU3(const Symbolic& lambda, dd::fp& phi, dd::fp& theta);
+        static OpType parseU3(dd::fp& lambda, const Symbolic& phi, dd::fp& theta);
+        static OpType parseU3(dd::fp& lambda, dd::fp& phi, const Symbolic& theta);
+        static OpType parseU3(const Symbolic& lambda, const Symbolic& phi, dd::fp& theta);
+        static OpType parseU3(const Symbolic& lambda, dd::fp& phi, const Symbolic& theta);
+        static OpType parseU3(dd::fp& lambda, const Symbolic& phi, const Symbolic& theta);
```
burgholzer, line 31: Again, I might be stupid, but isn't this missing a parsing method where all three parameters are symbolic?

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+        static OpType parseU2(const Symbolic& lambda, dd::fp& phi);
+        static OpType parseU2(dd::fp& lambda, const Symbolic& phi);
+
+        static OpType parseU1(const Symbolic& lambda);
+
+        void checkUgate();
+
+        void storeSymbolOrNumber(const SymbolOrNumber& param, std::size_t i);
+
+        [[nodiscard]] bool isSymbolicParameter(std::size_t i) const {
```
```suggestion
        [[nodiscard]] bool isSymbolicParameter(const std::size_t i) const {
```

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+    public:
+        SymbolicOperation() = default;
+
+        // [[nodiscard]] const std::array<Symbolic, MAX_PARAMETERS>& getSymbolicParameter() const {
+        //     return symbolicParameter;
+        // }
+
+        // std::array<Symbolic, MAX_PARAMETERS>& getSymbolicParameter() {
+        //     return symbolicParameter;
+        // }
```
burgholzer: left-over comments

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+        [[nodiscard]] SymbolOrNumber getParameter(std::size_t i) const {
```
```suggestion
        [[nodiscard]] SymbolOrNumber getParameter(const std::size_t i) const {
```

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+            if (symbolicParameter[i].has_value()) return symbolicParameter[i].value();
```
```suggestion
            if (symbolicParameter[i].has_value()) {
              return symbolicParameter[i].value();
            }
```

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+            return parameter[i];
+        }
+
+        void setSymbolicParameter(const Symbolic& par, std::size_t i) {
```
```suggestion
        void setSymbolicParameter(const Symbolic& par, const std::size_t i) {
```

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+        [[nodiscard]] inline bool isStandardOperation() const override {
+            return std::all_of(symbolicParameter.begin(), symbolicParameter.end(), [](const auto& sym) { return !sym.has_value(); });
+        }
+
+        // [[nodiscard]] bool equals(const Operation& op, const Permutation& perm1, const Permutation& perm2) const override {
+        //     return Operation::equals(op, perm1, perm2);
+        // }
+        // [[nodiscard]] bool equals(const Operation& operation) const override {
+        //     return equals(operation, {}, {});
+        // }
```
burgholzer: left-over comments

src/operations/SymbolicOperation.cpp
```diff
@@ -0,0 +1,286 @@
+#include "operations/SymbolicOperation.hpp"
+
+#include "Definitions.hpp"
+#include "operations/StandardOperation.hpp"
+
+#include <variant>
+
+namespace qc {
+
+    void SymbolicOperation::storeSymbolOrNumber(const SymbolOrNumber& param, std::size_t i) {
```
burgholzer, line 10:
```suggestion
    void SymbolicOperation::storeSymbolOrNumber(const SymbolOrNumber& param, const std::size_t i) {
```

src/operations/SymbolicOperation.cpp
```diff
@@ -0,0 +1,286 @@
...
+                type = parseU3(parameter[0], symbolicParameter[1].value(), parameter[2]);
+            else if (!isSymbolicParameter(1) && !isSymbolicParameter(2))
+                type = parseU3(symbolicParameter[0].value(), parameter[1], parameter[2]);
+            else if (!isSymbolicParameter(0))
+                type = parseU3(parameter[0], symbolicParameter[1].value(), symbolicParameter[2].value());
+            else if (!isSymbolicParameter(1))
+                type = parseU3(symbolicParameter[0].value(), parameter[1], symbolicParameter[2].value());
+            else if (!isSymbolicParameter(2))
+                type = parseU3(symbolicParameter[0].value(), symbolicParameter[1].value(), parameter[2]);
+        }
```
burgholzer, line 185: this does look a little messy to be honest. this also holds for all the `parseU...` functions with different signatures.
Wouldn't it be cleaner to have individual `parseU...` functions that all take `const SymbolOrNumber&`s and then correspondingly treat the different cases?
In my imagination that would safe some of the code duplication above.

src/operations/SymbolicOperation.cpp
```diff
@@ -0,0 +1,286 @@
...
+        storeSymbolOrNumber(par1, 1);
+        storeSymbolOrNumber(par2, 2);
+        startQubit = startingQubit;
+        checkUgate();
+        setName();
+    }
+
+    [[nodiscard]] dd::fp SymbolicOperation::getInstantiation(const SymbolOrNumber& symOrNum, const VariableAssignment& assignment) {
+        return std::visit(Overload{
+                                  [&](dd::fp num) { return num; },
```
burgholzer:
```suggestion
                                  [&](const dd::fp num) { return num; },
```

include/operations/SymbolicOperation.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include <variant>
+
+namespace qc {
+    // Overload pattern for std::visit
+    template<typename... Ts>
+    struct Overload: Ts... {
+        using Ts::operator()...;
+    };
+    template<class... Ts>
+    Overload(Ts...) -> Overload<Ts...>;
```
burgholzer, line 20: Didn't know that programming pattern. Nice one 👍🏼 

Just for reference: if we were to adopt C++20 at some point, the second line can actually be dropped, since the compiler deduces these rules automatically.

src/operations/SymbolicOperation.cpp
```diff
@@ -0,0 +1,286 @@
...
+                                  [&](const Symbolic& sym) { return sym.evaluate(assignment); }},
+                          symOrNum);
+    }
+
+    SymbolicOperation::SymbolicOperation::SymbolicOperation(dd::QubitCount nq, dd::Qubit target, OpType g, const SymbolOrNumber& lambda, const SymbolOrNumber& phi, const SymbolOrNumber& theta, dd::Qubit startingQubit) {
```
burgholzer:
```suggestion
    SymbolicOperation::SymbolicOperation(dd::QubitCount nq, dd::Qubit target, OpType g, const SymbolOrNumber& lambda, const SymbolOrNumber& phi, const SymbolOrNumber& theta, dd::Qubit startingQubit) {
```

src/operations/SymbolicOperation.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    // Instantiates this Operation
+    // Afterwards casting to StandardOperation can be done if assignment is total
+    void SymbolicOperation::instantiate(const VariableAssignment& assignment) {
+        parameter[0] = getInstantiation(getParameter(0), assignment);
+        symbolicParameter[0].reset();
+        parameter[1] = getInstantiation(getParameter(1), assignment);
+        symbolicParameter[1].reset();
+        parameter[2] = getInstantiation(getParameter(2), assignment);
+        symbolicParameter[2].reset();
```
burgholzer: this could easily be written in a more generic fashion so that it does not depend on the number of parameters being 3.

src/operations/SymbolicOperation.cpp
```diff
@@ -0,0 +1,286 @@
...
+    }
+    void SymbolicOperation::dumpQiskit([[maybe_unused]] std::ostream& of, [[maybe_unused]] const RegisterNames& qreg, [[maybe_unused]] const RegisterNames& creg, [[maybe_unused]] const char* anc_reg_name) const {
+        throw QFRException("Dumping Qiskit Circuit is not supported for parameterized gates.");
+    }
+
+    StandardOperation SymbolicOperation::getInstantiatedOperation(const VariableAssignment& assignment) const {
+        auto lambda = getInstantiation(getParameter(0), assignment);
+        auto phi    = getInstantiation(getParameter(1), assignment);
+        auto theta  = getInstantiation(getParameter(2), assignment);
+        return StandardOperation(nqubits, targets, type, lambda, phi, theta, startQubit);
```
burgholzer, line 271: while not as easy as the function below, this function could also benefit from a little more generality.
If this turns out to be harder, feel free to leave it as it is.

burgholzer (changes requested):
Many thanks for this contribution. I just went through all the code and you can find my comments inline. Most of them should be fairly easy to address (`const` 😈 ). Maybe a handful probably require a small to medium effort. 
I guess this should be ready to merge after that iteration.

include/QuantumComputation.hpp
pehamTom: I think I do have an explanation for that one. Seems to be a relic from some older clang-format version I had that messed it up. The newer one doesn't touch that comment, so it just stayed.

include/QuantumComputation.hpp
pehamTom: Same as above

include/QuantumComputation.hpp
pehamTom: Good point. I felt that `isSymbolic`might be misleading. What is a symbolic quantum computation? One that has only symbolic parameters? One that has at least one? I figured the `isVariableFree` approach allowed me to circumvent that.

include/QuantumComputation.hpp
pehamTom: I am a big fan of fold expressions :smile: . I didn't want to clutter the hpp file with template magic so I did it the old fashioned way. Also, so far, 3 is the maximum number of variables that need to be included at any given point. But I think doing it with folds is more adaptable for the future. Gonna change that.

mqt/qfr/bindings.cpp
pehamTom: My bad. Removed it again.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Left over from previous version. Removed now.

src/CMakeLists.txt
pehamTom: That probably comes from my internal CMake formatter. Should I revert that? Doesn't change anything really, right? I guess tabs were replaced.

src/QuantumComputation.cpp
pehamTom: Removed it.

src/QuantumComputation.cpp
pehamTom: Good catch. That is gone now.

src/QuantumComputation.cpp
pehamTom: Agree

src/QuantumComputation.cpp
pehamTom: Totally forgot that you can do that now.

src/operations/StandardOperation.cpp
pehamTom: A victim to my twitchy finger when browsing through files :laughing: .

test/CMakeLists.txt
pehamTom: Gone now.

test/unittests/test_symbolic.cpp
pehamTom: Copy paste leftover. Removed.

test/unittests/test_symbolic.cpp
pehamTom: Which it is now.

test/unittests/test_symbolic.cpp
pehamTom: Same as above.

test/unittests/test_symbolic.cpp
pehamTom: Seems to be a leftover from the time when I had a derived QuantumComputation for symbolic computations.

test/unittests/test_zx_functionality.cpp
pehamTom: True. Removed.

src/zx/FunctionalityConstruction.cpp
pehamTom: No one is safe from the const devil.

src/zx/FunctionalityConstruction.cpp
pehamTom: That one is properly consted now.

src/zx/FunctionalityConstruction.cpp
pehamTom: You better belief that that one is consted.

src/zx/FunctionalityConstruction.cpp
pehamTom: C&P from older commit or something. Not entirely sure what happened there.

src/zx/FunctionalityConstruction.cpp
pehamTom: Const'd

src/zx/FunctionalityConstruction.cpp
pehamTom: Running out of funny const replies....

src/zx/FunctionalityConstruction.cpp
pehamTom: See above. Changed.

src/zx/FunctionalityConstruction.cpp
pehamTom: Took care of those unconsted renegades.

src/zx/FunctionalityConstruction.cpp
pehamTom: Maybe just bad habits :musical_note:

src/zx/FunctionalityConstruction.cpp
pehamTom: They really thought they wouldn't get consted. But the const-police will prevail.

src/zx/FunctionalityConstruction.cpp
pehamTom: Actually, these have already been unconsted in previous versions. So we had a whole release with unconsted variables.

src/zx/FunctionalityConstruction.cpp
pehamTom: I really hope I could offer an explanation for that...

src/zx/FunctionalityConstruction.cpp
pehamTom: A truly sneaky const. Forgot that you can cast to const pointer, I think.

src/zx/FunctionalityConstruction.cpp
pehamTom: A younger me would probably just write `int` in this case. But `const std::size_t` just shows you how beautiful C++ can be :upside_down_face: .

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Removed them.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Does SonarLint tell you all that?

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: At this point it seems that const should just be the default.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: A const bool? Now that is a new one.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: That is embarrassing. That is freshmen-level coding.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Yes, especially for iterators. Changed it.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Changed

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: I wasn't aware. But if that is in accordance with our coding guidelines, it shall be changed.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Removed.

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Removed.

include/operations/SymbolicOperation.hpp
pehamTom: You are of course right.

include/operations/SymbolicOperation.hpp
pehamTom: Not really, although I understand where the question comes from. The `parseU3` method is called when constructing a `SymbolicOperation` with the `U3` type. As far as I understand it, the `parseU3` method takes care of figuring out whether a U3 gate is truly a U3 gate and not just a, say, Hadamard gate. If all parameters are symbolic then no reduction could ever be possible.

include/operations/SymbolicOperation.hpp
pehamTom: Haven't had a const in a while. Good refresher.

include/operations/SymbolicOperation.hpp
pehamTom: No longer.

include/operations/SymbolicOperation.hpp
pehamTom: This is purely to signify intent, right? The const doesn't actually do anything here.

include/operations/SymbolicOperation.hpp
pehamTom: Done

include/operations/SymbolicOperation.hpp
pehamTom: I'm thinking about making an emacs shortcout for just typing const.

include/operations/SymbolicOperation.hpp
pehamTom: Done

src/operations/SymbolicOperation.cpp
pehamTom: Reminds me of a famous Haiku:

The steam is running
The autumn leaves are falling
The const doesn't care

src/operations/SymbolicOperation.cpp
pehamTom: But wouldn't that just offset the problem to the next function call? Due to the different nature of symbols and numbers I always have to figure out which combination of symbols and numbers I have right now.

src/operations/SymbolicOperation.cpp
pehamTom: Const'd

include/operations/SymbolicOperation.hpp
pehamTom: Yeah only found out about it while working on this PR. Without it, Variants are just a bit awkward to work with (which they still are).

src/operations/SymbolicOperation.cpp
pehamTom: Didn't know that that even compiles.

include/QuantumComputation.hpp
burgholzer: Excellent, so issues like this should be fixed in the future 👍🏼

src/operations/SymbolicOperation.cpp
pehamTom: Right. Might have been during a late-night coding session.

include/QuantumComputation.hpp
burgholzer: Hm. also a good point. Would `isParameterFree` make it a little better. `Variable` is not really used when we speak about those kind of circuits. If this makes sense, then we should probably also change the other occurrences of `Variable` throughout this file.

src/operations/SymbolicOperation.cpp
pehamTom: That one I can't, right? The StandardOperation is not generic and has a set number of parameters. So it wouldn't make sense to construct it with a generic number of arguments. Correct me if my thinking is wrong on this one.

src/CMakeLists.txt
burgholzer: I guess these changes can be kept as-is. Once we introduce pre-commit to the QFR formatting will be streamlined anyway.

src/zx/FunctionalityConstruction.cpp
burgholzer: At some point, there will be no more `const`s to add. Until then, you better think of some good excuses ;o)

src/zx/FunctionalityConstruction.cpp
burgholzer: 😱 how could that have happened

src/zx/FunctionalityConstruction.cpp
burgholzer: 🙃 it certainly is...

mqt/qfr/qiskit/QuantumCircuit.hpp
burgholzer: it does. highly recommended.

mqt/qfr/qiskit/QuantumCircuit.hpp
burgholzer: A short and nice article on that: https://belaycpp.com/2021/06/30/should-every-variable-be-const-by-default/

mqt/qfr/qiskit/QuantumCircuit.hpp
burgholzer: The reasoning provided by SonarLint can be found here https://rules.sonarsource.com/java/RSPEC-1774 
I tend to agree with their assessment.

include/operations/SymbolicOperation.hpp
burgholzer: Yeah, agreed.

include/operations/SymbolicOperation.hpp
burgholzer: It still makes sure that `i` is not modified in the body of the function.

include/operations/SymbolicOperation.hpp
burgholzer: Just do it ✅

src/operations/SymbolicOperation.cpp
burgholzer: That would be perfect for a Twitter post ;o)

include/QuantumComputation.hpp
pehamTom: Also thought about that. The thing is that the word "Parameter" doesn't necessarily refer to a symbolic or variable parameter. That's why "Variable" in this context made the most sense to me. But if you disagree I will gladly change it.

src/operations/SymbolicOperation.cpp
burgholzer: Technically that is true. I still somehow feel that it is more straight-forward to follow the same structure as in the base class function and essentially break out of checks whenever a parameter is symbolic and, hence, the simplification condition cannot be decided.
In the end, I am fine with the current state. So if you are not convinced by my argumentation above, then feel free to leave this as it is.

include/QuantumComputation.hpp
burgholzer: I mean, in the Qiskit docs of the Parameter class they state
> A parameter is a variable value that is not required to be fixed at circuit definition.

I tend to agree with that definition.
Although I also see that we currently refer to the angles of rotation gates as `params` or `parameters`.
In the end, I don't have a sting opinion on that. On the one hand, consistency with Qiskit would be nice. On the other hand, consistency within our code base would be desirable.

src/operations/SymbolicOperation.cpp
burgholzer: I was surprised about that as well 🙃

src/operations/SymbolicOperation.cpp
burgholzer: The Operation class has a compile-time-set number of parameters. However, that number might not be `3` in the near to mid-term future (there is a global `U` gate in Qiskit that takes `4` parameters).
Due to the constructor taking a fixed number of arguments, it might actually be tough to make this generic. Could fold expressions be used here to expand the list of parameters?

pehamTom: First comment in this thread seems to solve the issue: https://stackoverflow.com/questions/60434033/how-do-i-expand-a-compile-time-stdarray-into-a-parameter-pack . However, that is quite some template boilerplate :smile:

burgholzer: Oh boy. Although I like fancy C++ concepts, that might be too much for such a simple use case.
Probably best to leave it as it is and live with the fact that this potentially has to be adapted in the future.

mqt/qfr/qiskit/QuantumCircuit.hpp
burgholzer: For some reason this is still here in the latest commit

burgholzer (approved):
LGTM. Ready to merge after all conversations have been resolved

mqt/qfr/qiskit/QuantumCircuit.hpp
pehamTom: Well... Now it is gone

PR #226, opened 2022-12-01
✨ Convenience features

burgholzer:
This PR adds some convenience functions that allow easier creation of `QuantumComputation`s.
Specifically, a `QuantumComputation` can now be easily converted to an `Operation`. This allows one to create compound operations by first building a quantum circuit and, then, turning it into an operation.
Furthermore, a convenience function for adding classically-controlled operations to circuits is provided. This avoids the verbose code of having to create the quantum operation, constructing the ClassicControlled operation out of that, and then emplacing the resulting operation in the circuit.

Furthermore, it adds a method to determine the circuit depth.

Most of the changes here are actually in the tests, which have been refactored to take advantage of the new convenience features and, imho, look much nicer and less verbose now.

### DIFF
--- a/include/QuantumComputation.hpp
+++ b/include/QuantumComputation.hpp
@@ -718,6 +718,24 @@ namespace qc {
         virtual void dumpOpenQASM(std::ostream& of);
         virtual void dumpTensorNetwork(std::ostream& of) const;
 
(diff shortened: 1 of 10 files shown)


hillmich (approved):
LGTM! Avoiding the boilerplate for classically controlled operations is really nice.

PR #244, opened 2022-12-23
♻️ DD-independent QFR-Core

burgholzer:
This (long-overdue) PR makes the core features of the QFR independent of the DD package.
Most importantly, this lifts any limitations on the number of qubits that can be handled by the QFR itself and creates a cleaner separation of concerns.
Note that this required touching almost every part of the library, so the number of lines changed within this PR is quite massive.

These changes allow to separate the `dd` and the `zx` functionality into separate CMake targets, so that they are only built if necessary. This allows, e.g., QMAP to only link to the core library part not require the DD or the ZX part at all.
The new library targets are named `qfr_dd` and `qfr_zx` (with corresponding test targets `qfr_test_dd` and `qfr_test_zx`).

In addition to the above, a significant amount of linter warnings has been fixed while going through all the files. This should ease a future integration of `clang-tidy` considerably.

There are a few aspects that are expected to improve considerably on the way towards the envisioned `mqt-core` library:
- The core QFR library unconditionally links to the `ZX` package library as of now since some of the used structs (e.g. `Symbolic`) are only defined there. That dependency should be easy to resolve once both codebases are combined.
- There is a redundant specification of controls in the QFR (`qc::Control`) and the DD package (`dd::Control`) which merely exists because the DD package has no way of knowing about the higher level QFR Control class as of now. This redundancy (and the accompanying translation effort between both types at the interface between the QFR and the DD package) can be eliminated completely once both codebases are combined.
- Tests have been mainly kept as-is. However, it might make sense to re-consider some of them and re-organize the overall structure so that the tests also more closely following the idea of having a `core` library and several data-structure extensions (`dd` and `zx` for now)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
         name: Coverage
         run: |
           cmake -S . -B buildCov -DCMAKE_BUILD_TYPE=Debug -DBUILD_QFR_TESTS=ON -DCOVERAGE=ON
-          cmake --build buildCov --config Debug --target qfr_test
+          cmake --build buildCov --config Debug
           cd buildCov/test
           ctest -C Debug --output-on-failure
       - if: runner.os == 'Linux'
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ It acts as an intermediate representation and provides the facilitites to
 ### System Requirements
 
 Building (and running) is continuously tested under Linux, MacOS, and Windows using the [latest available system versions for GitHub Actions](https://github.com/actions/virtual-environments). However, the implementation should be compatible
-with any current C++ compiler supporting C++17 and a minimum CMake version of 3.14.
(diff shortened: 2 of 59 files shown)


burgholzer:
@hillmich just to make you aware of this change. This will require a slight change in DDSIM as part of the next dependabot update (probably Friday). Instead of linking against `MQT::qfr`, DDSIM has to link against `MQT::qfr_dd`.
I wouldn't expect that many other required changes. Maybe some type conversions. I guess we'll see when the dependabot update hits.

burgholzer:
> @hillmich just to make you aware of this change. This will require a slight change in DDSIM as part of the next dependabot update (probably Friday). Instead of linking against `MQT::qfr`, DDSIM has to link against `MQT::qfr_dd`.
> I wouldn't expect that many other required changes. Maybe some type conversions. I guess we'll see when the dependabot update hits.

@SmaranTum almost the same goes for `SyReC`.

PR #257, opened 2023-01-26
Fixed a wrong property in CMake

hillmich:
- Make nlohmann_json a system include to suppress potential warnings (see https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html for documentation)


burgholzer (approved):
Thanks for catching that mistake 👍🏼

PR #266, opened 2023-02-13
🐛 Fix controlled u-gate OpenQASM output

JoachimMarin:
The qelib1.inc file from OpenQASM contains the functions u3, cu3, u and cu (https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/qasm/libs/qelib1.inc). For this particular case, it is important to explicitly use the 3 parameter version, because cu has 4 parameters, while u only has 3 parameters.
So prepending c to the regular u gate leads to incorrect OpenQASM code, because then one parameter is missing (gamma).

### DIFF
--- a/src/operations/StandardOperation.cpp
+++ b/src/operations/StandardOperation.cpp
@@ -267,16 +267,16 @@ namespace qc {
                 }
                 break;
             case V:
-                op << "u(pi/2, -pi/2, pi/2)";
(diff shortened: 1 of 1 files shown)


burgholzer (approved):
Thanks for catching this 👍🏼 
If I remember correctly, that change came in when Qiskit incorporated some aspects of OpenQASM 3.0 and I think we never adapted for that.

Issue #276, opened 2023-03-02
♻️ Make Version String in QASM Parser Optional


burgholzer:
According to https://openqasm.com/language/comments.html#version-string (emphasis is mine)
> The first non-comment line of an OpenQASM program may _**optionally**_ be OPENQASM M.m ; indicating a major version M and minor version m.

We should adapt our parser so that the version string is no longer mandatory.

Issue #275, opened 2023-03-02
🐛 Minor issues due to global phase difference of `phase` and `rz` gate


burgholzer:
The OpenQASM 2.0 specification lacks a notion of global phase tracking. The OpenQASM 2.0 standard library contains the following definitions:
```qasm
gate u1(lambda) q { U(0,0,lambda) q; }
gate p(lambda) q { U(0,0,lambda) q; }
gate rz(phi) a { u1(phi) a; }
```
This means that, the phase gate `p` and the Z-rotation gate `rz` boil down to the same gate. However, these two differ by a global phase factor, which is why the OpenQASM 3.0 spec defines these gates as
```qasm
gate u1(lambda) q { U(0, 0, lambda) q; }
gate p(lambda) a { ctrl @ gphase(lambda) a; }
gate rz(lambda) a { gphase(-lambda/2); U(0, 0, lambda) a; }
```

Now, a global phase is not a problem per-se and can be ignored. But once controls are added, this phase becomes relative and plays a role, i.e., a `cp` gate is different to a `crz` gate. However, in our QASM parser, we simply strip away leading `c`'s, and look for a definition of the underlying gate. As a consequence, both gates (`cp` and `crz`) end up as `cp` in the resulting quantum circuit.

I believe the underlying problem can be fixed by refactoring our parser a little bit, so that it does not look up definitions of gates we natively support in the QFR. That approach is also fairly close to what Qiskit is doing and should allow to reduce the complexity of some parts of the parser actually. Most likely, this requires adding further tokens to the parser or, alternatively , introducing some kind of dictionary of supported gates that can be used for mapping identifiers to `(Standard)Operation`s or `OpType`s with controls, targets and parameters.

Note: This is only a problem for the QASM import. The Qiskit QuantumCircuit import properly handles the distinction between phase and RZ gates.

PR #280, opened 2023-03-13
✨ Gate Extensions and ♻️ QASM Parser Refactor

burgholzer:
This (rather large) PR builds on cda-tum/dd_package#144 and extends the capabilities of the QFR by **_eight_** more gates:
 - `DCX` (double CNOT)
 - `ECR` (cross-resonance)
 - `RZZ(\theta)` (ZZ rotation gate)
 - `RXX(\theta)` (XX rotation gate)
 - `RYY(\theta)` (YY rotation gate)
 - `RZX(\theta)` (ZX rotation gate)
 - `XXMinusYY(\theta, \beta)` (XX-YY rotation gate)
 - `XXPlusYY(\theta, \beta)` (XX+YY rotation gate)
It also includes initial support for the OpenQASM 3.0 `gphase` global phase gate.

The addition of these new gates made it clear, that our QASM parser needs some major refactors to remain extendable and performant. This PR performs such a refactoring, with the following key take-aways:
 - 🐛 All gates that are natively supported by the QFR are now directly read in instead of recursing down until `U` and `CX`. This fixes the long-outstanding bug outlined in #275.
 - 🚸 Handling of gates and definitions is unified much more than before. The number of tokens has been reduced and lots of duplicate code and unnecessary special case handling has been removed.
 - ⚡ The default `qelib1.inc` has been reduced to a minimum due to most gates being natively supported now. This should speed up parsing of every single QASM file.
 - 🔥 The number of gate parameters is no longer fixed to a maximum of three. Over the course of this refactor, a breaking change was introduced regarding the parameter order in `U3` and `U2` gates, in order to align our definition with the one Qiskit uses. This requires calls to `.u3(...)`, `.u2(...)`, and constructions of the respective gates to be updated correspondingly.
 - 🚸 Implements #276


include/operations/ClassicControlledOperation.hpp
```diff
@@ -21,12 +21,12 @@ namespace qc {
...
-            name         = "c_" + op->getName();
-            parameter[0] = static_cast<fp>(controlRegister.first);
-            parameter[1] = static_cast<fp>(controlRegister.second);
-            parameter[2] = static_cast<fp>(expectedValue);
-            type         = ClassicControlled;
+            nqubits = op->getNqubits();
+            name    = "c_" + op->getName();
+            parameter.emplace_back(static_cast<fp>(controlRegister.first));
+            parameter.emplace_back(static_cast<fp>(controlRegister.second));
+            parameter.emplace_back(static_cast<fp>(expectedValue));
```
hillmich, line 29: Maybe `reserve` a capacity of `3`? I think most implementations start with a capacity of zero and I'm not sure how much memory will be allocated on the first `emplace_back`.

mqt/qfr/qiskit/QuantumCircuit.hpp
```diff
@@ -302,25 +319,20 @@ namespace qc::qiskit {
...
-                phi    = parseParam(params[0]);
-                lambda = parseParam(params[1]);
-            } else if (params.size() == 3) {
-                theta  = parseParam(params[0]);
-                phi    = parseParam(params[1]);
-                lambda = parseParam(params[2]);
+            std::vector<qc::SymbolOrNumber> parameters{};
+            for (std::size_t i = 0; i < params.size(); ++i) { // NOLINT(modernize-loop-convert)
+                parameters.emplace_back(parseParam(params[i]));
             }
```
hillmich: Doesn't this work as `for-each` loop?

src/operations/StandardOperation.cpp
```diff
@@ -347,68 +385,14 @@ namespace qc {
...
-        of << " " << qreg[targets[0]].second << ", " << qreg[targets[1]].second << ";\n";
-
-        for (const auto& c: controls) {
-            if (c.type == Control::Type::Neg) {
-                of << "x " << qreg[c.qubit].second << ";\n";
+        for (std::size_t i = 0; i < targets.size(); ++i) {
+            of << " " << qreg[targets[i]].second;
+            if (i < targets.size() - 1) {
+                of << ",";
             }
```
hillmich: Do we have a guarantee that there will be at least one element? For this concatenation I'd find it cleaner to write the first element before the loop and then prepren `","` before each element in the loop. (Feel free to just resolve this)

src/operations/SymbolicOperation.cpp
```diff
@@ -63,43 +72,48 @@ namespace qc {
...
     }
-    OpType SymbolicOperation::parseU3(const Symbolic& lambda,
-                                      const Symbolic& phi, fp& theta) {
+    OpType SymbolicOperation::parseU3(fp& theta, [[maybe_unused]] const Symbolic& phi,
+                                      [[maybe_unused]] const Symbolic& lambda) {
         if (std::abs(theta - PI_2) < PARAMETER_TOLERANCE) {
             theta = PI_2;
-            return parseU2(lambda, phi);
+            return parseU2(phi, lambda);
         }
```
hillmich: Do `phi` and `lambda` really need `[[maybe_unused]]`?

hillmich (changes requested):
Thank you very much for making the effort. I have a few suggestions/questions :sweat_smile:

include/operations/ClassicControlledOperation.hpp
burgholzer: Good suggestion. Done in https://github.com/cda-tum/qfr/pull/280/commits/e0edd5200e68efd6f6e7a86812089086277e6654.

mqt/qfr/qiskit/QuantumCircuit.hpp
burgholzer: I didn't get it to work at first because range-based for loops over pybind types are weird. 
Thanks to you poking me, I got it to work in https://github.com/cda-tum/qfr/pull/280/commits/4b36b199fe857eb6063b0debba426e71db3480fb.

src/operations/StandardOperation.cpp
burgholzer: Unfortunately, we don't have such a guarantee (the global phase gate has no targets). However, I think I found a neat solution in https://github.com/cda-tum/qfr/pull/280/commits/be655096613d6138c6b17a0dec430c71f7cae26d.

src/operations/SymbolicOperation.cpp
burgholzer: You are right. Removed in https://github.com/cda-tum/qfr/pull/280/commits/c2077d9819a283d7376abf10a6511db3c0886ab6.

hillmich (approved):
LGTM

PR #283, opened 2023-03-16
⬆️ Update DD package submodule

aaronleesander:
This PR just updates the DD package submodule with a fix in the `expectationValue` function of the DD package.


burgholzer:
That might be the wrong DD package branch. You probably want to switch to the `main` branch locally, update that by pulling the latest changes and then committing the result.

burgholzer (approved):
Now we are talking ✅

PR #286, opened 2023-03-22
Add opaque teleportation gate

hillmich:
If teleportation was used in mapping, add an opaque gate to the output.

### DIFF
--- a/src/QuantumComputation.cpp
+++ b/src/QuantumComputation.cpp
@@ -632,7 +632,7 @@ namespace qc {
 
         of << "OPENQASM 2.0;" << std::endl;
         of << "include \"qelib1.inc\";" << std::endl;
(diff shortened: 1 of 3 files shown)


src/QuantumComputation.cpp
```diff
@@ -632,6 +632,9 @@ namespace qc {
 
         of << "OPENQASM 2.0;" << std::endl;
         of << "include \"qelib1.inc\";" << std::endl;
+        if (std::any_of(std::begin(ops), std::end(ops), [](const auto& op) { return op->getType() == OpType::Teleportation })) {
```
burgholzer:
```suggestion
        if (std::any_of(std::begin(ops), std::end(ops), [](const auto& op) { return op->getType() == OpType::Teleportation; })) {
```

hillmich:
I fixed the error and added a test :+1:

burgholzer (approved):
LGTM. Many thanks 🙏

Issue #33, opened 2021-12-16
Support for OpenQASM 3.0 looping and branching


burgholzer:
The OpenQASM 3.0 standard provides the means to describe the flow of a program using `if` statements, `for` and `while` loops.
This allows to more easily write static quantum algorithms that consist of iterations (such as Grover's algorithm or Phase Estimation) but also allows for the construction of sequential quantum circuits (such as Repeat Until Success schemes).

As a starting point, the creation of static algorithms should be supported using these constructs, i.e., programs should be supported that only contain constants and hence make it possible to unroll all loops at compile-time. An example of such a circuit is shown in https://arxiv.org/pdf/2104.14722.pdf
```QASM
/*
 * Iterative phase estimation
 */
OPENQASM 3;
include "stdgates.inc";

const n = 3; // number of iterations 
const θ = 3*π/8; //phase angle on target qubit

qubit q; // phase estimation qubit
qubit r; // target qubit for the controlled-unitary gate 
angle[n] c = 0; // phase estimation bits

// initialize
reset q;
reset r;

// prepare uniform superposition of eigenvectors of phase
h r;

// iterative phase estimation loop
for i in [1:n] { // implicitly cast val to int
  reset q;
  h q;
  ctrl @ pow(2**i) @ phase(θ) q, r;
  inv @ phase(c) q;
  h q;
  measure q -> c[0];
  // newest measurement outcome is associated to a π/2 phase shift
  // in the next iteration, so shift all bits of c left 
  c <<= 1;
}
// Now c contains the n-bit estimate of φ in the
// eigenvalue e^{i*φ} and qreg r is projected to an // approximate eigenstate of the phase gate.
```

In theory, this could be implemented independently from the other OpenQASM 3.0 features, but it probably makes sense to add support for constant values first (#31).

Issue #32, opened 2021-12-16
Support for casting of OpenQASM 3.0 types


burgholzer:
The new OpenQASM standard introduces many additional datatypes besides quantum and classical registers. This allows to describe classical components of quantum algorithms in a very intuitive way. As a first step, support for all kinds of new types should be added to the QFR library and, correspondingly, to the QuantumComputation class. For details on all the new typed provided by the standard, see https://qiskit.github.io/openqasm/language/classical.html.

The standard provides the means for converting between the available types. For details, see https://qiskit.github.io/openqasm/language/types.html#casting-specifics.

This depends on #30 for the availability of all the types.

Issue #30, opened 2021-12-16
Support for OpenQASM 3.0 additional datatypes


burgholzer:
The new OpenQASM standard introduces many additional datatypes besides quantum and classical registers. This allows to describe classical components of quantum algorithms in a very intuitive way. As a first step, support for all kinds of new types should be added to the QFR library and, correspondingly, to the QuantumComputation class. For details on all the new typed provided by the standard, see https://qiskit.github.io/openqasm/language/classical.html.

Similar to #29, the QFR should support Integers, Floating point numbers, Fixed-point angles, complex numbers, and boolean types.
The `QuantumComputation` class should be capable of allocating and maintaining these types. The `ClassicalOperation` class introduced in #29, will most likely need to be expanded to incorporate all the possible operations on these classical types.

Depends on #29.

Issue #27, opened 2021-12-16
Support for the powering modifier of OpenQASM 3.0


burgholzer:
The new OpenQASM standard introduces gate modifiers in order to more efficiently describe quantum circuits. Any modifier `mod` can be applied to a gate `g` via `mod @ g`. For more details, see Section 4.2 in https://arxiv.org/pdf/2104.14722.pdf or the [Live Specification](https://qiskit.github.io/openqasm/language/gates.html#quantum-gate-modifiers).

The powering modifier `pow(r)` can be used to represent the r-th power of any gate `g` via `pow(r) @ g`, where `r` can either be an integer or a floating point number.

The case where `r` is a floating point number needs to be handled by computing the principal logarithm of the unitary that represents `g`. This might be tricky in general.

The case where `r` is an integer can be naively solved by repeatedly applying `g`. If `r <0`, `pow(r) @ g` can be translated to `inv @ pow(-r) @ g`. Several simplification might be possible here, e.g., the powers of phase gates can easily be computed by appropriately scaling the respective parameter. In its simplest form `pow(2) @ s = z`.

 At the very least, support for the case where `r` is an integer shall be provided which provides much greater flexibility for describing circuits in a standardized way.

lavanya-m-k:
@burgholzer Could you please assign this to me. The PR is https://github.com/munich-quantum-toolkit/core/pull/991

burgholzer:
> [@burgholzer](https://github.com/burgholzer) Could you please assign this to me. The PR is [#991](https://github.com/munich-quantum-toolkit/core/pull/991)

I'll assign you once the Pull Request has been merged.

PR #309, opened 2023-05-17
✨ OpenQASM 3.0 support

martin-fink:
This PR replaces the existing OpenQASM 2.0 parser with a new OpenQASM 3.0 parser.

The new parser now builds a syntax tree, where type checking, constant evaluation, and translation to the Quantum circuit.

The parser can handle the following new features:

### New Syntax

New syntax for declaring bits, qubit, measure operations. The old syntax (`creg`, `qreg`) is still supported.

```qasm
qubit[8] q;
bit[8] c;

c[0] = measure q[0];
measure q[1] -> c[1];

if (c[0] == 1) {
  x q[0];
}
```

### Gate modifiers

Gate modifiers (`inv`, `ctrl`, and `negctrl`) are now supported. This replaces the `c` prefix.
See the OpenQASM 3.0 specification for more information: https://openqasm.com/language/gates.html#quantum-gate-modifiers

```qasm
ctrl @ x q[0], q[1]; // Equivalent to cx q
ctrl(2) @ x q[0], q[1], q[2]; // Equivalent to ccx q;
```

### Classical constant values

The parser now supports classical computation with constant values. This can be used to e.g. define the number of quantum registers.

```qasm
const uint N = 4;
qubit[N * 2];
x qubit[N * 2 - 1];
```

Additionally, all features of the previous parser are still supported.

The big features from OpenQASM 3.0 still missing are:

- classical computational features such as loops, functions, etc. (see #33)
- types such as bools, floats, angles, complex types, etc. (see #30, #32)
- `pow` modifier (#27)


src/QuantumComputation.cpp
```diff
@@ -106,6 +106,10 @@ namespace qc {
                 updateMaxControls(2);
                 importOpenQASM(is);
                 break;
+            case Format::OpenQASM3:
+                updateMaxControls(2);
```
burgholzer: That statement can be safely removed and only exists for legacy reasons in the OpenQASM 2.0 import.

src/parsers/qasm3_parser/cmake/FindANTLR.cmake
burgholzer, line 1: These files in the `cmake` directory can be moved to a top-level `cmake` folder at the project root. That helps to keep all the CMake stuff separated from the sources.

README.md
```diff
@@ -24,7 +24,7 @@ The MQT Core library forms the backbone of the quantum software tools developed
 - [MQT DDVis](https://github.com/cda-tum/mqt-ddvis): A Web-Application visualizing Decision Diagrams for Quantum Computing.
 - [MQT SyReC](https://github.com/cda-tum/mqt-syrec): A Tool for Synthesis of Reversible Circuits/Quantum Computing Oracles.
 
-For a full list of tools and libraries, please visit the [MQT website](https://mqt.readthedocs.io/).
+For a full list of tools and libraries, please run the [MQT website](https://mqt.readthedocs.io/).
```
burgholzer:
```suggestion
For a full list of tools and libraries, please visit the [MQT website](https://mqt.readthedocs.io/).
```

docs/index.md
```diff
@@ -9,7 +9,7 @@ The MQT Core library forms the backbone of the quantum software tools developed
 - [MQT DDVis](https://github.com/cda-tum/mqt-ddvis): A Web-Application visualizing Decision Diagrams for Quantum Computing.
 - {doc}`MQT SyReC <syrec:index>`: A Tool for Synthesis of Reversible Circuits/Quantum Computing Oracles.
 
-For a full list of tools and libraries, please visit the {doc}`MQT website <mqt:index>`.
+For a full list of tools and libraries, please run the {doc}`MQT website <mqt:index>`.
```
burgholzer:
```suggestion
For a full list of tools and libraries, please visit the {doc}`MQT website <mqt:index>`.
```

include/QuantumComputation.hpp
```diff
@@ -746,7 +752,7 @@ class QuantumComputation {
     dump(std::move(of), format);
   }
   virtual void dump(std::ostream&& of, Format format);
-  virtual void dumpOpenQASM(std::ostream& of);
+  virtual void dumpOpenQASM(std::ostream& of, bool openQASM3 = false);
```
burgholzer:
```suggestion
  virtual void dumpOpenQASM(std::ostream& of, bool openQASM3);
```
clang-tidy complains here because default arguments on virtual methods are prohibited.
How bad (mostly in the sense of code duplication) would it be to just define a `dumpOpenQASM3` method? (if both are the same than one might just call the other after all)

include/operations/ClassicControlledOperation.hpp
```diff
@@ -124,10 +138,23 @@ class ClassicControlledOperation final : public Operation {
...
     of << "if(";
     of << creg[controlRegister.first].first;
-    of << " == " << expectedValue << ") ";
+    of << " " << comparisonKind << " " << expectedValue << ") ";
     op->dumpOpenQASM(of, qreg, creg);
   }
 
+  void dumpOpenQASM3(std::ostream& of, const RegisterNames& qreg,
+                     const RegisterNames& creg,
+                     uint32_t indent = 0) const override {
```
burgholzer:
```suggestion
                     uint32_t indent) const override {
```
no default arguments allowed unfortunately. Why's that indent necessary anyways? Was this just poorly handled in the OQ2 export?
Same warning holds for all other overrides of this method.

include/parsers/qasm3_parser/Parser.hpp
```diff
@@ -0,0 +1,194 @@
...
+                            "'";
+      if (context.has_value()) {
+        message += " " + context.value();
+      }
+      error(current(), message);
+    }
+
+    Token const token = current();
+    scan();
+    return token;
```
burgholzer:
```suggestion
    auto token = current();
    scan();
    return token;
```
Should resolve the clang-tidy warning

include/parsers/qasm3_parser/InstVisitor.hpp
```diff
@@ -0,0 +1,146 @@
+#pragma once
+
+#include <any>
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <utility>
```
burgholzer:
```suggestion
#include <any>
#include <iostream>
#include <map>
#include <memory>
#include <sstream>
#include <utility>
```
You are missing that include in this file

include/operations/SymbolicOperation.hpp
```diff
@@ -137,6 +137,10 @@ class SymbolicOperation final : public StandardOperation {
   [[noreturn]] void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
                                  const RegisterNames& creg) const override;
 
+  [[noreturn]] void dumpOpenQASM3(std::ostream& of, const RegisterNames& qreg,
+                                  const RegisterNames& creg,
+                                  uint32_t indent = 0) const override;
```
```suggestion
                                  uint32_t indent) const override;
```

include/operations/StandardOperation.hpp
```diff
@@ -104,7 +104,9 @@ class StandardOperation : public Operation {
 
   void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
                     const RegisterNames& creg) const override;
-
+  void dumpOpenQASM3(std::ostream& of, const RegisterNames& qreg,
+                     const RegisterNames& creg,
+                     uint32_t indent = 0) const override;
```
burgholzer:
```suggestion
                     uint32_t indent) const override;
```

include/operations/Operation.hpp
```diff
@@ -179,6 +179,10 @@ class Operation {
   virtual void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
                             const RegisterNames& creg) const = 0;
 
+  virtual void dumpOpenQASM3(std::ostream& of, const RegisterNames& qreg,
+                             const RegisterNames& creg,
+                             uint32_t indent = 0) const = 0;
```
burgholzer:
```suggestion
                             uint32_t indent) const = 0;
```

include/operations/CompoundOperation.hpp
```diff
@@ -156,6 +156,14 @@ class CompoundOperation final : public Operation {
     }
   }
 
+  void dumpOpenQASM3(std::ostream& of, const RegisterNames& qreg,
+                     const RegisterNames& creg,
+                     uint32_t indent = 0) const override {
```
burgholzer:
```suggestion
                     uint32_t indent) const override {
```

include/operations/ClassicControlledOperation.hpp
```diff
@@ -124,10 +138,23 @@ class ClassicControlledOperation final : public Operation {
...
+    for (uint32_t i = 0; i < indent; ++i) {
+      of << "  ";
+    }
```
burgholzer:
```suggestion
    of << std::string(indent, ' ');
```

include/parsers/qasm3_parser/NestedEnvironment.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <map>
+#include <string>
+#include <vector>
```
burgholzer:
```suggestion
#include <map>
#include <optional>
#include <string>
#include <vector>
```

include/parsers/qasm3_parser/Scanner.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <stack>
+
+namespace qasm3 {
+class Scanner {
+private:
+  std::istream& is;
```
burgholzer: Maybe prefer to make this either a plain `std::istream*` or a `std::shared_ptr`. Reference members prevent copy construction etc.

include/parsers/qasm3_parser/Statement.hpp
```diff
@@ -0,0 +1,435 @@
...
+
+  std::string getName() override {
+    return std::string{"IdentifierExpr ("} + identifier + ")";
+  }
+};
+
+class IdentifierList : public Expression,
+                       public std::enable_shared_from_this<IdentifierList> {
+public:
+  std::vector<std::shared_ptr<IdentifierExpression>> identifiers;
```
burgholzer:
```suggestion
  std::vector<std::shared_ptr<IdentifierExpression>> identifiers{};
```
fairly sure that should fix the clang-tidy warning.

include/parsers/qasm3_parser/Statement.hpp
```diff
@@ -0,0 +1,435 @@
+#pragma once
+
+#include "InstVisitor.hpp"
+#include "QuantumComputation.hpp"
+#include "Types.hpp"
+
+#include <any>
+#include <string>
+#include <utility>
+#include <vector>
```
burgholzer:
```suggestion
#include <any>
#include <cassert>
#include <string>
#include <utility>
#include <vector>
```

include/parsers/qasm3_parser/StdGates.hpp
```diff
@@ -0,0 +1,108 @@
...
+    "// controlled-phase\n"
+    "gate cphase(lambda) a, b { ctrl @ phase(lambda) a, b; }\n"
+    "// identity or idle gate\n"
+    "gate id a { U(0, 0, 0) a; }\n"
+    "// IBM Quantum experience gates\n"
+    "gate u1(lambda) q { U(0, 0, lambda) q; }\n"
+    "gate u2(phi, lambda) q { gphase(-(phi+lambda)/2); U(pi/2, phi, lambda) q; "
+    "}\n"
+    "gate u3(theta, phi, lambda) q { gphase(-(phi+lambda)/2); U(theta, phi, "
+    "lambda) q; }";
```
burgholzer: If you have a look at how this is handled in the OQ2 parser you should find that it only imports the definitions for gates that cannot be natively handled by the mqt-core library anyways. Could we adopt something similar here? That would also eliminate the commented-out code that cannot be handled at the moment.

Edit: now I see that something like that is already happening right below. I would have guessed that it is way more efficient to go the route of the `STANDARD_GATES` then to rely on the parsed definitions.

include/parsers/qasm3_parser/passes/ConstEvalPass.hpp
```diff
@@ -0,0 +1,104 @@
+#pragma once
+
+#include "../NestedEnvironment.hpp"
```
burgholzer: Please do not use relative imports! Holds for any file.

src/QuantumComputation.cpp
```diff
@@ -632,7 +635,7 @@ void QuantumComputation::dump(const std::string& filename) {
   }
 }
 
-void QuantumComputation::dumpOpenQASM(std::ostream& of) {
+void QuantumComputation::dumpOpenQASM(std::ostream& of, bool openQASM3) {
```
burgholzer, line 588: It feels to me as if it would be better to split this into two methods and extract the common functionality (such as consolidating the registers) into another separate function.

src/operations/NonUnitaryOperation.cpp
```diff
@@ -185,4 +185,32 @@ void NonUnitaryOperation::addDepthContribution(
...
+  auto classicsIt = classics.cbegin();
+  for (const auto& q : targets) {
+    if (type == Measure) {
+      of << creg[*classicsIt].second << " = ";
+      ++classicsIt;
+    }
+    of << toString(type) << " " << qreg[q].second;
+    of << ";\n";
+  }
+}
```
burgholzer: This seems to share quite some code with the OQ2 method. Any potential to unify?

src/operations/StandardOperation.cpp
```diff
@@ -558,4 +558,207 @@ void StandardOperation::invert() {
                        " is not supported.");
   }
 }
+void StandardOperation::dumpOpenQASM3(std::ostream& of,
```
burgholzer: Same as for the non-unitary operation goes here: any chance this can be consolidated with the OQ2 method? For example, by extracting common parts?

src/operations/SymbolicOperation.cpp
```diff
@@ -312,6 +312,14 @@ bool SymbolicOperation::equals(const Operation& op, const Permutation& perm1,
...
 }
 
+[[noreturn]] void
+SymbolicOperation::dumpOpenQASM3([[maybe_unused]] std::ostream& of,
+                                 [[maybe_unused]] const qc::RegisterNames& qreg,
+                                 [[maybe_unused]] const qc::RegisterNames& creg,
+                                 [[maybe_unused]] uint32_t indent) const {
+  throw QFRException("OpenQASM 2.0 doesn't support parametrized gates!");
+}
+
```
burgholzer: First of all, the error message is kind of wrong here.
Second, OQ3 does support symbolic gates. However, I guess supporting these might be a little tough at the moment (need to add parameters of the circuit to the overall OQ3 program, etc.)
Depending on how you feel about it, either just change the error message here to something like "... not yet supported ... " or try to implement the parameters.

src/parsers/qasm3_parser/Scanner.cpp
```diff
@@ -0,0 +1,578 @@
...
+  keywords["barrier"] = Token::Kind::Barrier;
+  keywords["true"] = Token::Kind::True;
+  keywords["false"] = Token::Kind::False;
+  keywords["im"] = Token::Kind::Imag;
+  //        keywords["dt"]         = Token::Kind::TimeUnitDt;
+  //        keywords["ns"]         = Token::Kind::TimeUnitNs;
+  //        keywords["us"]         = Token::Kind::TimeUnitUs;
+  //        keywords["mys"]        = Token::Kind::TimeUnitMys;
+  //        keywords["ms"]         = Token::Kind::TimeUnitMs;
+  //        keywords["s"]          = Token::Kind::TimeUnitS;
```
burgholzer: why are these not enabled here?

test/unittests/test_qfr_functionality.cpp
burgholzer, line 1: Just quick comment: I really like the tests here. Keep them coming! These are essential to make sure everything is working properly.
I believe it would be great though to move them to a separate `test_qasm3_io.cpp`. (maybe a good chance to also introduce a `test_qasm2_io.cpp` just to create a little separation of concerns.

burgholzer (changes requested):
Hey 👋🏼 

I know you are still working on this, but I just had some time to kill at the airport and thought I'd give you some feedback. You should be able to use GitHub's batch commit feature (go to files changed and then add suggestions to commit batch) to resolve most of the suggestions/comments. Hope all of that makes sense.

src/parsers/qasm3_parser/Scanner.cpp
martin-fink: The `s` keyword was a problem as it collided with the `s` gate, that's why I decided to disable all of these for the moment. I will re-enable them and handle this in the parser.

include/QuantumComputation.hpp
```diff
@@ -747,7 +752,10 @@ class QuantumComputation {
     dump(std::move(of), format);
   }
   virtual void dump(std::ostream&& of, Format format);
-  virtual void dumpOpenQASM(std::ostream& of);
+  // defaults to OpenQASM 2.0
+  void dumpOpenQASM(std::ostream& of) { dumpOpenQASM(of, false); }
+  void dumpOpenQASM3(std::ostream& of) { dumpOpenQASM(of, true); }
+  virtual void dumpOpenQASM(std::ostream& of, bool openQasm3);
```
burgholzer: I would be fine with making `OpenQASM 3.0` the default. Modern Qiskit versions are equipped with a sufficiently capable parser and our own tools rely on the functionality provided here. In that sense OQ2 can be considered legacy.

I also have a slight tendency to rename the first method from `dumpOpenQASM` to `dumpOpenQASM2` to be more precise.

include/operations/ClassicControlledOperation.hpp
```diff
@@ -121,11 +135,19 @@ class ClassicControlledOperation final : public Operation {
   }
 
   void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
-                    const RegisterNames& creg) const override {
-    of << "if(";
+                    const RegisterNames& creg, size_t indent,
+                    bool openQASM3) const override {
+    of << std::string(indent * 2, ' ');
```
burgholzer: I would assume the `2` here is just the constant being used for the number of spaces per indent.
Might be worth to put that as an explicit constant into the `Definitions.hpp` file to make sure it is consistent (and adjustable) everywhere.

include/operations/CompoundOperation.hpp
```diff
@@ -103,6 +103,8 @@ class CompoundOperation final : public Operation {
     return controls.erase(it);
   }
 
+  [[nodiscard]] inline bool isControlled() const override { return false; }
+
```
burgholzer: I think we discussed this at some point (when adding the control feature):
if a compound operation contains controls (i.e., its own `controls` member is non-empty), then the operation is controlled.
Back then, we made sure that, if a control is added to a compound gate using `addControl`, this is properly reflected in its `controls`.
I would be interested to know if there is any particular reason for making this straight up return false.

include/operations/Operation.hpp
```diff
@@ -176,8 +176,13 @@ class Operation {
     return op.print(os);
   }
 
+  void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
```
burgholzer: Just in order to make sure the line of thought is not lost: I would also argue that this should probably become `dumpOpenQASM2` (similar ro the proposed change in the QuantumComputation class)

src/operations/StandardOperation.cpp
```diff
@@ -239,22 +239,54 @@ StandardOperation::StandardOperation(const std::size_t nq, const Controls& c,
 /***
  * Public Methods
  ***/
-void StandardOperation::dumpOpenQASM(
-    std::ostream& of, const RegisterNames& qreg,
-    [[maybe_unused]] const RegisterNames& creg) const {
+void StandardOperation::dumpOpenQASM(std::ostream& of,
```
burgholzer, line 242: I would assume that would be beneficial for readability to split this method up into an OQ2 and an OQ3 one.
It would also reduce the length of the function itself (which would please CodeQL).
I don't believe that it would introduce too much code duplication just from an inspection of the diff

src/operations/StandardOperation.cpp
```diff
@@ -308,17 +340,37 @@ void StandardOperation::dumpOpenQASM(
     }
     break;
   case V:
-    op << "u3(pi/2,-pi/2,pi/2)";
+    if (openQASM3) {
+      op << "U";
+    } else {
+      op << "u3";
```
burgholzer: I am not 100% sure, but I believe the OQ2 standard already defines the `U` gate. So the distinction here shouldn't really be necessary. That should simplify some of the conditions below.

src/zx/FunctionalityConstruction.cpp
```diff
@@ -198,6 +198,10 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
   if (op->getType() == qc::OpType::Barrier) {
     return it + 1;
   }
+  // global phase is ignored
+  if (op->getType() == qc::OpType::GPhase && !op->isControlled()) {
+    return it + 1;
+  }
```
burgholzer: the ZX package actually supports global phases. You can simply call
```
diag.addGlobalPhase(...);
```
and that should make it work. You can find examples for that throughout this file.

src/zx/FunctionalityConstruction.cpp
```diff
@@ -474,6 +485,9 @@ bool FunctionalityConstruction::transformableToZX(const qc::Operation* op) {
   if (op->getType() == qc::OpType::Barrier) {
     return true;
   }
+  if (op->getType() == qc::OpType::GPhase && !op->isControlled()) {
+    return true;
+  }
```
burgholzer, line 491: This is really just a technical side comment and nothing that has to necessarily be realized in this PR:
A controlled global phase is also naturally supported as such a gate merely becomes a `Phase` gate on the respective controls. So a `global_phase(theta)` controlled on qubit `q[0]` becomes a `p(theta) q[0]`. a global phase controlled on two qubits `q[0], q[1]` becomes a `cp(theta) q[0], q[1]` gate.

test/circuits/test.qasm
```diff
@@ -38,22 +38,22 @@ u3(1,2,3) r[0];
 u2(1,2) r[1];
 u1(1) q[0];
 barrier q;
-show_probabilities;
+//show_probabilities;
```
burgholzer:
```suggestion
```
might as well

test/circuits/test.qasm
```diff
@@ -38,22 +38,22 @@ u3(1,2,3) r[0];
...
 measure q -> c;
 measure r[0] -> d[0];
 measure q -> d;
 reset q;
-snapshot(1) r[0],r[1];
+//snapshot(1) r[0],r[1];
```
burgholzer:
```suggestion
```
might as well

burgholzer (changes requested):
This is really looking amazing already! Kudos to that 🎉 
I just found some time to go over the integration parts (everything except the parser itself) and I only stumbled upon very few comments that came to my mind.
These should be really straightforward to address.

Once they are, I think the only thing left is for me to go through the actual parser code at least once. However, given the extensive tests you wrote (many thanks for that), I am quite certain that most of the code "just works".
Let me know if you believe that something still needs my attention!

include/operations/CompoundOperation.hpp
martin-fink: The reason for that is that I apparently forgot this -- I'll fix it. Thanks for noticing!

include/operations/CompoundOperation.hpp
```diff
@@ -103,6 +103,10 @@ class CompoundOperation final : public Operation {
     return controls.erase(it);
   }
 
+  [[nodiscard]] inline bool isControlled() const override {
+    return !controls.empty();
+  }
+
```
burgholzer:
```suggestion
```
That's the default implementation in the `Operation` class. So it can be removed here.

burgholzer (commented):
Just looked over your latest set of changes again. Those look great! Many thanks.
Is there anything left from your end that you would like to get into this PR before it gets in?

I would appreciate if you could comment on the question above on which of the overall issues that have been opened in mqt-core are resolved by this PR.

martin-fink:
From my side I'm happy -- I will address the CodeQL warnings though.
Regarding your questions which issues will be closed

- #29 yes
- #31 yes
- #30 partly there (type checker, constant eval is there for float, bool; not for angles, time)
- #32 no
- #33 no
- #27 no

burgholzer:
> From my side I'm happy -- I will address the CodeQL warnings though. Regarding your questions which issues will be closed
> 
> * [Support for OpenQASM 3.0 classical bits and registers #29](https://github.com/cda-tum/mqt-core/issues/29) yes
> * [Support for OpenQASM 3.0 constant values #31](https://github.com/cda-tum/mqt-core/issues/31) yes
> * [Support for OpenQASM 3.0 additional datatypes #30](https://github.com/cda-tum/mqt-core/issues/30) partly there (type checker, constant eval is there for float, bool; not for angles, time)
> * [Support for casting of OpenQASM 3.0 types #32](https://github.com/cda-tum/mqt-core/issues/32) no
> * [Support for OpenQASM 3.0 looping and branching #33](https://github.com/cda-tum/mqt-core/issues/33) no
> * [Support for the powering modifier of OpenQASM 3.0 #27](https://github.com/cda-tum/mqt-core/issues/27) no

Great, many thanks. Updated the list of linked issues accordingly.
I also took the liberty of adjusting the title from WIP to a proper title and update with the latest changes from `main`.
As a last step before merging: Could you please update the PR description so that it is a little more complete and elaborative on what is actually available now (and maybe also that the OQ2 parser is completely replaced by the new parser).
Afterwards, you can request a (final) review and I should be able to merge 🎉

martin-fink:
I've updated the description. Thanks!

src/parsers/QASM3Parser.cpp
```diff
@@ -0,0 +1,807 @@
...
+
+    // here we count the number of controls
+    std::vector<std::pair<std::shared_ptr<GateOperand>, bool>> controls{};
+    // since standard gates may define a number of control targets, we first
+    // need to handle those
+    size_t nControls{gate->getNControls()};
+    if (targets.size() < nControls) {
+      error("Gate '" + identifier + "' takes " + std::to_string(nControls) +
+                " controls, but only " + std::to_string(targets.size()) +
+                " qubit swere supplied.",
```
burgholzer:
```suggestion
                " qubits were supplied.",
```

src/parsers/QASM3Parser.cpp
```diff
@@ -0,0 +1,807 @@
...
+  [[nodiscard]] std::unique_ptr<qc::Operation>
+  getBarrierOp(const std::shared_ptr<BarrierStatement>& barrierStatement,
+               const qc::QuantumRegisterMap& qregs) const {
+    std::vector<qc::Qubit> qubits{};
+    for (const auto& gate : barrierStatement->gates) {
+      translateGateOperand(gate, qubits, qregs, barrierStatement->debugInfo);
+    }
+
+    return std::make_unique<qc::NonUnitaryOperation>(qc->getNqubits(), qubits,
+                                                     qc::Barrier);
```
burgholzer: `Barrier` is no longer a `NonUnitaryOperation` but has been promoted to a `StandardOperation` a while back.
Should be an easy fix though.
```suggestion
    return std::make_unique<qc::StandardOperation>(qc->getNqubits(), qubits,
                                                     qc::Barrier);
```

include/parsers/qasm3_parser/StdGates.hpp
burgholzer, line 1: Just something I was wondering about a little bit:
- Most of the gates defined in `STDGATES` (essentially everything but `cu`) is natively supported as a `StandardGate` definition. Couldn't we just add the gates to the `STANDARD_GATES` map down below right away and safe on parsing and applying the definitions repeatedly?
- Follow up question: In the existing OQ2 parser, we were able to parse any gate starting with a bunch of `c`'s and ending in a known standard gate without a definition for the gate. Does this still apply for the new parser? If so, couldn't the definitions of these controlled versions be skipped altogether?
- Last, but not least, do you see a way that we could support the `QASM2_COMPAT_GATES` by default (without being given a definition)? I saw that there is an error in place if a certain gate is defined multiple times. Couldn't we just circumvent that error for any built-in gates and just "ignore" the duplicate definition for these cases?

burgholzer (approved):
I now also briefly looked over the parser code itself and it is looking good ✨ 

I just left one last comment where I would like to get your opinion on a couple of things. Those things need not necessarily happen as part of this PR if they turn out to be to complicated.

burgholzer:
And I seem to have screwed up one of the tests with one of the changes above. Could you maybe fix that? 😅

include/parsers/qasm3_parser/StdGates.hpp
martin-fink: > Most of the gates defined in STDGATES (essentially everything but cu) is natively supported as a StandardGate definition. Couldn't we just add the gates to the STANDARD_GATES map down below right away and safe on parsing and applying the definitions repeatedly?

Yes, that should be possible. I'll follow up on that one.

> Follow up question: In the existing OQ2 parser, we were able to parse any gate starting with a bunch of c's and ending in a known standard gate without a definition for the gate. Does this still apply for the new parser? If so, couldn't the definitions of these controlled versions be skipped altogether?

No, this is no longer supported. Should I add added as a fallback if importing OpenQASM 2.0? I.e. if the parser sees a `OpenQASM 2.0` declaration, these are parsed as `ctrl(n) @ gate`?

> Last, but not least, do you see a way that we could support the QASM2_COMPAT_GATES by default (without being given a definition)? I saw that there is an error in place if a certain gate is defined multiple times. Couldn't we just circumvent that error for any built-in gates and just "ignore" the duplicate definition for these cases?

Should be possible as well. Should all declarations of builtin gates be ignored? Or just OpenQASM 2.0 gates?

burgholzer: > > Most of the gates defined in STDGATES (essentially everything but cu) is natively supported as a StandardGate definition. Couldn't we just add the gates to the STANDARD_GATES map down below right away and safe on parsing and applying the definitions repeatedly?
> 
> Yes, that should be possible. I'll follow up on that one.

That would be really cool.

> > Follow up question: In the existing OQ2 parser, we were able to parse any gate starting with a bunch of c's and ending in a known standard gate without a definition for the gate. Does this still apply for the new parser? If so, couldn't the definitions of these controlled versions be skipped altogether?
> 
> No, this is no longer supported. Should I add added as a fallback if importing OpenQASM 2.0? I.e. if the parser sees a `OpenQASM 2.0` declaration, these are parsed as `ctrl(n) @ gate`?

Such a fallback would be nice. Otherwise, code that works now will no longer work after this PR is merged. I think the key thing is to avoid the duplicate definition error if a user defines such a controlled gate (linked to the point below).
What was nice about the above, was that we natively supported any controlled version of the standard gates (like `x`, `y`, `h`, etc.) by just saying that the standard gate is native and no further configuration.
I think we might be able to emulate that behavior somehow also for the OQ3 parser, by adding these additional definitions (like `cx`, `cz`, etc.) to the `STANDARD_GATES`. What do you think?

> > Last, but not least, do you see a way that we could support the QASM2_COMPAT_GATES by default (without being given a definition)? I saw that there is an error in place if a certain gate is defined multiple times. Couldn't we just circumvent that error for any built-in gates and just "ignore" the duplicate definition for these cases?
> 
> Should be possible as well. Should all declarations of builtin gates be ignored? Or just OpenQASM 2.0 gates?

In the existing implementation (of the OQ2 parser) all definitions for builtin gates and versions that prefix a builtin gate with arbitrarily many `c`'s were simply skipped over. That was pretty convenient as it allowed to skip over large definitions (e.g., for `mcx` gates).
I would say that any definition for a builtin gate can be ignored. As for the `c` part, see above.

martin-fink: I've updated the standard gates, now only `cu` is parsed from stdgates.inc.

Now, if the parser encounters a `OpenQASM 2.0` declaration, it enters 'openqasm 2 compatibility mode'. If it encounters an unknown gates with leading `c`s, it pops those `c`s and interprets them as `ctrl @` modifiers. I.e., this
```qasm
cccx q[0], q[1], q[2], q[3];
```
is equivalent to
```qasm
ctrl @ x q[0], q[1], q[2], q[3];
```

The same is true for gate declarations. In OpenQASM 2.0 compatibility mode, the following gate definition will be ignored:
```qasm
gate cccx q1, q2, q3, q4 {
 // gate definition
}
```

burgholzer (approved):
Many many thanks again for this nice contribution! Great to see OpenQASM 3 support within the MQT finally happening 🎉 
Looking forward to some future additions to enhance the support even further.

PR #321, opened 2023-06-03
🎨 LLVM code format and Python binding restructure

burgholzer:
This PR unifies the `clang-format` configuration (closely following the LLVM style) and applies it to all files.
The ZX package was mostly formatted according to those rules anyway, so there are hardly any changes there.
Lots of line break and whitespace changes in the DD package and the QFR files though. Might need some cleanup later on.

This also moves the `qfr_python` target (which was previously a header-only library) to the main source tree and splits it into header and source files with the main goal of optimizing compile times.


burgholzer:
Ignoring the coverage decrease for now as the much bigger impact for coverage will probably come from moving the DD and ZX package out of the `extern` folder that is currently ignored.

PR #358, opened 2023-06-20
✨ Removal of identity nodes in matrix DDs

aaronleesander:
This pull request strips matrix DDs of their identities, representing them implicitly instead of explicitly.

This includes not constructing a gate DD for the full system size (i.e. a single qubit operation is expanded to n-qubit with identities), skipping nodes that resemble the identity, as well as handling skipped levels in operations such as multiplication and addition.

The identity is now represented by a single terminal node.


BertiFlorea:
Hey 👋🏼 
After long battles with Git, I finally managed to resolve as much as possible of the conflicts. However, @burgholzer you could take a closer look at some of the following functions in the DD package header:
1. `checkSpecialMatrices` (The symmetry check was removed in #442, however, if we remove the symmetry check and the identity check, then we would only have to check if it is a terminal the parameter node)
2. `add2` (Here I tried to keep the logic from the current branch)
3. `reduceAncillaeRecursion` (I kept the version from the identity-removal branch)
4. Lines from 2226 - 2538 (Most of the methods between these lines do not exist anymore in the current version of mqt-core, so I didn't know if they should be kept or not)

Additionally, in `Edge.cpp` the `isIdentity` method is commented out because I tried to implement it in the header file (Edge.hpp). Still, I wasn't sure what to return for the `else` case (therefore I left it commented where it initially was, in Edge.cpp).

There are still many issues and changes that need to be addressed, but for the moment if you could take a look and review the current status would be great! (I am sure there are flagrant mistakes or some urgent changes that need to be done)
Also, any hint or advice no matter how small is very welcome! 🙂

Lastly, apart from bringing the branch up to date, as discussed, there are some methods in the package that do not yet account for the missing identity nodes. Would it be possible to highlight them in your review to make it easier to track? 🙏🏼 

P.S. Many thanks in advance for your time to review, because I know there are quite a lot of places to investigate. (I think around 211 commits have been merged)

include/mqt-core/dd/FunctionalityConstruction.hpp
```diff
@@ -26,6 +26,10 @@ MatrixDD buildFunctionality(const qc::Grover* qc, Package<Config>& dd);
 template <class Config>
 MatrixDD buildFunctionalityRecursive(const qc::Grover* qc, Package<Config>& dd);
 
+template <class Config>
+MatrixDD buildFunctionality(const QuantumComputation* qc, Package<Config>& dd,
+                            const std::optional<std::size_t> ncycles);
+
```
burgholzer:
```suggestion
```
I am fairly sure this function has been removed on the `main` branch and is no longer necessary.

include/mqt-core/dd/Node.hpp
```diff
@@ -39,31 +39,30 @@ struct mNode {                        // NOLINT(readability-identifier-naming)
   RefCount ref{};                     // reference count
   Qubit v{};                          // variable index
   std::uint8_t flags = 0;
-  // 32 = unused (was used to mark a node with is symmetric).
-  // 16 = marks a node resembling identity
+  // 32 = marks a node with is symmetric.
+  // 16 = free (used to mark identity nodes)
```
burgholzer:
```suggestion
  // 32 = unused (was used to mark a node which is symmetric)
  // 16 = unused (was used to mark a node resembling the identity)
```
seems a little more consistent to me

include/mqt-core/dd/Node.hpp
```diff
@@ -39,31 +39,30 @@ struct mNode {                        // NOLINT(readability-identifier-naming)
...
   }
   [[nodiscard]] static constexpr mNode* getTerminal() noexcept {
     return nullptr;
   }
-
-  [[nodiscard]] inline bool isIdentity() const noexcept {
-    return (flags & static_cast<std::uint8_t>(16U)) != 0;
+  [[nodiscard]] inline bool isSymmetric() const noexcept {
+    return (flags & static_cast<std::uint8_t>(32U)) != 0;
   }
```
burgholzer:
```suggestion
```
No longer needed as this has been removed on the main branch

include/mqt-core/dd/Node.hpp
```diff
@@ -39,31 +39,30 @@ struct mNode {                        // NOLINT(readability-identifier-naming)
...
-  [[nodiscard]] static constexpr bool isIdentity(const mNode* p) noexcept {
-    return p == nullptr || p->isIdentity();
+  [[nodiscard]] static constexpr bool isSymmetric(const mNode* p) noexcept {
+    return p == nullptr || p->isSymmetric();
   }
```
burgholzer:
```suggestion
```
no longer needed

include/mqt-core/dd/Node.hpp
```diff
@@ -39,31 +39,30 @@ struct mNode {                        // NOLINT(readability-identifier-naming)
...
-  inline void setIdentity(const bool identity) noexcept {
-    if (identity) {
-      flags = (flags | static_cast<std::uint8_t>(16U));
+  inline void setSymmetric(const bool symmetric) noexcept {
+    if (symmetric) {
+      flags = (flags | static_cast<std::uint8_t>(32U));
     } else {
```
burgholzer:
```suggestion
```
No longer needed

include/mqt-core/dd/Node.hpp
```diff
@@ -39,31 +39,30 @@ struct mNode {                        // NOLINT(readability-identifier-naming)
...
-      flags = (flags & static_cast<std::uint8_t>(~16U));
+      flags = (flags & static_cast<std::uint8_t>(~32U));
     }
   }
```
burgholzer:
```suggestion
```
No longer needed

include/mqt-core/dd/Node.hpp
```diff
@@ -80,12 +79,12 @@ struct dNode {                        // NOLINT(readability-identifier-naming)
   RefCount ref{};                     // reference count
   Qubit v{};                          // variable index
   std::uint8_t flags = 0;
-  // 32 = unused (was used to mark a node with is symmetric).
-  // 16 = marks a node resembling identity
+  // 32 = marks a node with is symmetric.
+  // 16 = free (used to mark identity nodes)
```
burgholzer:
```suggestion
  // 32 = unused (was used to mark a node which is symmetric)
  // 16 = unused (was used to mark a node resembling the identity)
```

src/dd/Edge.cpp
```diff
@@ -24,6 +24,13 @@ namespace dd {
 ///                      \n General purpose methods \n
 ///-----------------------------------------------------------------------------
 
+// template <class Node> bool Edge<Node>::isIdentity() const {
+//   if constexpr (std::is_same_v<Node, mNode> || std::is_same_v<Node, dNode>) {
+//     return isTerminal() && w != Complex::zero();
+//   }
+//   return false;
+// }
```
burgholzer: This looks reasonable as it is. You can just uncomment it and use the respective function.
Potentially, it could be worth adding an additional flag `upToGlobalPhase` that defaults to `true`. If that flag is set to false, `w` has to be equal to `Complex::one()` for the method to return true.

src/dd/Edge.cpp
burgholzer, line 1: There are many function throughout this file that need adaptation to the new identity-less structure. These are all the routines that got moved from the `Package.hpp` file. So in some of the cases, you should be able to use the previous commits on this PR as a reference for how to adjust these functions.
This concerns almost every single function in here, since most rely on the fact that `p->v` can be used to get the number of qubits in the system. In most cases, some kind of workaround will be necessary here, such as adding a new parameter that passes in the number of qubits.
The alternative to the above would be to introduce a new struct `DD<Node>` that essentially looks something like
```c++
template<class Node>
struct DD {
Edge<Node> e{};
std::size_t numQubits{};
}
```
And use these new classes as top-level DD objects. Then most of the functions currently being defined here (`getVector`, etc.) could be moved to that class instead. This would imply quite a couple of interfaces changes, but could overall work out really nicely.

src/dd/FunctionalityConstruction.cpp
```diff
@@ -197,6 +197,30 @@ MatrixDD buildFunctionalityRecursive(const qc::Grover* qc,
...
+    }
+    auto g = dd->multiply(f, e);
+    dd->decRef(e);
+    dd->incRef(g);
+    e = g;
+    dd->garbageCollect();
+  }
+  return e;
+}
+
```
burgholzer:
```suggestion
```
This was removed on main

src/dd/FunctionalityConstruction.cpp
```diff
@@ -211,4 +235,7 @@ template MatrixDD buildFunctionality(const qc::Grover* qc,
                                      Package<DDPackageConfig>& dd);
 template MatrixDD buildFunctionalityRecursive(const qc::Grover* qc,
                                               Package<DDPackageConfig>& dd);
+template MatrixDD buildFunctionality(const qc::QuantumComputation* qc,
+                                     Package<DDPackageConfig>& dd,
+                                     const std::optional<std::size_t> ncycles);
```
burgholzer:
```suggestion
```

test/algorithms/test_grover.cpp
```diff
@@ -60,7 +60,6 @@ TEST_P(Grover, Functionality) {
   qc->printStatistics(std::cout);
   auto x = '1' + qc->expected;
   std::reverse(x.begin(), x.end());
-  std::replace(x.begin(), x.end(), '1', '2');
```
burgholzer: I am fairly sure this should not be removed

test/algorithms/test_grover.cpp
```diff
@@ -81,7 +80,6 @@ TEST_P(Grover, FunctionalityRecursive) {
...
```
burgholzer: I am fairly sure this should not be removed

test/dd/test_dd_functionality.cpp
```diff
@@ -329,10 +329,10 @@ TEST_F(DDFunctionality, nonUnitary) {
...
   auto barrier =
       qc::StandardOperation(nqubits, {0, 1, 2, 3}, qc::OpType::Barrier);
-  EXPECT_EQ(getDD(&barrier, *dd), dd->makeIdent(nqubits));
-  EXPECT_EQ(getInverseDD(&barrier, *dd), dd->makeIdent(nqubits));
-  EXPECT_EQ(getDD(&barrier, *dd, dummyMap), dd->makeIdent(nqubits));
-  EXPECT_EQ(getInverseDD(&barrier, *dd, dummyMap), dd->makeIdent(nqubits));
+  EXPECT_EQ(getDD(&barrier, *dd), dd->makeIdent());
+  EXPECT_EQ(getInverseDD(&barrier, *dd), dd->makeIdent());
+  EXPECT_EQ(getDD(&barrier, *dd, dummyMap), dd->makeIdent());
+  EXPECT_EQ(getInverseDD(&barrier, *dd, dummyMap), dd->makeIdent());
```
burgholzer: You could probably rewrite these to something like
```c++
EXPECT_TRUE(getDD(...).isIdentity());
```

test/dd/test_package.cpp
```diff
@@ -477,41 +478,55 @@ TEST(DDPackageTest, SerializationErrors) {
...
+  EXPECT_FALSE(local);
+  bellState.p->v = 1;
 
-  auto idCached = dd->makeIdent(4);
-  EXPECT_EQ(id4, idCached);
+  bellState.p->e[0].w.r->ref = 0;
+  local = dd->isLocallyConsistent(bellState);
+  EXPECT_FALSE(local);
+  bellState.p->e[0].w.r->ref = 1;
 }
```
burgholzer: Something went wrong with the merge here in these sections.
I believe the consistency methods have been removed from the main branch and the identity tests can be removed as part of this PR. So, in the end, none of these lines should be here anymore.

test/dd/test_package.cpp
```diff
@@ -607,15 +617,15 @@ TEST(DDPackageTest, GarbageMatrix) {
   dd->incRef(bellMatrix);
   reducedBellMatrix =
       dd->reduceGarbage(bellMatrix, {false, true, false, false});
-  auto mat = reducedBellMatrix.getMatrix();
+  auto mat = dd->getMatrix(reducedBellMatrix, 2);
```
burgholzer: This is calling a method that should no longer exist since it has been moved to the Edge.hpp file

test/dd/test_package.cpp
```diff
@@ -1082,7 +1093,7 @@ TEST(DDPackageTest, NormalizationNumericStabilityTest) {
     auto p = dd->makeGateDD(dd::pMat(lambda), 1, 0);
     auto pdag = dd->makeGateDD(dd::pMat(-lambda), 1, 0);
     auto result = dd->multiply(p, pdag);
-    EXPECT_TRUE(result.p->isIdentity());
+    EXPECT_TRUE(result.isOneTerminal());
```
burgholzer:
```suggestion
    EXPECT_TRUE(result.isIdentity());
```

test/dd/test_package.cpp
```diff
@@ -1579,7 +1590,7 @@ TEST(DDPackageTest, DDFromSingleQubitMatrix) {
   const auto dd = std::make_unique<dd::Package<>>(nrQubits);
   const auto matDD = dd->makeDDFromMatrix(inputMatrix);
 
-  const auto outputMatrix = matDD.getMatrix();
+  const auto outputMatrix = dd->getMatrix(matDD, nrQubits);
```
burgholzer: Similar to above, this is using a function that should no longer exist. Please also check all other occurrences within this file.

test/dd/test_package.cpp
```diff
@@ -2165,6 +2176,34 @@ TEST(DDPackageTest, InnerProductTopNodeConjugation) {
...
+  const auto nrQubits = 2U;
+
+  auto dd = std::make_unique<dd::Package<>>(nrQubits);
+
+  auto originalDD = dd->makeGateDD(dd::X_MAT, 2, 1_pc, 0);
+  auto matrix = dd->getMatrix(originalDD, nrQubits);
+  auto recreatedDD = dd->makeDDFromMatrix(matrix);
+
+  ASSERT_EQ(originalDD, recreatedDD);
+}
```
burgholzer: These are just rather rudimentary tests that have been used for developing. Before the PR is being merged, they should be extended to more elaborate tests that explicitly make sure the new functionality works as intended.

src/dd/Operations.cpp
```diff
@@ -88,7 +88,7 @@ void dumpTensor(qc::Operation* op, std::ostream& of,
     const auto localDD = getDD(localOp.get(), dd);
 
     // translate local DD to matrix
-    const auto localMatrix = localDD.getMatrix();
+    const auto localMatrix = dd.getMatrix(localDD, localQubits);
```
burgholzer: This is using a method that has been deleted on the main branch. You'll find more details in other comments.

include/mqt-core/dd/Package.hpp
```diff
@@ -757,40 +732,36 @@ template <class Config> class Package {
...
-    if (!mNode::isIdentity(e0.p) || !e1.w.exactlyZero() ||
-        !e2.w.exactlyZero() || !e0.w.exactlyOne() || !e3.w.exactlyOne() ||
-        !mNode::isIdentity(e3.p)) {
+    if (!mNode::isSymmetric(e0.p) || !mNode::isSymmetric(e3.p)) {
       return;
     }
-    p->setIdentity(true);
+
+    p->setSymmetric(true);
   }
```
burgholzer: This function can be completely removed now as it no longer has any function

include/mqt-core/dd/Package.hpp
```diff
@@ -1532,6 +1536,7 @@ template <class Config> class Package {
       @return a complex number representing the scalar product of the DDs
   **/
   ComplexValue innerProduct(const vEdge& x, const vEdge& y) {
+    // TODO: Adapt to identities
```
```suggestion
```
This was actually a nonsense todo as no change has been applied to vector DDs.

include/mqt-core/dd/Package.hpp
```diff
@@ -1740,6 +1737,7 @@ template <class Config> class Package {
       return CachedEdge<Node>::zero();
     }
 
+    // TODO: this does also not yet work with identities
```
burgholzer: Here this is a proper TODO as I am fairly sure that the current method does not handle identities well. Especially considering the variable index in the nodes.

include/mqt-core/dd/Package.hpp
```diff
@@ -1858,13 +1866,13 @@ template <class Config> class Package {
                                   const dd::fp tol,
                                   const std::vector<bool>& garbage,
                                   const bool checkCloseToOne) {
-    // immediately return if this node has already been visited
-    if (visited.find(m.p) != visited.end()) {
+    // immediately return of this node is identical to the identity
```
```suggestion
    // immediately return if this node is identical to the identity
```

include/mqt-core/dd/Package.hpp
```diff
@@ -2006,24 +1961,18 @@ template <class Config> class Package {
   ///
   mEdge reduceAncillae(mEdge& e, const std::vector<bool>& ancillary,
                        const bool regular = true) {
+    // TODO:these methods do not handle identities yet
```
burgholzer: This is also a valid TODO. I believe even the top-level logic currently in the PR does not really properly handle the situation and the version from the main branch is much closer to a full solution (at least here; not for the recursive function).

include/mqt-core/dd/Package.hpp
```diff
@@ -2006,24 +1961,18 @@ template <class Config> class Package {
   ///
   mEdge reduceAncillae(mEdge& e, const std::vector<bool>& ancillary,
```
burgholzer, line 2153: Unfortunately, GitHub won't let me comment on the respective function, but the `reduceGarbage` method (for matrices) should also be adapted to properly handle the new identity-less DD structure.

include/mqt-core/dd/Package.hpp
```diff
@@ -2122,6 +2071,77 @@ template <class Config> class Package {
...
+  //        edges[i] =
+  //            reduceAncillaeRecursion(f.p->e[i], ancillary, var - 1, regular);
+  //      }
+  //      f = makeDDNode(var, edges);
+  //    }
+  //
+  //    f.w = cn.lookup(f.w * e.w);
+  //    return f;
+  //  }
+
```
```suggestion
```
No need to keep this here. The method from the main branch (which is just above this large comment block) supersedes it.

include/mqt-core/dd/Package.hpp
```diff
@@ -2122,6 +2071,77 @@ template <class Config> class Package {
                                        mCachedEdge::zero()});
```
burgholzer: The 
```c++
mCachedEdge reduceAncillaeRecursion(mNode* p,
                                      const std::vector<bool>& ancillary,
                                      const Qubit lowerbound,
                                      const bool regular = true)
```
needs to properly handle the new structure.

include/mqt-core/dd/Package.hpp
```diff
@@ -2122,6 +2071,77 @@ template <class Config> class Package {
...
+  //            reduceAncillaeRecursion(f.p->e[i], ancillary, var - 1, regular);
+  //      }
+  //      f = makeDDNode(var, edges);
+  //    }
+  //
+  //    f.w = cn.lookup(f.w * e.w);
+  //    return f;
+  //  }
+
   vCachedEdge reduceGarbageRecursion(vNode* p, const std::vector<bool>& garbage,
```
burgholzer: GitHub won;t let me comment on the right line, but the 
```c++
mCachedEdge reduceGarbageRecursion(mNode* p, const std::vector<bool>& garbage,
                                     const Qubit lowerbound,
                                     const bool regular = true)
```
method also needs adaptation to the new structure.

include/mqt-core/dd/Package.hpp
```diff
@@ -2203,6 +2223,320 @@ template <class Config> class Package {
...
+      if (!e.p->e[3].w.approximatelyZero()) {
+        getMatrix(e.p->e[3], c, x, y, mat, level - 1);
+      }
+    } else if ((!e.isTerminal() && e.p->v < level) ||
+               (e.isTerminal() && level != -1)) {
+      getMatrix(e, c, i, j, mat, level - 1);
+      getMatrix(e, c, x, y, mat, level - 1);
+    }
+  }
+
```
```suggestion
```
These can be removed completely, as they have all been moved to the `Edge.hpp` file. 
Any adaptation should happen there.

burgholzer (changes requested):
Hey @BertiFlorea 👋🏼 

Many thanks for taking on this task and resolving all those conflicts with `main`. That must have been quite burdensome.
Nevertheless, I think the merge went rather well. I added comments to some leftover places, where something is not quite right yet. But these should be easy to fix.


Naturally, the most important thing right now is getting the code to compile again.
After that, some of the methods will need adjustments to account for the new identity-less structure.
The good thing here is that the gate construction, the addition, and the multiplication all seem to be in a good state already and look really clean now. This means that the main functionality should be working well already.
All it takes now is to "just" (that's a big "just") adapt the utility functions for translating DDs to vectors and matrices as well as some smaller mathematical functions such as the kronecker product and ancillary/garbage reduction. 
As you requested, I explicitly tried to add comments to the respective places that should help keeping track of what still needs to change.

We can really take this step. So let's try to first resolve all the easy comments and suggestions.
Let me know if you have any questions.

src/dd/Edge.cpp
BertiFlorea: I've just updated the `getValueByPath()` method which many functions in this file rely on (see [commit](https://github.com/cda-tum/mqt-core/pull/358/commits/e26a998f4a955894fc1cabc6e61876e44df2dd06)). At this point, I just wanted to mention for future reference that each time we call a printing method (regardless of the node type) we will have to set the number of total qubits beforehand. For example (`test_package.cpp: 335`):

``` 
bellMatrix.numQubits = 2; 
bellMatrix.printMatrix();
```
A couple of them are still to be refactored (e.g., the matrix traversal ones).
P.S. I didn't modify any vector-related functions, since the identity-less structure only impacts the matrices.

BertiFlorea: Done ✅

include/mqt-core/dd/Edge.hpp
```diff
@@ -51,7 +51,8 @@ using MatrixEntryFunc = std::function<void(const std::size_t, const std::size_t,
  */
 template <class Node> struct Edge {
   Node* p;
-  Complex w;
+  Complex w{};
+  std::size_t numQubits{};
```
burgholzer: Just a quick side comment here:
This change is not really acceptable as it increases the size of _every_ single node in the decision diagrams (the literal nodes, not the `Node` class) by 33%. For something that is not even needed for the main functionality (MxV multiplication, etc.), this is a tradeoff that we cannot make here.

An acceptable change would be to adjust the signature of the functions that explicitly need the number of qubits to take an additional argument. This is closely related to the other comment in this PR about the potential need for a new level of abstraction above the `Edge` class that, e.g., has the information on the number of qubits. Although I would assume that this new level of abstraction is more work than simply adjusting the function signatures and it might even make some things a little more challenging (e.g., dropping the number of qubits member from the `Operation` class).
Bottomline: I would just revise the signatures of the couple of functions that need the number of qubits.

BertiFlorea: I just fixed it right now. Thanks a lot for the advice!

include/mqt-core/dd/Package.hpp
```diff
@@ -1462,8 +1440,7 @@ template <class Config> class Package {
       return {y.p, rWeight};
     }
 
-    if constexpr (std::is_same_v<RightOperandNode, mNode> ||
-                  std::is_same_v<RightOperandNode, dNode>) {
+    if constexpr (std::is_same_v<RightOperandNode, mNode>) {
```
burgholzer: Fixing the density matrix multiplication failures might be as easy as reverting this change here.

burgholzer (approved):
All the checks are green, all the dependent PRs are green as well.
So this epic can finally be merged 🎉 
Thanks to everyone that contributed!

PR #361, opened 2023-06-23
🐛 Make MSVC work

burgholzer:
## Description

Before the move to MQT Core, the DD Package was designed as a header-only library. The way we had to handle terminal nodes via static inline member variables lead to the project not compiling with MSVC.
Now that the DD package is part of MQT Core, there really is no need for it to be header-only. 

This PR makes a first step into that direction and resolves the ackward handling of the various DD terminals. 
> **Warning**
> This is a breaking change. The previous solution declared a
> ```c++
>  static mNode terminalNode;
>  constexpr static mNode* terminal{&terminalNode};
> ```
> This now becomes
> ```c++
> static mNode terminal;
> ```
> Thus, it might be necessary to replace occurences of `mNode::terminal` with `&mNode::terminal`. (similarly for `vNode` and `dNode`)

For the first time in years, the change sin this PR allow the project to be compiled by MSVC again and we can skip all the nasty workarounds for that.
This PR also addresses some of the warnings raised by MSVC so that none of the compilers should be producing any warnings.

## Checklist:

### DIFF
--- a/include/dd/Package.hpp
+++ b/include/dd/Package.hpp
@@ -952,8 +952,8 @@ private:
       const auto& zeroWeight = cn.getCached(begin->real(), begin->imag());
       const auto& oneWeight =
           cn.getCached(std::next(begin)->real(), std::next(begin)->imag());
-      const auto zeroSuccessor = vEdge{&vNode::terminal, zeroWeight};
-      const auto oneSuccessor = vEdge{&vNode::terminal, oneWeight};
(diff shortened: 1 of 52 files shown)


include/dd/Package.hpp
```diff
@@ -7,11 +7,12 @@
 #include "ComplexValue.hpp"
 #include "ComputeTable.hpp"
 #include "Control.hpp"
+#include "DDpackageConfig.hpp"
```
```suggestion
```

PR #371, opened 2023-07-03
✨♻️ Refactor `QuantumComputation` API and Expose to Python

pehamTom:
## Description

To make mqt-core accessible from python - in particular to make commonly used mqt-core methods and datatypes available to other mqt tools in python - this PR aims to port the core functionality to python. 

This PR does not aim to provide every functionality available in the C++ library to python. This is left for future changes when the demand for such functionality arises.

In the process, some of the interfaces of the `QuantumComputation` class where adjusted drastically.
This constitutes a breaking change and will require consuming libraries to adapt.
However, it greatly simplifies some constructions and makes the interface way cleaner.
In particular,
- some operation names have changed to more closely map to the Qiskit names
- the order of function parameters in the `QuantumComputation` class was switched to be more natural and in-line with what other SDKs are providing. The same convenience functions were given more suitable names to indicate their functionality (e.g., `x`, `cx`, and `mcx`). Lots of boilerplate code was reduced and refactored.
- `Control` objects are now implicitly constructible from `Qubit` indices which, again, simplifies a lot of constructs
- Operations and Quantum Computations are now copyable by default. There never was a good reason for them to not be copyable and it limited us and the API in quite some ways. This makes use of the previously defined `clone` methods.
- The reliability of circuit printing was improved considerably.
- The new and native Qiskit import can properly handle Initial Layouts composed of multiple registers in not-necessarily alphabetical order. This should eliminate some error that people encountered in certain applications, e.g., using QCEC.

Last, but certainly not least, this PR also adds full API documentation for the exposed functionality and a Quickstart Guide to get started.

## Checklist:


burgholzer:
@pehamTom many thanks for working on this; much appreciated help!
Just wanted to write down some thoughts:

Once the `QuantumComputation` class is "fully" exposed to Python, I would argue that the existing binding functionality from the `.../qiskit/*.cpp` files should be directly ported to Python as the C++ implementation is cumbersome (and not even that efficient I believe, due to the frequent calls from C++ to Python).
One could think about adopting a similar naming to the TKET library, which would be something like `qiskit_to_mqt` and `mqt_to_qiskit`.

To this end, I would like to introduce a certain separation of concerns between the _core_ library and its extensions (like the adapter to Qiskit). 
Someone who does not depend or rely on Qiskit should not have to install it as part of MQT Core. I believe this should be fairly easy to accomplish by adding an optional `qiskit` extra to the package and handling all related code in a separate submodule `mqt.core.qiskit`. The functions of that module are _not_ re-exported as part of the `mqt.core.__init__.py`. Users have to explicitly import the submodule and/or the functions to use them.

This allows us to set a precedent for how to provide adapters from the MQT to other toolkits and should keep the code rather modular. I hope this makes sense. If you do not agree with something or something is not clear, let's discuss.

extern/googletest
burgholzer, line 1: This change seems unintended.

extern/json
burgholzer, line 1: This change seems unintended.

extern/pybind11
burgholzer, line 1: This change seems unintended.

pyproject.toml
```diff
@@ -34,6 +34,9 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dynamic = ["version"]
+dependencies = [
+             "qiskit-terra>=0.20.2",
+]
```
burgholzer: I'd really like to make the dependency management in mqt-core a little more modular.
One should not have to install qiskit if one is just using the core functionality.
I'd rather incorporate that as an optional `qiskit` dependency. (The `dev` extra should also install that dependency)
If then, tomorrow, a `cirq` extension comes along, it get's another extra.


All in all I would really like the core of mqt-core to remain dependency free for as long as possible.

pyproject.toml
burgholzer, line 1: GitHub won't let me comment on the right line, but since we drop 3.12 wheels with this PR, the corresponding metadata tag in line 31 should also be dropped.

src/mqt/core/qiskit_utils/__init__.py
burgholzer, line 1: I have a slight preference towards a shorter name for this module, e.g., `qiskit`. Any objections to that?

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+from __future__ import annotations
+
+from typing import cast
+
+from qiskit import QuantumCircuit
+from qiskit.circuit import AncillaRegister, ClassicalRegister, Parameter, QuantumRegister, Qubit
+from qiskit.circuit.library import MCXRecursive, MCXVChain, XXMinusYYGate, XXPlusYYGate
+from qiskit.transpiler import Layout, TranspileLayout
+
+from mqt.core._core import Expression, SymbolicOperation
```
burgholzer: If those are used in tests, i.e., user code, then these classes should be importable via the `core` module.

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+
+def test_mcx_recursive() -> None:
+    """Test import of large mcx gate."""
+    q = QuantumCircuit(9)
+    q.append(MCXRecursive(num_ctrl_qubits=7), range(9))
+    mqt_qc = qiskit_to_mqt(q)
+    assert mqt_qc.n_qubits == 9
+    assert mqt_qc.n_ops == 1
+    assert mqt_qc[0].name.strip() == "x"
+    assert mqt_qc[0].n_qubits == 9
```
burgholzer: Just looking at the test (have not looked at the corresponding import code), I would have expected this to only act on 8 qubits since one is an ancillary if I am not mistaken.

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+
+def test_mcx_vchain() -> None:
+    """Test import of mcx gate with v-chain."""
+    q = QuantumCircuit(9)
+    q.append(MCXVChain(num_ctrl_qubits=5), range(9))
+    mqt_qc = qiskit_to_mqt(q)
+    assert mqt_qc.n_qubits == 9
+    assert mqt_qc.n_ops == 1
+    assert mqt_qc[0].name.strip() == "x"
+    assert mqt_qc[0].n_qubits == 9
```
burgholzer: Similar to above. I am fairly sure this should report fewer qubits.

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+    assert mqt_qc.n_qubits == 3
+    assert mqt_qc.n_ops == 3
+    assert mqt_qc[0].name.strip() == "h"
+    assert mqt_qc[1].name.strip() == "x"
+    assert mqt_qc[2].name.strip() == "x"
+    assert mqt_qc[0].n_qubits == 3
+    assert mqt_qc[1].n_qubits == 3
+    assert mqt_qc[1].n_qubits == 3
+    assert {control.qubit for control in mqt_qc[1].controls} == {0}
+    assert {control.qubit for control in mqt_qc[2].controls} == {0}
```
burgholzer: This might be conforming with the old behavior, but I actually believe it might not be desirable to straight up flatten the quantum circuit. I'd rather have a custom instruction translate to a CompoundOperation.

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def test_layout() -> None:
+    """Test import of initial layout."""
+    qc = QuantumCircuit(3)
+    q_reg = QuantumRegister(3, "q")
+    qc._layout = TranspileLayout(  # noqa: SLF001
```
burgholzer: You can eliminate these kind of warnings by explicitly annotating the self type. A python Version range compatible way can be found here: https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/_compat/typing.py
Adopting these kind of compatibility wrappers is quickly becoming a best practice among the python community.

Note that you might also have to add https://github.com/scikit-build/scikit-build-core/blob/6e18ece11af457c322ebb1733e304aec473925d3/pyproject.toml#L41
And 
https://github.com/scikit-build/scikit-build-core/blob/6e18ece11af457c322ebb1733e304aec473925d3/pyproject.toml#L222

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+        Layout.from_intlist([1, 2, 0], q_reg),
+    )
+    qc.h(0)
+    qc.s(1)
+    qc.x(2)
+    mqt_qc = qiskit_to_mqt(qc)
+    assert mqt_qc.n_qubits == 3
+    assert mqt_qc.n_ops == 3
+    assert mqt_qc.initial_layout.apply([0, 1, 2]) == [2, 1, 0]
+    assert mqt_qc.output_permutation.apply([0, 2, 1]) == [2, 1, 0]
```
burgholzer: Couldn't you just read out the layouts here instead of applying them to something?

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+    qc.rxx(0.5, 0, 1)  # op 16
+    qc.rzz(0.5, 0, 1)  # op 17
+    qc.ryy(0.5, 0, 1)  # op 18
+    qc.append(XXMinusYYGate(0.1), [0, 1])  # op 19
+    qc.append(XXPlusYYGate(0.1), [0, 1])  # op 20
+
+    mqt_qc = qiskit_to_mqt(qc)
+    assert mqt_qc.n_qubits == 3
+    assert mqt_qc.n_ops == 20
+    assert not mqt_qc.is_variable_free()
```
burgholzer: Why would that circuit not be variable free?

test/python/test_qiskit.py
```diff
@@ -0,0 +1,197 @@
...
+    mqt_qc = qiskit_to_mqt(qc)
+
+    assert mqt_qc.n_qubits == 1
+    assert mqt_qc.n_ops == 1
+    assert mqt_qc[0].name.strip() == "rx"
+    assert isinstance(mqt_qc[0], SymbolicOperation)
+    assert isinstance(mqt_qc[0].get_parameter(0), Expression)
+    expr = cast(Expression, mqt_qc[0].get_parameter(0))
+    assert expr.num_terms() == 2
+    assert expr.constant == 0.0
```
burgholzer: This test could use the "variable_free" test.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+#include <nanobind/stl/variant.h>
+#include <nanobind/stl/vector.h>
+#include <ostream>
 #include <python/nanobind.hpp>
+#include <sstream>
+#include <string>
+#include <vector>
 
 namespace mqt {
+using qc::QuantumComputation;
```
burgholzer: Why is that statement needed here?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+      "Representation of quantum circuits within MQT Core")
+      .def(nb::init<>(), "Constructs an empty QuantumComputation.")
+      .def(nb::init<std::size_t>(), "nq"_a,
+           "Constructs an empty QuantumComputation with the given number of "
+           "qubits.")
+      .def(nb::init<std::string>(), "filename"_a,
+           "Read QuantumComputation from given file. Supported formats are "
+           "[OpenQASM, Real, GRCS, TFC, QC]")
+      // .def_rw("name", &qc::QuantumComputation::getName,
+      // &qc::QuantumComputation::setName)
```
burgholzer: Leftover code that can be removed

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+      .def_prop_ro("n_ancillae", &qc::QuantumComputation::getNancillae)
+      .def_prop_ro("n_qubits_without_ancillae",
+                   &qc::QuantumComputation::getNqubitsWithoutAncillae)
+      .def_prop_ro("n_cbits", &qc::QuantumComputation::getNcbits)
+      .def_prop_ro("n_ops", &qc::QuantumComputation::getNops)
+      .def_prop_ro("n_single_qubit_ops",
+                   &qc::QuantumComputation::getNsingleQubitOps)
+      .def_prop_ro("n_individual_ops",
+                   &qc::QuantumComputation::getNindividualOps)
+      .def_prop_ro("depth", &qc::QuantumComputation::getDepth)
```
burgholzer: Should this really be a property? Accessing that property requires a function call that's linear in the number of gates.
Might be worth revisiting that for the other properties as well.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+      .def_prop_rw("gphase", &qc::QuantumComputation::getGlobalPhase,
+                   &qc::QuantumComputation::gphase)
+      .def_rw("initial_layout", &qc::QuantumComputation::initialLayout)
+      .def_rw("output_permutation", &qc::QuantumComputation::outputPermutation)
+      .def("i", nb::overload_cast<qc::Qubit>(&qc::QuantumComputation::i))
+      .def("i", nb::overload_cast<qc::Qubit, const qc::Control&>(
+                    &qc::QuantumComputation::i))
+      .def("i", nb::overload_cast<qc::Qubit, const qc::Controls&>(
+                    &qc::QuantumComputation::i))
```
burgholzer: I guess there is really no generic way to unify this boilerplate code. For example, using templates... it's always three methods that take the same parameters and pass-through to the underlying circuit method with the same name.
Technically a solution here could maybe already work for the original circuit functions and unify them.
If you don't see an immediate solution, don't worry too much about it. I was just wishfully thinking.

src/python/module.cpp
burgholzer, line 1: Many of the functions here that are directly bound are missing annotations for their parameters.

burgholzer, line 1: Many of the functions that are bound here are missing corresponding docstrings. It would really be awesome to straight up adopt the usual docstrings used in python that include at least a short description, parameter and return type descriptions. That way, these core methods can easily be incorporated into the documentation.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+      .def("set_logical_qubit_ancillary",
+           &qc::QuantumComputation::setLogicalQubitAncillary)
+      .def("add_qubit_register", &qc::QuantumComputation::addQubitRegister)
+      .def("add_classical_bit_register",
+           &qc::QuantumComputation::addClassicalRegister)
+      .def("add_ancillary_register",
+           &qc::QuantumComputation::addAncillaryRegister)
+      .def("append_operation",
+           [](QuantumComputation& qc, const qc::Operation& op) {
+             qc.emplace_back(op.clone()); // not an ideal solution but it works
```
burgholzer: Couldn't you use std::move instead? Would invalidate the operation of course, but I guess that would be expected/could be documented as such.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+           [](qc::QuantumComputation& qc, const std::string& filename,
+              const std::string& format) {
+             if (format == "qasm") {
+               qc.import(filename, qc::Format::OpenQASM);
+             } else if (format == "real") {
+               qc.import(filename, qc::Format::Real);
+             } else if (format == "grcs") {
+               qc.import(filename, qc::Format::GRCS);
+             } else if (format == "tfc") {
+               qc.import(filename, qc::Format::TFC);
```
burgholzer: Are these still relevant?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+             } else if (format == "tensor") {
+               qc.import(filename, qc::Format::Tensor);
+             } else if (format == "qc") {
+               qc.import(filename, qc::Format::QC);
+             } else {
+               throw qc::QFRException("Unknown format: " + format);
+             }
+           })
```
burgholzer: Instead of documenting this in a code comment, move it to a corresponding docstring as a `note`.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+      .def("dump",
+           nb::overload_cast<const std::string&>(&qc::QuantumComputation::dump))
+      .def("dump",
+           [](qc::QuantumComputation& qc, const std::string& filename,
```
burgholzer: Still necessary?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+               qc.dump(filename, qc::Format::TFC);
+             } else if (format == "tensor") {
+               qc.dump(filename, qc::Format::Tensor);
+             } else if (format == "qc") {
+               qc.dump(filename, qc::Format::QC);
+             } else {
+               throw qc::QFRException("Unknown format: " + format);
+             }
+           })
+      .def("to_open_qasm",
```
burgholzer: Why are lambdas necessary here? Couldn't you just bind the respective functions?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+           [](qc::QuantumComputation& qc) {
+             std::ostringstream oss;
+             qc.dumpOpenQASM(oss);
+             return oss.str();
+           })
+      .def("__len__", &qc::QuantumComputation::getNindividualOps)
```
burgholzer: Same as above.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+      .def(
+          "__getitem__",
+          [](const QuantumComputation& qc,
+             std::size_t
+                 idx) { return qc.at(idx).get(); }, // Beware: this gives write
+                                                    // access to underlying
```
burgholzer: Shouldn't this also be exposed?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+           "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Targets&, qc::OpType,
+                    std::vector<qc::fp>, qc::Qubit>(),
+           "nq"_a, "targets"_a, "op_type"_a, "params"_a = std::vector<qc::fp>{},
+           "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, qc::Control, qc::Qubit, qc::OpType,
+                    const std::vector<qc::fp>&, qc::Qubit>(),
+           "nq"_a, "control"_a, "target"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, qc::Control, const qc::Targets&, qc::OpType,
```
burgholzer: Is the `&&` really intended here?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, qc::Qubit, qc::OpType,
+                    const std::vector<qc::fp>&, qc::Qubit>(),
+           "nq"_a, "controls"_a, "target"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, const qc::Targets&,
+                    qc::OpType, std::vector<qc::fp>, qc::Qubit>(),
+           "nq"_a, "controls"_a, "targets"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, qc::Qubit, qc::Qubit>(),
```
burgholzer: Is it really necessary to bind that function? Seems rather private to me.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+           "nq"_a, "controls"_a, "target"_a, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, qc::Qubit, qc::Qubit,
+                    qc::OpType, std::vector<qc::fp>, qc::Qubit>(),
+           "nq"_a, "controls"_a, "target0"_a, "target1"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def("is_standard_operation", &qc::StandardOperation::isStandardOperation)
```
burgholzer: Still relevant?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+                    const std::vector<qc::fp>&, qc::Qubit>(),
+           "nq"_a, "controls"_a, "target"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, const qc::Targets&,
+                    qc::OpType, std::vector<qc::fp>, qc::Qubit>(),
+           "nq"_a, "controls"_a, "targets"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, qc::Qubit, qc::Qubit>(),
+           "nq"_a, "controls"_a, "target"_a, "starting_qubit"_a = 0)
+      .def(nb::init<std::size_t, const qc::Controls&, qc::Qubit, qc::Qubit,
```
burgholzer: I'd rather remove that duplicate definition. __len__ is enough IMHO

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+                    qc::OpType, std::vector<qc::fp>, qc::Qubit>(),
+           "nq"_a, "controls"_a, "target0"_a, "target1"_a, "op_type"_a,
+           "params"_a = std::vector<qc::fp>{}, "starting_qubit"_a = 0)
+      .def("is_standard_operation", &qc::StandardOperation::isStandardOperation)
+      .def("clone", &qc::StandardOperation::clone)
+      .def("equals", nb::overload_cast<const qc::Operation&>(
```
burgholzer: Just wondering: what's the return value policy here? Does this just return a reference to the operation or does it transfer ownership?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+  nb::class_<qc::CompoundOperation, qc::Operation>(m, "CompoundOperation")
+      .def(nb::init<std::size_t,
+                    std::vector<std::unique_ptr<qc::Operation>>&&>())
+      .def("clone", &qc::CompoundOperation::clone)
+      .def("set_n_qubits", &qc::CompoundOperation::setNqubits)
+      .def("is_compound_operation", &qc::CompoundOperation::isCompoundOperation)
+      .def("is_non_unitary_operation",
+           &qc::CompoundOperation::isNonUnitaryOperation)
+      .def("equals",
+           nb::overload_cast<const qc::Operation&, const qc::Permutation&,
```
burgholzer: As above, probably not really necessary.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+
+  nb::class_<qc::NonUnitaryOperation, qc::Operation>(m, "NonUnitaryOperation")
+      .def(
+          nb::init<std::size_t, std::vector<qc::Qubit>, std::vector<qc::Bit>>())
+      .def(nb::init<std::size_t, qc::Qubit, qc::Bit>())
+      .def(nb::init<std::size_t, std::vector<qc::Qubit>, std::size_t>())
+      .def(nb::init<std::size_t>())
+      .def(nb::init<std::size_t, std::vector<qc::Qubit>, qc::OpType>())
+      .def("clone", &qc::NonUnitaryOperation::clone)
+      .def("is_unitary", &qc::NonUnitaryOperation::isUnitary)
```
burgholzer: Maybe move that to a docstring

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+                             const qc::Permutation&>(
+               &qc::CompoundOperation::equals, nb::const_))
+      .def("acts_on", &qc::CompoundOperation::actsOn)
+      .def("add_depth_contribution",
+           &qc::CompoundOperation::addDepthContribution)
+      .def("__len__", &qc::CompoundOperation::size)
+      .def("size", &qc::CompoundOperation::size)
+      .def("empty", &qc::CompoundOperation::empty)
+      .def("__getitem__", [](const qc::CompoundOperation& op,
+                             std::size_t i) { return op.at(i).get(); })
```
burgholzer: Isn't the permutation class used somewhere above already? Does nanobind not require used types to be bound before their first use?

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+           [](const sym::Expression<double, double>& expr, std::size_t idx) {
+             if (idx >= expr.numTerms()) {
+               throw nb::index_error();
+             }
+             return expr.getTerms()[idx];
+           })
+
+      .def("is_zero", &sym::Expression<double, double>::isZero)
+      .def("is_constant", &sym::Expression<double, double>::isConstant)
+      .def("num_terms", &sym::Expression<double, double>::numTerms)
```
burgholzer: This is self contained and should be extracted into a separate file to create a better separation of concerns.
Taking inspiration from boost-histogram, this could be wrapped in a function `register_symbolic` in a similarly named file.

src/python/module.cpp
```diff
@@ -1,10 +1,765 @@
...
+                              double lhs) { return rhs / lhs; })
+
+      .def(
+          "__eq__",
+          [](const sym::Expression<double, double>& lhs,
+             const sym::Expression<double, double>& rhs) { return lhs == rhs; })
+      .def("__str__", [](const sym::Expression<double, double>& expr) {
+        std::stringstream ss;
+        ss << expr;
+        return ss.str();
```
burgholzer: Could also be extracted I guess.

src/python/module.cpp
burgholzer, line 1: I would really love to create a bit more separation of concerns here. Mainly inspired by how bindings are handled in https://github.com/scikit-hep/boost-histogram/tree/develop/src
This allows for faster recompiles and better locality of code.

For example, all Operations could be moved to a separate `_core.operations` module. And the respective functions can be bound in a separate file `register_operations`. Lots of inspiration for how to implement that transparently from the above link. Also how to expose the `_core.operations` module as `core.operations` and providing type stubs in a separate file.

The QuantumComputation bindings should also be moved to a separate file for registering. Maybe there should even be a separate submodule for that.

In general, I'd try to split up the code as much as feasible.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+from mqt.core._core import (
+    Control,
+    Expression,
+    OpType,
+    QuantumComputation,
+    StandardOperation,
+    SymbolicOperation,
+    Term,
+    Variable,
+)
```
burgholzer: I have recently learned that it's actually not that bad to use relative imports for library internal imports.
So this could/should probably be something like `from ..core import (`. You can see examples of this in the boost-histogram repository.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+        QuantumComputation: The converted circuit.
+    """
+    mqt_computation = QuantumComputation()
+
+    if qiskit_circuit.name is not None:
+        mqt_computation.name = qiskit_circuit.name
+
+    qubit_index = 0
+    qubit_map = {}
+    qiskit_qregs = qiskit_circuit.qregs
```
burgholzer: Isn't that just an unnecessary copy that could be inlined in the loop?

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for reg in qiskit_qregs:
+        size = reg.size
+        name = reg.name
+        if isinstance(reg, AncillaRegister):
+            mqt_computation.add_ancillary_register(size, name)
+            for i in range(size):
+                qubit_map[AncillaQubit(reg, i)] = qubit_index
+                qubit_index += 1
```
burgholzer: Couldn't you just iterate over the qubits in the register here? Same below.
Something like 
```
for anc_reg in reg:
```

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+        else:
+            mqt_computation.add_qubit_register(size, name)
+            for i in range(size):
+                qubit_map[Qubit(reg, i)] = qubit_index
+                qubit_index += 1
+
+    clbit_index = 0
+    clbit_map = {}
+    qiskit_cregs = qiskit_circuit.cregs
```
burgholzer: Also looks like an unnecessary copy.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for reg in qiskit_cregs:
+        size = reg.size
+        name = reg.name
+        mqt_computation.add_classical_bit_register(size, name)
+        for i in range(size):
+            clbit_map[Clbit(reg, i)] = clbit_index
+            clbit_index += 1
```
burgholzer: Same as above.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+    mqt_computation.gphase = qiskit_circuit.global_phase
+
+    data = qiskit_circuit.data
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for inst in data:
+        instruction = inst[0]
+        qargs = inst[1]
+        cargs = inst[2]
```
burgholzer: Wouldn't that be way cleaner?
```suggestion
    for instruction, qargs, cargs in data:
```

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+        params = instruction.params
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+    mqt_computation.initialize_io_mapping()
+    return mqt_computation
+
+
+def _emplace_operation(
+    mqt_computation: QuantumComputation,
+    instr: Instruction,
+    qargs: tuple[Qubit] | list[Qubit],
+    cargs: tuple[Clbit] | list[Clbit],
```
burgholzer: I'd prefer to use something like `Iterable[Qubit]` to annotate these types since it is way more generic.
Might have to look up the typing compatibility necessary for having that work under all Python versions.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+        float_parameters = [cast(float, parameter) for parameter in parameters]
+        mqt_computation.append_operation(
+            StandardOperation(mqt_computation.n_qubits, controls, target1, target2, type_, float_parameters)
+        )
+    else:
+        mqt_computation.append_operation(
+            SymbolicOperation(mqt_computation.n_qubits, controls, target1, target2, type_, parameters)
+        )
+        for parameter in parameters:
+            mqt_computation.add_variable(parameter)
```
burgholzer: This seems like quite the code duplication to the above function. There should be a way to eliminate that.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def _get_logical_qubit_indices(mqt_computation: QuantumComputation, layout: Layout) -> dict[Qubit, int]:
+    registers = layout.get_registers()
```
burgholzer: Can just be inlined in the loop below.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+    for register in registers:
+        if register.name == "ancilla":
+            ancilla_register = register
+            continue
+
+        size = register.size
+        for physical_qubit_index in range(size):
+            logical_qubit_indices[Qubit(register, physical_qubit_index)] = logical_qubit_index
+            logical_qubit_index += 1
```
burgholzer: As above, why not just iterate over the register itself?

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+    if ancilla_register is not None:
+        ancilla_size = ancilla_register.size
+        for physical_qubit_index in range(ancilla_size):
+            logical_qubit_indices[Qubit(ancilla_register, physical_qubit_index)] = logical_qubit_index
+            mqt_computation.set_logical_qubit_ancillary(logical_qubit_index)
+            logical_qubit_index += 1
```
burgholzer: Maybe not as easy to simplify but I guess there is also some potential here.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+    return logical_qubit_indices
+
+
+def _import_layouts(mqt_computation: QuantumComputation, qiskit_circuit: QuantumCircuit) -> None:
+    initial_layout = qiskit_circuit._layout.initial_layout  # noqa: SLF001
+    final_layout = qiskit_circuit._layout.final_layout  # noqa: SLF001
```
burgholzer: This is missing the compatibility fallback of the original implementation. Was this a decision made on purpose?
This implies that the minimal Qiskit version we support has to be raised quite a bit.
The `final_layout` was added only quite recently and there was a restructuring of the classes a couple versions prior to that.

That reminds me: this PR should also introduce a test with the minimally supported Qiskit version. See QCEC and QMAP for how to set that up as a nox session and in CI. (Note that this creates an additional coverage report and hence you need to bump the number of expected reports in the codecov Config)

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+
+    initial_logical_qubit_indices = _get_logical_qubit_indices(mqt_computation, initial_layout)
+    final_logical_qubit_indices = _get_logical_qubit_indices(mqt_computation, final_layout)
+
+    initial_physical_qubits = initial_layout.get_physical_bits()
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for physical_qubit, logical_qubit in initial_physical_qubits.items():
+        if logical_qubit in initial_logical_qubit_indices:
+            mqt_computation.initial_layout[physical_qubit] = initial_logical_qubit_indices[logical_qubit]
+
+    final_physical_qubits = final_layout.get_physical_bits()
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for physical_qubit, logical_qubit in final_physical_qubits.items():
+        if logical_qubit in final_logical_qubit_indices:
+            mqt_computation.output_permutation[physical_qubit] = final_logical_qubit_indices[logical_qubit]
+
+
+def _import_definition(
+    mqt_computation: QuantumComputation,
+    qiskit_circuit: QuantumCircuit,
+    qargs: tuple[int] | list[int],
+    cargs: tuple[int] | list[int],
```
burgholzer: As above. Prefer to use iterate in these cases.

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    qubit_map: dict[Qubit, int],
+    clbit_map: dict[Clbit, int],
+) -> None:
+    qarg_map = {}
+    def_qubits = qiskit_circuit.qubits
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for i in range(len(qargs)):
+        qarg_map[def_qubits[i]] = qargs[i]
+    carg_map = {}
+    def_clbits = qiskit_circuit.clbits
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for i in range(len(cargs)):
+        carg_map[def_clbits[i]] = cargs[i]
+
+    data = qiskit_circuit.data
```
burgholzer: Unnecessary copy

src/mqt/core/qiskit_utils/imports.py
```diff
@@ -0,0 +1,337 @@
...
+    for inst in data:
+        instruction = inst[0]
+
+        mapped_qargs = [qarg_map[qarg] for qarg in inst[1]]
+        mapped_cargs = [carg_map[carg] for carg in inst[2]]
```
burgholzer: Can be rewritten more compactly.

src/mqt/core/__init__.py
```diff
@@ -2,7 +2,24 @@
...
 
-from ._core import add
+from ._core import (
+    CompoundOperation,
+    Control,
+    Expression,
+    OpType,
+    Permutation,
+    QuantumComputation,
+    StandardOperation,
```
burgholzer: This doesn't seem to expose all relevant classes.

src/mqt/core/__init__.py
```diff
@@ -2,7 +2,24 @@
...
-__all__ = ["__version__", "add"]
+__all__ = [
+    "__version__",
+    "OpType",
+    "StandardOperation",
+    "CompoundOperation",
+    "Permutation",
+    "QuantumComputation",
+    "Control",
+    "Expression",
```
burgholzer: Nitpick: the order here is not consistent with the includes above.

src/mqt/core/_core/__init__.pyi
```diff
@@ -1 +1,733 @@
-def add(a: int, b: int) -> int: ...
+from typing import ClassVar, Iterator, overload
+
+class Operation:
+    def __init__(self, arg0: int, arg1: list[Operation], /) -> None: ...
```
burgholzer: What are these `/` all about? Doesn't really seem necessary.

src/mqt/core/_core/__init__.pyi
burgholzer, line 1: I regret to say this, but I do believe you have to regenerate these after adding documentation and argument names to the module.cpp file so that the argument names are right here and appear correctly in suggestions.

src/mqt/core/_core/__init__.pyi
```diff
@@ -1 +1,733 @@
...
+    def acts_on(self, arg: int, /) -> bool: ...
+    def add_depth_contribution(self, arg: list[int], /) -> None: ...
+    def clone(self) -> Operation: ...
+    @property
+    def controls(self) -> set[Control]: ...
+    @controls.setter
+    def controls(self, arg: set[Control]) -> None: ...
```
burgholzer: This seems kind of odd. There should be a better way to declare readwrite properties.

src/mqt/core/_core/__init__.pyi
```diff
@@ -1 +1,733 @@
...
+    def equals(self, arg0: Operation, arg1: Permutation, arg2: Permutation, /) -> bool: ...
+    def get_starting_qubit(self) -> int: ...
+    def get_used_qubits(self) -> set[int]: ...
+    def is_classic_controlled_operation(self) -> bool: ...
+    def is_compound_operation(self) -> bool: ...
+    def is_controlled(self) -> bool: ...
+    def is_non_unitary_operation(self) -> bool: ...
+    def is_standard_operation(self) -> bool: ...
+    def is_symbolic_operation(self) -> bool: ...
+    def is_unitary(self) -> bool: ...
```
burgholzer: Aren't some of these missing the override specifier?

src/mqt/core/_core/__init__.pyi
```diff
@@ -1 +1,733 @@
...
+    def to_open_qasm(self, qreg: list[tuple[str, str]], creg: list[tuple[str, str]], /) -> str: ...
+
+class StandardOperation(Operation):
+    @overload
+    def __init__(
+        self,
+        nq: int,
+        controls: set[Control],
+        target0: int,
+        target1: int,
```
burgholzer: This is surely missing the default argument. I haven't checked all functions right now, but please make sure that every function that has default arguments also has them in the stubs.

src/mqt/core/_core/__init__.pyi
```diff
@@ -1 +1,733 @@
...
+    @overload
+    def __init__(self, coeff: float, /) -> None: ...
+    @overload
+    def __init__(self, expr: list[Term]) -> None: ...
+    @property
+    def terms(self) -> list[Term]: ...
+    @property
+    def constant(self) -> float: ...
+    @constant.setter
+    def constant(self, value: float) -> None: ...
```
burgholzer: Seeing these kind of overloads here, I am really wondering whether it would be feasible to unify them to functions that accept, in this example, a `float` or a `Term`. But that might not be to easy to handle on the bindings side, right?
Maybe with lambdas and variants?
The above applies to many functions in throughout the bindings.

burgholzer (changes requested):
Many thanks for all the work here. This is great progress and already a huge step towards being ready to merge.
However, since this will become a core feature within the MQT and its realization sets a precedent for all other repositories, I'd really like to get this right the first time.
You can find many comments inline. I hope they help in improving the overall code.
I am afraid some of that will require some time, but I do believe that time is better spent now than spent later trying to catch up.
Please feel free to disagree and/or discuss some of the points I raised.

test/python/test_qiskit.py
pehamTom: I guess that would be worth a comment. The `XXMinusYYGate` has an implicit symbolic parameter if not explicitly overriden.

test/python/test_qiskit.py
pehamTom: Agreed

src/python/module.cpp
pehamTom: Leftover code. Removed.

src/python/module.cpp
pehamTom: Done

src/python/module.cpp
pehamTom: I was experimenting around with this a bit. Without changing the according methods in `QuantumComputation` I don't think there is a possibility.

src/python/module.cpp
pehamTom: Yeah that is one of the reasons why I reverted this PR to a draft. There are so many small things...

src/python/module.cpp
pehamTom: No matter what I tried, this doesn't work.

Moving anything but a `unique_ptr` to `Operation` doesn't even compile

Moving a `unique_ptr` compiles but crashes when invoking the method.

src/python/module.cpp
pehamTom: Are you asking whether the file formats are relevant and should be exposed to Python?

src/python/module.cpp
pehamTom: Then I would also need to bind the format enum. Not a problem necessarily but it seems odd to have an enum that is only used here. I can change it if you disagree.

src/python/module.cpp
pehamTom: Do you mean the `at` method?

src/python/module.cpp
pehamTom: It exposes this constructor: 

```
  explicit CompoundOperation(
      const std::size_t nq,
      std::vector<std::unique_ptr<Operation>>&& operations)
      : CompoundOperation(nq) {
    // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer)
    ops = std::move(operations);
  }
```

If I don't use an r-value reference nanobind won't compile the code

src/python/module.cpp
pehamTom: I guess you are referring to the lambda. Since `dumpOpenQASM` needs a `stringstream` argument, I can't really avoid the lambda.

test/python/test_qiskit.py
burgholzer: What?! 🙃 we did not make use of that parameter in the old implementation for importing Qiskit circuits. There is a default argument for that parameter if I am not mistaken.

src/python/module.cpp
burgholzer: I would be very open to change the base methods if you have any ideas on how to refactor them. I hate that they create so much boilerplate code and code duplication.

src/python/module.cpp
burgholzer: Hm. Let's keep this thread open for now. There has to be some way to accomplish that. I probably won't have much time to dig into that myself the coming days but would love to at some point.

src/python/module.cpp
burgholzer: I'll resolve this comment for now. It should have been applied to a different line.

src/python/module.cpp
burgholzer: Nah. Should have also been a different line. Resolving for now.

src/python/module.cpp
burgholzer: Nah. That was one of the commented out code lines above or below.

In the lines here I meant to say that this comment should be in a docstring instead of in the source code.

src/python/module.cpp
burgholzer: That comment was meant for the depth contribution functions.

src/python/module.cpp
burgholzer: That was a comment on the `size` member of some class

src/python/module.cpp
burgholzer: That was meant as a general comment on the permutation class binding

src/mqt/core/_core/__init__.pyi
pehamTom: One could always leave out the setter, but then the information is missing from the stub. This seems to be the way to do it.

src/mqt/core/_core/__init__.pyi
pehamTom: There is no @override in python, right? Or do you mean @overload? I don't know too much about annotations in Python but @overload seems to only be used to indicate actual method overloading.

src/mqt/core/_core/__init__.pyi
pehamTom: Should all be in there now

src/mqt/core/_core/__init__.pyi
pehamTom: I tried to stay close to the C++ implementation. Do you think much is gained by unifying them?

src/mqt/core/_core/__init__.pyi
pehamTom: Stubs should be much better now

test/python/test_qiskit.py
pehamTom: Fixed

src/python/module.cpp
pehamTom: Should be much better now

src/mqt/core/qiskit_utils/imports.py
pehamTom: That just copies a reference for clarity of name. This shouldn't be that bad.

src/mqt/core/qiskit_utils/imports.py
pehamTom: I mean I need the index. But I can also enumerate, I guess

src/mqt/core/qiskit_utils/imports.py
pehamTom: Same as above

src/mqt/core/qiskit_utils/imports.py
pehamTom: Same as above :smile:

src/mqt/core/qiskit_utils/imports.py
pehamTom: Same as above

src/mqt/core/qiskit_utils/imports.py
pehamTom: Same as above

src/mqt/core/qiskit_utils/__init__.py
pehamTom: None whatsoever

test/python/test_qiskit.py
pehamTom: I ported the previous C++ implementation. No ancillary registers are allocated for this gate. Also the qiskit circuit acts on 9 non-ancillary qubits as long as I don't give it any.

src/python/module.cpp
pehamTom: Removed

test/python/test_qiskit.py
pehamTom: Ajdusted in the new version

test/python/test_qiskit.py
pehamTom: Changed it

src/python/module.cpp
pehamTom: I think for the depth it makes sense. The others should only involve constant sized calls

src/python/module.cpp
pehamTom: I can't think of a way how. I actually don't think the boilerplate is too bad here. One could possibly do some template magic.

src/python/module.cpp
pehamTom: There are now docstrings for most of the methods. The ones without are self-explenatory

src/python/module.cpp
pehamTom: pybind doesn't relinquish ownership of pointers so we cannot move. One could pass a raw pointer and create a unique_ptr from that. But that would break the intention of the unique_ptr and python would still delete the raw pointer anyway. Unless the *entire* implementation is switched to shared_ptr there is no way around the copy here.

src/python/module.cpp
pehamTom: Removed it

src/python/module.cpp
pehamTom: With pybind I have no trouble compiling this.

src/python/module.cpp
pehamTom: Done.

src/python/module.cpp
pehamTom: Also in the new symbolic submodule

src/python/module.cpp
pehamTom: There are three separate submodules now

src/mqt/core/qiskit_utils/imports.py
pehamTom: Changed it. Actually has to be just `..`since `core` is the root module.

src/mqt/core/qiskit_utils/imports.py
pehamTom: Changed accordingly

src/mqt/core/qiskit_utils/imports.py
pehamTom: Iterable doesn't really work since I need to index stuff which is not an intrinsic property of an Iterable. It seems odd, but there doesn't seem to be a satisfactory solution except manually defining something like an "Indexable".

src/mqt/core/qiskit_utils/imports.py
pehamTom: The methods only differ in small parts but they are also not long. I don't think much is gained from refactoring. I think it is probably rather more confusing.

src/mqt/core/qiskit_utils/imports.py
pehamTom: removed

src/mqt/core/qiskit_utils/imports.py
pehamTom: I think i just need the index here

src/mqt/core/qiskit_utils/imports.py
pehamTom: Removed the additional size variable

src/mqt/core/qiskit_utils/imports.py
pehamTom: The comment says that this was changed after qiskit terra 0.22. But it seems that this has been deprecated for a longer time. Maybe I saw that wrong. But I don't think we have to support that anymore. Especially since I can't find where the information about the output permutation is stored.

src/mqt/core/qiskit_utils/imports.py
pehamTom: Since this actually involves a method call, I agree

src/mqt/core/qiskit_utils/imports.py
pehamTom: removed

src/mqt/core/qiskit_utils/imports.py
pehamTom: see above

src/mqt/core/qiskit_utils/imports.py
pehamTom: removed

src/mqt/core/qiskit_utils/imports.py
pehamTom: removed

src/mqt/core/qiskit_utils/imports.py
pehamTom: removed

src/mqt/core/qiskit_utils/imports.py
pehamTom: done

src/mqt/core/__init__.py
pehamTom: adjusted in new version

src/mqt/core/qiskit_utils/imports.py
burgholzer: There is. It is called `Sequence` ;-) My suggestion of `Iterable` was a little bit too generic.

src/mqt/core/__init__.py
pehamTom: Is there maybe a pre-commit hook that does stuff like that? Seems like a automatable consistency thing

src/mqt/core/_core/__init__.pyi
pehamTom: gone in new version

src/mqt/core/qiskit_utils/imports.py
burgholzer: > The comment says that this was changed after qiskit terra 0.22. But it seems that this has been deprecated for a longer time. Maybe I saw that wrong. But I don't think we have to support that anymore. Especially since I can't find where the information about the output permutation is stored.

Previously, the information about the output permutation did not exist in Qiskit.
The `TranspileLayout` class was introduced in `0.22.0`.
`final_layout` was only introduced in `0.23.0`.
Starting from `0.24.0`, there is no need to access the private `_layout` member anymore, as a `layout` attribute was added.

I am fine with raising the minimum supported version to `0.22.0`.
That still means there should be some conditional logic here!

src/mqt/core/__init__.py
burgholzer: I bet at some point there will be a ruff check for that 🙃

pyproject.toml
pehamTom: Done

pyproject.toml
pehamTom: Is now managed correctly

src/mqt/core/__init__.py
pehamTom: Ordered it.

src/mqt/core/qiskit_utils/imports.py
pehamTom: Since this comment is still unresolved. I didn't change to enumerate in the end since I really only need the index here

src/mqt/core/qiskit_utils/imports.py
pehamTom: Ah, seems I googled badly. Changed it.

src/mqt/core/qiskit_utils/imports.py
pehamTom: I see. Then let's go with the minimum supported version. I added the according check for the `layout` attribute.

test/python/test_qiskit.py
pehamTom: By the way, same as above. It gives the same number of qubits that qiskit gives.

test/python/test_qiskit.py
pehamTom: Could you elaborate a bit further? The issue here is that I am accessing a private member. I don't quite get where the self type comes in. Or is there a better way to set the layout attribute to begin with? The `layout` property itself is read only.

test/python/test_qiskit.py
pehamTom: Since qiskit is insistent that the default is a `ParameterExpression` I amended the test by explicitly passing a float.

extern/googletest
pehamTom: Bumped the right version.

extern/json
pehamTom: Bumped to the right version

extern/pybind11
pehamTom: This has to be removed in any case, right?

burgholzer: Yes. I am fine with doing so as part of this PR.
This forces consuming projects to update their packaging strategy when they want to update.
DDSIM, QCEC, QMAP, SyReC, QECC are all already updated. QuSAT may be the only one still left with the old packaging system.

test/python/test_qiskit.py
burgholzer: A seven-controlled MCX gate only acts on 8 qubits. IIRC the existing implementation has stripped away that one extra qubit since we can natively represent MCX gates.
Same goes for the `v-chain` mode, where we strip the `ncontrols -2` ancillary qubits to get a more compact circuit.
I can investigate this further during my review if you'd like.

test/python/test_qiskit.py
burgholzer: I am going to resolve this conversation as the one above should be sufficient to track the underlying "issue"

test/python/test_qiskit.py
burgholzer: May bad in this case. I was thinking about a different case where these `SLF` warnings come in.
I just very briefly skimmed the Qiskit code, but haven't really found a better way. One could probably "abuse" the Qiskit transpiler in some way by running some layout passes on the circuit.

src/python/module.cpp
burgholzer: Doesn't sound too promising 😢  i'll still give it a look during my review.

src/mqt/core/qiskit_utils/imports.py
burgholzer: Fine! 👍🏼

src/mqt/core/_core/__init__.pyi
burgholzer: Not too sure. I think as a dynamically typed language, Python just makes it easy to provide these unified functions and it seems a little cleaner to me. However, in the end, it doesn't really matter too much I guess.

pehamTom (commented):
Given the size of the PR and the scope of the changes, this looks good to me. I am probably going to be the first user of this when I try to adapt QMAP so I will take care of any issues that arise in the process. If there is nothing further to discuss, I would say this is ready to merge. And thanks for all the help and refactoring!

burgholzer (approved):
Alright. In that case, let's go.

Thanks for this great and important contribution!

PR #373, opened 2023-07-06
♻️ Unified DD memory management and streamlined complex number handling

burgholzer:
## Description

This PR serves many purposes that are heavily intertwined:
- There is lots of code duplication currently in the code when it comes to managing memory within the DD package. `UniqueTable`, `ComplexTable`, and `ComplexCache` all maintain essentially the same functionality for storing and providing chunks of memory. This PR introduces a generalized `MemoryManager` that factors out the common functionality and drastically reduces code duplication
- Several concepts in the DD package are intertwined although they argueably shouldn't be. The `ComplexTable` should act as a unique table, but also manages the memory (same for the `UniqueTable`). These concepts are disentangled by the `MemoryManager` mentioned above. The class for representing numbers (`CTEntry`) is hidden in the complex table. This couples both of these classes together and requires the complex table to be imported in many files where the functionality is absolutely not needed. For DD nodes, a clear separation between the Table that stores the nodes and the node class itself exists via the `Node` class and file. This PR adopts a similar strategy for the numbers to be stored as part of DDs. This simplifies several import chains and allows for things such as forward declarations of certain structs,
- Several parts of the complex number handling in the package have been historically misnamed. The `ComplexTable`, for example, could only ever store real numbers. This PR eliminates these misnomers and makes sure that things that represent real numbers are now also called that way and not misnamed as _complex_. 
- This PR adds several convenience functions to the `ComplexNumbers` class that allowed simplifying a significant portion of code throughout the DD package that handles complex numbers. These changes revealed quite some inefficiency across several functions in the DD package, which have been addressed.
- This PR also streamlines how reference counting is handled and distributes the respective code where appropriate. The unique table for numbers now also keeps track of its active entries. This brings both unique table data-structures closer together.

As of now, this is a WIP and not 100% finished, but I hope it will be soon.
@hillmich if you have any time on your hands over the next couple of days, I'd appreciate your opinion on some of these changes.

## Checklist:


cmake/CompilerOptions.cmake
```diff
@@ -13,7 +13,7 @@ function(enable_project_options target_name)
   endif()
 
   if(MSVC)
-    target_compile_options(${target_name} INTERFACE /utf-8)
+    target_compile_options(${target_name} INTERFACE /utf-8 /Zm10)
```
hillmich, line 16: Is there a reason to change the default amount of memory that MSVC allocates for precompiled headers?

include/dd/RealNumberUniqueTable.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * @brief The hash function for the hash table.
+   * @details The hash function for the table is a simple linear (clipped) hash
+   * function. The hash function is used to map floating point numbers to the
+   * buckets of the table.
+   * @param val The floating point number to hash. Must be non-negative.
+   * @returns The hash value of the floating point number.
+   */
+  static std::int64_t hash(fp val) noexcept;
```
hillmich, line 65: Wouldn't  an unsigned data type be more useful for this function? (We mostly use the hashes to access some offset in memory.)

hillmich (approved):
Thank you very much for your effort! There are only two minor questions/comments from my end.

cmake/CompilerOptions.cmake
burgholzer: It is one of the recommended options for reducing MSVC's `compiler is out of heap space` errors we (unfortunately) frequently hit in Debug mode builds. See https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1060?view=msvc-170
I am not 100% convinced it works reliably, but we'll see.

include/dd/RealNumberUniqueTable.hpp
burgholzer: At first, I was totally on board with your conclusion. Then I remembered that I actually implemented it like that back in the days because the signed version leads to more efficient assembly (especially when using gcc).
See https://godbolt.org/z/hbEWh4vsh.
So I'd rather keep it the way it is.

hillmich: Wow, that is a good catch! Would you mind adding this argument to the comments? (Feel free to resolve the conversation then.)

cmake/CompilerOptions.cmake
hillmich: Yeah, that seems more like a quick workaround than an actual fix, but it's recommended by Microsoft, so I guess we can stick with it as long as it works.

include/dd/RealNumberUniqueTable.hpp
burgholzer: Done in https://github.com/cda-tum/mqt-core/pull/373/commits/f28fb26086d0b49b6fedb5365f31c5ba2ce7d34d.
I hope the Godbolt link stays active for a while. Fortunately it's quite easy to replicate if it should become dead.

PR #381, opened 2023-07-15
♻️ Eliminate DD terminal nodes

### LINKED ISSUE #374: ✨ Eliminate global variables from DD package

The DD package contains a “fine collection” of (non-const) global variables. This includes the terminal nodes for DDs as well as the real number constants for 0., 1., 1./sqrt(2) and the complex number constants zero and one. All of these are defined as (non-const) static variables with global scope. This has a number of drawbacks:
- it is a bad coding practice to have non-const global variables (clang-tidy even warns about it).
- It leads to all kinds of problems with static initialization and the order of initialization (clang-tidy even warns about that)
- It was one of the main reasons why the MQT had problems building under Windows with MSVC. And took quite some while to get working.
- These “constant” entries are never really constant; they are mutable. Not only does this convey the wrong intentions, but it also increases the potential for introducing errors when these entries are erroneously modified.
- Due to these special entities actually being static instances of the respective class, their memory address differs across runs and, in the worst case (without proper care), between translation units.
- Due to these special entities being regular class instances, it is mostly assumed that any pointer member may just be dereferenced (as nullptr should never really occur). While this might seem beneficial at first, it promotes not using the member functions of the respective classes that safeguard the access and rather directly manipulating the dereferenced instances. Especially since we make use of unaligned pointers, this can be quite dangerous.
- Minor: all the data members of these special entities (such as ref count, next pointer, etc.) are never actually used. So it is kind of a waste to allocate them.

This list of undesirable behavior is arguably quite long. So what could be done about it?
Since we are heavily dealing with pointer types, there is one rather straight forward solution: replace these global (non-)constants with clever uses of nullptr.

For DD terminal nodes, that is fairly straightforward. Up until now, a terminal node was a pointer to a global static Node that had variable index -1. The successors of that node were never accessed and any check for a terminal either compared addresses with the pointer to the terminal node or checked the variable index. This kind of usage of the variable index necessitated a signed type for that variable and, hence, halved the maximal number of qubits that could be supported by the package from 255 to 128.
The notion of a terminal can simply be replaced by the nullptr without any apparent downsides. 
- Checking whether a node is a terminal remains a memory address comparison. It now, would just be a simple nullptr check. This should really be a one to one replacement.
- After that change, the data type of the variable index might be switched to an unsigned one, so that more qubits become available as part of the package. This might require some more changes in the code as I would suspect that there are quite some checks in the code base that have the form ˋif(p->v == -1)ˋ. These should be adjusted accordingly to benefit from the enhanced qubit range.

For the real number constants, the story is not so easy. The main problem here is that there is not a single, but three different constants that are being used (0., 1., and 1/sqrt(2)). Thus, these constants can’t simply all be replaced by using nullptr. Furthermore, the value of these entries are actually used in computations. However, there is a potential solution for that:
We already make use of the LSB of pointers to numbers to indicate that they are negative. Why not turn that up a notch and make use of the available scratch space in the lower bits of the pointers? (Due to alignment reasons, the lower three bits of any valid pointer are zero, which leaves 3 bits for hiding some stuff in pointers). The following hinges on one assumption: nullptr lives at address 0x0. Instead of creating global (non-)constants, the nullptr and its least significant three bits can be used to encode these special values:
- As before, the least significant bit encodes the sign of the number.
- 0x0 encodes 0.
- 0x2 encodes 1.
- 0x4 encodes 1/sqrt(2)
- This leaves room for one more number to encode. Here, it might make a lot of sense to use 0.5 (which is then encoded as 0x6). The number 0.5 is currently unconditionally added to the real number table and never collected. Might as well make it a constant.
- Checking for a particular constant remains as easy as before and is a simple pointer comparison.
- Checking wether a given pointer points to any constant is as simple as masking out the three least significant bits and comparing to 0x0.
- Getting the value for a constant number can be efficiently handled by a size 8 lookup table that is based on the three least significant bits of a pointer.
- The big question is: how high is the overhead of having to add an additional if condition in the access function to the value of a number (for checking whether the number is a constant). And, can this overhead sometimes be avoided if it is clear that none of the entries is a constant? This has to be benchmarked an investigated.

The good thing is: these two changes are independent from one another and can happen in parallel (or the complex number change might not happen at all). Even the change from a signed to an unsigned integer type for the variable index can happen separately.

burgholzer:
## Description

This PR completely eliminates the static DD node terminals and replaces them by using `nullptr` instead.
For one, this eliminates some global static variables.. Yay 😃
Furthermore, DD terminal nodes are no longer associated with the index `-1`. This allows changing the qubit index to an unsigned type and unlocks double the number of qubits (while also allowing to get rid of some `static_cast`s).
Finally, analysis has shown that the DD node size does not increase when using 16 bits for the qubit index (due to alignment). Hence, the package now directly supports up to `65536` qubits 🥳.

Note that this, naturally, is a breaking change. Qubit indices are now unsigned and checks for a terminal node should now use the respective functions of the Node classes instead of relying the variable index to be `-1`.
Something to look out for: Any decreasing loops that used a Qubit index for iteration and relied on `>=0` stopping conditions have to be revised since they now just wrap around at zero.

## Checklist:

### DIFF
--- a/include/dd/Package.hpp
+++ b/include/dd/Package.hpp
@@ -1782,6 +1782,14 @@ public:
   RightOperand
   multiply(const LeftOperand& x, const RightOperand& y, const Qubit start = 0,
            [[maybe_unused]] const bool generateDensityMatrix = false) {
+    static_assert(std::disjunction_v<std::is_same<LeftOperand, mEdge>,
+                                     std::is_same<LeftOperand, dEdge>>,
+                  "Left operand must be a matrix or density matrix");
+    static_assert(std::disjunction_v<std::is_same<RightOperand, vEdge>,
+                                     std::is_same<RightOperand, mEdge>,
+                                     std::is_same<RightOperand, dEdge>>,
+                  "Right operand must be a vector, matrix or density matrix");
+
     [[maybe_unused]] const auto before = cn.cacheCount();
 
     Qubit var{};
--- a/include/dd/Package.hpp
+++ b/include/dd/Package.hpp
@@ -200,9 +200,12 @@ public:
     cn.incRef(e.w);
     const auto& p = e.p;
     const auto inc = getUniqueTable<Node>().incRef(p);
-    if (inc && p->ref == 1U) {
-      for (const auto& child : p->e) {
-        incRef(child);
+    if (inc) {
+      assert(p != nullptr);
+      if (p->ref == 1U) {
+        for (const auto& child : p->e) {
+          incRef(child);
+        }
       }
     }
   }
@@ -221,9 +224,12 @@ public:
     cn.decRef(e.w);
     const auto& p = e.p;
     const auto dec = getUniqueTable<Node>().decRef(p);
-    if (dec && p->ref == 0U) {
-      for (const auto& child : p->e) {
-        decRef(child);
+    if (dec) {
+      assert(p != nullptr);
+      if (p->ref == 0U) {
+        for (const auto& child : p->e) {
+          decRef(child);
+        }
       }
     }
   }
@@ -1811,9 +1817,11 @@ public:
       dEdge::revertDmChangesToEdges(xCopy, yCopy);
     } else {
       if (!x.isTerminal()) {
+        assert(x.p != nullptr);
         var = x.p->v;
       }
       if (!y.isTerminal() && (y.p->v) > var) {
+        assert(y.p != nullptr);
         var = y.p->v;
       }
       e = multiply2(x, y, var, start);
--- a/include/dd/Package.hpp
+++ b/include/dd/Package.hpp
@@ -836,7 +836,8 @@ public:
     auto e = makeDDNode(z, em0);
 
     // process lines above the larger target (by creating identity structures)
-    for (++z; z < n + start; ++z) {
+    const auto end = static_cast<Qubit>(n + start);
+    for (++z; z < end; ++z) {
       e = makeDDNode(z, std::array{e, mEdge::zero, mEdge::zero, e});
     }
 
@@ -2609,7 +2610,7 @@ private:
     f = makeDDNode(f.p->v, edges);
 
     // something to reduce for this qubit
-    if (f.p->v >= 0 && ancillary[f.p->v]) {
+    if (ancillary[f.p->v]) {
       if (regular) {
         if (f.p->e[1].w != Complex::zero || f.p->e[3].w != Complex::zero) {
           f = makeDDNode(f.p->v, std::array{f.p->e[0], mEdge::zero, f.p->e[2],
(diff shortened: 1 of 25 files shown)


include/dd/UniqueTable.hpp
```diff
@@ -128,12 +134,9 @@ template <class Node, std::size_t NBUCKET = 32768> class UniqueTable {
...
     const auto inc = ::dd::incRef(p);
-    if (inc) {
-      assert(p != nullptr);
-      if (p->ref == 1U) {
-        stats.trackActiveEntry();
-        ++active[static_cast<std::size_t>(p->v)];
-      }
+    if (inc && p->ref == 1U) {
+      stats.trackActiveEntry();
+      ++active[p->v];
```
burgholzer:
```suggestion
    if (inc) {
      assert(p != nullptr);
      if (p->ref == 1U) {
        stats.trackActiveEntry();
        ++active[p->v];
      }
```

include/dd/UniqueTable.hpp
```diff
@@ -150,12 +153,9 @@ template <class Node, std::size_t NBUCKET = 32768> class UniqueTable {
...
     const auto dec = ::dd::decRef(p);
-    if (dec) {
-      assert(p != nullptr);
-      if (p->ref == 0U) {
-        --stats.activeEntryCount;
-        --active[static_cast<std::size_t>(p->v)];
-      }
+    if (dec && p->ref == 0U) {
+      --stats.activeEntryCount;
+      --active[p->v];
```
burgholzer:
```suggestion
    if (dec) {
      assert(p != nullptr);
      if (p->ref == 0U) {
        --stats.activeEntryCount;
        --active[p->v];
      }
```

include/dd/Package.hpp
```diff
@@ -205,9 +200,12 @@ template <class Config> class Package {
...
     const auto inc = getUniqueTable<Node>().incRef(p);
-    if (inc && p->ref == 1U) {
-      for (const auto& child : p->e) {
-        incRef(child);
+    if (inc) {
+      assert(p != nullptr);
+      if (p->ref == 1U) {
+        for (const auto& child : p->e) {
+          incRef(child);
+        }
```
```suggestion
    if (inc && p->ref == 1U) {
      for (const auto& child : p->e) {
        incRef(child);
```

include/dd/Package.hpp
```diff
@@ -226,9 +224,12 @@ template <class Config> class Package {
...
     const auto dec = getUniqueTable<Node>().decRef(p);
-    if (dec && p->ref == 0U) {
-      for (const auto& child : p->e) {
-        decRef(child);
+    if (dec) {
+      assert(p != nullptr);
+      if (p->ref == 0U) {
+        for (const auto& child : p->e) {
+          decRef(child);
+        }
```
```suggestion
    if (dec && p->ref == 0U) {
      for (const auto& child : p->e) {
        decRef(child);
```

hillmich (approved):
Thanks a lot for this effort! The increase in available qubits is really nice.

You correctly point out that decrementing loops that check for `>=0` require attention. Please wait a bit before merging so I can make this work correctly in the simulator :)

hillmich:
It took some effort, but the simulator works now :)

burgholzer:
> It took some effort, but the simulator works now :)

Nice. Thanks for your work on that!
I'll probably merge today or tomorrow 😌

PR #385, opened 2023-07-27
🐛 Fix DD node leak

### LINKED ISSUE #384: 🐛 DD Node Leak
### Environment information

Might date back to rather old versions of the DD package. Still trying to figure that out.

### Description

The following simple example (that has been reduced from a much bigger real use case) consistently leaks decicion diagram nodes
```c++
#include "gtest/gtest.h"

#include "dd/Package.hpp"

using namespace qc;

TEST(DebugTest, debug) {
  const auto nqubits = 1U;
  auto matrix = dd::GateMatrix {
      dd::complex_one,
      dd::complex_zero,
      dd::complex_zero,
      dd::complex_zero
  };
  auto dd = std::make_unique<dd::Package<>>(nqubits);

  auto dd1 = dd->makeGateDD(matrix, nqubits, 0U);
  matrix[0] = dd::complex_zero;
  matrix[3] = dd::complex_one;
  auto dd2 = dd->makeGateDD(matrix, nqubits, 0U);
  const auto repetitions = 10U;
  for (auto i=0U; i<repetitions; ++i) {
    dd->multiply(dd1, dd2);
  }
  dd->garbageCollect(true);
  EXPECT_EQ(dd->mMemoryManager.getUsedCount(), 0U);
}
```
Since no reference counting is performed, every DD node should be reclaimed in the end.
However, with the current code base, every repetition of the main multiplication leaks a matrix DD node.

A simple example where computations as the above frequently arise  is during the multiplication of CNOT gate DDs, e.g., `CNOT(0, 1) * CNOT(0, 1)`.

After some painful digging, I found the error here:
https://github.com/cda-tum/mqt-core/blob/4819bcd73d486ee11d9e42dd189831f81e366acf/include/dd/Package.hpp#L603-L611
If a cached node only has zero successors, then that part of the code will not free the node but happily return the zero terminal.
I still don't quite understand why that hasn't surfaced earlier and will dig into that a bit more.

### Expected behavior

The DD package should not leak memory.

### How to Reproduce

Execute the above test and watch it fail.

burgholzer:
## Description

Fixes #384 

## Checklist:

### DIFF
--- a/include/dd/Package.hpp
+++ b/include/dd/Package.hpp
@@ -602,9 +602,7 @@ public:
 
       // all equal to zero
       if (argmax == -1) {
-        if (!cached && !e.isTerminal()) {
-          // If it is not a cached computation, the node has to be put back into
-          // the chain
+        if (!e.isTerminal()) {
           getMemoryManager<Node>().returnEntry(e.p);
         }
         return Edge<Node>::zero;
--- a/test/dd/test_package.cpp
+++ b/test/dd/test_package.cpp
@@ -1825,3 +1825,25 @@ TEST(DDPackageTest, InnerProductTopNodeConjugation) {
   // it will result in +0.416.
   EXPECT_NEAR(dd->expectationValue(op, evolvedState), -0.416, 0.001);
 }
+
+/**
+ * @brief This is a regression test for a long lasting memory leak in the DD
+ * package.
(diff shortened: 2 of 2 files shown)


burgholzer:
clang-tidy install is currently failing due to https://github.com/llvm/llvm-project/issues/64120
Should resolve itself over time.

PR #400, opened 2023-08-16
♻️ [WIP] Remove redundant memory manager for complex numbers

### LINKED ISSUE #375: ✨ Potentially unify memory manager of number cache and table in DD package

At the moment, there are two memory managers for real numbers: one that manages the cache and one that provides entries for the respective unique table. Since both effectively provide the same type of entry, the question arises whether they can be unified to a single memory manager for numbers.

There are multiple sides to this:
- the access patterns of the cache and the unique table are rather different. For the cache, many entries are taken from it during the course of an operation and are returned after the operation (putting them on the (linked) list of entries available for reuse. On the other hand, the whole purpose of the unique table is to persist numbers and only release them as part of garbage collection.
- It seems wasteful to grow both memory managers, when one could simply use one. So this might reduce the overall memory footprint.
- Since the list of entries available for reuse is always checked first when requesting a new entry, combining the memory managers might increase the average speed of allocating new nodes since the cache usage puts many entries there. Of course, new entries have to be taken from the allocated chunks by the cache usage if many nodes are used as unique table entries, but that trade off might as well be worth it.
- There might actually be some potential for optimization in the lookup method: if a cached entry is looked up and is not already found in the unique table, then there would be no need to allocate a new entry and return the original one to the cache since the original entry could just be used and not returned. This might make it a little harder to check that the cache is maintained properly (there is no cache count before and cache count after any more). Most likely this can be resolved by keeping track of the number of entries in the unique table (which we should do anyway) and comparing that to the number of totally allocated numbers. Specifically, the cache count should be the number of total used numbers minus the total number of entries in the unique table.

burgholzer:
## Description

This is a draft PR that tries to evaluate whether unifying the memory managers for numbers in the DD package increases the performance of the package. See #375 for more details.
This required rather few changes overall to get working. However, as of now, it seems that it leads to worse performance. Preliminary tests on my workstation show runtimes of roughly 31s vs. 37s for the `mqt-core-test-dd` target on `main` vs. this branch. I am fairly sure this can be improved if one carefully benchmarks/profiles the code.

#375 also hints at a possible performance improvement when looking up cached numbers where returning the number and allocating a new number could be skipped completely. This is not implemented yet and I do not have the time for it at the moment. If someone is up for it, please go ahead 😎 

Closes #375

## Checklist:

### DIFF
--- a/include/dd/Package.hpp
+++ b/include/dd/Package.hpp
@@ -104,12 +104,6 @@ public:
    * separately. Hence, it suffices for the manager to only manage real numbers.
    */
   MemoryManager<RealNumber> cMemoryManager{};
-  /**
-   * @brief The cache manager for complex numbers
-   * @note Similar to the memory manager, the cache only maintains real entries,
-   * but typically gives them out in pairs to form complex numbers.
-   */
-  MemoryManager<RealNumber> cCacheManager{};
 
   /**
    * @brief Get the memory manager for a given type
@@ -139,7 +133,6 @@ public:
     mMemoryManager.reset(resizeToTotal);
     dMemoryManager.reset(resizeToTotal);
     cMemoryManager.reset(resizeToTotal);
-    cCacheManager.reset(resizeToTotal);
   }
 
   /// The unique table used for vector nodes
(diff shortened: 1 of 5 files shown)


burgholzer:
Closing in favour of #444 which completely removes the need for the cache.

PR #411, opened 2023-09-02
Add `addControl`, `clearControls`, `removeControl` functions on Operation

martin-fink:
## Description

This PR replaces the `setControls` function on `Operation` with `addControls` and `clearControls`.
Calling `clearControls` on `CompoundOperation`s will throw an Exception, since doing so is not possible at the moment (we don't know if a control was pushed down from a `CompoundOperation` or was there from the beginning.

## Checklist:

### DIFF
--- a/include/operations/CompoundOperation.hpp
+++ b/include/operations/CompoundOperation.hpp
@@ -54,33 +54,31 @@ public:
     });
   }
 
-  void addControls(const Controls& c) override {
-    for (auto ctrl : c) {
-      controls.insert(ctrl);
-    }
+  void addControl(const Control c) override {
+    controls.insert(c);
     // we can just add the controls to each operation, as the operations will
     // check if they already act on the control qubits.
     for (auto& op : ops) {
-      op->addControls(c);
+      op->addControl(c);
     }
   }
 
   void clearControls() override {
     // we remove just our controls from nested operations
+    // We need to copy controls here as we will modify the set while iterating
     removeControls(Controls{controls});
   }
 
-  void removeControls(const Controls& c) override {
+  void removeControl(const Control c) override {
     // first we iterate over our controls and check if we are actually allowed
     // to remove them
-    for (const auto& ctrl : c) {
-      if (controls.erase(ctrl) == 0) {
-        throw QFRException("Cannot remove control from compound operation.");
-      }
+    if (controls.erase(c) == 0) {
+      throw QFRException("Cannot remove control from compound operation as it "
+                         "is not a control.");
     }
 
     for (auto& op : ops) {
-      op->removeControls(c);
+      op->removeControl(c);
     }
   }
 
--- a/include/operations/CompoundOperation.hpp
+++ b/include/operations/CompoundOperation.hpp
@@ -54,6 +54,36 @@ public:
     });
   }
 
+  void addControls(const Controls& c) override {
+    for (auto ctrl : c) {
+      controls.insert(ctrl);
+    }
+    // we can just add the controls to each operation, as the operations will
+    // check if they already act on the control qubits.
+    for (auto& op : ops) {
+      op->addControls(c);
+    }
+  }
+
+  void clearControls() override {
+    // we remove just our controls from nested operations
+    removeControls(Controls{controls});
+  }
+
+  void removeControls(const Controls& c) override {
+    // first we iterate over our controls and check if we are actually allowed
+    // to remove them
+    for (const auto& ctrl : c) {
+      if (controls.erase(ctrl) == 0) {
+        throw QFRException("Cannot remove control from compound operation.");
+      }
+    }
+
+    for (auto& op : ops) {
+      op->removeControls(c);
+    }
+  }
+
   [[nodiscard]] bool equals(const Operation& op, const Permutation& perm1,
                             const Permutation& perm2) const override {
     if (const auto* comp = dynamic_cast<const CompoundOperation*>(&op)) {
--- a/include/operations/CompoundOperation.hpp
+++ b/include/operations/CompoundOperation.hpp
@@ -65,8 +65,7 @@ public:
 
   void clearControls() override {
     // we remove just our controls from nested operations
-    // We need to copy controls here as we will modify the set while iterating
-    removeControls(Controls{controls});
+    removeControls(controls);
   }
 
   void removeControl(const Control c) override {
@@ -80,6 +79,14 @@ public:
     for (auto& op : ops) {
       op->removeControl(c);
     }
+  }
+
+  Controls::iterator removeControl(const Controls::iterator it) override {
+    for (auto& op : ops) {
+      op->removeControl(*it);
+    }
+
+    return controls.erase(it);
   }
 
   [[nodiscard]] bool equals(const Operation& op, const Permutation& perm1,
--- a/include/operations/NonUnitaryOperation.hpp
+++ b/include/operations/NonUnitaryOperation.hpp
@@ -40,6 +40,12 @@ public:
   std::vector<Bit>& getClassics() { return classics; }
   [[nodiscard]] size_t getNclassics() const { return classics.size(); }
 
+  Controls& getControls() const override {
+    throw QFRException("Cannot get controls from non-unitary operation.");
+  }
+
+  void addDepthContribution(std::vector<std::size_t>& depths) const override;
+
   void addControl(const Control /*c*/) override {
     throw QFRException("Cannot add control to non-unitary operation.");
   }
--- a/include/operations/NonUnitaryOperation.hpp
+++ b/include/operations/NonUnitaryOperation.hpp
@@ -48,7 +48,6 @@ public:
     throw QFRException("Cannot clear controls from non-unitary operation.");
   }
 
-
   void removeControls(const Controls& /*c*/) override {
     throw QFRException("Cannot remove controls from non-unitary operation.");
   }
--- a/include/operations/NonUnitaryOperation.hpp
+++ b/include/operations/NonUnitaryOperation.hpp
@@ -40,7 +40,7 @@ public:
   std::vector<Bit>& getClassics() { return classics; }
   [[nodiscard]] size_t getNclassics() const { return classics.size(); }
 
-  Controls& getControls() const override {
+  [[nodiscard]] Controls& getControls() const override {
     throw QFRException("Cannot get controls from non-unitary operation.");
   }
 
--- a/include/operations/NonUnitaryOperation.hpp
+++ b/include/operations/NonUnitaryOperation.hpp
@@ -58,6 +58,10 @@ public:
     throw QFRException("Cannot remove controls from non-unitary operation.");
   }
 
+  Controls::iterator removeControl(const Controls::iterator /*it*/) override {
+    throw QFRException("Cannot remove controls from non-unitary operation.");
+  }
+
   [[nodiscard]] bool equals(const Operation& op, const Permutation& perm1,
                             const Permutation& perm2) const override;
   [[nodiscard]] bool equals(const Operation& operation) const override {
--- a/include/operations/NonUnitaryOperation.hpp
+++ b/include/operations/NonUnitaryOperation.hpp
@@ -40,6 +40,11 @@ public:
   std::vector<Bit>& getClassics() { return classics; }
   [[nodiscard]] size_t getNclassics() const { return classics.size(); }
 
+  [[nodiscard]] std::set<Qubit> getUsedQubits() const override {
+    const auto& opTargets = getTargets();
+    return {opTargets.begin(), opTargets.end()};
+  }
+
   [[nodiscard]] Controls& getControls() const override {
     throw QFRException("Cannot get controls from non-unitary operation.");
   }
--- a/include/operations/NonUnitaryOperation.hpp
+++ b/include/operations/NonUnitaryOperation.hpp
@@ -40,6 +40,19 @@ public:
   std::vector<Bit>& getClassics() { return classics; }
   [[nodiscard]] size_t getNclassics() const { return classics.size(); }
 
+  void addControls(const Controls& /*c*/) override {
+    throw QFRException("Cannot add controls to non-unitary operation.");
+  }
+
+  void clearControls() override {
+    throw QFRException("Cannot clear controls from non-unitary operation.");
+  }
+
+
+  void removeControls(const Controls& /*c*/) override {
+    throw QFRException("Cannot remove controls from non-unitary operation.");
+  }
+
   [[nodiscard]] bool equals(const Operation& op, const Permutation& perm1,
                             const Permutation& perm2) const override;
   [[nodiscard]] bool equals(const Operation& operation) const override {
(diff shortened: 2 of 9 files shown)


martin-fink:
This factors out the functionality previously implemented in #309 to the `Operation` class.

include/operations/CompoundOperation.hpp
```diff
@@ -40,6 +40,14 @@ class CompoundOperation final : public Operation {
     }
   }
 
+  void setControls(const Controls& c) override {
+    for (auto& op : ops) {
+      auto existingControls = op->getControls();
+      existingControls.insert(c.begin(), c.end());
```
burgholzer:
```suggestion
      existingControls.emplace(c.cbegin(), c.cend());
```
Should also work here and be slightly more efficient.

burgholzer (changes requested):
Many thanks for the PR.

Can you maybe split this up even further? 
I think both concepts are rather independent and still require some work on their own.

As for the control-part, I feel that you could take this a step further. Consider the following as a rough checklist
- [ ] I think the current concept of how this is implemented does not work in the general case. Imagine the following case
```c++
auto op = CompoundOperation{3};
op.emplace_back<StandardOperation>(3, 0, OpType::X);
op.setControls({1_pc});
op.setControls({2_pc});
```
The resulting gate would have two controls instead of just one. I believe it will become quite complicated to fix that general scenario. You would probably use the `controls` member of the CompoundOperation to keep track of the controls and modify the methods that work on CompoundOperations to additionally factor those controls in. That's a rather big, non-local change though that I would like to avoid for now.
Depending on how you feel about it, I would maybe replace the current version of the new function with one that throws an error indicating that this is not handled currently and might lead to surprises.
- [ ] If not already, a similar warning should probably be raised in the `NonUnitaryOperatoin` class and you should carefully check whether this is properly handled in the `ClassicControlledOperation` class.
- [ ] What you might want to do instead of the above is to add `addControl(const Control c)` and `addControls(const Controls& controls)` functions to the `Operation` class and _all_ derived classes. The second method should just reuse the first one to avoid code duplication. This more closely resembles the semantics of the `control` modifier from OQ3. The `addControl` method should definitely contain some sanity checking, that none of the new controls is already a control or target of any existing gate. You can check that by running over the new controls `c` and calling `op->actsOn(c)`. An error should be thrown in that case. Again, I would argue that these functions should return an error for `NonUnitaryOperation`s as that concept does not really make sense there.
- [ ] It would be great to test these various scenarios and make sure that everything works properly. This includes testing all different Operation types that should work and those that should not, including scenarios where the control to be added is already part of the gate.


As for the `inverse` method, the current implementation is lacking quite some functionality. Again, a small checklist (you might want to use as starter for that second PR):
- [ ] The inverse of an operation is not simply derived by changing the size of the parameters. Instead, there are some operations that are self-inverse (e.g., I, X, Y, Z, H),  some (non-parametrised) gates where the inverse is another gate we support (e.g., S, T, SX), and some where the inverse can be derived from some parameter changes (e.g., any rotation gate). This mapping needs to be provided somewhere and I believe the `Operation` class is the wrong place for that. I would argue that the base class method should be abstract (`=0`)
- [ ] In the `StandardOperation` class, there probably needs to be a rather long switch-case per operation (or type of operation) that handles inverting the individual operations. You should be able to group quite some of them together to reduce the number of switch cases that actually need an implementation. You should be able to derive these from https://github.com/cda-tum/mqt-core/blob/main/include/dd/Operations.hpp. In the best case, some of the potential code duplication between that file and the newly introduced `inverse` method can be reduced.
- [ ] In the `SymbolicOperation` class, there probably needs to be a little specialisation for parametrised gates that takes inspiration from the `StandardOperation` class
- [ ] I think you already nailed the `invert` operation for the `CompoundOperation`. You might want to add a similar method to the `QuantumComputation` class though.
- [ ] In the `NonUnitaryOperation` class, an error should be thrown
- [ ] In the `ClassicControlledOperation` class, you probably just want to defer to the underlying operation's inverse method
- [ ] All of these scenarios should be tested to ensure that they work properly.

I hope this gives you some guidelines on how to proceed.

martin-fink:
Please let me know if I should move the `inverse` method to a separate PR! :)

burgholzer:
> Please let me know if I should move the `inverse` method to a separate PR! :)

Hey 👋🏻 
Yeah. That would be great. In the end, they are separate features. Don't forget to update the PR descriptions and titles in that case.

include/operations/NonUnitaryOperation.hpp
```diff
@@ -40,6 +40,18 @@ class NonUnitaryOperation final : public Operation {
...
+    throw QFRException("Cannot add controls to non-unitary operation.");
+  }
+
+  void clearControls() override {
+    throw QFRException("Cannot clear controls from non-unitary operation.");
+  }
+
+  void removeControls(const Controls& /*c*/) override {
+    throw QFRException("Cannot remove controls from non-unitary operation.");
+  }
```
burgholzer: Would probably also make sense for consistency reasons to override the `getControls()` method to throw an exception here, right?

include/operations/Operation.hpp
```diff
@@ -88,7 +88,20 @@ class Operation {
...
 
-  virtual void setControls(const Controls& c) { controls = c; }
+  virtual void setControls(const Controls& c) {
+    clearControls();
+    addControls(c);
+  }
+
+  virtual void addControl(const Control c) { addControls({c}); }
+
+  virtual void addControls(const Controls& c) = 0;
```
burgholzer: It's a minor detail, but I would have rather expected the `addControl` method to be abstract and the `addControls` method to just be
```c++
for (const auto& control: c) {
  addControl(control);
}
```
Kind of feels more natural to be to describe how a single control is added and then just reusing that method for the multiple control case. That should also safe some code duplication of the loops over controls.

include/operations/Operation.hpp
```diff
@@ -88,7 +88,20 @@ class Operation {
...
+
+  virtual void clearControls() = 0;
+
+  virtual void removeControl(const Control c) { removeControls({c}); }
+
+  virtual void removeControls(const Controls& c) = 0;
```
burgholzer: Same as above applies here.

include/operations/SymbolicOperation.hpp
```diff
@@ -125,6 +125,26 @@ class SymbolicOperation final : public StandardOperation {
...
+  }
+
+  void clearControls() override { controls.clear(); }
+
+  void removeControls(const Controls& c) override {
+    for (auto ctrl : c) {
+      controls.erase(ctrl);
+    }
+  }
+
```
```suggestion
```
I am fairly sure this can just be removed as the `SymbolicOperation` class inherits from the `StandardOperation` class and, hence, inherits all these methods directly.

include/operations/StandardOperation.hpp
```diff
@@ -72,6 +72,26 @@ class StandardOperation : public Operation {
...
 
+  void addControls(const Controls& c) override {
+    for (auto ctrl : c) {
+      if (actsOn(ctrl.qubit)) {
+        throw QFRException(
+            "Cannot add control to operation as it already acts on "
+            "the control qubit.");
+      }
+
+      controls.insert(ctrl);
```
burgholzer: Again, a minor detail that probably doesn't make the biggest difference here, but I believe the following should be preferred in almost any case
```suggestion
      controls.emplace(ctrl);
```

include/operations/StandardOperation.hpp
```diff
@@ -72,6 +72,26 @@ class StandardOperation : public Operation {
...
+    }
+  }
+
+  void clearControls() override { controls.clear(); }
+
+  void removeControls(const Controls& c) override {
+    for (auto ctrl : c) {
+      controls.erase(ctrl);
+    }
+  }
```
burgholzer: This method should probably also contain a sanity check that is the inverse of the `addControl` case, i.e., it should throw an exception if the gate does not act on the qubit to be removed. Even more specific, the qubit to be removed should be an existing control.

include/operations/CompoundOperation.hpp
```diff
@@ -54,6 +54,36 @@ class CompoundOperation final : public Operation {
...
+    // we can just add the controls to each operation, as the operations will
+    // check if they already act on the control qubits.
+    for (auto& op : ops) {
+      op->addControls(c);
+    }
+  }
+
+  void clearControls() override {
+    // we remove just our controls from nested operations
+    removeControls(Controls{controls});
```
burgholzer:
```suggestion
    removeControls(controls);
```
That should work as well, right? No need to construct a new object here.

test/unittests/test_qfr_functionality.cpp
```diff
@@ -1937,3 +1937,133 @@ TEST_F(QFRFunctionality, dumpAndImportTeleportation) {
...
+  ASSERT_EQ(op.getNcontrols(), 2);
+  auto expectedControls =
+      Controls{Control{1, Control::Type::Pos}, Control{2, Control::Type::Pos}};
+  EXPECT_EQ(op.getControls(), expectedControls);
+  op.removeControl(Control{1, Control::Type::Pos});
+  auto expectedControlsAfterRemove = Controls{Control{2, Control::Type::Pos}};
+  EXPECT_EQ(op.getControls(), expectedControlsAfterRemove);
+  op.clearControls();
+  EXPECT_EQ(op.getNcontrols(), 0);
+}
```
burgholzer: This test (and also the ones below) can be simplified a little to be less verbose.
Also, if you want to add 2 controls to a single-qubit operation, the operation has to be allocated with 3 qubits. It's kind of unfortunate that we actually don't throw an exception in case you try to use more qubits. However, just as a side note, I expect the `nqubits` argument of the Operation class to go away in the near term future as it is only a technical dept of our decision diagram package.
```suggestion
TEST_F(QFRFunctionality, addControlStandardOperation) {
  auto op = StandardOperation(3, 0, OpType::X);

  op.addControl(1_pc);
  op.addControl(2_pc});

  ASSERT_EQ(op.getNcontrols(), 2);
  const auto expectedControls = Controls{1_pc, 2_pc};
  EXPECT_EQ(op.getControls(), expectedControls);
  op.removeControl(1_pc);
  const auto expectedControlsAfterRemove = Controls{2_pc};
  EXPECT_EQ(op.getControls(), expectedControlsAfterRemove);
  op.clearControls();
  EXPECT_EQ(op.getNcontrols(), 0);
}
```

burgholzer (changes requested):
Many thanks for this clean PR. I just went through everything and I only have minor feedback left that should be very easy to incorporate.

Small side note: You don't necessarily need to squash your commits and force push for your PRs. If PRs are small enough (such as this one), we simplify squash all commits upon merging.

include/operations/CompoundOperation.hpp
martin-fink: This is needed as otherwise the iterator in `removeControls` will be modified while we iterate through it.

martin-fink: I should probably add a comment in there why I'm doing this :)

include/operations/NonUnitaryOperation.hpp
martin-fink: I think yes, but this breaks one test (`CircuitDepthNonUnitaryOperations`, it calls `getControls` from the `Operation::addDepthContribution` function). I guess this can be fixed there by explicitly checking if an operation may have controls, but it might be cleaner to leave it as is. What do you think?

burgholzer: I would rather argue that the proper way to fix this is to override the `addDepthContribution` method for the `NonUnitaryOperation` class so that it only ever iterates over the targets.
That should make the test work again, right?

include/operations/CompoundOperation.hpp
burgholzer: Ah, that makes sense.
Although it is a only a very small cost, you could technically avoid that cost by provifing an overload for `removeControl` that takes an iterator to a control and returns an iterator.
That way you could replicate the behavior of `erase`, which returns a valid iterator to the element after the last removed element (see https://en.cppreference.com/w/cpp/container/set/erase).
Overall, iterator validity is an important concern though. This also directly relates to my suggestion of making the `removeControl` function abstract and simply calling it repeatedly in the `removeControls` function.

include/operations/StandardOperation.hpp
```diff
@@ -72,6 +72,29 @@ class StandardOperation : public Operation {
 
   [[nodiscard]] bool isStandardOperation() const override { return true; }
 
+  void addControl(const Control c) override {
+    if (actsOn(c.qubit)) {
+      throw QFRException(
+          "Cannot add control to operation as it already acts on "
+          "the control qubit.");
```
burgholzer: I seem to be in a nitpicky mood today, but I believe this error message could be a little more informative.
In the sense that it could mention the control that it aims to add.

include/operations/StandardOperation.hpp
```diff
@@ -72,6 +72,29 @@ class StandardOperation : public Operation {
...
+
+    controls.emplace(c);
+  }
+
+  void clearControls() override { controls.clear(); }
+
+  void removeControl(const Control c) override {
+    if (controls.erase(c) == 0) {
+      throw QFRException("Cannot remove control from operation as it is not a "
+                         "control.");
```
burgholzer: Same as above. A slightly more informative error message could be beneficial.

src/operations/NonUnitaryOperation.cpp
```diff
@@ -191,4 +191,20 @@ void NonUnitaryOperation::printReset(std::ostream& os,
     }
   }
 }
+
+void NonUnitaryOperation::addDepthContribution(
+    std::vector<std::size_t>& depths) const {
+  if (type == Barrier) {
+    return;
+  }
+
```
burgholzer:
```suggestion
```
Barriers are no longer considered non-unitary operations since #394, so this part can be removed.

src/operations/NonUnitaryOperation.cpp
```diff
@@ -191,4 +191,20 @@ void NonUnitaryOperation::printReset(std::ostream& os,
...
+  std::size_t maxDepth = 0;
+  for (const auto& target : getTargets()) {
+    maxDepth = std::max(maxDepth, depths[target]);
+  }
+  maxDepth += 1;
+  for (const auto& target : getTargets()) {
+    depths[target] = maxDepth;
+  }
```
burgholzer: I would argue that this is probably not really what one would expect here.
Even if a measurement or reset operation has multiple targets, each individual one is independent from the others; so it's just for convenience and compactness reasons that multiple measurements/resets might be combined in a single operation. as such, this should probably be

```suggestion
  for (const auto& target : getTargets()) {
    depths[target] += 1;
  }
```

burgholzer (changes requested):
Thanks for the changes. One last round of feedback. Then, this should be good to merge.

Edit: It would be nice to get the patch coverage of the changes in this PR above the mark so that the CI check turns green ✅

burgholzer (approved):
Many thanks! Looks great now ✅

martin-fink:
Thank you for taking the time for the reviews!

PR #414, opened 2023-09-06
set a new func for TQDepth

avelsh:
add a new feuture to count the TQDepth for TwoQubitEncoding

### DIFF
--- a/include/operations/Operation.hpp
+++ b/include/operations/Operation.hpp
@@ -135,6 +135,7 @@ public:
   }
 
   virtual void addDepthContribution(std::vector<std::size_t>& depths) const;
+  virtual void addTQDepthContribution(std::vector<std::size_t>& depths) const;
 
   [[nodiscard]] virtual bool equals(const Operation& op,
                                     const Permutation& perm1,
--- a/include/operations/Operation.hpp
+++ b/include/operations/Operation.hpp
@@ -135,7 +135,8 @@ public:
   }
 
   virtual void addDepthContribution(std::vector<std::size_t>& depths) const;
-  virtual void addTwoQubitDepthContribution(std::vector<std::size_t>& depths) const;
+  virtual void
+  addTwoQubitDepthContribution(std::vector<std::size_t>& depths) const;
 
   [[nodiscard]] virtual bool equals(const Operation& op,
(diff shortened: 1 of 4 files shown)


src/operations/Operation.cpp
```diff
@@ -249,4 +249,29 @@ void Operation::addDepthContribution(std::vector<std::size_t>& depths) const {
...
+
+  std::size_t maxDepth = 0;
+  for (const auto& target : getTargets()) {
+    maxDepth = std::max(maxDepth, depths[target]);
+  }
+  for (const auto& control : getControls()) {
+    maxDepth = std::max(maxDepth, depths[control.qubit]);
+  }
+
+  if (!getControls().empty()) {
```
pehamTom, line 266: There are two-qubit gates that do not have a control (e.g. SWAP gate) but are still two-qubit gates.

src/operations/Operation.cpp
```diff
@@ -249,4 +249,29 @@ void Operation::addDepthContribution(std::vector<std::size_t>& depths) const {
   }
 }
 
+void Operation::addTwoQubitDepthContribution(
```
pehamTom, line 252: This doesn't just count two-qubit gates but any gate with a control. So a multi-controlled x-gate would still be considered a two-qubit gate in this case. I actually don't think this is a huge problem but the feature and the methods should be renamed accordingly (i.e. multiQubitGateDepth or something).

include/operations/Operation.hpp
```diff
@@ -135,6 +135,8 @@ class Operation {
   }
 
   virtual void addDepthContribution(std::vector<std::size_t>& depths) const;
+  virtual void
+  addTwoQubitDepthContribution(std::vector<std::size_t>& depths) const;
```
pehamTom, line 139: This feature doesn't work correctly for a `CompoundOperation`. For these the method would have to be overridden so as to recursively compute the depth.

pehamTom (changes requested):
Thank you for this feature. In addition to the changes below, please add some test cases to check if everything works correctly.

burgholzer:
Closing this for now, as no one seems to be actively working on it.
Feel free to re-open if this changes at any point in time.

PR #415, opened 2023-09-06
Fix makeDDFromMatrix Issue in Asymmetrical Matrices

rahimiparham:
## Description

The function ```makeDDFromMatrix``` incorrectly swaps the pointers for the second and third nodes in level 1 (2x2) input submatrices. This leads to a mismatch between the input and reconstructed matrices if the input matrix is not symmetrical. All the existing tests of this function don't contain a matrix where this problem would be visible. This PR contains two commits: The first one adds a new test to reproduce and show the problem, and the second one is the actual fix.

## Checklist:


burgholzer (approved):
Thanks for the nice and concise PR description and code. LGTM 👍🏼

PR #416, opened 2023-09-07
Add invert functionality for quantum operations

martin-fink:
## Description

Add an `invert` function to `Operation`. This mirrors the functionality of the [OpenQASM 3.0 `inv` modifier](https://openqasm.com/language/gates.html#inverse-modifier).

For Compound gates, this calls `invert` on each nested operation and reverses the order of the operations.
For Standard gates, this checks which type of gate is being inverted and handles each case.

## Checklist:

### DIFF
--- a/include/operations/Operation.hpp
+++ b/include/operations/Operation.hpp
@@ -83,6 +83,12 @@ public:
     return usedQubits;
   }
 
+  [[nodiscard]] std::unique_ptr<Operation> getInverted() const {
+    auto op = clone();
+    op->invert();
+    return op;
+  }
+
   // Setter
   virtual void setNqubits(const std::size_t nq) { nqubits = nq; }
 
@@ -177,5 +183,7 @@ public:
 
   virtual void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
                             const RegisterNames& creg) const = 0;
+
+  virtual void invert() = 0;
 };
 } // namespace qc
--- a/src/operations/StandardOperation.cpp
+++ b/src/operations/StandardOperation.cpp
@@ -467,4 +467,95 @@ void StandardOperation::dumpOpenQASMTeleportation(
   of << "teleport " << qreg[targets[0]].second << ", "
      << qreg[targets[1]].second << ", " << qreg[targets[2]].second << ";\n";
 }
+
+void StandardOperation::invert() {
+  switch (type) {
+  // self-inverting gates
+  case I:
+  case X:
+  case Y:
+  case Z:
+  case H:
+  case SWAP:
+  case ECR:
+  case Barrier:
+    break;
+  // gates where we just update parameters
+  case GPhase:
+  case Phase:
+  case RX:
+  case RY:
+  case RZ:
+  case RXX:
+  case RYY:
+  case RZZ:
+  case RZX:
+    parameter[0] = -parameter[0];
+    break;
+  case U2:
+    std::swap(parameter[0], parameter[1]);
+    parameter[0] = -parameter[0] + PI;
+    parameter[1] = -parameter[1] - PI;
+    break;
+  case U3:
+    parameter[0] = -parameter[0];
+    parameter[1] = -parameter[1];
+    parameter[2] = -parameter[2];
+    std::swap(parameter[1], parameter[2]);
+    break;
+  case XXminusYY:
+  case XXplusYY:
+    parameter[0] = -parameter[0];
+    break;
+  case DCX:
+    std::swap(targets[0], targets[1]);
+    break;
+  // gates where we have specialized inverted operation types
+  case S:
+    type = Sdag;
+    break;
+  case Sdag:
+    type = S;
+    break;
+  case T:
+    type = Tdag;
+    break;
+  case Tdag:
+    type = T;
+    break;
+  case V:
+    type = Vdag;
+    break;
+  case Vdag:
+    type = V;
+    break;
+  case SX:
+    type = SXdag;
+    break;
+  case SXdag:
+    type = SX;
+    break;
+  case Peres:
+    type = Peresdag;
+    break;
+  case Peresdag:
+    type = Peres;
+    break;
+  // Tracking issue for iSwap: https://github.com/cda-tum/mqt-core/issues/423
+  case iSWAP:
+  case None:
+  case Compound:
+  case Measure:
+  case Reset:
+  case Teleportation:
+  case ClassicControlled:
+  case ATrue:
+  case AFalse:
+  case MultiATrue:
+  case MultiAFalse:
+  case OpCount:
+    throw QFRException("Inverting gate" + toString(type) +
+                       " is not supported.");
+  }
+}
 } // namespace qc
--- a/src/operations/SymbolicOperation.cpp
+++ b/src/operations/SymbolicOperation.cpp
@@ -332,4 +332,64 @@ void SymbolicOperation::instantiate(const VariableAssignment& assignment) {
   }
   checkUgate();
 }
+
+void SymbolicOperation::negateSymbolicParameter(const std::size_t index) {
+  if (isSymbolicParameter(index)) {
+    // NOLINTBEGIN(bugprone-unchecked-optional-access) - we check for this
+    symbolicParameter.at(index) = -symbolicParameter.at(index).value();
+    // NOLINTEND(bugprone-unchecked-optional-access)
+  } else {
+    parameter.at(index) = -parameter.at(index);
+  }
+}
+
+void SymbolicOperation::addToSymbolicParameter(const std::size_t index,
+                                               const fp value) {
+  if (isSymbolicParameter(index)) {
+    // NOLINTBEGIN(bugprone-unchecked-optional-access) - we check for this
+    symbolicParameter.at(index) = symbolicParameter.at(index).value() + value;
+    // NOLINTEND(bugprone-unchecked-optional-access)
+  } else {
+    parameter.at(index) += value;
+  }
+}
+
+void SymbolicOperation::invert() {
+  switch (type) {
+  case GPhase:
+  case Phase:
+  case RX:
+  case RY:
+  case RZ:
+  case RXX:
+  case RYY:
+  case RZZ:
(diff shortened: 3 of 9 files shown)


src/operations/StandardOperation.cpp
```diff
@@ -467,4 +467,76 @@ void StandardOperation::dumpOpenQASMTeleportation(
...
+    parameter[1] = -parameter[1] - PI;
+    break;
+  }
+  case U3: {
+    parameter[0] = -parameter[0];
+    parameter[1] = -parameter[1];
+    parameter[2] = -parameter[2];
+    std::swap(parameter[1], parameter[2]);
+    break;
+  }
```
martin-fink: These parameter changes were taken from `Operations.hpp`

include/operations/Operation.hpp
```diff
@@ -154,5 +154,7 @@ class Operation {
 
   virtual void dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
                             const RegisterNames& creg) const = 0;
+
+  virtual void invert() = 0;
```
martin-fink, line 187: An alternative would be a `const` function returning a new `Operation` instance. This might be helpful to reduce the code duplication that this PR introduces between `Operations.hpp` and `StandardOperation`.
For the `const` function, `Operations.hpp` could call `invert` without changing the operation itself. Not sure what is better here.

burgholzer: I thought a bit about this and I would love to somehow get rid of the code duplication with the `dd/Operations.hpp` file. However, I do not think this comes for free as it would most likely require a copy of the operation in one way or the other. I'd rather leave the code in that file untouched for now and leave a potential refactoring for later.
But it would still be nice to have the respective interfaces for the inverse method ready at that point.

Could we actually offer both methods; an in-place non-const version that does not cause a copy and an const version that essentially calls `op->clone()`, and then `inverse()` on the resulting clone?
It might be a bit tricky to get this working with C++ as C++ does not permit two methods that only differ in their return type last time I checked. But maybe the `const` on one function is enough for the compiler to not complain about it.

src/operations/StandardOperation.cpp
```diff
@@ -467,4 +467,76 @@ void StandardOperation::dumpOpenQASMTeleportation(
...
+  case Peres:
+    type = Peresdag;
+    break;
+  case Peresdag:
+    type = Peres;
+    break;
+  default:
+    throw QFRException("Inverting gate" + toString(type) +
+                       " is not supported.");
+  }
```
burgholzer: Could you please carefully check whether this list is complete. It, for sure, is missing the `ECR` gate (which is self-inverse) and the `DCX` gate (a two-target operation whose inverse is given by swapping both targets). It also seems to be missing the `XXminusYY` and `XXplusYY` gates from the `dd/Operations.hpp` file.

src/operations/SymbolicOperation.cpp
```diff
@@ -332,4 +332,54 @@ void SymbolicOperation::instantiate(const VariableAssignment& assignment) {
...
+      parameter.at(index) = -parameter.at(index);
+    }
+  };
+  auto add = [this](const size_t index, const fp value) {
+    if (isSymbolicParameter(index)) {
+      symbolicParameter.at(index) = symbolicParameter.at(index).value() + value;
+    } else {
+      parameter.at(index) += value;
+    }
+  };
```
burgholzer: no real need to define these as lambdas here. Simply declare these as private methods of the `SymbolicOperation` class (probably using a slightly more descriptive name).

src/operations/SymbolicOperation.cpp
```diff
@@ -332,4 +332,54 @@ void SymbolicOperation::instantiate(const VariableAssignment& assignment) {
...
+    std::swap(symbolicParameter[0], symbolicParameter[1]);
+    break;
+  case U3:
+    negate(0);
+    negate(1);
+    negate(2);
+
+    std::swap(parameter[1], parameter[2]);
+    std::swap(symbolicParameter[1], symbolicParameter[2]);
+    break;
```
burgholzer: just for safety: also make sure to cover all uncovered operations (with parameters) from the `StandardOperation` file.

test/unittests/test_qfr_functionality.cpp
```diff
@@ -1937,3 +1937,106 @@ TEST_F(QFRFunctionality, dumpAndImportTeleportation) {
...
+TEST_F(QFRFunctionality, invertUnsupportedOperation) {
+  auto op = NonUnitaryOperation(1, Targets{0}, OpType::Measure);
+
+  ASSERT_THROW(op.invert(), QFRException);
+}
+
+TEST_F(QFRFunctionality, invertStandardOpSelfInverting) {
+  auto opTypes = {
+      OpType::I, OpType::X, OpType::Y, OpType::Z, OpType::H, OpType::SWAP,
+  };
```
burgholzer:
```suggestion
  const auto opTypes = {
      OpType::I, OpType::X, OpType::Y, OpType::Z, OpType::H, OpType::SWAP,
  };
```
small nitpick (also holds for the other test cases)

test/unittests/test_qfr_functionality.cpp
```diff
@@ -1937,3 +1937,106 @@ TEST_F(QFRFunctionality, dumpAndImportTeleportation) {
...
+
+  for (auto opType : opTypes) {
```
burgholzer, line 2101:
```suggestion
  for (const auto opType : opTypes) {
```
Same as above and throughout the remaining tests

test/unittests/test_qfr_functionality.cpp
```diff
@@ -1937,3 +1937,106 @@ TEST_F(QFRFunctionality, dumpAndImportTeleportation) {
...
+    auto op = StandardOperation(1, Targets{0}, opType);
```
burgholzer:
```suggestion
    auto op = StandardOperation(1U, 0U, opType);
```
another tiny nitpick where the test does not need to be as verbose. Also holds for the other tests.

burgholzer (changes requested):
Also here, many thanks for this nice and neatly organized contribution.
As in the other PR I have a few minor nitpicks and a reply to one of the general questions you posted in a comment. Should be pretty straight-forward to realize.

include/operations/Operation.hpp
martin-fink: I think adding a `inverted` function that clones and calls `invert` might work.

burgholzer: Sounds worth trying. Maybe it even works when the functions are called the same 🤷🏼

martin-fink: I've added an `inverted` function -- for my taste this makes the intent a bit clearer, but I can change it to `invert` if you want!

burgholzer: Like that! Might want to call it `getInverse` though in order to be a little more consistent with the rest of the code base and clearly indicate that this is a Getter.

src/operations/StandardOperation.cpp
```diff
@@ -467,4 +467,94 @@ void StandardOperation::dumpOpenQASMTeleportation(
...
+    parameter[1] = -parameter[1] - PI;
+    break;
+  case U3:
+    parameter[0] = -parameter[0];
+    parameter[1] = -parameter[1];
+    parameter[2] = -parameter[2];
+    std::swap(parameter[1], parameter[2]);
+    break;
+  case DCX:
+    std::swap(parameter[0], parameter[1]);
```
burgholzer: It's not about swapping the parameters here, but really about swapping the targets, i.e., the inverse of DCX(0, 1) is DCX(1, 0).
Could be as easy as
```suggestion
    std::swap(targets[0], targets[1]);
```

src/operations/StandardOperation.cpp
```diff
@@ -467,4 +467,94 @@ void StandardOperation::dumpOpenQASMTeleportation(
...
+    type = SX;
+    break;
+  case Peres:
+    type = Peresdag;
+    break;
+  case Peresdag:
+    type = Peres;
+    break;
+  case None:
+  case Barrier:
```
burgholzer: The Barrier gate is self-inverse and can be moved to the top (it merely acts as an identity)

src/operations/StandardOperation.cpp
```diff
@@ -467,4 +467,94 @@ void StandardOperation::dumpOpenQASMTeleportation(
...
+  case iSWAP:
```
burgholzer, line 545: As for the iSWAP gate, we currently lack the appropriate inverse type.
For now, I believe this is fine, as adding such a type would probably be out of scope for this PR.
But this is kind of a show stopper for getting rid of the code duplication in the `dd/Operations.hpp` file.
I see two ways forward:
- try to cram the new type of gate into this PR
- create a separate tracking issue for this problem and maybe add a comment here.

src/operations/SymbolicOperation.cpp
```diff
@@ -332,4 +332,67 @@ void SymbolicOperation::instantiate(const VariableAssignment& assignment) {
...
+    negateSymbolicParameter(0);
+    negateSymbolicParameter(1);
+    negateSymbolicParameter(2);
+
+    std::swap(parameter[1], parameter[2]);
+    std::swap(symbolicParameter[1], symbolicParameter[2]);
+    break;
+  case DCX:
+    std::swap(parameter[0], parameter[1]);
+    break;
```
burgholzer:
```suggestion
```
As a consequence of the comment in the other file, this can be removed

test/unittests/test_qfr_functionality.cpp
burgholzer, line 1: Maybe you can add some further tests for the changes in this PR so that the patch coverage check turns green ✅

burgholzer (changes requested):
Also here, some finishing touches before this is ready to merge.

burgholzer (approved):
LGTM ✅  Many thanks!
Would you mind quickly resolving the conflicts with the main branch that arose from the other PR just getting merged? Then this can be merged as well!

martin-fink:
Rebased! Thanks!

PR #444, opened 2023-10-18
♻️ Remove complex cache

burgholzer:
## Description

This PR completely refactors the DD package to not rely on a separate complex cache and make better use of the `CachedEdge` data structure.
This saves a level of indirection in many cases and makes the overall package just a little simpler as it now clearly separates DDs with edge weights that are not yet in the complex unique table from those that are in there already.

## Checklist:

- [x] The pull request only contains commits that are related to it.
- [x] I have added appropriate tests and documentation.
- [x] I have made sure that all CI jobs on GitHub pass.
- [x] The pull request introduces no new warnings and follows the project's style guidelines.

PR #445, opened 2023-10-19
:sparkles: added state preparation routines for GHZ and W state

BertiFlorea:
## Description

In #[341](https://github.com/cda-tum/mqt-core/issues/341#issue-1758878364) the DD package provided methods for directly constructing DD for computational basis states or any product state that can be constructed from the single-qubit states `|0>, |1>, |+>, |->, |L>, |R>`.

## Updates
There are now 2 routines which implement the 
- GHZ state
- W state,
also, there are unit tests which check that the preparation routines provide the desired states for a smaller number of qubits.

Fixes #[341](https://github.com/cda-tum/mqt-core/issues/341#issue-1758878364) 

## Notes
In the beginning, I tried adding the routines to a different header file, however, I noticed that it was a lot easier to stick to the current structure and add the functions to the Package.hpp header inside the DD package. Additionally, I tried to use the existing functions in order to create & normalize the decision diagrams.

## Checklist:


include/dd/Package.hpp
```diff
@@ -484,6 +484,52 @@ template <class Config> class Package {
...
+    }
+
+    std::vector<BasisStates> oneBasisState =
+        std::vector<BasisStates>(n, BasisStates::one);
+    auto zeroState = makeZeroState(n);
+    auto oneState = makeBasisState(n, oneBasisState);
+    auto f = normalize(add(zeroState, oneState), false);
+
+    return f;
+  }
```
burgholzer: This is mainly for future reference: We actively plan to remove the `start` parameter of all of these functions as it is quite cumbersome to handle and currently only used in the HSF Simulator in our DDSIM project.

Your implementation here actually does not handle the starting qubit (lowest index in the DD before the final terminal).

This is fine and I would be willing to actually drop the `start` parameter completely in these new state preparation routines!

burgholzer: This is not quite what I had in mind for these kind of routines here.
What you are doing is constructing two states and then conducting a (rather costly) addition operation.
Addition is kind of the worst kind of operation you can do on DDs.
Instead, the goal here would be to really build the DDs from the ground up in a similar fashion to the existing state-preparation methods.
That means, that the routines here should only rely on the `makeDDNode` method of the DD package and that's it.
Realising functions that way makes the method scale linearly in the number of qubits and just requires a single pass for the construction.

include/dd/Package.hpp
```diff
@@ -484,6 +484,52 @@ template <class Config> class Package {
...
+      std::vector<BasisStates> state =
+          std::vector<BasisStates>(n, BasisStates::zero);
+      state[i] = BasisStates::one;
+
+      auto newBasisState = makeBasisState(n, state);
+      f = add(newBasisState, f);
+    }
+    f = normalize(f, false);
+    return f;
+  }
```
burgholzer: Essentially, what I said above also applies here.
There is a programmatic way to construct these DDs from the ground up and that is what we should be aiming for here.

test/dd/test_package.cpp
```diff
@@ -107,6 +107,31 @@ TEST(DDPackageTest, BellState) {
...
+  // build state vector for (|001> + |010> + |100>) * 1/sqrt(3)
+  const dd::CVec state = {
+      0, std::sqrt(3) / 3, std::sqrt(3) / 3, 0, std::sqrt(3) / 3, 0, 0, 0};
+  const auto stateDD = dd->makeStateFromVector(state);
+
+  wState.printVector();
+
+  EXPECT_EQ(wState, stateDD);
+}
+
```
burgholzer: Thanks for already adding some tests.
Would it be possible to create a new `.cpp` file for these state preparation tests and also move all existing tests on the basis state generation to that new file?
I would really like to make the existing library a little more modular.

Also, please try to cover the corner cases of each of the methods in separate tests. Coverage for these kinds of changes should be very close to 100%.

burgholzer (changes requested):
Hey 👋🏼 

Many thanks for this nice initial contribution. I just went through your changes and left some comments on things that I believe still need a more detailed look/some changes.

Note that #444 will hopefully land soon and that might require some changes within this PR. No action item for now though.

include/dd/Package.hpp
BertiFlorea: Understood, I dropped the ```start``` parameter completely. ✅

test/dd/test_package.cpp
BertiFlorea: I moved all the tests created with respect to the state preparation routines into a new ```.cpp``` file. Additionally, I made sure to cover all corner cases and test for the desired behaviour.

include/dd/Package.hpp
BertiFlorea: I tried to construct the DDs from the ground up as you suggested, I hope that this newer version is better! 👍🏼 
Also, my intention was to create the left subtree, then the right subtree and lastly to merge them (hence the name of the variables ```leftSubtree```, ```rightSubtree```).

include/dd/Package.hpp
BertiFlorea: I created some examples by hand and some using [https://www.cda.cit.tum.de/app/ddvis/](url) and tried to do the same as before (i.e. create the left and right subtree in parallel). The only note here is that the first and the last nodes must be created separately outside the loop. 

I know that having 3 variables might be much, however, I tried keeping it as simple as possible so that it can be readable and easier to understand. Lastly, as far as I know, the normalization step should also be linear in the number of qubits, right?

include/dd/Package.hpp
```diff
@@ -485,6 +485,80 @@ template <class Config> class Package {
...
+          "Requested state with " + std::to_string(n) +
+          " qubits, but current package configuration only supports up to " +
+          std::to_string(nqubits) +
+          " qubits. Please allocate a larger package instance."};
+    }
+
+    if (n == 0) {
+      throw std::runtime_error{
+          "Cannot create GHZ state for 0 qubits. Please choose n > 0."};
+    }
```
burgholzer: No need to throw here. The state for 0 qubits is just a 1-terminal.
```suggestion
    if (n == 0U) {
      return vEdge::one();
    }
```

include/dd/Package.hpp
```diff
@@ -485,6 +485,80 @@ template <class Config> class Package {
...
+      rightSubtree = makeDDNode(static_cast<Qubit>(p),
+                                std::array{vEdge::zero(), rightSubtree});
+    }
+
+    auto f = makeDDNode(static_cast<Qubit>(n - 1),
+                        std::array<vEdge, RADIX>{
+                            {{leftSubtree.p, cn.lookup(dd::SQRT2_2, 0)},
+                             {rightSubtree.p, cn.lookup(dd::SQRT2_2, 0)}}});
+
+    return f;
```
burgholzer: This can be optimized a little further to something like
```suggestion
    return makeDDNode(static_cast<Qubit>(n - 1),
                        std::array<vEdge, RADIX>{
                            {{leftSubtree.p, {&constants::sqrt2over2, &constants::zero}},
                             {rightSubtree.p, {&constants::sqrt2over2, &constants::zero}}}});
```

include/dd/Package.hpp
```diff
@@ -485,6 +485,80 @@ template <class Config> class Package {
...
+          "Requested state with " + std::to_string(n) +
+          " qubits, but current package configuration only supports up to " +
+          std::to_string(nqubits) +
+          " qubits. Please allocate a larger package instance."};
+    }
+
+    if (n == 0) {
+      throw std::runtime_error{
+          "Cannot create W state for 0 qubits. Please choose n > 0."};
+    }
```
burgholzer: Similar as above, this should not error, but just return the 1-terminal
```suggestion
    if (n == 0U) {
      return vEdge::one();
    }
```

include/dd/Package.hpp
```diff
@@ -485,6 +485,80 @@ template <class Config> class Package {
...
+                               std::array{leftSubtree, rightSubtree});
+      rightSubtree = makeDDNode(static_cast<Qubit>(p),
+                                std::array{rightSubtree, vEdge::zero()});
+    }
+
+    auto f = makeDDNode(static_cast<Qubit>(n - 1),
+                        std::array{leftSubtree, rightSubtree});
+    f = normalize(f, false);
+
+    return f;
```
burgholzer: First of all: I am really surprised that the logic here is so simple. Wouldn't have expected that!
I even think you can simplify that further; see the suggestion below.
Lastly, the final normalisation call should not be necessary as every node coming from `makeDDNode` should be normalised. Is the top edge weight of the DD correctly set to 1 here? I would have expected some kind of modification to be necessary.

```suggestion
    auto leftSubtree = vEdge::zero();
    auto rightSubtree = vEdge::one();
    for (size_t p = 0; p < n - 1; ++p) {
      leftSubtree = makeDDNode(static_cast<Qubit>(p),
                               std::array{leftSubtree, rightSubtree});
      rightSubtree = makeDDNode(static_cast<Qubit>(p),
                                std::array{rightSubtree, vEdge::zero()});
    }
    return makeDDNode(static_cast<Qubit>(n - 1),
                        std::array{leftSubtree, rightSubtree});
```

test/dd/test_state_preparation.cpp
```diff
@@ -0,0 +1,41 @@
...
+  ASSERT_THROW(dd->makeGHZState(6), std::runtime_error);
+}
+
+TEST(DDPackageTest, WStateEdgeCasesTest) {
+  auto dd = std::make_unique<dd::Package<>>(3);
+
+  EXPECT_EQ(dd->makeWState(1), dd->makeBasisState(1, {dd::BasisStates::one}));
+  ASSERT_THROW(dd->makeWState(0), std::runtime_error);
+  ASSERT_THROW(dd->makeWState(6), std::runtime_error);
+}
```
burgholzer: These tests need updates after the changes requested above.

include/dd/Package.hpp
```diff
@@ -485,6 +485,80 @@ template <class Config> class Package {
...
+    }
+
+    if (n == 0) {
+      throw std::runtime_error{
+          "Cannot create W state for 0 qubits. Please choose n > 0."};
+    }
+
+    if (n == 1) {
+      return makeBasisState(n, {BasisStates::one});
+    }
```
burgholzer: If you adopt the changes below, I believe you can actually drop this special case here as it is natively handled.
```suggestion
```

test/dd/test_state_preparation.cpp
```diff
@@ -0,0 +1,41 @@
...
+  auto dd = std::make_unique<dd::Package<>>(3);
+  auto wState = dd->makeWState(3);
+
+  // build state vector for (|001> + |010> + |100>) * 1/sqrt(3)
+  const dd::CVec state = {
+      0, std::sqrt(3) / 3, std::sqrt(3) / 3, 0, std::sqrt(3) / 3, 0, 0, 0};
+  const auto stateDD = dd->makeStateFromVector(state);
+
+  EXPECT_EQ(wState, stateDD);
+}
```
burgholzer: Would it be possible to make these tests a little more generic?
Specifically, test a wider range of qubits. You can create parametrised tests and test fixtures using googletest pretty easily. You can have a look at https://github.com/cda-tum/mqt-core/blob/main/test/algorithms/test_wstate.cpp for example.
Would be nice to test an equal range of qubits here.
It is quite straight-forward to formulate the checks for the correct state in a generic way
- for the GHZ state the amplitudes of the 0...0 and 1...1 both have to be 1/sqrt(2)
- for the W-state the amplitudes of all states with exactly one 1 entry must have the same amplitude (should be `sqrt(n)/n` if I am not mistaken)

In addition to the above, it would be nice to extend the tests
- https://github.com/cda-tum/mqt-core/blob/main/test/algorithms/test_wstate.cpp
- https://github.com/cda-tum/mqt-core/blob/main/test/algorithms/test_entanglement.cpp

with an additional test that makes sure the DD generate by simulating the algorithm is the same as the DD generated by the newly added routines. For that, you can just 
- construct the algorithm
- construct a DD package
- construct a zero state
- call the simulate method (from https://github.com/cda-tum/mqt-core/blob/be7c5eca7d22e285388ffb62c4478b8edb97569c/include/dd/Simulation.hpp#L10-L32), yielding a DD `e`
- construct the respective state using the new routine, yielding a DD `f`
- assert that `e == f` (pointers and edge weights should match).

burgholzer (changes requested):
Many thanks for the changes here! Looking way better already 👍🏼 
I left some comments inline on how to improve it even further and a minor request for some more generic tests that I think would make sense.

include/dd/Package.hpp
burgholzer: See also the other comment from the review just now. Your implementation looked pretty reasonable already. Just a couple of further improvements.

Normalisation actually happens on every call to `makeDDNode` and only touches the node it is called on and the edge weights to its successors. So it is actually O(1).

test/dd/test_state_preparation.cpp
BertiFlorea: Done 👌🏼

test/dd/test_state_preparation.cpp
BertiFlorea: I have added the requested changes, please let me know if there are any changes needed! 😃

extern/boost/config
burgholzer, line 1: Please revert the changes here. Submodules are not supposed to change.

extern/googletest
burgholzer, line 1: Please revert the changes here. Submodules are not supposed to change.

extern/json
burgholzer, line 1: Please revert the changes here. Submodules are not supposed to change.

include/dd/Package.hpp
```diff
@@ -485,6 +485,69 @@ template <class Config> class Package {
...
+    if (n == 0U) {
+      return vEdge::one();
+    }
+
+    auto leftSubtree = vEdge::zero();
+    // This is just a temporary trick to get the scaling right.
+    // It is inherently limited by the tolerance we employ and will crash once
+    // 1/sqrt(n) < RealNumber::eps.
+    // The trick to resolve this is to spread the 1/sqrt(n) across the various
+    // levels of the DD during construction.
```
burgholzer: Based on the discussion we had last time, this comment is outdated and can be removed.
Instead of that, IIRC, we discussed to add a check that warns whenever `1./sqrt(n) < RealNumber::eps` because that's when the method falls apart. A potential warning message for the exception too throw could include something like "Requested qubit size for generating W-state would lead to an underflow due to `1 / sqrt(n)` being smaller than the currently set tolerance (<actual tolerance>). If you still wanna run the computation, please lower the tolerance accordingly".

test/dd/test_state_preparation.cpp
```diff
@@ -0,0 +1,56 @@
+#include "dd/Package.hpp"
+
+#include <gtest/gtest.h>
+
+class StatePreparation : public testing::TestWithParam<qc::Qubit> {};
+
+extern std::vector<std::string> generateWStateStrings(const std::size_t length);
```
burgholzer:
```suggestion
extern std::vector<std::string> generateWStateStrings(std::size_t length);
```
I guess this is the only warning left to fix here.

test/dd/test_state_preparation.cpp
burgholzer, line 1: Now that I see how small this kind of file is getting: Could we just move the Corner Case tests to the respective algorithm test files?
The other tests are kind of covered already by the functionality tests you added there.
Then you can delete this file.

burgholzer (changes requested):
Many thanks for the latest set of changes. We are really getting there. I think this should be the last round of changes requested. After that, we are good to merge.

extern/boost/config
BertiFlorea: Done ✅

extern/googletest
BertiFlorea: Done ✅

extern/json
BertiFlorea: Done ✅

include/dd/Package.hpp
BertiFlorea: Done ✅

test/dd/test_state_preparation.cpp
BertiFlorea: Done ✅

burgholzer (approved):
Many thanks for this nice addition!

PR #452, opened 2023-10-23
Benchmarking the DD package

tyi1025:
## Description

Adding benchmark evaluation of the dd package as a component. The routine runs a series of benchmarks and generates a JSON file. A Python submodule was implemented to aggregate and visualize the JSON data.

Besides, enhances tests with new benchmark methods.

Fixes [#(447)](https://github.com/cda-tum/mqt-core/issues/447)


burgholzer:
Many thanks for the initial draft here. I just pushed some changes that should bring the PR on track to get merged. Note that I didn't have enough time to fix all the errors that now pop up in the evaluation file. Would be nice if you could adapt the file accordingly.

I also kind thought that the changes to the algorithms tests would look a little different. I would have expected them to more or less replace the existing tests. You have access to the DD package as well as the final state/matrix DD from the `Experiment` result object.

docs/eval/compare_default.png
burgholzer, line 1: Instead of adding all these pictures for the documentation, just show and run the actual code on very small examples. The code is being run on every commit to main and really shows the underlying functionality and how to use it.

docs/json_evaluation.ipynb
burgholzer, line 1: This is a solid start. However, as pointed out above, this shouldn't just be a single cell notebook, but rather a sequence of step by step demonstrations.

Some comments triggered by the way things are described here:
- The current functionality is purely benchmarking the DD package. This should be reflected in the description here.
- `mqt-core/eval/eval-dd-package.cpp` is not a script. it is a source file. and you do not execute source files. what you really have to do is to configure CMake with the proper new option, and then build and run the right target.
- add a proper reference/link to source code lines that are supposed to be modified.
- state that running the target will produce a result file in JSON data
- there is no need to keep two files for running the benchmarks. It should be enough to just get the full results.json file and use that as an input to the `compare` method
- no changes on the evaluation script should ever be necessary to run a comparison. It should just be a CLI call similar to ASV. You pass it two results files and some options and that is all. Every other information should already be encoded in the results file
- link to the reference documentation for the `compare` method instead of describing the options here.

include/dd/Benchmark.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+namespace qc {
+class QuantumComputation;
+}
+
+namespace dd {
+
+namespace constants {
+static const std::size_t DEFAULT_SHOTS = 1024U;
+}
```
burgholzer: Just inline this parameter down below. It is only used once anyways.

pyproject.toml
```diff
@@ -36,7 +36,7 @@ requires-python = ">=3.8"
 dynamic = ["version"]
 
 [project.optional-dependencies]
-test = ["pytest>=7.0", "mqt.core[qiskit]"]
+test = ["pytest>=7.0", "mqt.core[qiskit]", "pandas>=1.3.0"]
```
burgholzer: Please move this to a separate extra `benchmark` or `evaluation` extra and add that extra to the test dependencies. You can literally copy and paste from the way the qiskit extra is handled.

test/algorithms/test_bernsteinvazirani.cpp
```diff
@@ -143,3 +136,19 @@ TEST_P(BernsteinVazirani, DynamicEquivalenceSimulation) {
...
+  EXPECT_NE(out->sim.p, nullptr);
+}
+
+TEST_F(BernsteinVazirani, BVBenchmarkFunctionality) {
+  auto nqubits = 20;
+  auto qc = std::make_unique<qc::BernsteinVazirani>(nqubits);
+  qc::CircuitOptimizer::removeFinalMeasurements(*qc);
+  const auto out = dd::benchmarkFunctionalityConstruction(*qc);
+  EXPECT_NE(out->func.p, nullptr);
+}
```
burgholzer: These tests are redundant, aren't they? the behaviour is inherently already tested by the other methods.
Please also adjust for the other files. Anyhow: these tests do not test anything besides the fact that the simulation runs and produces _something_. This might give a false sense of security from the resulting good coverage.

burgholzer: Furthermore, these tests are not parametrised for some reason.

test/python/example_results.json
burgholzer, line 1: as already hinted at in the documentation feedback: there should really be one file per results. and not multiple results in a single file.

burgholzer (changes requested):
Hey 👋🏼 

Many many thanks for this nice addition. I haven't yet made it through the whole PR yet, but wanted to still give you some feedback right away as I do believe that some work is still required to get this done right. You can find that feedback down below. I'll do a full review once the respective change requests have been accounted for.

Many thanks 🙏🏼

docs/eval/results.json.png
burgholzer, line 1: Instead of an image, you can print the actual content of a JSON file as part of the docs. Just place the respective JSON files in this subfolder instead. There shouldn't be a need for including screenshots when you are executing a Jupyter notebook anyway.

docs/json_evaluation.ipynb
burgholzer, line 1: This file is currently not linked in the documentation. As a result, IIRC, it is not even executed during the docs build.
You can always see a preview of the documentation as part of the RtD CI check when you click on "Details".
Please make sure to check that the docs look as you intended them to.

burgholzer, line 1: It's rather hard to comment on individual lines in Jupyter files unfortunately.
But please cross-reference the actual documentation when you link to the documentation of the `compare` method.
You can find examples for that in the existing docs or the Sphinx documentation on cross-referencing functions.

eval/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+if(NOT TARGET gtest OR NOT TARGET gmock)
+  # Prevent overriding the parent project's compiler/linker settings on Windows
+  set(gtest_force_shared_crt # cmake-lint: disable=C0103
+      ON
+      CACHE BOOL "" FORCE)
+  add_subdirectory("${PROJECT_SOURCE_DIR}/extern/googletest" "extern/googletest" EXCLUDE_FROM_ALL)
+  set_target_properties(gtest gtest_main gmock gmock_main PROPERTIES FOLDER extern)
+endif()
+
+package_add_test(${PROJECT_NAME}-dd-eval ${PROJECT_NAME}-dd eval_dd_package.cpp)
```
burgholzer: I believe this might do more than you intended it to do.
For example the above CMake function also links against `gtest_main`, but you define your own main method.
Consider explicitly constructing the target here by hand-picking the appropriate lines from the `package_add_test` macro.

eval/eval_dd_package.cpp
```diff
@@ -0,0 +1,365 @@
...
+static const std::string FILENAME_END = ".json";
+
+static constexpr std::size_t SEED = 42U;
+
+// a function that parses a nlohmann::json from a file "results.json", populates
+// it with the results of the current run and writes it back to the file
+
+void verifyAndSave(const std::string& name, const std::string& type,
+                   qc::QuantumComputation& qc, const Experiment& exp) {
+  EXPECT_TRUE(exp.success());
```
burgholzer:
```suggestion
  ASSERT_TRUE(exp.success());
```
If the run was not successful, you probably want to abort.

eval/eval_dd_package.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+#include <string>
+#include <utility>
+
+namespace dd {
+
+std::string
+    inputFilename; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
```
burgholzer: The clang-tidy warning is there for a reason..
This file contains quite a lot of boilerplate code that is only in place due to googletest being used.
Seeing this, I guess it would be way simpler and straight-forward to just turn all of this into a single-class, that
- takes the `filename` as an input in the constructor
- has static constexpr members for all of the constants and benchmark settings
- individual methods for running the tests (essentially replace googletest test with method/function of the class)
- a `run` method that starts all the evaluations
- provides some kind of logging which benchmarks have been executed already (can be very rudimentary)

Then this file is way less chaotic and more organised.

eval/eval_dd_package.cpp
```diff
@@ -0,0 +1,365 @@
...
+  computeTables.erase("matrix_kronecker");
+  computeTables.erase("vector_kronecker");
+  computeTables.erase("vector_inner_product");
+  computeTables.erase("matrix_conjugate_transpose");
+
+  if (type == "Functionality") {
+    entry["dd"].erase("vector");
+    computeTables.erase("vector_add");
+    computeTables.erase("matrix_vector_mult");
+  }
```
burgholzer: I would rather not erase all of these entries. They might be useful in the future and they do not hurt much now.
If a benchmark does not use them, they are not expected to change. So the `compare` method wouldn't highlight them if one is only looking at things that changed.

eval/eval_dd_package.cpp
```diff
@@ -0,0 +1,365 @@
...
+  verifyAndSave("RandomClifford", "Functionality", *qc, *out);
+}
+
+} // namespace dd
+
+int main(int argc, char** argv) {
+  if (argc == 1) {
+    std::cerr << "Need specified file name" << '\n';
+    return 1;
+  }
```
burgholzer: The input checking here could probably be a little more elaborate. What if people pass more than one argument?
Something like checking whether exactly one argument is being provided and stating that the script expects exactly one argument for the name of the file, otherwise; seems to be sufficient to me.

pyproject.toml
```diff
@@ -36,7 +36,7 @@ requires-python = ">=3.8"
 dynamic = ["version"]
 
 [project.optional-dependencies]
-test = ["pytest>=7.0", "mqt.core[qiskit]"]
+test = ["pytest>=7.0", "mqt.core[qiskit]", "mqt.core[evaluation]"]
```
```suggestion
test = ["pytest>=7.0", "mqt.core[qiskit, evaluation]"]
```

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+import pandas as pd
+
+pd.set_option("display.max_colwidth", None)
+pd.set_option("display.max_rows", None)
+pd.set_option("display.width", None)
+
+sort_options = ["ratio", "experiment"]
+
+
+def higher_better(key: str) -> bool:
```
burgholzer:
```suggestion
def __higher_better(key: str) -> bool:
```
functions that are intended to be private should be marked as such (using `__` to start the function name)
please also apply to other functions that shouldn't be public. (guess every one except the `compare` method)

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+) -> None:
+    """Compare the results of two benchmarking runs from the generated json file.
+
+    :param baseline_filepath: Path to the baseline json file.
+    :param feature_filepath: Path to the feature json file.
+    :param factor: How much does the new metric have to differ from the baseline to be considered as a significant change?
+    :param only_changed: If False, a table with the benchmarks that haven't changed significantly will be shown.
+    :param sort: Sort the table by this column. Valid options are "ratio" and "experiment".
+    :param no_split: If True, the output tables (improved and worsened. Also the results that stay the same if `only_changed` in False) will be merged into one table.
+    """
```
burgholzer: Please follow the docstring style from the rest of the project. See https://github.com/cda-tum/mqt-core/blob/1ee8f70084aa49e312a5d09423665d0582e43dd0/src/mqt/core/io.py#L15-L27 for example

This should also include potential exceptions

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    factor: float = 0.1,
+    only_changed: bool = True,
+    sort: str = "ratio",
+    no_split: bool = False,
+) -> None:
+    """Compare the results of two benchmarking runs from the generated json file.
+
+    :param baseline_filepath: Path to the baseline json file.
+    :param feature_filepath: Path to the feature json file.
+    :param factor: How much does the new metric have to differ from the baseline to be considered as a significant change?
```
burgholzer:
```suggestion
    :param factor: How much a result has to change to be considered significant.
```

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    :param only_changed: If False, a table with the benchmarks that haven't changed significantly will be shown.
```
burgholzer:
```suggestion
    :param only_changed: Whether to only show results that changed significantly.
```

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    :param sort: Sort the table by this column. Valid options are "ratio" and "experiment".
+    :param no_split: If True, the output tables (improved and worsened. Also the results that stay the same if `only_changed` in False) will be merged into one table.
```
burgholzer:
```suggestion
    :param no_split: Whether to merge all results together in one table or to separate the results into benchmarks that improved, stayed the same, or worsened.
```

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+        if isinstance(value, dict):
+            items.update(flatten_dict(value, new_key, sep=sep))
+        else:
+            items[new_key] = value
+    return items
+
+
+def compare(
+    baseline_filepath: str,
+    feature_filepath: str,
```
burgholzer: you should be able to use `os.PathLike` here to be a little more permissive on the actual input

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    worsened_before = []
+    worsened_after = []
+    worsened_ratio = []
+    worsened_exp = []
+
+    df_same = pd.DataFrame()
+    same_before = []
+    same_after = []
+    same_ratio = []
+    same_exp = []
```
burgholzer: why all the auxiliary lists if you could directly write to the data frame?

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+        print(df_all)  # noqa: T201
+        return
+
+    if no_split and only_changed:
+        df_all = pd.concat([df_improved, df_worsened], ignore_index=True)
+        df_all.index = pd.Index([""] * len(df_all.index))
+        df_all = df_all.sort_values(by=sort)
+        print("All changed benchmarks:")  # noqa: T201
+        print(df_all)  # noqa: T201
+        return
```
burgholzer: quite some code duplication between the two if's here.
that logic can be simplified and combined

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    print("Benchmarks that have improved:")  # noqa: T201
+    print(df_improved)  # noqa: T201
+
+    if not only_changed:
+        print("Benchmarks that have stayed the same:")  # noqa: T201
+        print(df_same)  # noqa: T201
+    print("Benchmarks that have worsened:")  # noqa: T201
+    print(df_worsened)  # noqa: T201
```
burgholzer: you can ignore the `T201` warning for this file in the `pyproject.toml` configuration. You'll find an example of that already there in the ruff configuration section.

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    :param factor: How much does the new metric have to differ from the baseline to be considered as a significant change?
+    :param only_changed: If False, a table with the benchmarks that haven't changed significantly will be shown.
+    :param sort: Sort the table by this column. Valid options are "ratio" and "experiment".
+    :param no_split: If True, the output tables (improved and worsened. Also the results that stay the same if `only_changed` in False) will be merged into one table.
+    """
+    if factor < 0:
+        msg = "Factor must be positive!"
+        raise ValueError(msg)
+    if sort not in sort_options:
+        msg = "Invalid sort option!"
```
burgholzer: A good error message would include what the available options are.

src/mqt/core/evaluation.py
burgholzer, line 1: What is still missing here is exposing this as a console script so that people that have `mqt-core[evaluation]` installed can just call `compare <...> <..>` on the command line.
You can find more details on how to set that up here: https://setuptools.pypa.io/en/latest/userguide/entry_point.html#console-scripts
(you probably need to add a simple main function for that here. should be straight-forward)
You can add tests for that executable using the script runner from https://pypi.org/project/pytest-console-scripts/
An examples of how to do that are available as part of our QECC repo: https://github.com/cda-tum/mqt-qecc/blob/d063e4790ad1bb421effef86453cc5b9f2c421da/test/python/test_ecc_framework.py#L31-L53
This addition should also be reflected in the documentation.

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+
+import json
+from pathlib import Path
+from typing import Any
+
+import pandas as pd
+
+pd.set_option("display.max_colwidth", None)
+pd.set_option("display.max_rows", None)
+pd.set_option("display.width", None)
```
burgholzer, line 19: Is there any reason for setting these options here?

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+        if k in flattened_feature:
+            ls = [v, flattened_feature[k]]
+            flattened_data[k] = ls
+            del flattened_feature[k]
+        else:
+            ls = [v, "skipped"]
+            flattened_data[k] = ls
+    for k, v in flattened_feature.items():
+        ls = ["skipped", v]
+        flattened_data[k] = ls
```
burgholzer: Maybe I am stupid, but does this not kind of overwrite anything you set in the for loop above?

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+        msg = "Invalid sort option!"
+        raise ValueError(msg)
+    base_path = Path(baseline_filepath)
+    with base_path.open(mode="r", encoding="utf-8") as f:
+        d = json.load(f)
+    flattened_data = flatten_dict(d)
+    feature_path = Path(feature_filepath)
+    with feature_path.open(mode="r", encoding="utf-8") as f:
+        d_feature = json.load(f)
+    flattened_feature = flatten_dict(d_feature)
```
burgholzer: Do you see a way to not flatten the JSON data and work with the structured data instead?
I feel that one loses quite some information when considering everything as a flat hierarchy.

Some suggestions for the code below (this might make some of the other comments obsolete):
- pandas has a method to read JSON files: https://pandas.pydata.org/docs/reference/api/pandas.read_json.html#pandas-read-json 
- pandas also has many database options for joining/merging/comparing data frames. See https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#merge-join-concatenate-and-compare
- The worsened, improves, unchanged results are just filters of the original DF. So you should be able to straight up use filters for getting the appropriate Subframes. See: https://pandas.pydata.org/docs/dev/getting_started/intro_tutorials/03_subset_data.html#how-do-i-filter-specific-rows-from-a-dataframe

I would propose to build on these existing concepts and their powerful implementation.
That should make the code here way easier and "standard".

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    df_same["experiment"] = same_exp
+    df_same = df_same.sort_values(by=sort)
+    df_same.index = pd.Index([""] * len(df_same.index))
+
+    if no_split and not only_changed:
+        df_all = pd.concat([df_improved, df_same, df_worsened], ignore_index=True)
+        df_all.index = pd.Index([""] * len(df_all.index))
+        df_all = df_all.sort_values(by=sort)
+        print("All benchmarks:")  # noqa: T201
+        print(df_all)  # noqa: T201
```
burgholzer: Just a little shorter: (also relevant for the other print statements
```suggestion
        print(f"All benchmarks:\n{df_all}")  # noqa: T201
```

test/algorithms/test_bernsteinvazirani.cpp
```diff
@@ -35,10 +36,8 @@ TEST_P(BernsteinVazirani, FunctionTest) {
   qc.printStatistics(std::cout);
 
   // simulate the circuit
-  auto dd = std::make_unique<dd::Package<>>(qc.getNqubits());
   const std::size_t shots = 1024;
-  auto measurements =
-      simulate(&qc, dd->makeZeroState(qc.getNqubits()), dd, shots);
+  auto measurements = dd::benchmarkSimulateWithShots(qc, shots);
```
burgholzer, line 40:
```suggestion
  const auto measurements = dd::benchmarkSimulateWithShots(qc, shots);
```

test/algorithms/test_bernsteinvazirani.cpp
```diff
@@ -57,10 +56,8 @@ TEST_P(BernsteinVazirani, FunctionTestDynamic) {
...
```
burgholzer, line 60:
```suggestion
  const auto measurements = dd::benchmarkSimulateWithShots(qc, shots);
```

test/algorithms/test_bernsteinvazirani.cpp
```diff
@@ -76,10 +73,8 @@ TEST_F(BernsteinVazirani, LargeCircuit) {
...
```
burgholzer, line 77:
```suggestion
  const auto measurements = dd::benchmarkSimulateWithShots(qc, shots);
```

test/algorithms/test_bernsteinvazirani.cpp
```diff
@@ -95,10 +90,8 @@ TEST_F(BernsteinVazirani, DynamicCircuit) {
...
```
burgholzer, line 94:
```suggestion
  const auto measurements = dd::benchmarkSimulateWithShots(qc, shots);
```

test/algorithms/test_entanglement.cpp
```diff
@@ -39,8 +39,9 @@ TEST_P(Entanglement, GHZRoutineFunctionTest) {
   const auto nq = GetParam();
 
   auto qc = qc::Entanglement(nq);
-  auto dd = std::make_unique<dd::Package<>>(nq);
-  const dd::VectorDD e = simulate(&qc, dd->makeZeroState(qc.getNqubits()), dd);
+  auto exp = dd::benchmarkSimulate(qc);
+  auto e = exp->sim;
+  auto& dd = exp->dd;
   const auto f = dd->makeGHZState(nq);
```
burgholzer:
```suggestion
  const auto f = exp->dd->makeGHZState(nq);
```
Might as well safe the line

test/algorithms/test_grover.cpp
```diff
@@ -109,10 +110,9 @@ TEST_P(Grover, Simulation) {
   ASSERT_NO_THROW({ qc = std::make_unique<qc::Grover>(nqubits, seed); });
 
   qc->printStatistics(std::cout);
-  auto in = dd->makeZeroState(nqubits + 1U);
   // there should be no error simulating the circuit
   const std::size_t shots = 1024;
-  auto measurements = simulate(qc.get(), in, dd, shots);
+  auto measurements = dd::benchmarkSimulateWithShots(*qc, shots);
```
burgholzer, line 107:
```suggestion
  const auto measurements = dd::benchmarkSimulateWithShots(*qc, shots);
```

test/algorithms/test_qft.cpp
```diff
@@ -69,7 +59,13 @@ TEST_P(QFT, Functionality) {
   // there should be no error constructing the circuit
   ASSERT_NO_THROW({ qc = std::make_unique<qc::QFT>(nqubits, false); });
   // there should be no error building the functionality
-  ASSERT_NO_THROW({ func = buildFunctionality(qc.get(), dd); });
+  auto exp = std::make_unique<dd::FunctionalityConstructionExperiment>();
+  exp->dd = std::make_unique<dd::Package<>>(nqubits);
+  initialCacheCount = exp->dd->cn.cacheCount();
+  initialComplexCount = exp->dd->cn.realCount();
```
burgholzer: No real need to construct a duplicate package instance here.
The initial Cache and Complex Count should be compile time constants.
If I am not mistaken, the initial complex count should be `1` (the 0.5 entry) and the initial cache count should be `0`.
So, technically, you can fully remove these `initial...` member from this file (and also the Grover file, where similar comments apply).

test/algorithms/test_qft.cpp
```diff
@@ -176,34 +181,47 @@ TEST_P(QFT, Simulation) {
...
-  ASSERT_NO_THROW({ func = buildFunctionalityRecursive(qc.get(), dd); });
+  exp = dd::benchmarkFunctionalityConstruction(*qc);
+  auto func = exp->func;
 
   // there should be no error building the functionality regularly
-  qc::MatrixDD funcRec{};
-  ASSERT_NO_THROW({ funcRec = buildFunctionality(qc.get(), dd); });
+  auto funcRec = buildFunctionalityRecursive(qc.get(), exp->dd);
 
+  dd = std::move(exp->dd);
```
burgholzer: move that statement up a bit and you can reduce even more of the changes from above.

test/algorithms/test_qft.cpp
```diff
@@ -176,34 +181,47 @@ TEST_P(QFT, Simulation) {
...
   ASSERT_EQ(func, funcRec);
+
```
burgholzer:
```suggestion
```

test/algorithms/test_qft.cpp
```diff
@@ -176,34 +181,47 @@ TEST_P(QFT, Simulation) {
...
   // there should be no error constructing the circuit
   ASSERT_NO_THROW({ qc = std::make_unique<qc::QFT>(nqubits, true, true); });
 
+  auto exp = std::make_unique<dd::SimulationExperiment>();
+  exp->dd = std::make_unique<dd::Package<>>(nqubits);
+  initialCacheCount = exp->dd->cn.cacheCount();
+  initialComplexCount = exp->dd->cn.realCount();
+  auto sim = exp->sim;
+  dd = std::move(exp->dd);
+
```
burgholzer:
```suggestion
```
should not be needed.

test/algorithms/test_qft.cpp
```diff
@@ -220,4 +238,6 @@ TEST_P(QFT, DynamicSimulation) {
 
   // the number of unique entries should be close to the number of shots
   EXPECT_GE(ratio, 0.7);
+
+  dd->decRef(sim);
```
burgholzer:
```suggestion
```
should also not be needed.

test/algorithms/test_qpe.cpp
```diff
@@ -117,7 +118,6 @@ TEST_P(QPE, QPETest) {
   qc::VectorDD e{};
   ASSERT_NO_THROW(
       { e = simulate(&qc, dd->makeZeroState(qc.getNqubits()), dd); });
```
burgholzer: why was this not changed to use the simulation method?

test/algorithms/test_wstate.cpp
```diff
@@ -45,8 +43,9 @@ TEST_P(WState, RoutineFunctionTest) {
   const auto nq = GetParam();
 
   auto qc = qc::WState(nq);
-  auto dd = std::make_unique<dd::Package<>>(qc.getNqubits());
-  const dd::VectorDD e = simulate(&qc, dd->makeZeroState(qc.getNqubits()), dd);
+  auto exp = dd::benchmarkSimulate(qc);
+  auto& dd = exp->dd;
+  auto e = exp->sim;
   const auto f = dd->makeWState(nq);
```
burgholzer:
```suggestion
  auto e = exp->sim;
  const auto f = exp->dd->makeWState(nq);
```

test/python/example_baseline_results.json
burgholzer, line 1: The naming of this file does not resemble the expected filenames that are produced by the evaluation script. (`results_baseline.json`). Same goes for the other file.

burgholzer, line 1: Another small comment here. But for illustrative purposes: couldn't you drastically reduce these files to just contain enough so that the various options of the `compare` method show different results?
I am imagining something below 50 LoC for each JSON file that allows to cover all the functionality (also in tests).

test/python/test_eval_json.py
```diff
@@ -0,0 +1,92 @@
+"""Test the evaluation of JSON results."""
+
+from __future__ import annotations
+
+import pytest
+
+from mqt.core.evaluation import compare, flatten_dict, higher_better
+
+path_base = "test/python/example_baseline_results.json"
+path_feature = "test/python/example_feature_results.json"
```
burgholzer: this will not work in general if I am not mistaken as it is dependent on the directory in which you call pytest.
a simple
```
cd test
pytest
```
probably already fails. There are ways to mitigate this that involve `importlib.resources`.
A quick google search should bring up a few results.

test/python/test_eval_json.py
```diff
@@ -0,0 +1,92 @@
...
+        raise AssertionError(msg) from e
+
+
+def test_compare_sort_by_experiment() -> None:
+    """Test sort by experiment."""
+    try:
+        compare(path_base, path_feature, factor=0.2, only_changed=True, sort="experiment", no_split=True)
+    except Exception as e:
+        msg = "compare() should not raise exception!"
+        raise AssertionError(msg) from e
```
burgholzer: These tests do not really test _anything_ despite the fact that some code is being executed without error.
As pointed out in the comment on the actual dummy result files, it would be great to actually test the proper behaviour here.

This is probably easily accomplished by making the `compare` method return the combined pandas data frame (or, potentially better, the three already separated data frames for unchanged, improved, worsened) and add an additional method (the one that is also called from the command line then) to print the respective results.

src/mqt/core/evaluation.py
burgholzer: You can also look into something like `click` for creating that CLI which could come in handy for specifying all the options.

burgholzer (changes requested):
Many thanks for the changes. This is beginning to take shape. You can find detailed comments on further steps in the inline comments.
Let me know if you have any questions.

src/mqt/core/evaluation.py
tyi1025: If we don't set them, a big part of the CLI output will be truncated.

src/mqt/core/evaluation.py
tyi1025: Since I already deleted the values above from the feature dict that can also be found in the baseline dict, the ones remaining in the feature dict are gonna be the ones not present in baseline

src/mqt/core/evaluation.py
burgholzer: Ah, I see! Makes sense. Maybe add a code comment that explains that 😃

src/mqt/core/evaluation.py
burgholzer: Ah. I see. That confused me a little.
Got it now.

src/mqt/core/evaluation.py
tyi1025: Because I want all the entries to be rows under corresponding headers. It's only possible to add them to lists one by one, so creating empty lists here is necessary.

test/algorithms/test_qft.cpp
tyi1025: Move it where exactly?

burgholzer: Just look at the changelog on GitHub ;o) there aren't that many lines to choose from.

test/algorithms/test_qft.cpp
tyi1025: I removed both of the `initial` members from both of the classes.
It wasn't possible to make exp a member in these classes. A smart pointer of `SimulationExperiment` (or `FunctionalityConstructionExperiment`) and a smart pointer of `Experiment` weren't compatible. exp->sim or ->func gave errors because `Experiment`s aren't supposed to have those.

src/mqt/core/evaluation.py
tyi1025: After numerous careful verifications, I can assure you that this flattening approach does not lose information. The reason is that none of the (flattened-out) keys are the same. Otherwise, they would be the same metrics.
`pandas.read_json` was the first thing I tried. The truth is that it couldn't even read our json right. I had to remove invalid information manually pandas somehow squeezed in there.
I also read the functions in the second link. Even if we fully use pandas here, we'd still need a flattened dict, because it is required in the CLI output in the last column.

src/mqt/core/evaluation.py
burgholzer: What I am trying to get at here is: why do you not simply create empty data frames up here and row-by-row add the individual entries?
Right now you are populating lists with data only to turn them into a data frame right after.

test/algorithms/test_qft.cpp
burgholzer: Such pointers _are_ compatible, it just requires runtime casting using `dynamic_cast` to the appropriate class.
If you have a base class pointer `std::unique_ptr<Base> p` and a derived class `Derived`, and you know that `p` is pointing to a derived class instance, you can do
```c++
auto* derived = dynamic_cast<Derived*>(p.get());
assert(derived != nullptr); // to make sure the cast succeeded
```

docs/eval/results_baseline.json
burgholzer, line 1: Do we _really_ need to duplicate these files for the docs? Seems redundant to me. You should be able to access the test folder content from the docs build

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,177 @@
...
+        if isinstance(value, dict):
+            items.update(__flatten_dict(value, new_key, sep=sep))
+        else:
+            items[new_key] = value
+    return items
+
+
+def compare(
+    baseline_filepath: PathLike[Any],
+    feature_filepath: PathLike[Any],
```
burgholzer: fairly sure these should be
```suggestion
    baseline_filepath: str | PathLike[str],
    feature_filepath: str | PathLike[str],
```

src/mqt/core/evaluation.py
burgholzer: That fact that you are telling me that pandas can't read our JSONs right tells me that something is off.
Do we need to fix something in the JSON we output?

I do not really see a reason for the flattening overall.
Why can't the current last column not be replaced by way better structured columns that are already in the JSON? For example:
```
before after ratio name  task  num_qubits component metric 
0.002989 0.000929 0.310786 BV sim 256 matrix_unique_table total_col_ratio
```
The above definitely needs some fleshing out (e.g., proper column names, etc.), but would
- look more structrured
- requires no flattening
- should be directly implementable on top of pandas.
- wouldn't require to decypher flattened names for benchmarks
- ...

src/mqt/core/evaluation.py
tyi1025: It suggests [here](https://stackoverflow.com/questions/10715965/create-a-pandas-dataframe-by-appending-one-row-at-a-time) that growing a DataFrame by row is not a good idea for performance reasons.

burgholzer: Thanks for the link. I see the point being made there.
However, with something like the solution (template) I shared with you yesterday on discord, this is no longer really relevant. As it is just a concatenation and then filtering.

docs/index.md
```diff
@@ -29,6 +29,13 @@ self
 quickstart
 ```
 
+```{toctree}
+:maxdepth: 2
+:caption: Evaluation Guide
```
burgholzer:
```suggestion
:caption: DD Package Evaluation
```

docs/index.md
```diff
@@ -29,6 +29,13 @@ self
...
+
+json_evaluation
```
burgholzer:
```suggestion
dd_package_evaluation
```
Seems like a more suitable name for that file

docs/json_evaluation.ipynb
burgholzer, line 1: I started this review by looking at the rendered documentation page. A couple things I noticed (purposely denoted as a tasklist):
- [x] The title could rather be `Benchmarking the DD Package` for now, as this seems a bit more appropriate.
- [x] The wording throughout the notebook isn't ideal at times, but I can take care of that in a final iteration.
- [x] I'd pretty print the JSON in the first evaluation cell. The JSON module has a method for that.
- [x] I would argue that the order of results should be "Improved", "Worsened", "Stayed the same". In most of the cases most results will stay the same and the interesting part is what really changed.
- [x] the `sort="algorithm"` option does not seem to do anything just judging from the output. I would have expected the order of benchmarks to follow the original order in the JSON file when selecting this option. This means, I would have expected to see the entries grouped together and sorted the way they appear in the original JSON. First sorted by algortihm, then by task, by number of qubits, by component, by metric.
- [x] the CLI example currently shows an error that should be fixed. Furthermore, flags in CLIs should be provided as such and not as options, i.e., it should be `--only-changed` instead of `--only-changed=true`.
- [x] overall, the differing numerical precision in some of these values makes the tables rather unpleasant to read. It should be enough to print at most 3 significant digits (after the comma). Maybe even two are enough. In general, the vertical spacing could be optimized a little by optimizing column names.
- [x]  I am kind of missing an option where all the DD details are not shown and it only shows the runtimes and the overall DD memory data. That would make it easier to get a first grasp of what is going on and where improvements happened. 
- [x] Based on that it might also be interesting to add a filter for a certain benchmark. Say once you identified that a certain benchmark has gone worse and you want to know what exactly is going on and you only want to focus on that particular benchmark.
- [x] Alignment in the column headers is inconsistent. Some of them are left aligned, some right. I'd probably stick to right alignment.
- [x] It seems weird that the `gate_count` is under `metric`. I wouldn't expect the gate count of a benchmark to ever change. Especially not when benchmarking the DD package. I'd suggest using a column with the header `|G|` after the `num_qubits` column (which could be abbreviated to `n`)
- [x] The "higher is better" metrics really make reading the results confusing. Can we do something about that? For the `hit_ratio` it would be as simple as to output `1. - hit_ratio`. For the number of `hits` that does not work so well. Any ideas are welcome. Technically, one could compute the `misses` and the `miss_ratio` instead. One just has to figure out the right results to use as a basis for the computation.
- [x] The order in the results that got worse seems to be mixed up. I would argue that results that got worse the most should be the first entries in the list. (just as the entries with the smallest ratio/biggest improvement come first for the benchmarks that improved). This is slighlty linked to the previous point, where it becomes obvious that the higher-are-better metrics are not ordered well yet.
- [x] If `before=0`, then the script seems to currently print `1e9` while the proper answer here would supposedly be `inf`

pyproject.toml
```diff
@@ -53,13 +53,24 @@ docs = [
     "nbsphinx",
     "sphinx-autodoc-typehints",
     "sphinx-autoapi",
-    "qiskit-terra[visualization]"
+    "qiskit-terra[visualization]",
+    "colorama>=0.4.4",
+    "tabulate>=0.8.9"
```
burgholzer: Why are these required again here? (while `pandas` is not)
If you require the `evaluation` extra to be installed for building the docs, then just state that as depending on
```
mqt.core[evaluation]
```

src/mqt/core/evaluation.py
```diff
@@ -0,0 +1,236 @@
...
+        [],
+        [],
+        [],
+        [],
+        [],
+    )
+
+    for k, v in flattened_data.items():
+        after = v[1]
+        before = v[0]
```
burgholzer:
```suggestion
        before = v[0]
        after = v[1]
```
triggers my inner monk. sorry for that.

pyproject.toml
```diff
@@ -53,13 +53,24 @@ docs = [
...
 ]
 qiskit = [
    "qiskit-terra>=0.22.0",
 ]
+evaluation = [
+    "pandas>=1.3.0",
+    "pytest-console-scripts>=1.4",
+    "colorama>=0.4.4",
```
burgholzer: There is definitely no need for that extra dependency just to produce two lines of colored terminal output.
See https://stackoverflow.com/a/287944 for the pretty simple and straight-forward solution.

pyproject.toml
```diff
@@ -53,13 +53,24 @@ docs = [
...
+    "tabulate>=0.8.9"
```
burgholzer: I do not see this dependency being used anywhere. Is there a reason for it still being here?

pyproject.toml
```diff
@@ -53,13 +53,24 @@ docs = [
...
+    "qiskit-terra[visualization]",
+    "colorama>=0.4.4",
+    "tabulate>=0.8.9"
 ]
 qiskit = [
    "qiskit-terra>=0.22.0",
 ]
+evaluation = [
+    "pandas>=1.3.0",
+    "pytest-console-scripts>=1.4",
```
burgholzer: This is not a dependency of the evaluation component. It is a test dependency. As such, it should be added to the test dependencies (preferably between pytest and the recursive mqt.core dependency).

test/algorithms/test_qft.cpp
```diff
@@ -1,40 +1,25 @@
...
-  }
+  void TearDown() override { dd->garbageCollect(true); }
 
   void SetUp() override {
     nqubits = GetParam();
-    dd = std::make_unique<dd::Package<>>(nqubits);
-    initialCacheCount = dd->cn.cacheCount();
-    initialComplexCount = dd->cn.realCount();
+    exp = std::make_unique<dd::Experiment>();
+    exp->dd = std::make_unique<dd::Package<>>(nqubits);
```
burgholzer: why does the setup routine instantiate an experiment when every test just overwrites it anyways?

test/algorithms/test_qft.cpp
```diff
@@ -1,40 +1,25 @@
...
   }
 
   std::size_t nqubits = 0;
-  std::unique_ptr<dd::Package<>> dd;
   std::unique_ptr<qc::QFT> qc;
-  std::size_t initialCacheCount = 0;
-  std::size_t initialComplexCount = 0;
-  qc::VectorDD sim{};
-  qc::MatrixDD func{};
+  std::unique_ptr<dd::Package<>> dd;
```
burgholzer: I do not see how this member should be necessary as the experiment inherently contains a package instance that is adequately sized.

test/algorithms/test_qft.cpp
```diff
@@ -1,40 +1,25 @@
...
-    dd->decRef(sim);
-    dd->decRef(func);
-    dd->garbageCollect(true);
-
-    // number of complex table entries after clean-up should equal initial
-    // number of entries
-    EXPECT_EQ(dd->cn.realCount(), initialComplexCount);
-    // number of available cache entries after clean-up should equal initial
-    // number of entries
-    EXPECT_EQ(dd->cn.cacheCount(), initialCacheCount);
```
burgholzer, line 21: The tests here lose this sanity check. I already annotated this some time back, but this seems to have gotten lost. I believe that also applies to the Grover test file.

test/python/test_eval_json.py
```diff
@@ -0,0 +1,232 @@
...
+path_base = Path(__file__).resolve().parent / "results_baseline.json"
+path_feature = Path(__file__).resolve().parent / "results_feature.json"
+
+
+def test_flatten_dict() -> None:
+    """Test flatten_dict."""
+    d1 = {"a": {"b": {"main": 1}}}
+    assert __flatten_dict(d1) == {"a.b.main": 1}
+    d2 = {"a": {"b": {"main": 1, "447-add-benchmark-suite-in-mqt-core": 2}}, "d": {"main": 2}}
+    assert __flatten_dict(d2) == {"a.b.main": 1, "a.b.447-add-benchmark-suite-in-mqt-core": 2, "d.main": 2}
```
burgholzer: Also triggers the inner monk in me: can we please get rid of the cryptic branch name here?
```suggestion
    d2 = {"a": {"b": {"main": 1, "feature": 2}}, "d": {"main": 2}}
    assert __flatten_dict(d2) == {"a.b.main": 1, "a.b.feature": 2, "d.main": 2}
```

test/python/test_eval_json.py
```diff
@@ -0,0 +1,232 @@
...
+        "component": "",
+        "metric": "feature",
+    }
+    assert __post_processing("a.b.main.feature.algorithm") == {
+        "algorithm": "a",
+        "task": "b",
+        "num_qubits": "main",
+        "component": "feature",
+        "metric": "algorithm",
+    }
```
burgholzer: How do these tests make any sense at all? 🙃

test/python/test_eval_json.py
```diff
@@ -0,0 +1,232 @@
...
+            "compare",
+            "./test/python/results_baseline.json",
+            "./test/python/results_feature.json",
+            "--sort=algorithm",
+            "--only_changed=true",
+            "--no_split=true",
+        ]
+    )
+    assert "All changed benchmarks:" in ret.stdout
+    assert ret.success
```
burgholzer: Why do all of these functions have the same docstring? 🙃  at that point the docstring does not add any value.

.pre-commit-config.yaml
```diff
@@ -70,6 +70,8 @@ repos:
         args: []
         additional_dependencies:
           - pytest
+          - pandas-stubs
+          - types-colorama
```
burgholzer:
```suggestion
```
Should not be necessary per comment below

burgholzer (changes requested):
Many thanks for another iteration. I just went over the changes and purposely started from the documentation page (https://mqt--452.org.readthedocs.build/projects/core/en/452/json_evaluation.html) to see which kind of things are really noticeable from a user side. I tried to not worry too much about the internal implementation details.
You can see the results of that in one of the big comments down below that you can also use as a task list.
And then, there are some further comments on the actual code itself that should be rather straight forward to handle.

pyproject.toml
tyi1025: I'm not sure about the cause, because frankly, it's kind of hard to debug jupyter notebooks.
However, if I don't add colorama and tabulate, the docs build will fail with a `ModuleNotFoundError`. It doesn't even if I don't add pandas.
On the other hand, I'm not sure if we need the entirety of `mqt.core[evaluation]`, since the point was to separate the evaluation module, right?
I suggest that we keep tabulate, and remove colorama since per the other suggestion I should remove it from the project anyway.

docs/json_evaluation.ipynb
tyi1025: Thanks for getting to it so quickly despite being sick! I'm on it.

pyproject.toml
tyi1025: tabulate is implicitly required for `df.to_markdown()`. See [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_markdown.html)

test/algorithms/test_qft.cpp
tyi1025: The three dd calls above have been moved to the individual tests, because we have removed `sim` and `func` as members (also `dd` per the other suggestion).
As for the rest two `expect` statements, I thought [you wanted the two `counts` members removed](https://github.com/cda-tum/mqt-core/pull/452#discussion_r1406127145)?

pyproject.toml
burgholzer: You can just use the `output-formatting` extra of pandas to automatically get that dependency.

pyproject.toml
burgholzer: I am fairly sure you will need the evaluation extra here if you want to use evaluation code as part of the docs. Anything else would be pretty strange.
I could imagine that the only reason pandas was not required here is that it might be a dependency of Qiskit in some way.

test/algorithms/test_qft.cpp
burgholzer: I wanted them to be replaced with static constants (as the initial numbers are always the same). Although this is really not the most important part of the PR.

docs/json_evaluation.ipynb
tyi1025: Would the column `|G|` only be used for `gate_counts`?

burgholzer: Yeah. Although I do not even know if we'd want to "waste" a column on that. It just felt strange to see it in the metrics

tyi1025: Would it be better under `components`? I personally don't think adding a column just for stuff like gate_counts would be a good idea.
Or, if the `gate_count` is really not supposed to change, I could:
- Remove `gate_count` and `runtime` from these tables, since they are the only metrics not under `dd`. 
- Then, I make a separate table or something equally clear to note how the `gate_counts` and the `runtime` changed. This way I can also remove `dd` from all the tables, which makes the tables clearer to read as well.

tyi1025: Is it important to you to see if a metric is `unused` or `skipped`? If not, I could change it all to `nan` or `?`, which would incidentally help the formatting of the number columns.

tyi1025: Would you be okay with `algo` instead of `algorithm`? It is a somewhat confusing abbreviation but could potentially save some space😅

tyi1025: > Based on that it might also be interesting to add a filter for a certain benchmark. Say once you identified that a certain benchmark has gone worse and you want to know what exactly is going on and you only want to focus on that particular benchmark.

Would this filter be, say for "hits" then we see tables with only "hits" in it, but all across the original json (i.e. across algorithm, num_qubits etc.)? Or do you want this filter to be as specific as "<algorithm>.<task>.<num_qubits>.<component>.hits"?

tyi1025: > The "higher is better" metrics really make reading the results confusing. Can we do something about that? For the hit_ratio it would be as simple as to output 1. - hit_ratio. For the number of hits that does not work so well. Any ideas are welcome. Technically, one could compute the misses and the miss_ratio instead. One just has to figure out the right results to use as a basis for the computation.

If I understood correctly, `misses` should be simply `lookups` - `hits`. We could modify the dd-package implementation so that our stats will have `misses` and `miss_ratio` instead of `hits` and `hit_ratio`.

burgholzer: > Would it be better under `components`? I personally don't think adding a column just for stuff like gate_counts would be a good idea. Or, if the `gate_count` is really not supposed to change, I could:
> 
> * Remove `gate_count` and `runtime` from these tables, since they are the only metrics not under `dd`.
> * Then, I make a separate table or something equally clear to note how the `gate_counts` and the `runtime` changed. This way I can also remove `dd` from all the tables, which makes the tables clearer to read as well.

I guess you could just filter out the `gate_count`. It's not really important for the benchmarks.
the runtime is a complete different story. It is really important to track that for the benchmarks.
It could be worth separating it from the other DD specific results.
In the end, one would probably first look at the runtimes and how they changed and then look at the DD specifics for detailed explanations.

burgholzer: > Is it important to you to see if a metric is `unused` or `skipped`? If not, I could change it all to `nan` or `?`, which would incidentally help the formatting of the number columns.

No, not at all. If no data is available, I don't really care why.

burgholzer: > Would you be okay with `algo` instead of `algorithm`? It is a somewhat confusing abbreviation but could potentially save some space😅

Fine with me.

tyi1025: > I guess you could just filter out the `gate_count`. It's not really important for the benchmarks. the runtime is a complete different story. It is really important to track that for the benchmarks. It could be worth separating it from the other DD specific results. In the end, one would probably first look at the runtimes and how they changed and then look at the DD specifics for detailed explanations.

Just to be clear by filtering out you mean removing it right?

burgholzer: > > Based on that it might also be interesting to add a filter for a certain benchmark. Say once you identified that a certain benchmark has gone worse and you want to know what exactly is going on and you only want to focus on that particular benchmark.
> 
> Would this filter be, say for "hits" then we see tables with only "hits" in it, but all across the original json (i.e. across algorithm, num_qubits etc.)? Or do you want this filter to be as specific as "..<num_qubits>..hits"?

I would imagine it would be something like "only show the data for the "GHZ, 1024 qubit, simulation" experiment."
Maybe these could be CLI options such as `--algorithm=ghz`, `--task=simulation`, `--num_qubits=1024`.

tyi1025: > I would imagine it would be something like "only show the data for the "GHZ, 1024 qubit, simulation" experiment." Maybe these could be CLI options such as `--algorithm=ghz`, `--task=simulation`, `--num_qubits=1024`.

Do you think these options should be individually supported (i.e. only show data for GHZ, or show the data for all the 1024-qubit experiments, or only show simulation), or should all of the three be specified, if the user wants to specify one of them?

burgholzer: > > The "higher is better" metrics really make reading the results confusing. Can we do something about that? For the hit_ratio it would be as simple as to output 1. - hit_ratio. For the number of hits that does not work so well. Any ideas are welcome. Technically, one could compute the misses and the miss_ratio instead. One just has to figure out the right results to use as a basis for the computation.
> 
> If I understood correctly, `misses` should be simply `lookups` - `hits`. We could modify the dd-package implementation so that our stats will have `misses` and `miss_ratio` instead of `hits` and `hit_ratio`.

That should be true. I'd like to keep the `hit_ratio` and `hits` in the DD package as these are the established metrics. Really kind of strange that these are the only ones where higher is better. Maybe stupid, but couldn't we just output `1 / ratio` for these metrics (and potentially add a `*` or something to the number to somehow communicate that these values are to be treated differently?

burgholzer: > > I guess you could just filter out the `gate_count`. It's not really important for the benchmarks. the runtime is a complete different story. It is really important to track that for the benchmarks. It could be worth separating it from the other DD specific results. In the end, one would probably first look at the runtimes and how they changed and then look at the DD specifics for detailed explanations.
> 
> Just to be clear by filtering out you mean removing it right?

Yes, exactly.

burgholzer: > > I would imagine it would be something like "only show the data for the "GHZ, 1024 qubit, simulation" experiment." Maybe these could be CLI options such as `--algorithm=ghz`, `--task=simulation`, `--num_qubits=1024`.
> 
> Do you think these options should be individually supported (i.e. only show data for GHZ, or show the data for all the 1024-qubit experiments, or only show simulation), or should all of the three be specified, if the user wants to specify one of them?

Good question. Just an idea: i think algorithm and num_qubits are kind of hierarchical; i.e., you need to specify the algorithm to be able to specify the number of qubits. The task filter is independent of the others.

tyi1025: > I guess you could just filter out the `gate_count`. It's not really important for the benchmarks. the runtime is a complete different story. It is really important to track that for the benchmarks. It could be worth separating it from the other DD specific results. In the end, one would probably first look at the runtimes and how they changed and then look at the DD specifics for detailed explanations.

Do you think we should do two sets of tables, one just for runtime, and the other for the dd details, and provide an option not to show the dd details?

burgholzer: > > I guess you could just filter out the `gate_count`. It's not really important for the benchmarks. the runtime is a complete different story. It is really important to track that for the benchmarks. It could be worth separating it from the other DD specific results. In the end, one would probably first look at the runtimes and how they changed and then look at the DD specifics for detailed explanations.
> 
> Do you think we should do two sets of tables, one just for runtime, and the other for the dd details, and provide an option not to show the dd details?

That sounds very reasonable!

tyi1025: > > > The "higher is better" metrics really make reading the results confusing. Can we do something about that? For the hit_ratio it would be as simple as to output 1. - hit_ratio. For the number of hits that does not work so well. Any ideas are welcome. Technically, one could compute the misses and the miss_ratio instead. One just has to figure out the right results to use as a basis for the computation.
> > 
> > 
> > If I understood correctly, `misses` should be simply `lookups` - `hits`. We could modify the dd-package implementation so that our stats will have `misses` and `miss_ratio` instead of `hits` and `hit_ratio`.
> 
> That should be true. I'd like to keep the `hit_ratio` and `hits` in the DD package as these are the established metrics. Really kind of strange that these are the only ones where higher is better. Maybe stupid, but couldn't we just output `1 / ratio` for these metrics (and potentially add a `*` or something to the number to somehow communicate that these values are to be treated differently?

I'm not sure if we should add a character to the number since it kind of disrupts the sorting and other data processing steps. How about we add a `*` to the end of the metric, like `hits*` and `hit_ratio*`?

burgholzer: > > > > The "higher is better" metrics really make reading the results confusing. Can we do something about that? For the hit_ratio it would be as simple as to output 1. - hit_ratio. For the number of hits that does not work so well. Any ideas are welcome. Technically, one could compute the misses and the miss_ratio instead. One just has to figure out the right results to use as a basis for the computation.
> > > 
> > > 
> > > If I understood correctly, `misses` should be simply `lookups` - `hits`. We could modify the dd-package implementation so that our stats will have `misses` and `miss_ratio` instead of `hits` and `hit_ratio`.
> > 
> > 
> > That should be true. I'd like to keep the `hit_ratio` and `hits` in the DD package as these are the established metrics. Really kind of strange that these are the only ones where higher is better. Maybe stupid, but couldn't we just output `1 / ratio` for these metrics (and potentially add a `*` or something to the number to somehow communicate that these values are to be treated differently?
> 
> I'm not sure if we should add a character to the number since it kind of disrupts the sorting and other data processing steps. How about we add a `*` to the end of the metric, like `hits*` and `hit_ratio*`?

Sounds good to me

burgholzer (approved):
I just applied some finishing touches and this is now ready to merge 🎉 
Many thanks for all the effort! Looking forward to using the benchmark library for the first time.

tyi1025:
> I just applied some finishing touches and this is now ready to merge 🎉 Many thanks for all the effort! Looking forward to using the benchmark library for the first time.

Thanks a lot for the changes and merging, and of course the guidance and supervision! Hopefully customer satisfaction will be high :D

PR #453, opened 2023-10-23
Fixing `QuantumComputation::dump` file name dot count check

### LINKED ISSUE #446: 🐛 Multiple dots in file path passed to QuantumComputation::dump results in assertion fail
### Environment information

- mqt.core version: commit #433

### Description

When passing a file name containing more than 1 dot (e.g. `test.circuit.qasm`) to `QuantumComputation::dump` (or by extension any of its wrapper functions like `QuantumComputation::dumpOpenQASM`), the assertion in line 716 of `QuantumComputation.cpp` fails:
```java
assert(std::count(filename.begin(), filename.end(), '.') == 1);
```

If I understand this line correctly, it tries to assert, that the supplied file name includes a file extension (though I'm not quite sure, why this would even be necessary in line 716, where the dumping format is already clear; maybe this line actually belongs before current line 622 `const std::size_t dot = filename.find_last_of('.');`?).
In any case, if the only objective is to test for the existence of a file extension, shouldn't it be:
```java
assert(std::count(filename.begin(), filename.end(), '.') >= 1);
```

### Expected behavior

I would expect to be able to pass any valid file name (or even file path) to QuantumComputation::dump containing a valid quantum circuit file extension but no other constraints (and in case I supply an explicit `qc::Format` even the file extension to be irrelevant)

### How to Reproduce

```java
qc::QuantumComputation qc{4, 4};
qc.x(0, qc::Control{1});
qc.x(2, qc::Control{3});
qc.dump("test.circuit.qasm", qc::Format::OpenQASM);
// or
qc.dump("./test/circuit.qasm", qc::Format::OpenQASM); // second point not even in file name itself but only in parent path
```

EliasLF:
## Description

Removing the superfluous assertion in `QuantumComputation::dump(filename, format)`, that the file name needs to contain exactly 1 dot.
Instead, added an equivalent check (for >= 1 dot) in `QuantumComputation::dump(filename)`, where it is actually necessary.

Fixes #446 

## Checklist:

### DIFF
--- a/src/QuantumComputation.cpp
+++ b/src/QuantumComputation.cpp
@@ -611,6 +611,7 @@ std::ostream& QuantumComputation::printStatistics(std::ostream& os) const {
 
 void QuantumComputation::dump(const std::string& filename) {
   const std::size_t dot = filename.find_last_of('.');
+  assert(dot != std::string::npos);
   std::string extension = filename.substr(dot + 1);
   std::transform(
       extension.begin(), extension.end(), extension.begin(),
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
LGTM! Thanks for the nice and clean PR :tada:

PR #460, opened 2023-10-25
:sparkles: added inverse gate for iSWAP

### LINKED ISSUE #423: ✨ Inverse gate for `iSwap`

Currently, there is no inverse gate type for `iSwap`.

Add a new gate type `iSwapDag` that represents the inverse of `iSwap`.

BertiFlorea:
## Description

The DD Package was missing the ```iSWAPDag``` (iSWAP dagger) gate. However, the DD Package already had in some place a ```ISWAPDG_MAT``` which implemented the desired gate in a matrix form.

## Status
Currently, the `iSWAPdagDumpIsValid` test produces the following error:
```
gate type iswapdg could not be converted to OpenQASM

i:   0   1
1:iswdiswd
o:   0   1

~/mqt-core/test/unittests/test_io.cpp:389: Failure
Expected: qc->import(ss, qc::Format::OpenQASM); doesn't throw an exception.
  Actual: it throws std::runtime_error with description "[qasm parser] l:6 c:2 msg: Unknown gate q".


i:   0   1
o:   0   1
```

After reading through [https://qiskit.org/documentation/apidoc/circuit_library.html](url), my assumption is that OpenQASM does not support at the moment the iSWAP dagger gate natively, so, therefore, it doesn't have implementation for it.

(I know that the previous assumption might not be the reason why, but I tried to point out a tiny direction why this might not work. Anyway, I will try to look further into this issue in the near future)

Fixes #423 

## Checklist:


BertiFlorea:
Important note: `iSWAP` is supported by `OpenQASM`, however, the inverse isn't, therefore I didn't know precisely how to proceed further. (Also, I wasn't sure what files I was/wasn't allowed to modify)

include/operations/OpType.hpp
```diff
@@ -163,6 +166,8 @@ inline std::string shortName(const OpType& opType) {
     return "sw";
   case iSWAP:
     return "isw";
+  case iSWAPdg:
+    return "iswd";
```
burgholzer: As implied by the docstring of this method, short names are supposed to be 3 characters at most.
Otherwise, formatting during printing is all kinds of messed up 🙃 
What about using `"is"` for the iSWAP gate ad `"isd"` for the inverse for consistency reasons?

src/mqt/core/_core/__init__.pyi
```diff
@@ -217,6 +217,9 @@ class QuantumComputation(MutableSequence[Operation]):
     def iswap(self: Self, target1: int, target2: int) -> None: ...
     def ciswap(self: Self, control: Control, target1: int, target2: int) -> None: ...
     def mciswap(self: Self, controls: set[Control], target1: int, target2: int) -> None: ...
+    # def iswapdg(self: Self, target1: int, target2: int) -> None: ...
+    # def ciswapdg(self: Self, control: Control, target1: int, target2: int) -> None: ...
+    # def mciswapdg(self: Self, controls: set[Control], target1: int, target2: int) -> None: ...
```
burgholzer: These should not be commented. The library exposes them, so hey should be in the stub file.

src/mqt/core/_core/operations.pyi
```diff
@@ -35,42 +35,43 @@ class Control:
...
-    dcx: ClassVar[OpType]  # value = <OpType.dcx: 26>
-    ecr: ClassVar[OpType]  # value = <OpType.ecr: 27>
+    classic_controlled: ClassVar[OpType]  # value = <OpType.classic_controlled: 39>
+    compound: ClassVar[OpType]  # value = <OpType.compound: 35>
+    dcx: ClassVar[OpType]  # value = <OpType.dcx: 27>
+    ecr: ClassVar[OpType]  # value = <OpType.ecr: 28>
     gphase: ClassVar[OpType]  # value = <OpType.gphase: 1>
     h: ClassVar[OpType]  # value = <OpType.h: 4>
     i: ClassVar[OpType]  # value = <OpType.i: 2>
     iswap: ClassVar[OpType]  # value = <OpType.iswap: 23>
```
burgholzer, line 45: I believe you missed adding the new gate type here.

src/python/operations/register_optype.cpp
```diff
@@ -30,6 +30,7 @@ void registerOptype(py::module& m) {
       .value("rz", qc::OpType::RZ)
       .value("swap", qc::OpType::SWAP)
       .value("iswap", qc::OpType::iSWAP)
+      //      .value("iswapdg", qc::OpType::iSWAPdg)
```
burgholzer: Should be un-commented. There is no reason to not expose this gate to Python.

burgholzer (changes requested):
Also here: many thanks for this addition! This is looking really good already and I believe you managed to cover almost all the places.
I left some comments inline.
Other than that, it's really only the C++ coverage check that is complaining. You can directly inspect the coverage report by clicking on the respective link in the comment posted by codecov. I believe there should be further existing tests that you can add this new gate to in order to cover the missing lines.

src/python/operations/register_optype.cpp
BertiFlorea: Done ✅

src/mqt/core/_core/operations.pyi
BertiFlorea: Indeed, there was a commented line after iswap, which should have been the inverse.
Done ✅

src/mqt/core/_core/__init__.pyi
BertiFlorea: Done ✅

include/operations/OpType.hpp
BertiFlorea: I think we should still keep `"isw"` for iSWAP and `"isd"` for the inverse. The argument would be that `is` might be mistaken for **i**nverse S gate, but in the end, it's not that big of a deal. (At the moment it is `isw` and `isd` but I can change it if you think we should stick with the shorter one)

burgholzer (approved):
Looking good! many thanks!

Now that the new gate is in, a natural follow-up would be to consolidate the following methods:

PR #466, opened 2023-10-30
Added RZZ, RXX, RZX, and RYY to ZX-diagrams

pehamTom:
## Description

Added two-qubit rotations to ZX-diagrams.

## Checklist:


burgholzer:
Hey @pehamTom 👋🏻

Many thanks for these additions!
How hard would it be to also cover the remaining missing gates?
Especially the rotation gates rzx, xx-plus-yy, and xx-minus-yy, but also the ECR and DCX gate.

burgholzer:
Another small side question: is there any way to actually verify that these indeed realize the gate they intend to realize?
The current tests just assert that the gates cancel with their inverse, which isn't that much of a test after all.

pehamTom:
Don't know how to reply to your overall review directly. So here a quick reply:

Regarding the other gates: It is not hard to do but I just wanted to quickly add the gates Aaron needs.

Regarding tests: I am putting some tests into QCEC right now.

burgholzer:
> Don't know how to reply to your overall review directly. So here a quick reply:
> 
> 
> 
> Regarding the other gates: It is not hard to do but I just wanted to quickly add the gates Aaron needs.

Makes sense! Just thought it would be easiest if you are touching the code anyways 🙂 you know, context switching and so 😌

> Regarding tests: I am putting some tests into QCEC right now.

Great!

src/zx/FunctionalityConstruction.cpp
```diff
@@ -223,6 +224,43 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
...
+      addCnot(diag, ctrl, target2, qubits);
+      break;
+    }
+    case qc::OpType::RXX: {
+      const auto ctrl = static_cast<zx::Qubit>(p.at(op->getTargets()[1]));
+      const auto param = parseParam(op.get(), 0);
+      addCnot(diag, ctrl, target, qubits);
+      addXSpider(diag, ctrl, qubits, param);
+      diag.addGlobalPhase(-param / 2.0);
+      addCnot(diag, ctrl, target, qubits);
```
burgholzer: Just out of curiosity: is this really the most compact ZX Diagram you could construct for this gate? Aren't there any rules that allow you to combine some spiders here?
The same naturally holds for the other additions.
Quick googling showed me:
image
From https://arxiv.org/pdf/2012.13966.pdf which also seems to contain a more efficient realization for the CZ gate than we use currently. Am I missing something?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -223,6 +224,43 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
...
+      addXSpider(diag, target2, qubits, PiExpression(PiRational(1, 2)));
+      diag.addGlobalPhase(PiExpression{-PiRational(1, 2)});
+      addCnot(diag, ctrl, target2, qubits);
+      addZSpider(diag, target2, qubits, param);
+      diag.addGlobalPhase(-param / 2.0);
+      addCnot(diag, ctrl, target2, qubits);
+      addXSpider(diag, ctrl, qubits, PiExpression(-PiRational(1, 2)));
+      diag.addGlobalPhase(PiExpression{-PiRational(1, 2)});
+      addXSpider(diag, target2, qubits, PiExpression(-PiRational(1, 2)));
+      diag.addGlobalPhase(PiExpression{-PiRational(1, 2)});
```
burgholzer: Instead of multiple calls to the "addGlobalPhase" method, couldn't you just accumulate them in a single call at the beginning or the end of the function? (Also applies to the other methods)

Or, alternatively, couldn't you encapsulate those global phases in separate functions similar to the "addCnot" method? Like an "addPhaseGate" 🤷🏽‍♂️

burgholzer (changes requested):
Many thanks. Just some minor comments/questions. Overall LGTM, although the lack of proper testing is worrying me a bit here.

src/zx/FunctionalityConstruction.cpp
pehamTom: No you are not missing anything. The more compact representation basically saves one rewrite during equivalence checking. But I guess we can also use the more compact representations.

src/zx/FunctionalityConstruction.cpp
pehamTom: Sure. Guess this is a left over from debugging. Forgot to do that.

src/zx/FunctionalityConstruction.cpp
burgholzer: Yeah. Guess I'd prefer the more compact representation, even if it is only a minor performance boost.

src/zx/FunctionalityConstruction.cpp
```diff
@@ -223,6 +226,65 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
...
     }
+    case qc::OpType::RZZ: {
+      const auto ctrl = target;
+      const auto target2 = static_cast<zx::Qubit>(p.at(op->getTargets()[1]));
+      const auto param = parseParam(op.get(), 0);
+
+      addZSpider(diag, ctrl, qubits);
+      addZSpider(diag, target2, qubits);
+
+      auto mid_x = diag.addVertex(-1, -1, PiExpression(PiRational(0, 1)),
```
burgholzer: I know I am probably just picky again but it seems kind of strange to me that you "have" to add an artificial PiExpression that is just 0 here.
Wouldn't it be way nicer interface-wise, that I could just say `diag.addXVertex()` or something similar like that?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -223,6 +226,65 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
...
+                                  zx::VertexType::X);
+      auto mid_z = diag.addVertex(-1, -1, param, zx::VertexType::Z);
```
burgholzer: Couldn't both of these be const? Or are they modified by the `addEdge` method?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -223,6 +226,65 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
       addZSpider(diag, target2, qubits, PiExpression(), EdgeType::Hadamard);
       break;
     }
+    case qc::OpType::RZZ: {
+      const auto ctrl = target;
```
burgholzer: Now that there is no CNOT call here any more there is also no need for this copy any more (not that it was necessary from the beginning).

src/zx/FunctionalityConstruction.cpp
```diff
@@ -223,6 +226,65 @@ FunctionalityConstruction::parseOp(ZXDiagram& diag, op_it it, op_it end,
...
+
+      addZSpider(diag, ctrl, qubits);
+      addZSpider(diag, target2, qubits);
+
+      auto mid_x = diag.addVertex(-1, -1, PiExpression(PiRational(0, 1)),
+                                  zx::VertexType::X);
+      auto mid_z = diag.addVertex(-1, -1, param, zx::VertexType::Z);
+      diag.addEdge(qubits[static_cast<std::size_t>(ctrl)], mid_x);
+      diag.addEdge(qubits[static_cast<std::size_t>(target2)], mid_x);
+      diag.addEdge(mid_x, mid_z);
```
burgholzer: Now it's getting really picky: if you extract the RZZ code into a method, then you can just reuse that method here and safe some code duplication.

I believe it would generally be nicer to have more methods like the `addCnot` one because it creates a better separation of concerns and smaller methods.

burgholzer (changes requested):
Besides the clang-tidy warnings, I still have some minor nitpicks/requests. Please apply the reasoning for the RZZ gate also to all of the other ones.

Would it be too hard to also add the RZX gate here? I would suspect it's just a mixture of the two other ones.
Not so sure about the XX+-YY gates.

src/zx/FunctionalityConstruction.cpp
pehamTom: It was necessary because otherwise the linter complains that I use a variable called `target` for a parameter called `control`.

src/zx/FunctionalityConstruction.cpp
pehamTom: I would consider two instances of a couple of lines not a bad case of code duplication. Especially since this code is not expected to change anymore.

src/zx/FunctionalityConstruction.cpp
pehamTom: Yes.

src/zx/FunctionalityConstruction.cpp
pehamTom: I see all your points but I only provided this PR so that Aaron could compare against ZX for certain Ising models. I don't want this to turn into a clean-up of the ZX library.

src/zx/FunctionalityConstruction.cpp
pehamTom: Extracted everything into methods now.

src/zx/FunctionalityConstruction.cpp
burgholzer: Agreed. Could we still note that down somewhere so it is not lost in the void?

src/zx/FunctionalityConstruction.cpp
burgholzer: Just to clarify: clang-tidy is not authoritative in the sense that every warning it spits out has to be taken as gods word. It is totally fine to disagree with some of the warnings and ignore them on individual lines (as long as that does not become the norm and there is a reason for it).

burgholzer (approved):

PR #482, opened 2023-11-14
✨ Add `XX-plus-YY` and `XX-minus-YY` gate support to ZX library

### LINKED ISSUE #343: ✨ Enhance ZX gate support with newly introduced gates
#280 has brought support for many new gates. The ZX library is yet to be updated to natively support these gates.
As such, circuits containing these operations cannot be translated to ZXDiagrams at the moment.

All that's required to fix this, is to provide the corresponding transformations in the ZX library.

### LINKED ISSUE #486: 🐛 ZX PiExpression arithmetic not working properly
### Environment information

Any version of the ZX library.

### Description

The arithmetic for PiExpressions in the ZX package seems to be quite buggy in certain aspects.
For example
```c++
const auto beta = zx::PiRational(1, 2);

std::cout << beta * 2 << "\n";
std::cout << beta * 2. << "\n";

const auto betaExpr = zx::PiExpression{beta};
std::cout << betaExpr * 2. << "\n";
// std::cout << betaExpr * 2 << "\n"; // does not compile due to missing overload
```
yields
```console
1/1
1/1
159154943/500000000
```
The last entry is completely unexpected.

So that's the one thing. This also means something like `betaExpr / 2.` will result in complete garbage.
Now, #482 adds support for divisions by integers (which allows to write the above as `betaExpr / 2`). 
However, that revealed quite a problem in the way arithmetic expressions are handled. 
Namely, that 
```c++
const auto beta = zx::PiRational(-1, 2);
std::cout << (beta + beta) / 2 << "\n";  // and, if the above is fixed most likely also `(beta+beta)/2.` which currently returns garbage
```
prints
```console
1/2
```
instead of the desired result of `-1/2`.
The reason for that is `beta+beta` evaluating to `-1` which is `1 mod pi`.

### Expected behavior

I would expect division and multiplication by floating point numbers for `PiExpression` to work exactly as they do for `PiRational`.
Furthermore, I would expect moderately complex arithmetic expressions to produce the correct result.

### How to Reproduce

Run the examples from above

burgholzer:
## Description

This PR adds the last remaining unsupported gates to the ZX library.
In the process, it also introduces a new convenience function for integer division of parameter expressions that simplifies some code. In conjunction with #549, this also fixes #486.

Fixes #343 

## Checklist:

### DIFF
--- a/src/zx/FunctionalityConstruction.cpp
+++ b/src/zx/FunctionalityConstruction.cpp
@@ -168,6 +168,44 @@ void FunctionalityConstruction::addDcx(ZXDiagram& diag, const Qubit qubit1,
   addCnot(diag, qubit2, qubit1, qubits);
 }
 
+void FunctionalityConstruction::addXXplusYY(
+    ZXDiagram& diag, const PiExpression& theta, const PiExpression& beta,
+    const Qubit qubit0, const Qubit qubit1, std::vector<Vertex>& qubits) {
+  addRz(diag, beta, qubit1, qubits);
+  addRz(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRx(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addRz(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addZSpider(diag, qubit1, qubits, PiExpression(PiRational(1, 2)));
+  addCnot(diag, qubit0, qubit1, qubits);
+  addRy(diag, theta / 2, qubit0, qubits);
+  addRy(diag, theta / 2, qubit1, qubits);
+  addCnot(diag, qubit0, qubit1, qubits);
+  addZSpider(diag, qubit1, qubits, PiExpression(PiRational(-1, 2)));
+  addRz(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRx(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRz(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addRz(diag, -beta, qubit1, qubits);
+}
+
+void FunctionalityConstruction::addXXminusYY(
+    ZXDiagram& diag, const PiExpression& theta, const PiExpression& beta,
+    const Qubit qubit0, const Qubit qubit1, std::vector<Vertex>& qubits) {
+  addRz(diag, -beta, qubit1, qubits);
+  addRz(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRx(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addRz(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addZSpider(diag, qubit1, qubits, PiExpression(PiRational(1, 2)));
+  addCnot(diag, qubit0, qubit1, qubits);
+  addRy(diag, -theta / 2, qubit0, qubits);
+  addRy(diag, theta / 2, qubit1, qubits);
+  addCnot(diag, qubit0, qubit1, qubits);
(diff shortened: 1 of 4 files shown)


src/zx/FunctionalityConstruction.cpp
```diff
@@ -128,23 +149,60 @@ void FunctionalityConstruction::addRzx(ZXDiagram& diag,
...
+  addZSpider(diag, qubit1, qubits, PiExpression(PiRational(1, 2)));
+  addCnot(diag, qubit0, qubit1, qubits);
+  addRy(diag, theta / 2, qubit0, qubits);
+  addRy(diag, theta / 2, qubit1, qubits);
+  addCnot(diag, qubit0, qubit1, qubits);
+  addZSpider(diag, qubit1, qubits, PiExpression(PiRational(-1, 2)));
+  addRz(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRx(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRz(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addRz(diag, -beta, qubit1, qubits);
```
pehamTom: The Sdg and the RZ gate could be fused already in the construction.

src/zx/FunctionalityConstruction.cpp
```diff
@@ -128,23 +149,60 @@ void FunctionalityConstruction::addRzx(ZXDiagram& diag,
...
 }
 
+void FunctionalityConstruction::addXXplusYY(
+    ZXDiagram& diag, const PiExpression& theta, const PiExpression& beta,
+    const Qubit qubit0, const Qubit qubit1, std::vector<Vertex>& qubits) {
+  addRz(diag, beta, qubit1, qubits);
+  addRz(diag, PiExpression(PiRational(-1, 2)), qubit0, qubits);
+  addRx(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addRz(diag, PiExpression(PiRational(1, 2)), qubit0, qubits);
+  addZSpider(diag, qubit1, qubits, PiExpression(PiRational(1, 2)));
```
pehamTom: The S and the RZ on qubit1 can be fused by constructing the RZ with an additional phase of pi/2 and removing the `addZSpider`.

pehamTom (approved):
LGTM. I checked with the qiskit decompositions and everything seems alright with me. Just a tiny optimization but for me its fine to omit it. 

Another thing that I thought of is that Z rotations can be fused with the Control of CNOTs before or after (same for X rotations and targets). But I think that happens in more places than the added gates so it can be handled separately.

src/zx/FunctionalityConstruction.cpp
burgholzer: Done in https://github.com/cda-tum/mqt-core/pull/482/commits/5b061782e6394cfd7e08fc1638b088962848fbd8

src/zx/FunctionalityConstruction.cpp
burgholzer: Done in https://github.com/cda-tum/mqt-core/pull/482/commits/5b061782e6394cfd7e08fc1638b088962848fbd8

burgholzer:
> LGTM. I checked with the qiskit decompositions and everything seems alright with me. Just a tiny optimization but for me its fine to omit it.
> 
> Another thing that I thought of is that Z rotations can be fused with the Control of CNOTs before or after (same for X rotations and targets). But I think that happens in more places than the added gates so it can be handled separately.

Thanks for the feedback! I remember that I even tried that from the very beginning but I was getting some kind of weird results and couldn't get tests to pass.
Now that investigated a little more, I stumbled upon #486, which is kind of unfortunate and explains the errors I got on my first try.

burgholzer (commented):
@pehamTom many thanks for the changes here. Nice to see this go in!
Looking forward to the next QCEC release with the additional gate support and these fixes.

PR #487, opened 2023-11-27
[WIP] Partial equivalence checking

reb-ddm:
## Description

Added functionality for partial equivalence checking as described in the paper ["Partial Equivalence Checking of Quantum Circuits"](https://arxiv.org/abs/2208.07564) by Chen et al. It uses the decision diagram representation of the quantum circuit to check the equivalence conditions as described in the paper.

Some helper functions are in the file `Package.hpp` and the main partial equivalence checking function is in `Verification.hpp`.

## Checklist:


include/dd/FunctionalityConstruction.hpp
```diff
@@ -9,11 +9,15 @@ using namespace qc;
 
 template <class Config>
 MatrixDD buildFunctionality(const QuantumComputation* qc,
-                            std::unique_ptr<dd::Package<Config>>& dd);
+                            std::unique_ptr<dd::Package<Config>>& dd,
+                            bool reduceAncillaeQubits = true,
```
burgholzer: Nitpick: I believe the following would be a bit more "correct". Could you please adjust all occurrences?
```suggestion
                            bool reduceAncillaQubits = true,
```

include/dd/Verification.hpp
```diff
@@ -0,0 +1,66 @@
...
+                             std::unique_ptr<dd::Package<Config>>& dd) {
+
+  auto d1 = c1.getNqubitsWithoutAncillae();
+  auto d2 = c2.getNqubitsWithoutAncillae();
+  auto m1 = c1.getNqubits() -
+            static_cast<std::size_t>(std::count(c1.getGarbage().begin(),
+                                                c1.getGarbage().end(), true));
+  auto m2 = c2.getNqubits() -
+            static_cast<std::size_t>(std::count(c2.getGarbage().begin(),
+                                                c2.getGarbage().end(), true));
```
burgholzer: I think it would make sense to extract this into a method of the `QuantumComputation` class. Something like `getNgarbageQubits()`.

include/dd/Verification.hpp
```diff
@@ -0,0 +1,66 @@
...
+  // find the first garbage qubit at the end
+  for (Qubit i = 0U; i < n - static_cast<Qubit>(m1); i++) {
+    if (!garbage.at(i)) {
+      // swap it to the end
+      c1.swap(i, nextGarbage);
+      c2.swap(i, nextGarbage);
+      --nextGarbage;
+      nextGarbage = getNextGarbage(nextGarbage, garbage);
+    }
+  }
```
burgholzer: I haven't thought this through fully yet, but I believe it is (unfortunately) not as simple as that.
Any kind of adjustment (which should probably be encapsulated in a separate function) would have to make sure that
- the data qubits in both circuits are in the same place (I believe this should be already properly handled by the `buildFunctionality` method since it considers the initial layout.)
- that any measured qubit is in the same position in both circuits (this is not yet handled anyhow, but corresponding information can be inferred from the output permutation)

I'll keep it short here and refer to the big overall comment on the general mental model and qubit ordering.
For the beginning, we should focus on getting the case right, where we know that everything is perfectly set up.

include/dd/Package.hpp
```diff
@@ -2890,6 +2890,266 @@ template <class Config> class Package {
...
+  mEdge shiftAllRowsRecursive(const mEdge& e, std::int64_t m,
+                              std::int64_t offset) {
+
+    if (e.isTerminal()) {
+      return e;
+    }
+    if (m == 0 && offset == 0) {
+      return e;
+    }
+    mEdge eCopy{e.p, Complex::one()};
```
burgholzer: There are a couple of warnings reported by clang-tidy throughout the added code. Would be nice if they were fixed.
You can either see them on the "Files Changed" tab on GitHub or when you click on the "Details" button for the C++ linter CI run.

include/dd/Package.hpp
```diff
@@ -2890,6 +2890,266 @@ template <class Config> class Package {
...
+        resize(u.p->v + 1U + extra);
+      }
+      auto idExtra = makeIdent(extra);
+      u1 = kronecker(u, idExtra);
+    }
+    auto u2 = setColumnsToZero(u1, k);
+    auto u3 = shiftAllRows(u2, m);
+    auto u4 = multiply(conjugateTranspose(u1), u3);
+    auto u5 = setRowsToZero(u4, k);
+    return u5;
```
```suggestion
    return setRowsToZero(u4, k);
```

include/dd/Package.hpp
```diff
@@ -2890,6 +2890,266 @@ template <class Config> class Package {
...
+  mEdge partialEquivalenceCheckSubroutine(mEdge& u, Qubit m, Qubit k,
+                                          Qubit extra) {
+    auto u1{u};
+    // add extra ancillary qubits
+    if (extra > 0) {
+      if (u.p->v + 1U + extra > nqubits) {
+        resize(u.p->v + 1U + extra);
+      }
+      auto idExtra = makeIdent(extra);
+      u1 = kronecker(u, idExtra);
```
```suggestion
      u1 = kronecker(u, makeIdent(extra));
```

include/dd/Package.hpp
```diff
@@ -2890,6 +2890,266 @@ template <class Config> class Package {
...
+    // add qubits such that u1 and u2 have the same dimension
+    if (u1.isTerminal()) {
+      u1 = kronecker(u1, makeIdent(u2.p->v + 1));
+    } else if (u2.isTerminal()) {
+      u2 = kronecker(u2, makeIdent(u1.p->v + 1));
+    } else if (u1.p->v < u2.p->v) {
+      u1 = kronecker(u1, makeIdent(u2.p->v - u1.p->v));
+    } else if (u1.p->v > u2.p->v) {
+      u2 = kronecker(u2, makeIdent(u1.p->v - u2.p->v));
+    }
```
burgholzer: Just to point it out. I believe this is one of the places where qubits are added "on the wrong side".

include/dd/Package.hpp
```diff
@@ -2890,6 +2890,266 @@ template <class Config> class Package {
...
+
+    const Qubit h = u1.p->v + 1;
```
burgholzer: Nitpick: I'd prefer to stay closer to the paper terminology and refer to this as `n`.

include/dd/Package.hpp
```diff
@@ -2890,6 +2890,266 @@ template <class Config> class Package {
...
+      if (u.p->v + 1U + extra > nqubits) {
+        resize(u.p->v + 1U + extra);
+      }
+      auto idExtra = makeIdent(extra);
+      u1 = kronecker(u, idExtra);
+    }
+    auto u2 = setColumnsToZero(u1, k);
+    auto u3 = shiftAllRows(u2, m);
+    auto u4 = multiply(conjugateTranspose(u1), u3);
+    auto u5 = setRowsToZero(u4, k);
```
burgholzer: just wondering: you pass `k` to both of the methods that set something to zero. In the paper, they pass `d` to those methods. Any particular reason for the difference here?

burgholzer (changes requested):
Hey 👋🏼 

Many thanks for this initial draft! I now took some time to think this all through and to also think about how the algorithm from the original paper matches with the conventions employed throughout the MQT.
My current conclusion is that there is a slight mismatch between the conventions we typically use and assume and the paper and that this mismatch leads to the overall inefficiency you saw in your results.

Let me first lay out the mental model of how we consider circuits and the qubit ordering, which qubits go where, etc.:
- Qubits in an $n$-qubit circuit are indexed $q_{n-1}, q_{n-2}, \dots, q_1, q_0$, with $q_{n-1}$ denoting the "most-significant" qubit.
- When we draw circuits, then $q_0$ is always the bottom-most qubit while $q_{n-1}$ is the top-most qubit. Adding further qubits to a circuit, thus, expands the circuit by additional lines towards the top
- If an $n$-qubit circuit contains $d$ data qubits and $k=n-d$ ancillary qubits, then the default assumption is that $q_0,\dots,q_{d-1}$ are the data qubits and $q_d,\dots,q_{n-1}$ are the ancillary qubits.
- If an $n$-qubit circuit contains $m$ measured qubits and $g=n-m$ garbage qubits, then the default assumption is that $q_0,\dots,q_{m-1}$ are the measured qubits and $q_m,\dots,q_{n-1}$ are the garbage qubits.
- In any kind of tensor product (whether between vectors or matrices), the leftmost entry corresponds to the most-significant qubit. For example, $X\otimes Y \otimes Z$ means $X$ is applied to qubit $q_2$, $Y$ to qubit $q_1$, and $Z$ to qubit $q_0$.
- Any $n$-qubit computational basis state $|i\rangle$ is to be read as $|i\rangle=|i_{n-1}\rangle\otimes \dots |i_{0}\rangle = \vert\sum_{j=1,\dots,n-1} i_j \cdot 2^j\rangle$
- In an $n$-qubit decision diagram (DD), the default qubit ordering is $q_{n-1}, q_{n-2}, \dots, q_1, q_0$, that is, $q_{n-1}$ is the top-most qubit of the DD, while $q_0$ is the bottom-most qubit in the DD.

Having said the above, I believe that the current state of the PR mixes up some of the conventions.
- Circuits should be extended with additional qubits "towards the top", i.e., tensor products should always have the form $I \otimes U$. The current code repeatedly uses $U \otimes I$, which, coincidentally, is also significantly slower than the former.
- Step 5 in Alg.1 of the paper then translates to just keeping the first $2^d$ columns of the original matrix and setting everything else to zero. That is precisely what the `reduceAncillae` method is doing, so you should be able to reuse precisely that. What's more is that I believe this step is implicitly already taken care of since https://github.com/cda-tum/mqt-core/blob/d181d4576f40e13509bf2d52e25c48d6f34d8bfc/src/dd/FunctionalityConstruction.cpp#L17 constructs an initial matrix, where the ancillary qubits are already zeroed out (given that they are set as ancillary in the circuit). I haven't thought this through perfectly, but it think that even makes the `reduceAncillae` call at the end superfluous since a column that is initially zero can not become non-zero via applying gates. I might be wrong here though, so please double check.
- Now comes the tricky part. For Step 6, some of the rows in the matrix should be shifted. As far as I understood the algorithm, the purpose of this is to separate the individual contributions of the garbage parts. It might be that the realization of that based on our conventions is also kind of flipped so that you shift the $i^{th}$ group of rows by $i*2^d$ and form the groups a little different.
- The same probably goes for Step 8.

As you see, my thoughts get a lot more vague towards the end. But, hopefully, this gives you some food for thought on how to change things up. The inline comments below contain some small details and other miscellaneous feedback.

include/dd/FunctionalityConstruction.hpp
reb-ddm: I called it "ancillae" in order to be consistent with the name of the "reduceAncillae" function. Maybe it would make more sense to instead rename the "zeroAncillaPartialEquivalence" to "zeroAncillaePartialEquivalence"?

burgholzer: "Ancillae" is just the (latin) plural of "Ancilla". 
One can either refer to "Ancillae" or to "Ancilla Qubits" (or "Ancillary Qubits").
I believe at the moment we use all different versions.
Bringing some consistency into this would be appreciated.
I still think it should be either "reduceAncillaQubits" or "reduceAncillae" for the parameter above, but I agree with your proposal of "zeroAncillaePartialEquivalence".

reb-ddm: Ah ok I understand, "ancilla" or "ancillary" are adjectives and "ancillae" is the plural noun. Then I think it would be even more consistent with the already existing functions if I call it "reduceAncillaryQubits" (I can't call it "reduceAncillae", because there is already a function with this name).

include/dd/Verification.hpp
reb-ddm: For now I just assumed that these two preconditions are already satisfied (and I also wrote it as a comment over the function), because I thought that some code from [mqt-qcec](https://github.com/cda-tum/mqt-qcec) can be reused to enforce these preconditions.

include/dd/FunctionalityConstruction.hpp
burgholzer: That suggestion sounds good to me 👍🏼

include/dd/Verification.hpp
burgholzer: I guess it is fine to assume the pre-conditions for now.
As stated above, the first one is kind of handled naturally in mqt-core as well as mqt-qcec via the `initialLayout`.
The latter pre-condition is not enforced at all, but i think it should hold for most practical cases (where the first `n` logical qubits are the data qubits).
If these pre-confitions are assumed anyway, then I guess it is fair to remove the above code snippet.

include/mqt-core/QuantumComputation.hpp
```diff
@@ -306,6 +306,14 @@ class QuantumComputation {
   [[nodiscard]] std::size_t getNqubitsWithoutAncillae() const {
     return nqubits;
   }
+  [[nodiscard]] std::size_t getNmeasuredQubits() const {
+    return static_cast<std::size_t>(
+        std::count(getGarbage().begin(), getGarbage().end(), false));
```
burgholzer: I think it might be safer to compute this as `getNqubits()-getNgarbageQubits()` since the garbage array might be longer than the number of qubits.

include/mqt-core/dd/Package.hpp
```diff
@@ -2439,6 +2439,260 @@ template <class Config> class Package {
...
+      **/
+  bool zeroAncillaePartialEquivalenceCheck(const mEdge& u1, const mEdge& u2,
+                                           const Qubit m) {
+    auto u1u2 = multiply(u1, conjugateTranspose(u2));
+    const Qubit n = u1.p->v + 1;
+    std::vector<bool> garbage(n, false);
+    for (size_t i = m; i < n; i++) {
+      garbage[i] = true;
+    }
+    return isCloseToIdentity(u1u2, 1.0E-10, garbage, false);
```
burgholzer: I think I'd rather move these methods to the Verification header. The Package file is already quite cluttered and growing with almost every PR. 
Technically, these functions could even go directly into QCEC. However, that might make testing hard here. Interested in your opinion here.

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+  const fp randomParameter3 =
+      randomParameters[static_cast<size_t>(rand()) % randomParameters.size()];
+  return convertToStandardOperation(
+      n, nrQubits, randomOpType, randomTarget1, randomTarget2, randomParameter1,
+      randomParameter2, randomParameter3, randomControls);
+}
+
+std::pair<qc::QuantumComputation, qc::QuantumComputation>
+generateRandomBenchmark(const size_t n, const Qubit d, const Qubit m) {
+  if (d > n) {
```
burgholzer: This just checks the number of data qubits. It should probably also check the number of measured qubits.

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+  // 1) H gates
+  for (Qubit i = 0U; i < d; i++) {
+    circuit1.h(i);
+    circuit2.h(i);
+  }
+
+  for (Qubit i = 0U; i < static_cast<Qubit>(n); i++) {
+    circuit1.barrier(i);
+    circuit2.barrier(i);
+  }
```
burgholzer: I think you should be able to just call `qc.barrier()` instead of the loop to add a barrier to all qubits.

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+  // 4) Arbitrary gates
+  //    arbitrary gates are added to not measured qubits
+  if (d > m) {
+    const Qubit notMQubits = d - m;
```
burgholzer: Shouldn't this be `n` instead of `d`?

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+    for (Qubit i = 0U; i < notMQubits; i++) {
+      const auto op = makeRandomStandardOperation(n, notMQubits, m);
+      addStandardOperationToCircuit(circuit1, op, false);
+    }
+    for (Qubit i = 0U; i < notMQubits; i++) {
+      const auto op = makeRandomStandardOperation(n, notMQubits, m);
+      addStandardOperationToCircuit(circuit2, op, false);
+    }
```
burgholzer: These two loops could just be combined.
```suggestion
    for (Qubit i = 0U; i < notMQubits; i++) {
      addStandardOperationToCircuit(circuit1, makeRandomStandardOperation(n, notMQubits, m), false);
      addStandardOperationToCircuit(circuit2, makeRandomStandardOperation(n, notMQubits, m), false);
    }
```

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  for (Qubit i = 0U; i < static_cast<Qubit>(n); i++) {
+    circuit1.barrier(i);
+    circuit2.barrier(i);
+  }
+
+  // 5) CNOT gates (if there are ancilla qubits)
+
+  if (d > 0) {
```
burgholzer: Wouldn't `n>d` work better here?

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+  for (auto t : op.getTargets()) { // the targets are at most 2
+    targets.push_back(static_cast<Qubit>(t));
+  }
+
+  if (op.getType() == X && controls.size() == 2 && decomposeCcx) {
+    // decompose toffoli gate
+    addDecomposedCcxGate(circuit, controls[0], controls[1], targets[0]);
+  } else {
+    circuit.emplace_back<StandardOperation>(op);
+  }
```
burgholzer: The logic here can be improved a little by flipping the if condition.
If the decomposition flag is false, one can just add the operation. Also, If the operation is not an X or does not have two controls, one can return early.

In the remaining case, there should be no reason to construct extra target an controls vectors. You can simply change the interface of the decomposition method to take Controls and a target.

src/dd/Verification.cpp
```diff
@@ -0,0 +1,327 @@
...
+}
+
+std::vector<Qubit> fiveDiffferentRandomNumbers(const Qubit min,
+                                               const Qubit max) {
+  std::vector<Qubit> numbers;
+
+  for (Qubit i = min; i < max; i++) {
+    numbers.push_back(i);
```
burgholzer:
```suggestion
    numbers.emplace_back(i);
```
Always prefer `emplace_back` over `push_back` when inserting elements into stl containers.

test/algorithms/test_partial_equivalence.cpp
```diff
@@ -0,0 +1,420 @@
...
+  const qc::QuantumComputation c1{
+      "./circuits/grover-noancilla_nativegates_ibm_qiskit_opt0_3.qasm"};
+  const qc::QuantumComputation c2{
+      "./circuits/grover-noancilla_indep_qiskit_3.qasm"};
+
+  // 3 measured qubits and 3 data qubits, full equivalence
+  EXPECT_TRUE(dd->partialEquivalenceCheck(
+      buildFunctionality(&c1, *dd, false, false),
+      buildFunctionality(&c2, *dd, false, false), 3, 3));
+}
```
burgholzer: I think I would prefer to move all the tests that involve files to QCEC once the corresponding feature is added there.
It should be enough to test the basic functionality here and leave more elaborate tests/benchmarks to the top-level repository.
This also means, that you can remove the `circuits` directory.

test/algorithms/test_partial_equivalence.cpp
```diff
@@ -0,0 +1,420 @@
...
+                  static_cast<double>(reps) / 1000000.)
+              << " seconds; average number of gates = "
+              << (static_cast<double>(totalGates) / static_cast<double>(reps))
+              << "\n";
+  }
+}
+
+TEST(PartialEquivalenceTest, Benchmark) {
+  const auto dd = std::make_unique<dd::Package<>>(20);
+  srand(55);
```
burgholzer: I am really surprised that clang-tidy is not complaining throughout this file about the direct use of `rand` and friends.
I'd much rather prefer to use the modern C++ equivalents of these constructs form `<random>`. Most of the things here can be solved with `uniform_int_distribution`.
Note that every `QuantumComputation` that is initiated has a properly seeded RNG generator. You can just use `qc.getGenerator()` to get it.

This comment naturally also applies to the code where these random benchmarks are generated.

test/algorithms/test_partial_equivalence.cpp
```diff
@@ -0,0 +1,420 @@
...
+  const auto dd = std::make_unique<dd::Package<>>(30);
+  const size_t minN = 3;
+  const size_t maxN = 15;
+  const size_t reps = 10;
+  std::cout << "Zero-ancilla partial equivalence check\n";
+  partialEquivalencCheckingBenchmarks(dd, minN, maxN, reps, false);
+}
+
+TEST(PartialEquivalenceTest, InvalidInput) {
+  const auto dd = std::make_unique<dd::Package<>>(30);
```
burgholzer:
```suggestion
  const auto dd = std::make_unique<dd::Package<>>(4);
```
No need to allocate a package larger than necessary.

test/algorithms/test_partial_equivalence.cpp
```diff
@@ -0,0 +1,420 @@
...
+              << (static_cast<double>(totalTime.count()) /
+                  static_cast<double>(reps) / 1000000.)
+              << " seconds; average number of gates = "
+              << (static_cast<double>(totalGates) / static_cast<double>(reps))
+              << "\n";
+  }
+}
+
+TEST(PartialEquivalenceTest, Benchmark) {
+  const auto dd = std::make_unique<dd::Package<>>(20);
```
burgholzer: The DD package for the benchmarks should actually be created inside the loop in the function above instead of here.
Reusing the same DD package multiple times can lead to all kinds of performance degradation due to memory fragmentation and other phenomena.

Furthermore, you only ever need to allocate a package with as many qubits as you maximally need, which you can always compute in the loop above.

Side note: I would put the Package creation outside of the time measurement.

burgholzer (changes requested):
Hey 👋🏼 

Sorry for the long wait on the review.
This is just a first pass through the changes. I haven't really managed to look through the methods themselves in detail. I'll do that in a follow-up review.
However, this should un-stall your progress and give you some pointers on aspects that might still need some improvement.
Let me know if anything is unclear 👍🏼

burgholzer:
Ah, and btw., that remaining clang-tidy warning is a frequent false positive that should hopefully be resolved by now if you rebase your changes on the main branch or merge the main branch in.

src/dd/Verification.cpp
reb-ddm: In the paper, these random gates are only applied on the data qubits that are not measured, not to the ancillary qubits. Therefore I added the condition `d > m`. I could add this information in the comment.

src/dd/Verification.cpp
reb-ddm: If `n>d` then the loop would simply be executed zero times, so it is not necessary to check this beforehand. This loop adds a CNOT gate for each ancilla qubit with the ancilla qubit as a control qubit and any data qubit as a target. However, if there are no data qubits, there can't be any CNOT gates with a data qubit as a target. More concretely, the modulo operation a few lines later would result in an exception, if `d==0`.

src/dd/Verification.cpp
burgholzer: Ah, ok. In that case, this makes sense. A comment would be nice 👍🏻

src/dd/Verification.cpp
burgholzer: Ok, then maybe an update to the comment might help.
The comment above refers to there being ancillary qubits, and then the first check is whether there are any data qubits. That seems confusing to me.

test/algorithms/test_partial_equivalence.cpp
reb-ddm: I assume that you mean that I should remove the circuits from the directory, not remove the whole directory, as it was already there beforehand 🤔

burgholzer: Exactly 😌

reb-ddm:
I noticed that I needed to use `static_cast` very often in my code, mainly because `dd::Qubit` is defined as `uint16_t` and it often needs to be compared to variables of type `size_t`. But then I found out that there is a type `qc::Qubit`, which is defined as `uint32_t`. Changing my code to utilize `qc::Qubit` instead of `dd::Qubit` made me remove most of the casts. Is there a reason that these two types with the same name are different? And which one is supposed to be used when?

burgholzer:
> I noticed that I needed to use `static_cast` very often in my code, mainly because `dd::Qubit` is defined as `uint16_t` and it often needs to be compared to variables of type `size_t`. But then I found out that there is a type `qc::Qubit`, which is defined as `uint32_t`. Changing my code to utilize `qc::Qubit` instead of `dd::Qubit` made me remove most of the casts. Is there a reason that these two types with the same name are different? 

Yeah and it's kind of implied by the different namespace. In the general `qc::` namespace, we want to be able to represent rather large circuits and space is not really too much of a concern. Thus, we use unsigned 32bit integers there, which allows for up to roughly 4 billion qubits.
In the `dd::` namespace we want to be as efficient as possible since a variable of that Qubit type is part of every single DD Node. In the past, we used to define that qubit type as `int8_t`, which would actually limit us to 128 qubits. Since there was some demand for realizing larger simulations than that, we decided to with `uint16_t` for the DD Qubit type.

> And which one is supposed to be used when?

Any code that is part of `mqt-core` and does not concern DDs should be using the wider `qc::Qubit` type. Everything that involves DDs can/should be using the narrower `dd::Qubit` type. However, that's not really dogmatic.

include/mqt-core/dd/Package.hpp
reb-ddm: I think the `zeroAncilla` function is probably not even necessary, given that QCEC directly calls `isCloseToIdentity` in the AlternatingChecker. I removed the function `zeroAncillaePartialEquivalenceCheck` in `Package.hpp`, which basically just performed a multiplication and then called `isCloseToIdentity`. I rewrote the corresponding tests to be tests for `isCloseToIdentity`. 

I moved the `partialEquivalenceCheck` to `Verification.hpp`. I'm not sure how it will be used in QCEC, but for now it mostly calls the functions `setColumnsToZero` and `shiftAllRows` (after a bit of preprocessing). Therefore I think it would be possible to only test these two functions in MQT-Core and move the functions and tests for `partialEquivalenceCheck` to QCEC. At least that's my opinion, but I'm not very familiar with the QCEC code yet.

burgholzer: > I think the `zeroAncilla` function is probably not even necessary, given that QCEC directly calls `isCloseToIdentity` in the AlternatingChecker. I removed the function `zeroAncillaePartialEquivalenceCheck` in `Package.hpp`, which basically just performed a multiplication and then called `isCloseToIdentity`. I rewrote the corresponding tests to be tests for `isCloseToIdentity`.

Yeah. That makes a lot of sense.

> I moved the `partialEquivalenceCheck` to `Verification.hpp`. I'm not sure how it will be used in QCEC, but for now it mostly calls the functions `setColumnsToZero` and `shiftAllRows` (after a bit of preprocessing). Therefore I think it would be possible to only test these two functions in MQT-Core and move the functions and tests for `partialEquivalenceCheck` to QCEC. At least that's my opinion, but I'm not very familiar with the QCEC code yet.

That also sounds very reasonable and I would be happy with that change 👍🏼

burgholzer:
Closing this in favor of #563 for now. We can always re-open if necessary.

PR #489, opened 2023-11-28
:construction: added draft implementation for gates interface

### LINKED ISSUE #484: :recycle: Refactor handling of gate matrices and inverses
After #460, all gates in MQT Core have a proper inverse now.
A natural follow-up would be to consolidate the following methods:

- Maybe the above actually makes some of the operations (such as `iSWAPdg` redundant again.

_Originally posted by @burgholzer in https://github.com/cda-tum/mqt-core/pull/460#pullrequestreview-1738494988_

BertiFlorea:
## Description

In #484, the need for a new interface for gate matrices and inverses was presented in order to reduce the redundancy created by calling the inverse of an operation. (In other words, instead of having the switches and decide which operation type it is, we want to call something like `<our-operation/gate>.getInverseGateMatrix()`.

Fixes #484 

## Checklist:

- [X] The pull request only contains commits that are related to it.


include/operations/gates/GateMatrixInterface.hpp
```diff
@@ -0,0 +1,20 @@
...
+class GateMatrixInterface {
+public:
+  virtual ~GateMatrixInterface() = default;
+
+  virtual dd::GateMatrix getGateMatrix() = 0;
+  //  virtual dd::TwoQubitGateMatrix getGateMatrix() = 0;
+  virtual bool isSingleTargetGate() = 0;
+  virtual bool isTwoTargetGate() = 0;
+  virtual bool isThreeOrMoreTargetGate() = 0;
+  virtual dd::GateMatrix getInverseGateMatrix() = 0;
```
burgholzer: In order to solve the problem with the single-target and two-target gates you could make this class a template class that takes the type of the matrix as a template parameter. Something like:
```suggestion
namespace qc {
template<typename MatrixType>
class GateMatrixInterface {
public:
  virtual ~GateMatrixInterface() = default;

  virtual MatrixType getGateMatrix() = 0;
  virtual MatrixType getInverseGateMatrix() = 0;
  virtual bool isSingleTargetGate() = 0;
  virtual bool isTwoTargetGate() = 0;
  virtual bool isThreeOrMoreTargetGate() = 0;
```
Then, `HGate` would become
```c++
class HGate : public GateMatrixInterface<GateMatrix>, StandardOperation { ... };
```
while the `SWAPGate` would become
```c++
class SWAPGate : public GateMatrixInterface<TwoQubitGateMatrix>, StandardOperation { ... };
```

burgholzer (commented):
First of all, thanks for taking the time to look into this and play around with various ideas.
I looked through it and made up my mind a little.
Sorry for the long "essay" 🙃 

It is definitely not a simple challenge.
Hence, I decided to try and ask ChatGPT for some insight and it came up with something rather interesting (see the two screenshots below).
95620733-4d3a-47a4-af99-18847f980aed
da65292a-27d6-42fc-8b22-8a271c07b69b

I don't quite like the naming and all of that, but the coding pattern itself seems rather reasonable to me. It would avoid the multiple inheritance (which is frequently frowned upon) and also create a clear separation of concerns.
Maybe something to think about.

I also added an inline comment on how to potentially solve the single- vs. multi-target gate matrices. Maybe it is helpful.

The problem I see with the dictionary approach is that some of these methods (like getting the inverse) need almost all information of the Operation to properly return the right result. At that point, these functions might as well be implemented in the classes themselves (which some of them are already).

Technically, I would be fine with making `StandardOperation` an abstract base class and deferring the details to the respective derived Operation classes. However, I would expect that this requires quite some updates throughout the code base.
Although, I have to say that if it turns out that those changes are mostly in the test files and not in the library itself, this could be an effort that is very much worth it.

Just some more context to why getting this right might become even more important in the future: OpenQASM 3.0 brought along quite a few new modifiers that can be used to construct gates. This includes the `ctrl` modifier (for adding controls to a gate), the `inv` modifier (for inverting gates), and the `pow` modifier (for creating powers of a certain gate). All of these have reasonable default implementations that can be applied to arbitrary operations.
Some, however, can be implemented much more efficiently if the specific gate is known. For example, an integer power of a rotation gate is just the same rotation gate but with an angle that is multiplied by that integer.
Finding a way how to specify such special cases and performance improvements without creating ever growing functions with increasing complexity and branching would be really great.

All of the above is really just food for thought. One has to keep in mind that the intent of this change is to make the code more maintainable, extendable, readable, etc. and definitely not to make it more complicated, harder to work with or less performant.

burgholzer:
Thanks @BertiFlorea for the nice summary.
As discussed, I am closing this for now.
Looking forward to whatever you come up with as a follow up.

PR #491, opened 2023-12-01
🧹 Noise-Aware Simulator Cleanup 

33Gjl1Xe:
## Description

This PR relates to this [PR](https://github.com/cda-tum/mqt-ddsim/pull/321) in the Simulator.

More specifically the PR refactors some functionality related to noise-aware quantum circuit simulation, i.e.:
* I removed legacy functionality that allowed running the simulator in a less optimized version.
* I added a function for collapsing measurements of density matrices.

## Checklist:


burgholzer:
Hey 👋🏼 
#444 just landed (finally). This simplified quite some code throughout the DD package.
Could you please rebase your changes on top of the current main branch?
Happy to take a look then!

include/mqt-core/dd/Package.hpp
```diff
@@ -1165,6 +1165,71 @@ template <class Config> class Package {
...
+    for (dd::Qubit p = 1; p < static_cast<dd::Qubit>(qubits()); p++) {
+      if (p == index) {
+        f = makeDDNode(static_cast<dd::Qubit>(0), measureMatrix);
+      } else {
+        f = makeDDNode(static_cast<dd::Qubit>(0), identityMatrix);
+      }
+      measOp = kronecker(f, measOp);
+    }
+    return measOp;
+  }
```
burgholzer: This method should not be needed. You can just use a code sequence similar to https://github.com/cda-tum/mqt-core/blob/fc7d8660e6d2007befbcdf70abdad0e845e67ae8/include/mqt-core/dd/Package.hpp#L1181-L1184
to implement this in a way more compact fashion.

include/mqt-core/dd/Package.hpp
```diff
@@ -1165,6 +1165,71 @@ template <class Config> class Package {
...
+    e = tmp2;
+    dEdge::setDensityMatrixTrue(e);
+
+    // Normalize density matrix
+    auto result = e.w / densityMatrixTrace;
+    cn.decRef(e.w);
+    e.w = cn.lookup(result);
+    cn.incRef(e.w);
+    return {e, measuredResult};
+  }
```
burgholzer: This method has a slightly different interface then the regular method for vectors.
AFAICT, this need not be the case and you could just be returning the measurement result. `e` is anyway captures by reference and modified as part of the method.

include/mqt-core/dd/Package.hpp
```diff
@@ -1786,7 +1845,7 @@ template <class Config> class Package {
     auto r = trace(a, eliminate);
     return cn.lookup(r);
   }
-  ComplexValue trace(const mEdge& a) {
+  template <class Edge> ComplexValue trace(const Edge& a) {
```
burgholzer: I would slightly prefer to make this a `template <class Node>` and use `Edge<Node>` in order to get better IDE and code completion support because then the compiler at least knows that the parameter is some kind of Edge.

include/mqt-core/dd/Package.hpp
```diff
@@ -1814,17 +1873,19 @@ template <class Config> class Package {
 
 private:
   /// TODO: introduce a compute table for the trace?
-  mCachedEdge trace(const mEdge& a, const std::vector<bool>& eliminate,
-                    std::size_t alreadyEliminated = 0) {
+  template <class OperandNode>
+  CachedEdge<OperandNode> trace(const Edge<OperandNode>& a,
+                                const std::vector<bool>& eliminate,
+                                std::size_t alreadyEliminated = 0) {
```
burgholzer: Purely a consistency thing, but I think throughout most of the library, this template parameter is just referred to as `Node`

test/dd/test_dd_noise_functionality.cpp
```diff
@@ -90,44 +90,101 @@ TEST_F(DDNoiseFunctionalityTest, DetSimulateAdder4TrackAPD) {
...
+    for (const auto& j : result) {
       for (const auto& [key, value] : j) {
         EXPECT_NEAR(value, reference.at(key), TOLERANCE);
       }
     }
   }
 }
 
+TEST_F(DDNoiseFunctionalityTest, testingMeasure) {
+  std::mt19937_64 mt{0}; // NOLINT(cert-msc51-cpp)
```
burgholzer: No real need to define that here. Every QuantumComputation comes with a properly seeded RNG that you can get via the `getGenerator` method (https://github.com/cda-tum/mqt-core/blob/fc7d8660e6d2007befbcdf70abdad0e845e67ae8/include/mqt-core/QuantumComputation.hpp#L314).

test/dd/test_dd_noise_functionality.cpp
```diff
@@ -90,44 +90,101 @@ TEST_F(DDNoiseFunctionalityTest, DetSimulateAdder4TrackAPD) {
...
+  for (auto const& op : qcOp) {
+    dd->applyOperationToDensity(rootEdge, dd::getDD(op.get(), *dd));
+    deterministicNoiseFunctionality.applyNoiseEffects(rootEdge, op);
+  }
+  char result = '2';
+
+  const double tolerance = 1e-10;
+
+  auto tmp = rootEdge.getSparseProbabilityVectorStrKeys();
+  auto prob = 1.0 / 8;
```
burgholzer:
```suggestion
  auto prob = 0.125;
```
Just because it's so easy to write down exactly.

test/dd/test_dd_noise_functionality.cpp
```diff
@@ -90,44 +90,101 @@ TEST_F(DDNoiseFunctionalityTest, DetSimulateAdder4TrackAPD) {
...
+  EXPECT_NEAR(tmp["011"], prob, tolerance);
+  EXPECT_NEAR(tmp["100"], prob, tolerance);
+  EXPECT_NEAR(tmp["101"], prob, tolerance);
+  EXPECT_NEAR(tmp["110"], prob, tolerance);
+  EXPECT_NEAR(tmp["111"], prob, tolerance);
+
+  std::tie(rootEdge, result) = dd->measureOneCollapsing(rootEdge, 0, mt);
+
+  auto tmp0 = rootEdge.getSparseProbabilityVectorStrKeys();
+  prob = 1.0 / 4;
```
burgholzer:
```suggestion
  prob = 0.25;
```

test/dd/test_dd_noise_functionality.cpp
```diff
@@ -90,44 +90,101 @@ TEST_F(DDNoiseFunctionalityTest, DetSimulateAdder4TrackAPD) {
...
+
+  EXPECT_TRUE(fabs(tmp0["000"] + tmp0["001"] - prob) < tolerance);
+  EXPECT_TRUE(fabs(tmp0["010"] + tmp0["011"] - prob) < tolerance);
+  EXPECT_TRUE(fabs(tmp0["100"] + tmp0["101"] - prob) < tolerance);
+  EXPECT_TRUE(fabs(tmp0["110"] + tmp0["111"] - prob) < tolerance);
+
+  std::tie(rootEdge, result) = dd->measureOneCollapsing(rootEdge, 1, mt);
+
+  auto tmp1 = rootEdge.getSparseProbabilityVectorStrKeys();
+  prob = 1.0 / 2;
```
burgholzer:
```suggestion
  prob = 0.5;
```

burgholzer (changes requested):
This is looking really good and makes the code quite a lot simpler. So, many thanks for that!
I just have a couple of very minor comments that should be pretty straight-forward to address. After that, this should be good to go!

include/mqt-core/dd/Package.hpp
33Gjl1Xe: I changed the code as suggested.
However, I use nqubits instead of rootEdge.p->v + 1U.

include/mqt-core/dd/Package.hpp
33Gjl1Xe: You are correct, the interface could be changed to be identical to the regular measure method. However, I don't make use of the variables assumeProbabilityNormalization and epsilon. While I could mark these variables unused, it feels wrong to introduce unused variables simply for similarity. 

Are you sure, you want me to make this change?

include/mqt-core/dd/Package.hpp
33Gjl1Xe: I changed the code as suggested.

include/mqt-core/dd/Package.hpp
33Gjl1Xe: I changed the code as suggested.

test/dd/test_dd_noise_functionality.cpp
33Gjl1Xe: I changed the code as suggested.

test/dd/test_dd_noise_functionality.cpp
33Gjl1Xe: I changed the code as suggested.

test/dd/test_dd_noise_functionality.cpp
33Gjl1Xe: I changed the code as suggested.

test/dd/test_dd_noise_functionality.cpp
33Gjl1Xe: I changed the code as suggested.

include/mqt-core/dd/Package.hpp
burgholzer: That will not yield the correct result in general. You might have instantiated a package with 5 qubits, but only use 3 qubits in a simulation.
You'd have to use the variable index of the DD to get the proper qubit number.

In the same spirit, you can just completely inline this method as you are only using it in a single place.

include/mqt-core/dd/Package.hpp
burgholzer: The `assumeProbabilityNormalization` parameter should just be removed IMO. It is _always_ true. But this does not happen to need here.
What I actually meant here is to "just" change the return type of the function here to `char`.

33Gjl1Xe: Oh, I see. I changed the interface as suggested.

include/mqt-core/dd/Package.hpp
33Gjl1Xe: Good catch. I fixed the code as you suggested.

burgholzer:
Ah, and before I forget it: could you please add an appropriate PR description here so that one can roughly understand what has been changed in this PR?
Once you pushed your changes and updated the description, this should be good to go!

burgholzer:
Note that the Python failures here are unrelated to your changes. They stem from the recent Qiskit update to 0.46.0.
I will quickly create a PR that fixes them.

burgholzer:
> Note that the Python failures here are unrelated to your changes. They stem from the recent Qiskit update to 0.46.0. I will quickly create a PR that fixes them.

See https://github.com/cda-tum/mqt-core/pull/544. Once that is in, you should be able to merge that into your branch here and everything should be working again.

33Gjl1Xe:
> > Note that the Python failures here are unrelated to your changes. They stem from the recent Qiskit update to 0.46.0. I will quickly create a PR that fixes them.
> 
> See #544. Once that is in, you should be able to merge that into your branch here and everything should be working again.

Thanks!

33Gjl1Xe:
I believe the PR can now be merged.

burgholzer (approved):
Yeah 🎉  thanks for all the effort. This is really looking great now.

PR #498, opened 2023-12-15
✨ Add support for OpenQASM 3.0 `else`

martin-fink:
## Description

This PR adds support for the `else` branch in `if` statements when parsing OpenQASM 3.0.

Example:

```qasm
if (c == 1) {
  x q[1];
} else {
  x q[0];
  x q[1];
}
```

This is implemented by adding a second ClassicalControlledOperation to the Quantum Computation with a flipped condition.
When dumping the circuit above, the following is printed:

```qasm
if (c == 1) {
  x q[1];
}
if (c != 1) {
  x q[0];
  x q[1];
}
```

## Checklist:

### DIFF
--- a/src/parsers/QASM3Parser.cpp
+++ b/src/parsers/QASM3Parser.cpp
@@ -342,7 +342,7 @@ public:
                    const std::string& identifier,
                    const std::vector<std::shared_ptr<Expression>>& parameters,
                    std::vector<std::shared_ptr<GateOperand>> targets,
-                   qc::QuantumRegisterMap& qregs) {
+                   const qc::QuantumRegisterMap& qregs) {
     auto iter = gates.find(identifier);
     std::shared_ptr<Gate> gate;
     size_t implicitControls{0};
@@ -773,7 +773,7 @@ public:
         error("Only quantum statements are supported in blocks.",
               statement->debugInfo);
       }
-      auto qregs = qc->getQregs();
+      const auto& qregs = qc->getQregs();
 
       auto op =
           evaluateGateCall(gateCall, gateCall->identifier, gateCall->arguments,
--- a/src/parsers/QASM3Parser.cpp
+++ b/src/parsers/QASM3Parser.cpp
@@ -725,36 +725,11 @@ public:
             ifStatement->debugInfo);
     }
 
-    qc::ComparisonKind comparisonKind = qc::ComparisonKind::Eq;
-    qc::ComparisonKind flippedComparisonKind = qc::ComparisonKind::Eq;
-    switch (condition->op) {
-    case BinaryExpression::Op::LessThan:
-      comparisonKind = qc::ComparisonKind::Lt;
-      flippedComparisonKind = qc::ComparisonKind::Geq;
-      break;
-    case BinaryExpression::Op::LessThanOrEqual:
-      comparisonKind = qc::ComparisonKind::Leq;
-      flippedComparisonKind = qc::ComparisonKind::Gt;
-      break;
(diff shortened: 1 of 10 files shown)


src/parsers/QASM3Parser.cpp
```diff
@@ -726,24 +726,31 @@ class OpenQasm3Parser final : public InstVisitor {
...
     case BinaryExpression::Op::Equal:
       comparisonKind = qc::ComparisonKind::Eq;
+      flippedComparisonKind = qc::ComparisonKind::Neq;
       break;
     case BinaryExpression::Op::NotEqual:
       comparisonKind = qc::ComparisonKind::Neq;
+      flippedComparisonKind = qc::ComparisonKind::Eq;
       break;
     default:
       error("Unsupported comparison operator.", ifStatement->debugInfo);
```
burgholzer: This logic looks like it could be extracted into a function somewhere.
Maybe as a function `getInvertedComparsionKind(const ComparisonKind kind)` somewhere close to the `ComparisonKind` and a method `getComparisonKind(const BinaryExpression::Op op)` somewhere close to the `BinaryExpression`.

Could be a nice separation of concerns.

src/parsers/QASM3Parser.cpp
```diff
@@ -768,12 +775,26 @@ class OpenQasm3Parser final : public InstVisitor {
...
+      auto thenOps = translateBlockOperations(ifStatement->thenStatements);
+      qc->emplace_back(std::make_unique<qc::ClassicControlledOperation>(
+          thenOps, creg->second, rhs->getUInt(), comparisonKind));
+    }
+    if (!ifStatement->elseStatements.empty()) {
+      auto elseOps = translateBlockOperations(ifStatement->elseStatements);
+      qc->emplace_back(std::make_unique<qc::ClassicControlledOperation>(
+          elseOps, creg->second, rhs->getUInt(), flippedComparisonKind));
+    }
+  }
```
burgholzer: I believe this is the right way to go for now and a quick win. 👍🏼 
In the long run, it might be worth to just revisit the design of the `ClassicControlledOperation` class; especially given that we plan to support further control flow logic soon.
At that point, we might just relabel/rename the class to something like `IfElseOperation`.
And, then, we can also simply print it as an `if ... else`.

src/parsers/qasm3_parser/passes/ConstEvalPass.cpp
```diff
@@ -333,8 +333,7 @@ std::optional<ConstEvalValue> ConstEvalPass::visitBinaryExpression(
     rhsVal->type = ConstEvalValue::Type::ConstInt;
   } else if (lhsVal->type == ConstEvalValue::Type::ConstUint &&
              rhsVal->type == ConstEvalValue::Type::ConstFloat) {
-    lhsVal->value =
-        static_cast<double>(std::get<0>(lhsVal->value));
+    lhsVal->value = static_cast<double>(std::get<0>(lhsVal->value));
```
burgholzer: I have no idea how that warning got here, but there seem to be two identical `else if` branches here. 
One of them can simply be removed.

burgholzer (commented):
That was quick. Many thanks!
I just added some small little comments. Nothing too major.

I'll also re-run the coverage tests (they seem to be failing more frequently again recently). That's not on us though, but on the codecov service.

burgholzer (approved):
Many thanks for the PR. I just updated a few things and optimized the code a little.

The remaining lines are, per definition, uncoverable. So I am ignoring the failing patch coverage check.

martin-fink:
Thanks!

PR #500, opened 2023-12-15
🚨 proper use of `std::enable_if` for SFINAE

burgholzer:
## Description

clang-tidy v18 includes some new warnings that just landed in CLion. These revealed that the whole `Expression` file was using the `std::enable_if` pattern to realize SFINAE in the wrong fashion.
This PR adjusts for that deficiency.

## Checklist:

### DIFF
--- a/include/operations/Expression.hpp
+++ b/include/operations/Expression.hpp
@@ -30,11 +30,9 @@ public:
 };
 
 struct Variable {
-  // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
   static inline std::unordered_map<std::string, std::size_t> registered{};
   static inline std::unordered_map<std::size_t, std::string> names{};
   static inline std::size_t nextId{};
-  // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
 
   explicit Variable(const std::string& name);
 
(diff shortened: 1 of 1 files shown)


include/operations/Expression.hpp
```diff
@@ -30,11 +30,9 @@ class SymbolicException : public std::invalid_argument {
 };
 
 struct Variable {
-  // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
   static inline std::unordered_map<std::string, std::size_t> registered{};
```
burgholzer:
```suggestion
  // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
  static inline std::unordered_map<std::string, std::size_t> registered{};
```

include/operations/Expression.hpp
```diff
@@ -30,11 +30,9 @@ class SymbolicException : public std::invalid_argument {
...
   static inline std::unordered_map<std::size_t, std::string> names{};
   static inline std::size_t nextId{};
```
burgholzer:
```suggestion
  static inline std::size_t nextId{};
  // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
```

PR #520, opened 2024-01-04
For builtin gates, print `c` prefix instead of `ctrl @ `

### LINKED ISSUE #519: 🚸 Better OpenQASM 3 output

While updating the project documentation to use the newly added OpenQASM 3 functionality, I noticed that the corresponding output could be improved a little bit. Currently, any controlled gate is dumped using the control modifier.
However, some controlled gates actually have a definition in `stdgates.inc` and it would be nice if they could be exported as such. For example, a CNOT gate is currently exported as
```qasm
ctrl @ x q[0], q[1];
```

It would be a lot nicer to have it as
```qasm
cx q[0], q[1];
```
I suppose this is mostly possible for any kind of controlled gate with positive controls and can probably be solved using some kind of lookup table of known gates (maybe even re-using some parts from the parser).

A couple of examples from the quickstart documentation that I expected to work: `cx`, `ch`, `ccx`, `cs`. These would make nice test cases.

martin-fink:
## Description

For builtin gates, print a `c` prefix instead of `ctrl @ gate`.

### Example

Before:
```qasm
ctrl @ x q0, q1;
ctrl(2) @ x q0, q1, q2;
ctrl(3) @ x q0, q1, q2, q3;
```

After:
```qasm
cx q0, q1;
ccx q0, q1, q2;
ctrl(3) @ x q0, q1, q2, q3;
```

Gates with negative controls, more controls than natively supported or non-supported gates are printed as before, with the `ctrl @ gate` notation.

Fixes #519 

## Checklist:

### DIFF
--- a/include/operations/StandardOperation.hpp
+++ b/include/operations/StandardOperation.hpp
@@ -117,7 +117,6 @@ protected:
   void dumpGateType(std::ostream& of, std::ostringstream& op,
                     const RegisterNames& qreg) const;
 
-private:
   void dumpControls(std::ostringstream& op) const;
 };
 
--- a/include/operations/StandardOperation.hpp
+++ b/include/operations/StandardOperation.hpp
@@ -116,6 +116,9 @@ protected:
 
   void dumpGateType(std::ostream& of, std::ostringstream& op,
                     const RegisterNames& qreg) const;
+
+private:
+  void dumpControls(std::ostringstream& op) const;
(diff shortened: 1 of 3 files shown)


burgholzer (approved):
Many thanks. That was fast ⚡ 
I just refactored the function a little bit, but otherwise this LGTM 👍🏼

PR #532, opened 2024-01-15
Zero-ancilla partial equivalence checking

reb-ddm:
## Description

Changed `isCloseToIdentityRecursive` such that it supports garbage qubits and a flag can be added to only check if it's close to a diagonal matrix, without checking that the diagonal entries are close to one.
With this function, zero-ancilla partial equivalence can be checked as described in the paper ["Partial Equivalence Checking of Quantum Circuits"](https://arxiv.org/abs/2208.07564) by Chen et al. 

## Checklist:


include/dd/Package.hpp
```diff
@@ -1903,10 +1903,16 @@ template <class Config> class Package {
     const auto eliminate = std::vector<bool>(nqubits, true);
     return trace(a, eliminate).w;
   }
-  bool isCloseToIdentity(const mEdge& m, const dd::fp tol = 1e-10) {
+
+  // if checkCloseToOne is false, it only checks if it's close to a diagonal
+  // matrix
```
burgholzer: Lately I started making it a best practice that whenever I touch a certain function in the project or add a new one, I also add a proper docstring to that function.
This would also be great here.

include/dd/Package.hpp
```diff
@@ -1952,62 +1958,82 @@ template <class Config> class Package {
...
-    // check whether  m = [ ~1 0 0 y ]
-    const auto mag0 = dd::ComplexNumbers::mag2(m.p->e[0U].w);
-    if (std::abs(mag0 - 1.0) > tol) {
-      visited.insert(m.p);
-      return false;
-    }
-    const auto arg0 = dd::ComplexNumbers::arg(m.p->e[0U].w);
-    if (std::abs(arg0) > tol) {
-      visited.insert(m.p);
+      visited.insert(m.p); // is that not superfluous?
```
burgholzer: I see your point. It could be worth removing this to keep the visited set smaller.
Such a node that is close to zero wouldn't be visited anyway, I guess.
Does anything seem to break if you remove the statement?

test/dd/test_package.cpp
```diff
@@ -1140,6 +1140,69 @@ TEST(DDPackageTest, CloseToIdentity) {
...
+
+  auto c5MultipliedWithC6 = dd->multiply(c5, dd->conjugateTranspose(c6));
+
+  EXPECT_FALSE(dd->isCloseToIdentity(c5MultipliedWithC6, tol,
+                                     {false, true, true}, false));
+  EXPECT_FALSE(dd->isCloseToIdentity(c5MultipliedWithC6, tol,
+                                     {true, true, false}, false));
+  EXPECT_TRUE(dd->isCloseToIdentity(c5MultipliedWithC6, tol,
+                                    {true, false, true}, false));
+}
```
burgholzer: Instead of a single large test, these could be three (?) smaller tests. That makes it a little easier to spot potential errors when something goes wrong.

burgholzer (changes requested):
Many thanks for the pull request! Overall this is looking really good already. Just a couple of very minor comments that would be nice to address. After that, this is good to go ✅

include/dd/Package.hpp
reb-ddm: It seems to work when I remove it. Anyway after this insert, the function returns immediately, and the visited set goes out of scope, so I don't think this can cause probems.

burgholzer (approved):
Many thanks for the additional changes. Looking all good now! 😌

PR #534, opened 2024-01-16
:zap: Added native construction of two-target controlled-gate DDs

### LINKED ISSUE #334: ⚡ Natively construct two-target controlled-gate DDs
In cda-tum/dd_package#144, a function was added to the DD package that allows the native construction of DDs for two-target gates. This is a significant improvement compared to the old-fashioned way of constructing these DDs from the DDs of a gate decomposition for the respective gate (e.g., a SWAP DD can be constructed from 2 CNOT DDs).

However, the implementation in cda-tum/dd_package#144 only works for non-controlled two-target gates. Thus, controlled versions of these gates still require the old-fashioned workaround. This can be seen in https://github.com/cda-tum/dd_package/blob/7c9644a2a65742bb580e3e5fefeac075c9534351/include/dd/Package.hpp#L629-L810
All of that code could be avoided by making the construction routine handle (arbitrary) controls as well. The single-target method already does this as shown in https://github.com/cda-tum/dd_package/blob/7c9644a2a65742bb580e3e5fefeac075c9534351/include/dd/Package.hpp#L480-L539

This requires some careful handling of all possible cases (control before first target, control between both targets, control after second target) and respective testing.

I believe this is a nice little project as it is very localized and would eliminate quite some code here and in our top-level modules.
A best case scenario would be to find a generic solution for multi-target gates, so that the `makeGateDD` and `makeTwoQubitGateDD` methods can be unified. Hypothetically that further abstraction shouldn't be too hard once one has figured out how to handle the "control between targets" case.

BertiFlorea:
## Description
The purpose of this PR is to add the extra logic for two-target controlled-gate DDs mentioned in https://github.com/cda-tum/mqt-core/issues/334#issue-1758497551. The problem was that the current implementation of the `makeTwoQubitGateDD` only worked for non-controlled two-target gates. Additionally, what we also want to achieve with this PR is to get rid of all the extra methods using the previous workaround (i.e.https://github.com/cda-tum/dd_package/blob/7c9644a2a65742bb580e3e5fefeac075c9534351/include/dd/Package.hpp#L629-L810).

At the moment the updated version of `makeTwoQubitGateDD` is passing the CI, but there are still some TODOs left:
- [x] test for equivalence between old-fashioned workaround methods and `makeTwoQubitGateDD`
- [x] update the code across the package (e.g. getStandardOperationDD)
- [x] verify the functionality by testing for edge cases

Fixes #334 

## Checklist:


include/dd/Package.hpp
```diff
@@ -635,15 +654,48 @@ template <class Config> class Package {
...
     auto z = static_cast<Qubit>(start);
     const auto smallerTarget = std::min(target0, target1);
     for (; z < smallerTarget; ++z) {
       for (auto& row : em) {
-        for (auto& entry : row) {
-          entry = makeDDNode(
-              z, std::array{entry, mEdge::zero(), mEdge::zero(), entry});
+        for (auto i1 = 0U; i1 < RADIX; ++i1) {
+          for (auto i2 = 0U; i2 < RADIX; ++i2) {
+            auto i = i1 * RADIX + i2;
```
```suggestion
            const auto i = i1 * RADIX + i2;
```
Just a very minor addition.

include/dd/Package.hpp
```diff
@@ -671,22 +723,65 @@ template <class Config> class Package {
...
+        for (auto i2 = 0U; i2 < RADIX; ++i2) {
+          auto i = i1 * RADIX + i2;
+          if (it != controls.end() && it->qubit == z) {
+            auto edges = std::array{mEdge::zero(), mEdge::zero(), mEdge::zero(),
+                                    mEdge::zero()};
+            if (it->type == qc::Control::Type::Neg) { // negative control
+              edges[0] = em0[i];
+              if (i1 == i2) {
+                if (z == 0U) {
+                  edges[3] = mEdge::one();
```
burgholzer: I believe this case is impossible to hit. There was at least one target before this control. So `z` can never be zero.
This should allow to simplify the code here a little.

include/dd/Package.hpp
```diff
@@ -671,22 +723,65 @@ template <class Config> class Package {
...
 
-    // process lines between the two targets (by creating identity structures)
+    // process lines between the two targets
     for (++z; z < std::max(target0, target1); ++z) {
-      for (auto& entry : em0) {
-        entry = makeDDNode(
-            z, std::array{entry, mEdge::zero(), mEdge::zero(), entry});
+      for (auto i1 = 0U; i1 < RADIX; ++i1) {
+        for (auto i2 = 0U; i2 < RADIX; ++i2) {
+          auto i = i1 * RADIX + i2;
```
```suggestion
          const auto i = i1 * RADIX + i2;
```

burgholzer (changes requested):
Overall the changes here are looking good and it seems like this should be working.
The biggest thing that is missing is tests of the new functionality. That's also why the code coverage check is failing. Almost none of the new lines (part of the patch) are covered by tests.

This should be quickly resolved though once the existing workaround methods are removed.
I would propose to turn the existing workarounds for constructing controlled two-target DDs into tests for the new functionality. That way, the existing code is not lost and is useful for testing.

BertiFlorea:
At the moment, the following progress has been made:
1. The `makeTwoQubitGateDD` constructs the DD bottom-up correctly. 🎉 
2. I have updated the `makePeresDD` and `makePeresInvDD` workaround methods because they provided the inverse output instead of the desired one. (i.e. makePeresDD was producing the inverse DD of the Peres gate)
3. I have updated the code across the code base to rely on the native construction of two-target controlled-gate DDs instead of using the old workaround methods. Additionally, I used them to verify the functionality of our method by comparing the outputs (e.g. compared the output generated by `makeTwoQubitGateDD` for a swap matrix with `makeSWAPDD`).
Note: Here there were some already existing methods which were checking this, however, there was only one missing test case for the Peres gate.
4. To improve the code coverage I have added one test to verify all the edge cases that were not covered by the existing tests. (Update after the CI has finished: the test still needs some improvement, I'll come back with an update for this one)

Apart from point 4 (which is on my TODO list), the only question left from my side is if we still intend to unify `makeTwoQubitGateDD` and `makeGateDD` since it was mentioned in #334.

include/dd/Operations.hpp
```diff
@@ -107,137 +107,57 @@ getStandardOperationDD(const qc::StandardOperation* op, Package<Config>& dd,
...
   case qc::SWAP:
-    // SWAP is self-inverse
-    return dd.makeSWAPDD(nqubits, controls, target0, target1, startQubit);
+    gm = SWAP_MAT;
+    break;
   case qc::iSWAP:
-    if (inverse) {
-      return dd.makeiSWAPinvDD(nqubits, controls, target0, target1, startQubit);
-    }
-    return dd.makeiSWAPDD(nqubits, controls, target0, target1, startQubit);
```
burgholzer, line 167: The corresponding routines in the `Package.hpp` file can all be deleted now. That's actually what I meant with one of my earlier comments that all these functions for creating controlled two-target functions can be extracted into tests.
Just remove all that code from the main package (`Package.hpp`) and move it to the tests. You can just copy (and slighlty adjust) the definitions from the `Package.hpp` file and move them to the `test_dd_functionality.cpp` test file.
This should reduce the number of functions in the main package file considerably.

test/dd/test_dd_functionality.cpp
```diff
@@ -64,7 +64,7 @@ TEST_P(DDFunctionality, standardOpBuildInverseBuild) {
   using namespace qc::literals;
   auto gate = static_cast<qc::OpType>(GetParam());
 
-  qc::StandardOperation op;
+  qc::StandardOperation op, op_neg;
```
burgholzer: First of all, you can see some of the clang-tidy warnings here in the "Files Changed" tab.
These should be easy to address.
In this case here, you do not seem to be using `op_neg` for the test at all, so it could be removed entirely.

test/dd/test_dd_functionality.cpp
```diff
@@ -121,7 +121,7 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
```
burgholzer: I'd slightly prefer to refactor this into two separate tests. Let's keep the original test and extract the negative control one into a separate function.
Then, you can also directly test the single-target gates with negative controls, which should yield even better coverage.

test/dd/test_dd_functionality.cpp
```diff
@@ -172,6 +177,41 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
+  ASSERT_NO_THROW({
+    d = newDD->multiply(getDD(&op_neg, *newDD), getInverseDD(&op_neg, *newDD));
+  });
+  newDD->incRef(d);
+
+  // the DD of the identity needs to be reconstructed as well
+  ident = newDD->makeIdent(newNQubits);
+
+  EXPECT_EQ(ident, e);
+  EXPECT_EQ(ident, d);
```
burgholzer: This test seems oddly specific to me. What part of that is not already covered by existing tests?
If it's the part about the negative controls, then this should be fairly easy to incorporate into the existing tests that test out all the possibilities of combining controls and targets by adding another for loop that influences the type of control.

burgholzer (changes requested):
Overall this is looking really good already. Many Thanks 🎉 
The Python failures should be fixed after #536 is in. Then you can just merge the main branch in or rebase your branch on main.
There still are some warnings left to be fixed and I also added some comments that would be good to address before this gets merged.
Most importantly, the main library code in `Package.hpp` can be significantly reduced by moving the respective special purpose construction methods to the tests. This should slim down the library quite a bit.

As for your question on combining the two gate DD creation methods: I do not think that is necessary as part of this PR. Content-wise, this PR is great as it is and definitly enough to close the respective issue 👍🏼

include/dd/Operations.hpp
BertiFlorea: Done 👍🏼

test/dd/test_dd_functionality.cpp
BertiFlorea: Done 👍🏼

test/dd/test_dd_functionality.cpp
BertiFlorea: Done 👍🏼

test/dd/test_dd_functionality.cpp
BertiFlorea: Indeed, there was no need for this extra test, handling the request above did the job for the code coverage. 
Removed and done👍🏼

BertiFlorea:
I have just added all the requested changes. The code coverage failed but according to the error message, it seems that it might be a connection problem (I might be wrong still).

test/dd/test_dd_functionality.cpp
```diff
@@ -174,6 +174,63 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
 }
 
+TEST_P(DDFunctionality, controlledStandardNegOpBuildInverseBuild) {
+  using namespace qc::literals;
+  auto gate = static_cast<qc::OpType>(GetParam());
+
+  qc::StandardOperation op;
+  switch (gate) {
+  case qc::GPhase:
+    op = qc::StandardOperation(nqubits, Controls{0}, Targets{}, gate,
```
burgholzer:
```suggestion
    op = qc::StandardOperation(nqubits, Controls{0_nc}, Targets{}, gate,
```

test/dd/test_dd_functionality.cpp
```diff
@@ -174,6 +174,63 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
+                               std::vector{dist(mt)});
+    break;
+  case qc::U:
+    op = qc::StandardOperation(nqubits, 0, 1, gate,
```
burgholzer:
```suggestion
    op = qc::StandardOperation(nqubits, Controls{0_nc}, 1, gate,
```

test/dd/test_dd_functionality.cpp
```diff
@@ -174,6 +174,63 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
+                               std::vector{dist(mt), dist(mt), dist(mt)});
+    break;
+  case qc::U2:
+    op = qc::StandardOperation(nqubits, 0, 1, gate,
```
burgholzer:
```suggestion
    op = qc::StandardOperation(nqubits, Controls{0_nc}, 1, gate,
```

test/dd/test_dd_functionality.cpp
```diff
@@ -174,6 +174,63 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
+                               std::vector{dist(mt), dist(mt)});
+    break;
+  case qc::RX:
+  case qc::RY:
+  case qc::RZ:
+  case qc::P:
+    op = qc::StandardOperation(nqubits, 0, 1, gate, std::vector{dist(mt)});
```
burgholzer:
```suggestion
    op = qc::StandardOperation(nqubits, Controls{0_nc}, 1, gate, std::vector{dist(mt)});
```

test/dd/test_dd_functionality.cpp
```diff
@@ -174,6 +174,63 @@ TEST_P(DDFunctionality, controlledStandardOpBuildInverseBuild) {
...
+    op = qc::StandardOperation(nqubits, Controls{0_nc}, 1, 2, gate,
+                               std::vector{dist(mt)});
+    break;
+  case qc::XXminusYY:
+  case qc::XXplusYY:
+    op = qc::StandardOperation(nqubits, Controls{0_nc}, 1, 2, gate,
+                               std::vector{dist(mt), dist(mt)});
+    break;
+  default:
+    op = qc::StandardOperation(nqubits, 0, 1, gate);
```
burgholzer:
```suggestion
    op = qc::StandardOperation(nqubits, Controls{0_nc}, 1, gate);
```

burgholzer (approved):
Great. Many thanks for another nice contribution! 🎉 
I'll just apply some finishing touches and then proceed to merge 🔀

Issue #530, opened 2024-01-12
✨ Smaller Python wheels


burgholzer:
### What's the problem this feature will solve?

Since #529, `mqt-core` can finally be installed like a regular project.
However, the installation of all the main project libraries has bloated the binary Python wheels that we ship quite a bit. They went from a couple hundred KB to roughly 70MB. This isn't really great in the long run.

### Describe the solution you'd like

A potential solution would be to go down a similar path as `nanobind` and ship the complete source code instead of the compiled targets.
Nanobind accomplished this by moving all the target creation code to the `nanobind-config.cmake` file and also installing the `src` folder in a predictable location so that the installation layout is the same as the source layout.

This would allow us to ship the source code with the Python package, but require the dependent package to build the respective targets.

For general installation of the C++ project, I believe it still makes sense to keep the targets install.
So my current proposal would be to make the current behavior the default and the proposed behavior opt-in (and automatically enabled for building the python package).

PR #538, opened 2024-01-24
🚸 Building Shared Libraries

burgholzer:
## Description

This PR started out as a PR that should "just" improve the project installation before top-level projects eventually switch to the new `mqt-core` version. As part of that, it tried to address #530 by switching the library to build dynamic libraries by default. This considerably decreases the binary size and would, in general, be nice to support.

The corresponding changes need appropriate testing across all different projects to make sure that nothing is missed.
I will open corresponding PRs in all top-level projects that use mqt-core to make sure that the new library structure works for all those projects.

In particular:
- mqt-qcec: cda-tum/mqt-qcec#355
- mqt-qmap: cda-tum/mqt-qmap#418
- mqt-ddsim: cda-tum/mqt-ddsim#336

## Checklist:


burgholzer:
Ok. Getting shared libraries to work properly is more work than anticipated.
I am going to split this PR apart and will focus on getting the changes that are not related to shared libraries merged as soon as possible so that development in the top-level projects can resume.

burgholzer:
Just rebased this PR on top of main and cleaned up the commits a little bit.
This should be a solid start once there is more time to look into this.

burgholzer:
Ok. Further progress here. To keep the history of this PR, I just squashed the previous commits and added another commit that reverts the respective changes.
It turns out, that it might actually not be as hard as expected to get this working since hidden symbol visibility apparently is not a must.

The current state of the PR just changes very few things and, at least locally, that was enough to allow building all of our top-level Python packages based on the installed mqt.core package.

burgholzer:
Adding some interesting articles on C++ ABI compatibility here:
- https://uwekorn.com/2019/09/15/how-we-build-apache-arrows-manylinux-wheels.html
- https://discuss.python.org/t/how-to-set-glibcxx-use-cxx11-abi-for-manylinux2014-and-manylinux2010-wheels/10551

burgholzer:
The Python errors on Windows are interesting as I can't quite make sense of them. This needs some further investigation. It has something to do with the `mqt-core-python` target. So maybe (finally) removing that resolves the error.
Interestingly, the C++ tests work just fine on Windows.

The package check failure is due to the SONAME and Version symlinks as well as the shared libraries in a non-importable location. I am fairly sure it is fine to simply ignore these, but this also needs further investigation.

burgholzer:
Alright. The C++ tests working under Windows were simply because no shared libraries were built there.
The reason for the failure in the Python module was symbol visibility, which is hidden by default on Windows.
However, there is a convenient CMAKE option to export all symbols under Windows.
Once that was working, Python could not finde the mqt-core DLLs, which was fixed by adding the `bin` directory of the package installation to the DLL path.

burgholzer:
Closing in favor of #662

PR #539, opened 2024-01-26
🚸 Installation Improvements

burgholzer:
## Description

This PR improves a couple of bits and pieces throughout the library and should increase the overall usability and DX.

Most notable breaking changes:
- The `MQT::CorePython` target is removed. After this PR is merged, top-level packages are expected to use the `mqt.core` Python package for translating from Qiskit to MQT. Furthermore, the top-level packages are responsible for bringing in the `pybind11_json` dependency if they require it, as it is no longer shipped with `mqt-core`.
- The `MQT::ProjectOptions` and `MQT::ProjectWarnings` targets are now linked privately into the libraries in order to not pollute other CMake targets. This means that top-level projects now also need to explicitly link against these targets if they want to benefit from the common settings.
- Includes are now placed in a dedicated `include/mqt-core` subdirectory to create a cleaner installation structure.

## Checklist:


burgholzer:
That one cpp-linter failure cannot be fixed here since `pybind11` is not available in the C++ linting environment.

PR #543, opened 2024-01-31
 ✨ Add State Preparation Algorithm

M-J-Hochreiter:
## Description

As a part of my bachelor thesis I implemented the Qiskit State Preparation algorithm (https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/library/data_preparation/state_preparation.py#) in C++ using the MQT Framework. 
My advisor Yannick Stade suggested I open a pull request, as it might be interesting to add into MQT.

The state preparation allows a user to create a circuit that initializes a quantum state from a list of complex amplitudes, sometime needed in various quantum algorithms.

I am not sure how/where to add this class, thus I just created a new file.
If you could give me feedback whether this functionality might be interesting to add/what to change I would gladly try my best to fulfill your requests.

I have not yet written any tests, apart from manually comparing outputs of qiskit and my algorithm, and simulating both circuits with qiskit. (can be added in the future)

## Checklist:


burgholzer:
Hey 👋🏼 
Many thanks for the initial PR. This is definitely interesting and in scope for mqt-core! Looking forward to having this included.

The best place to put this would probably be in the `algorithms` submodule of `mqt-core` (https://github.com/cda-tum/mqt-core/tree/main/include/mqt-core/algorithms).
You can take lots of inspiration from the existing classes there. (https://github.com/cda-tum/mqt-core/blob/main/include/mqt-core/algorithms/Entanglement.hpp as one of the simpler ones, https://github.com/cda-tum/mqt-core/blob/main/include/mqt-core/algorithms/QPE.hpp as one of the more complicated examples).
The goal there would be to just create a new subclass of QuantumComputation called `StatePreparation` that takes the appropriate input in its constructor and then constructs the respective circuit using your methods. That shouldn't take too much adaptation from your existing code.

I would advise you to also create a `.cpp` file in the `src/algorithms` directory and move all the definitions from the header file there (just keeping the declarations themselves in the header). This makes compilation faster and helps to separate the interface of your methods from the implementation itself.
If any method is only ever used in one of your functions internally, you might as well move the declaration to the cpp file so the method is not exposed publicly. (This probably applies to most of the matrix manipulation stuff).

After the initial setup, it would be great to add some tests for the new functionality. These should probably placed here: https://github.com/cda-tum/mqt-core/tree/main/test/algorithms. Again, you can take lots of inspiration from the existing tests and should be able to use our own simulator for some experiments.

Please don't hesitate to ask questions if anything pops up!

burgholzer:
Closing for now. Please feal free to reopen if you plan to continue working on this.

burgholzer:
> Hi,
> 
> sorry for the long wait, I finally got around to working on this again!
> 
> I do have a question on the new constructor that should be implemented: As my previous methods created a QuantumComputation from scratch, do you believe it makes sense to just call the QuantumComputing constructor in my StatePreparation constructor with the QuantumComputation I obtained from my prepareState function?
> 
> I do not see any option to reopen this pull request and continue working on it.

Hey 👋🏼 

No worries. I just reopened the PR for you.
Have a look at the new versions of the algorithm implementations in MQT Core, e.g., 
- https://github.com/cda-tum/mqt-core/blob/main/include/mqt-core/algorithms/QPE.hpp and 
- https://github.com/cda-tum/mqt-core/blob/main/src/algorithms/QPE.cpp.

I'd imagine that your State Preparation functionality code can be implemented and integrated very similarly.

M-J-Hochreiter:
I refactored both files to match the other algorithms style if you would like to give it a look!

Next step would then be tests from what I can see right?

burgholzer:
> I refactored both files to match the other algorithms style if you would like to give it a look!
> 
> Next step would then be tests from what I can see right?

Getting the CI to a state where it is all-green would be desirable.
That includes writing tests that cover all the changes. 👍🏼

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+  if (!isNormalized(amplitudes)) {
+    throw std::invalid_argument{
+        "Using State Preparation with Amplitudes that are not normalized"};
+  }
+
+  // get number of qubits needed
+  double const numQubits = std::log2(amplitudes.size());
+
+  if (std::abs(numQubits) < EPS || std::floor(numQubits) != numQubits) {
```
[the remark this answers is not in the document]
M-J-Hochreiter: This code fails either when numQubits is smaller than an epsilon meaning its 0 or when its not an integer, as intended.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,287 @@
...
+    throw std::invalid_argument{
+        "Using State Preparation with vector size that is not a power of 2"};
+  }
+
+  QuantumComputation toZeroCircuit =
+      gatesToUncompute(amplitudes, static_cast<size_t>(numQubits));
```
burgholzer: You can get around the warning by relying on some bit-twiddling hacks, which are more performant as well.

```suggestion
  // check if the number of elements in the vector is a power of two
  if (amplitudes.size() == 0 || (amplitudes.size() & (amplitudes.size() - 1)) != 0) {
    throw std::invalid_argument{
        "Using State Preparation with vector size that is not a power of 2"};
  }
  const auto numQubits = static_cast<size_t>(std::log2(amplitudes.size()));
  QuantumComputation toZeroCircuit = gatesToUncompute(amplitudes, numQubits);
```

M-J-Hochreiter: That's a very elegant solution, thanks!

include/mqt-core/algorithms/StatePreparation.hpp
```diff
@@ -0,0 +1,38 @@
...
+ *      Shende, Bullock, Markov. Synthesis of Quantum Logic Circuits (2004)
+        [`https://ieeexplore.ieee.org/document/1629135`]
+ * */
+
+/**
+ * @throws invalid_argument when amplitudes are not normalized or length not
+ * power of 2
+ * @param list of complex amplitudes to initialize to
+ * @return MQT Circuit that initializes a state
+ * */
```
burgholzer:
```suggestion
/**
 * @brief Prepares a generic quantum state from a list of normalized
 *        complex amplitudes
 *
 * Adapted implementation of IBM Qiskit's State Preparation:
 * https://github.com/Qiskit/qiskit/blob/e9ccd3f374fd5424214361d47febacfa5919e1e3/qiskit/circuit/library/data_preparation/state_preparation.py
 * based on the following paper:
 *      V. V. Shende, S. S. Bullock and I. L. Markov, "Synthesis of quantum-logic circuits," in IEEE Transactions on Computer-Aided Design of Integrated Circuits and Systems, vol. 25, no. 6, pp. 1000-1010, June 2006, doi: 10.1109/TCAD.2005.855930.
 *
 * @param amplitudes state (vector) to prepare. Must be normalized and have a size that is a power of two
 * @return quantum computation that prepares the state
 * @throws invalid_argument @p amplitudes is not normalized or its length is not a
 * power of two
 **/
```

A suggestion for a slightly improved docstring here. Note that this will be reformatted by clang-format if you simply accept the suggestion.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  double const arg = -std::arg(std::accumulate(
+      amplitudes.begin(), amplitudes.end(), std::complex<double>(0, 0)));
+  if (arg != 0) {
+    disentangler.gphase(arg);
+  }
+  return disentangler;
+}
+
+auto createStatePreparationCircuit(
+    std::vector<std::complex<double>>& amplitudes) -> QuantumComputation {
```
burgholzer: Right now, the function signature suggests that the function will modify the `amplitudes` vector (since the argument is not a const reference).
This would be fine, as it is more efficient then copying, but it also implies that the amplitude vector is useless after calling this function.
The function docstring should note that the function changes the underlying vector in that case.

I see two options here:
- either make the parameter a constant reference and perform an explicit copy of the state.
- or update the docstring to reflect the behavior

(note that his comment depends on some of the other comments in the review)

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  if (amplitudes.empty() ||
+      (amplitudes.size() & (amplitudes.size() - 1)) != 0) {
+    throw std::invalid_argument{
+        "Using State Preparation with vector size that is not a power of 2"};
+  }
+  const auto numQubits = static_cast<size_t>(std::log2(amplitudes.size()));
+  QuantumComputation toZeroCircuit = gatesToUncompute(amplitudes, numQubits);
+
+  // invert circuit
+  CircuitOptimizer::flattenOperations(toZeroCircuit, false);
```
burgholzer: I'd argue that there is no direct need to flatten the circuit here. You might as well preserve the structure of the algorithm.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+    // minus sign because we move it to zero
+    thetas.push_back(-theta);
+    phis.push_back(-phi);
+  }
+  return {remainingVector, thetas, phis};
+}
+
+// creates circuit that takes desired vector to zero
+[[nodiscard]] auto
+gatesToUncompute(std::vector<std::complex<double>> amplitudes, size_t numQubits)
```
burgholzer: This method should not take its argument by value as this would cause a copy for each invocation. Pass this by reference and modify it in-place.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  for (size_t i = 0; i < (amplitudes.size() / 2); ++i) {
+    auto [remains, theta, phi] =
+        blochAngles(amplitudes[2 * i], amplitudes[2 * i + 1]);
+    remainingVector.push_back(remains);
+    // minus sign because we move it to zero
+    thetas.push_back(-theta);
+    phis.push_back(-phi);
+  }
+  return {remainingVector, thetas, phis};
+}
```
burgholzer: Do not pass the `amplitudes` vector by value as this will always incur a copy.
Pass it by reference and modify it in-place instead of creating a new vector and returning the new vector.
This will require some adaptations to the general code logic probably.

Also just a general remark: prefer to reserve containers with the right size if you know the size upfront using `vec.reserve(X)`. Furthermore, prefer to use `emplace_back` over `push_back` in general.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+    auto [remainingParams, thetas, phis] = rotationsToDisentangle(amplitudes);
+    amplitudes = remainingParams;
+    // perform required rotations
+    bool addLastCnot = true;
+    double const phisNorm = twoNorm(phis);
+    double const thetasNorm = twoNorm(thetas);
+    if (phisNorm != 0 && thetasNorm != 0) {
+      addLastCnot = false;
+    }
+    if (phisNorm != 0) {
```
burgholzer: Shouldn't these use `EPS`?

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+      // call multiplex with RZGate
+      QuantumComputation rzMultiplexer =
+          multiplex(OpType{RZ}, phis, addLastCnot);
+      // append rzMultiplexer to disentangler, but it should only attach on
+      // qubits i-numQubits, thus "i" is added to the local qubit indices
+      for (auto& op : rzMultiplexer) {
+        for (auto& target : op->getTargets()) {
+          target += static_cast<unsigned int>(i);
```
burgholzer: Most of the `unsigned int` here should (probably) be `Qubit`.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+        }
+        for (auto control : op->getControls()) {
+          // there were some errors when accessing the qubit directly and
+          // adding to it
+          op->setControls(
+              Controls{Control{control.qubit + static_cast<unsigned int>(i)}});
```
burgholzer: What were these kind of errors?
This should work out of the box. 
```c++
for (auto& control: op->getControls()) {
  control.qubit += static_cast<Qubit>(i);
}
```
Note the `&` in the for loop

Applies to other places in the code as well

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+        }
+      }
+      disentangler.emplace_back<Operation>(rzMultiplexer.asOperation());
+    }
+    if (thetasNorm != 0) {
+      // call multiplex with RYGate
+      QuantumComputation ryMultiplexer =
+          multiplex(OpType{RY}, thetas, addLastCnot);
```
burgholzer:
```suggestion
          multiplex(RY, thetas, addLastCnot);
```
simpler and less code to type. applies throughout the whole file.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+      // append reversed ry_multiplexer to disentangler, but it should only
+      // attach on qubits i-numQubits, thus "i" is added to the local qubit
+      // indices
+      std::reverse(ryMultiplexer.begin(), ryMultiplexer.end());
```
burgholzer: you can simply call the `reverse` method of the `QuantumComputation` class here.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+              Controls{Control{control.qubit + static_cast<unsigned int>(i)}});
+        }
+      }
+      disentangler.emplace_back<Operation>(ryMultiplexer.asOperation());
+    }
+  }
+  // adjust global phase according to the last e^(it)
+  double const arg = -std::arg(std::accumulate(
+      amplitudes.begin(), amplitudes.end(), std::complex<double>(0, 0)));
+  if (arg != 0) {
```
burgholzer: EPS?

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+#include <cmath>
+#include <complex>
+#include <cstddef>
+#include <cstdint>
+#include <iterator>
+#include <numeric>
+#include <tuple>
+#include <vector>
+
+static const double EPS = 1e-10;
```
burgholzer: should this be a configurable parameter of the method?

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+    for (size_t i = 0; i < matrixVec.size(); ++i) {
+      sum += matrixVec[i] * vector[i];
+    }
+    result.push_back(sum);
+  }
+  return result;
+}
+
+// recursive implementation that returns multiplexer circuit
+/**
```
burgholzer: the top comment should probably be the `@brief` in the regular docstring

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+[[nodiscard]] auto multiplex(OpType targetGate, std::vector<double> angles,
+                             bool lastCnot) -> QuantumComputation {
+  size_t const listLen = angles.size();
+  double const localNumQubits =
+      std::floor(std::log2(static_cast<double>(listLen))) + 1;
+  QuantumComputation multiplexer{static_cast<size_t>(localNumQubits)};
+  // recursion base case
+  if (localNumQubits == 1) {
+    multiplexer.emplace_back<StandardOperation>(Controls{}, 0, targetGate,
+                                                std::vector{angles[0]});
```
burgholzer:
```suggestion
    multiplexer.emplace_back<StandardOperation>(0, targetGate,
                                                std::vector{angles[0]});
```
should work just fine

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return multiplexer;
+  }
+
+  Matrix const matrix{std::vector<double>{0.5, 0.5},
+                      std::vector<double>{0.5, -0.5}};
```
burgholzer: might be worth defining this as static constexpr outside of the function to avoid creating it on every invocation

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  Matrix const identity =
+      createIdentity(static_cast<size_t>(pow(2., localNumQubits - 2.)));
```
burgholzer: use bit shifts instead of `std::pow` for efficiency

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  Matrix identity{
+      std::vector<std::vector<double>>(size, std::vector<double>(size, 0))};
+  for (size_t i = 0; i < size; ++i) {
+    identity[i][i] = 1;
+  }
+  return identity;
+}
+
+[[nodiscard]] auto matrixVectorProd(const Matrix& matrix,
+                                    std::vector<double> vector)
```
burgholzer: vector should be a const reference here to avoid a copy on each invocation

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  }
+  return sqrt(norm);
+}
+
+template <typename T>
+[[nodiscard]] auto isNormalized(std::vector<T> vec) -> bool {
+  return std::abs(1 - twoNorm(vec)) < EPS;
+}
+
+[[nodiscard]] auto kroneckerProduct(Matrix matrixA, Matrix matrixB) -> Matrix {
```
burgholzer: A nd B should be passed as const references here to avoid copying them on every invocation

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+template <typename T> [[nodiscard]] auto twoNorm(std::vector<T> vec) -> double {
+  double norm = 0;
+  for (auto elem : vec) {
+    norm += std::norm(elem);
+  }
+  return sqrt(norm);
+}
+
+template <typename T>
+[[nodiscard]] auto isNormalized(std::vector<T> vec) -> bool {
```
burgholzer: `vet` should be a const reference in all of these cases to avoid a copy.

src/algorithms/StatePreparation.cpp
```diff
@@ -0,0 +1,285 @@
...
+    multiplexer.emplace_back<Operation>(multiplex2.asOperation());
+  } else {
+    multiplexer.emplace_back<Operation>(multiplex2.asOperation());
+  }
+
+  if (lastCnot) {
+    multiplexer.cx(0, static_cast<Qubit>(localNumQubits - 1));
+  }
+
+  CircuitOptimizer::flattenOperations(multiplexer, false);
```
burgholzer: similar to below you might actually decide to skip this flatten operation to maintain the structure of the circuit.

burgholzer (changes requested):
Hey 👋🏼 

I was not quite sure whether you were done with your implementation, but I figured I'd do a quick review of the main library code and provide some suggestions for improvements.
I hope these make sense.
You probably also need to merge the latest changes from main so that everything runs smoothly.

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: using your for loop and compiling I get this error:
error: assignment of member ‘qc::Control::qubit’ in read-only object
             control.qubit += static_cast<Qubit>(i);
seems like getControls returns const, however I need to modify the control qubit ...

I do however not really understand why I get const qualified control qubits, because in the outer loop rzMultiplexer was created by multiplex which returns a non const QuantumComputation, thus the iterator over the QuantumComputation should also return non const operations?

With the targets which are implemeted equally this problem also does not seem to occur ...

What am I missing here? Seems to also be a cause of a segmentation fault in my current implementation, which makes sense as I am modifying the operation controls vector while looping over it

burgholzer: To be honest, this one is pretty weird. For some reason, it is deciding that the range-based for loop captures the values that it iterates over as `const Control`, which does not make sense. I even tried adding a separate method for getting the mutable controls with a different name, but that still resulted in the same compilation error.
Given that I tried with 3 different compilers on my system and also the CI fails consistently on all systems, this cannot be a simple compiler bug, and it is probably something more systematic.

Right now, I do not have much time to investigate this further, so I would propose to revert the proposed change and stick with what you had before, that was working.
Maybe after all the other comments have been resolved, this one will resolve on its own.

M-J-Hochreiter: Problem is, that my solution that worked a year ago now results in segmentation faults ... which I believe to be due to me actually modifying the container (the vector of controls) while looping over it.

thus i decided to collect the controls and set the new controls afterwards in commit 15d4535

include/mqt-core/algorithms/StatePreparation.hpp
M-J-Hochreiter: Done in 84a32fa

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commit 9a7dc3e

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commit 9a7dc3e

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: This would be a design decision, both would work. Qiskit does not provide that option ...

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Not 100% sure what you mean with this, do you want me to delete the comments of the multiplex function as it is not publicly visible?
I think I added that to make future adaption/understanding easier

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commit b3209bb

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commit a90ada4

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commit fa35de4

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commit e22906a

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in commits 3d657b1 and 6e96ded, also changed the type of localNumQubits to be of type Qubit, thus avoiding all later static_casts

src/algorithms/StatePreparation.cpp
burgholzer: That makes sense and would have also been what I would have suggested in general as a workaround. The code as it is can be made slightly more efficient though.
Instead of constructing a vector of controls and then turning this into a set afterwards, you may directly construct the `Controls` set.

Furthermore, clang-tidy still seems to have some complaints, which can find in the comment here https://github.com/cda-tum/mqt-core/actions/runs/13156744594/attempts/1#summary-36715533775 (the CI workflow cannot post direct comments on PRs from forks unfortunately, which is why you have to resort to the workflow summary to see these messages).

src/algorithms/StatePreparation.cpp
burgholzer: I'd argue that it doesn't really hurt to make this a configurable parameter in general. It should still have the current value as a default for general ease of use.

src/algorithms/StatePreparation.cpp
burgholzer: Literally just taking the 
```c++
// recursive implementation that returns multiplexer circuit
```
comment and moving it two lines down as
```c++
 * @brief Recursive implementation that returns multiplexer circuit
```
into the docstring.

burgholzer:
@M-J-Hochreiter thanks for your continued work on this! I see some comments in the PR that are still not yet resolved. I am happy to take a stab at the PR itself and bring it across the finish line once the open comments have been addressed.

The current Python failures you are seeing are most certainly because you are still using the CircuitOptimizer module in you code while it is not properly linked into the `MQT::CoreAlgorithms` target. However, based on some of the open comments, the flattening itself might be removed anyway, which makes this a "non-problem".
The reason you are not seeing this in the C++ build tests is because they use static libraries while the Python libs build shared libraries. This changes symbol visibility and vert much affects linking in general.

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Added it as a configurable parameter in d526a9a and 93a2374

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in 4f36ece

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Using C++17 vectors are not constexpr constructable as far as I know. Meaning I would need to make my Matrix definition an array, however I cannot do that as my identity and thus the kronecker product depends on the Number of local Qubits.

Not sure if there is an elegant solution to make this specific matrix a constexpr array and later use it as an vector?

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in 84a32fa

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in 84a32fa

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in 84a32fa

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Here the flattening indeed does not seem to affect the circuit. Thus removed in 50f7a60

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Removing the flattening here affects the circuit, as the subcircuits are not correctly reversed later on, in the recursive call if not flattened

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Done in 84a32fa

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Exchanged all push_back calls to emplace_back
Made amplitudes a const reference
in commit 4ce0865

M-J-Hochreiter: Reserved vector space where I saw fit/it's possible. 6858cd0

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: If i understood correctly what you meant implemented in e574dc9

Used an unnecessary temporary object. fixed in d93cec0

all the linting I will fix when all these conversations are resolved!

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: I am not sure which option would be better:

on the one hand performance wise we could add it in docstring

However if the normal usecase would be to still use the amplitudes after creating the circuit to get to these amplitudes then copying it in the function would be helpful to the user. I do not really know how the usecase looks like normally ...

burgholzer:
Hey 👋🏻 

Vacation is great, but sadly almost over 🥲 but that also means I'll be back at my computer soon and can look a little bit more into this.

I am fairly sure that we are simply not handling the global phase of the circuit correctly. As global phases typically don't play a role for simulating a quantum circuit (they don't change measurement outcomes), we never really cared too much about accurately reflecting them in the simulation.

The simulation method you are using is pretty straight forward: https://github.com/cda-tum/mqt-core/blob/main/include/mqt-core/dd/Simulation.hpp#L26-L44
And there you can see that the Quantum computation's `globalPhase` is never factored into the simulation at all.
So, assuming that the circuits you have created include a global phase, this will expectedly be missing from the final statevector.
This should be fairly easy to fix though. Before returning `e`, the circuits global phase needs to be multiplied to the DDs top edge weight.
It's a bit hard to get everything right from memory without access to my computer, but something like the following should work:
```c++
// create temporary copy for reference counting
auto oldW = e.w;
// adjust for global phase
e.w *= qc.globalPhase;
// adjust reference count
dd.cn.incRef(e.w);
dd.cn.decRef(oldW);
```

M-J-Hochreiter:
Ok, I will try this first thing tomorrow.

So you say I should just adjust the statevector manually according to the global phase (like you indicated with your code snippet) in the tests, right?

burgholzer:
> Ok, I will try this first thing tomorrow.
> 
> 
> 
> So you say I should just adjust the statevector manually according to the global phase (like you indicated with your code snippet) in the tests, right?

Nah, I'd rather just fix the problem at its origin and patch up the simulation method with the code snippet above (or a functioning version thereof; the code above is likely missing a lookup `e.w = dd.cn.lookup(e.w);` before adjusting the refcount).
It's a simple multiplication and one lookup + reference count adjustment. So it's not computationally expensive. Which is why we should probably not only apply this correction in tests, but during the actual simulation itself.
Hope that makes sense 🙂

M-J-Hochreiter:
Makes sense, I think ahah 🙂
However I do not believe that this change should be made withing this pull request? Or do you want me to include the change here, if I manage to make it work

burgholzer:
> Makes sense, I think ahah 🙂
> 
> However I do not believe that this change should be made within this pull request? Or do you want me to include the change here, if I manage to make it work

In a perfect world, the fix would be in a separate PR together with at least one test that failed before the PR and passes with the fix in the PR. If you are up for that, it would be most welcome.
However, I'd also be fine to include the fox as part of this PR given how it's the first time this has come up as an issue.

M-J-Hochreiter:
> ```c++
> // create temporary copy for reference counting
> auto oldW = e.w;
> // adjust for global phase
> e.w *= qc.globalPhase;
> // adjust reference count
> dd.cn.incRef(e.w);
> dd.cn.decRef(oldW);
> ```

Hi,

so i just tried this, however there seems to be a problem. It is not possible to change e.w as far as I see as 'e' (Edge) does not have a setter for w the Complex does not provide an implementation for *= operator ...

Trying to do it without the *= operator I have the problem that the * operator returns a ComplexValue not a Complex so I cannot assign it to e.w?

This is my code currently, but it fails at the multiplication assignment
```c++
  if (std::abs(qc.getGlobalPhase()) > 1e-10) {
    // create temporary copy for reference counting
    auto oldW = e.w;
    // adjust for global phase
    e.w = e.w * qc.getGlobalPhase();
    e.w = dd.cn.lookup(e.w);
    // adjust reference count
    dd.cn.incRef(e.w);
    dd.cn.decRef(oldW);
  }
```

burgholzer:
> > ```c++
> > // create temporary copy for reference counting
> > auto oldW = e.w;
> > // adjust for global phase
> > e.w *= qc.globalPhase;
> > // adjust reference count
> > dd.cn.incRef(e.w);
> > dd.cn.decRef(oldW);
> > ```
> 
> Hi,
> 
> so i just tried this, however there seems to be a problem. It is not possible to change e.w as far as I see as 'e' (Edge) does not have a setter for w and Complex does not provide an implementation for *= operator ...
> 
> Trying to do it without the *= operator I have the problem that the * operator returns a ComplexValue not a Complex so I cannot assign it to e.w?
> 
> This is my code currently, but it fails at the multiplication assignment
> 
> ```c++
>   if (std::abs(qc.getGlobalPhase()) > 1e-10) {
>     // create temporary copy for reference counting
>     auto oldW = e.w;
>     // adjust for global phase
>     e.w = e.w * qc.getGlobalPhase();
>     e.w = dd.cn.lookup(e.w);
>     // adjust reference count
>     dd.cn.incRef(e.w);
>     dd.cn.decRef(oldW);
>   }
> ```

Back at my computer again. Just pushed https://github.com/cda-tum/mqt-core/pull/543/commits/33d4cedd1dcaa02a089428034be86af56f283ae6 which should properly resolve this issue.

M-J-Hochreiter:
Ah thanks!

However there still seems to be a problem with applying the global phase. E.g. expecting vector 
```
[
   (0,  0),
   (0.707107,  0),
   (-0.707107,  0),
   (0,  0)
]``` 
and getting
```
[
    (0,  0), 
    (0, -0.707107), 
    (0,  0.707107), 
    (0,  0)
]
```
It still seems to be rotating the vector weirdly but I am not sure where this is coming from. Is the global phase applied correctly?

**Comment by burgholzer** [comment]
Hm. Looks like it might be off by a factor of `pi` here.
Which kind of gates are you using in your circuit construction? 
It could very well be that one of our internal representations for the gates is only correct up to global phase. Although I'd still find this quite strange.
Qiskit is reporting the "correct" state vector when you simulate the resulting circuit?

**Comment by M-J-Hochreiter** [comment]
My circuit only uses RZ,RY and controlled not gates. 
The most basic when trying to create the vector 
``` 
[
(1 / sqrt(2),  0)
(-1 / sqrt(2),  0)
]
 ```

gives the circuit: 
```
1:  ry  p: (1.570796326794896558)
2:  rz  p: (3.141592653589793116)
```
with global phase -pi/2 which in your implementation gets normalized to (-pi/2) + 2*pi

simulating this exact circuit with qiskit gives me the state-vector:
```
[
(-0.70710678,  0)
(0.70710678,  0)
]
```
Exactly the vector I am trying to create. With simulation in the test i get the vector:
```
[
(-0.707107,  0)
(0.707107,  -0)
]
```

burgholzer:
> My circuit only uses RZ,RY and controlled not gates. The most basic when trying to create the vector
> 
> ```
> [
> (1 / sqrt(2),  0)
> (-1 / sqrt(2),  0)
> ]
> ```
> 
> gives the circuit:
> 
> ```
> 1:  ry  p: (1.570796326794896558)
> 2:  rz  p: (3.141592653589793116)
> ```
> 
> with global phase -pi/2 which in your implementation gets normalized to (-pi/2) + 2*pi
> 
> simulating this exact circuit with qiskit gives me the state-vector:
> 
> ```
> [
> (-0.70710678,  0)
> (0.70710678,  0)
> ]
> ```
> 
> Exactly the vector I am trying to create. With simulation in the MQT I get the vector:
> 
> ```
> [
> (-0.707107,  0)
> (0.707107,  -0)
> ]
> ```

Maybe I am overlooking things now, but in the example you show, everything is correct, i.e., both simulations yield the same state vector (`-0 == 0`!)

burgholzer:
Alright. Glad that this could be resolved 👍🏼 

For some reason I currently cannot push to your fork anymore. However, below is a patch for resolving the linking errors
```patch
Index: src/algorithms/CMakeLists.txt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/algorithms/CMakeLists.txt b/src/algorithms/CMakeLists.txt
--- a/src/algorithms/CMakeLists.txt	(revision d521ec6f4f6a251ad5591894628a5a10af1b8a57)
+++ b/src/algorithms/CMakeLists.txt	(revision 948ec2b18cb510a987a4aa09fb1024efb65791c0)
@@ -22,7 +22,7 @@
   target_link_libraries(
     ${MQT_CORE_TARGET_NAME}-algorithms
     PUBLIC MQT::CoreIR
-    PRIVATE MQT::ProjectOptions MQT::ProjectWarnings)
+    PRIVATE MQT::CoreCircuitOptimizer MQT::ProjectOptions MQT::ProjectWarnings)
 
   # set include directories
   target_include_directories(
```

I'll quickly go through the open questions and subsequently will review the PR.
Thanks for all your work on this!

src/algorithms/StatePreparation.cpp
burgholzer: Everything seems to compile now and clang-tidy seems to be happy, so I am resolving this conversation.

src/algorithms/StatePreparation.cpp
burgholzer: Ah, that makes sense. Did not consider this before. Then let's just leave this here 👍🏼

src/algorithms/StatePreparation.cpp
burgholzer: I just looked into this again and I believe there is really not much to do here at the moment. This could be slightly improved when we move the project to C++20, but that's not happening anytime soon I suppose.

include/mqt-core/algorithms/StatePreparation.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *Integrated Circuits and Systems, vol. 25, no. 6, pp. 1000-1010, June 2006,
+ *doi: 10.1109/TCAD.2005.855930.
+ *
+ * @param amplitudes state (vector) to prepare. Must be normalized and have a
+ *size that is a power of two
+ * @param Eps precision wanted for computations, default 1e-10
+ * @return quantum computation that prepares the state
+ * @throws invalid_argument @p amplitudes is not normalized or its length is not
+ *a power of two
+ **/
```
burgholzer:
```suggestion
/**
 * @brief Prepares a generic quantum state from a list of normalized
 *        complex amplitudes
 *
 * Adapted implementation of IBM Qiskit's State Preparation:
 * https://github.com/Qiskit/qiskit/blob/e9ccd3f374fd5424214361d47febacfa5919e1e3/qiskit/circuit/library/data_preparation/state_preparation.py
 * based on the following paper:
 *
 *      V. V. Shende, S. S. Bullock and I. L. Markov,
 *      "Synthesis of quantum-logic circuits", in IEEE Transactions on
 *      Computer-Aided Design of Integrated Circuits and Systems,
 *      vol. 25, no. 6, pp. 1000-1010, June 2006,
 *      doi: 10.1109/TCAD.2005.855930.
 *
 * @param amplitudes State (vector) to prepare.
 * Must be normalized and have a size that is a power of two.
 * @param eps Precision wanted for computations, default 1e-10
 * @return Quantum computation that prepares the state
 * @throws invalid_argument If @p amplitudes is not normalized or its length is
 * not a power of two.
 **/
```
Just some small reformatting and rewording so the docstring looks a little prettier.

src/algorithms/StatePreparation.cpp
burgholzer, line 1: Unfortunately, I don't seem to have push access to your fork from my laptop.
Hence, here's a small patch that improves small parts of the implementation and fixes some more recent clang-tidy suggestions. You should be able to simply apply that patch.
Alternatively, you can add me as a contributor to your fork, which also allows me to push there.

```patch
Index: src/algorithms/StatePreparation.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/algorithms/StatePreparation.cpp b/src/algorithms/StatePreparation.cpp
--- a/src/algorithms/StatePreparation.cpp	(revision d521ec6f4f6a251ad5591894628a5a10af1b8a57)
+++ b/src/algorithms/StatePreparation.cpp	(date 1739897153127)
@@ -29,6 +29,7 @@
 namespace qc {
 using Matrix = std::vector<std::vector<double>>;
 
+namespace {
 template <typename T>
 [[nodiscard]] auto twoNorm(const std::vector<T>& vec) -> double {
   double norm = 0;
@@ -39,7 +40,8 @@
 }
 
 template <typename T>
-[[nodiscard]] auto isNormalized(const std::vector<T>& vec, double eps) -> bool {
+[[nodiscard]] auto isNormalized(const std::vector<T>& vec, const double eps)
+    -> bool {
   return std::abs(1 - twoNorm(vec)) < eps;
 }
 
@@ -62,7 +64,8 @@
           // Each element of matrix A is
           // multiplied by whole Matrix B
           // resp and stored as Matrix C
-          newMatrix[i * rowB + k][j * colB + l] = matrixA[i][j] * matrixB[k][l];
+          newMatrix[(i * rowB) + k][(j * colB) + l] =
+              matrixA[i][j] * matrixB[k][l];
         }
       }
     }
@@ -70,9 +73,8 @@
   return newMatrix;
 }
 
-[[nodiscard]] auto createIdentity(size_t size) -> Matrix {
-  Matrix identity{
-      std::vector<std::vector<double>>(size, std::vector<double>(size, 0))};
+[[nodiscard]] auto createIdentity(const size_t size) -> Matrix {
+  auto identity{std::vector(size, std::vector<double>(size, 0))};
   for (size_t i = 0; i < size; ++i) {
     identity[i][i] = 1;
   }
@@ -96,14 +98,14 @@
 
 /**
  * @brief recursive implementation that returns multiplexer circuit
- * @param target_gate : Ry or Rz gate to apply to target qubit, multiplexed
+ * @param targetGate : Ry or Rz gate to apply to target qubit, multiplexed
  * over all other "select" qubits
  * @param angles : list of rotation angles to apply Ry and Rz
- * @param lastCnot : add last cnot if true
+ * @param lastCnot : add last CNOT if true
  * @return multiplexer circuit as QuantumComputation
  */
 [[nodiscard]] auto multiplex(OpType targetGate, std::vector<double> angles,
-                             bool lastCnot) -> QuantumComputation {
+                             const bool lastCnot) -> QuantumComputation {
   size_t const listLen = angles.size();
   auto const localNumQubits = static_cast<Qubit>(
       std::floor(std::log2(static_cast<double>(listLen))) + 1);
@@ -115,8 +117,7 @@
     return multiplexer;
   }
 
-  Matrix const matrix{std::vector<double>{0.5, 0.5},
-                      std::vector<double>{0.5, -0.5}};
+  Matrix const matrix{std::vector{0.5, 0.5}, std::vector{0.5, -0.5}};
   Matrix const identity = createIdentity(1 << (localNumQubits - 2));
   Matrix const angleWeights = kroneckerProduct(matrix, identity);
 
@@ -155,7 +156,8 @@
 }
 
 [[nodiscard]] auto blochAngles(std::complex<double> const complexA,
-                               std::complex<double> const complexB, double eps)
+                               std::complex<double> const complexB,
+                               const double eps)
     -> std::tuple<std::complex<double>, double, double> {
   double theta{0};
   double phi{0};
@@ -170,14 +172,14 @@
     finalT = aAngle + bAngle;
     phi = bAngle - aAngle;
   }
-  return {finalR * exp(std::complex<double>{0, 1} * finalT / 2.), theta, phi};
+  return {std::polar(finalR, finalT / 2.), theta, phi};
 }
 
 // works out Ry and Rz rotation angles used to disentangle LSB qubit
 // rotations make up block diagonal matrix U
 [[nodiscard]] auto
 rotationsToDisentangle(const std::vector<std::complex<double>>& amplitudes,
-                       double eps)
+                       const double eps)
     -> std::tuple<std::vector<std::complex<double>>, std::vector<double>,
                   std::vector<double>> {
   size_t const amplitudesHalf = amplitudes.size() / 2;
@@ -191,7 +193,7 @@
 
   for (size_t i = 0; i < amplitudesHalf; ++i) {
     auto [remains, theta, phi] =
-        blochAngles(amplitudes[2 * i], amplitudes[2 * i + 1], eps);
+        blochAngles(amplitudes[2 * i], amplitudes[(2 * i) + 1], eps);
     remainingVector.emplace_back(remains);
     // minus sign because we move it to zero
     thetas.emplace_back(-theta);
@@ -203,7 +205,8 @@
 // creates circuit that takes desired vector to zero
 [[nodiscard]] auto
 gatesToUncompute(std::vector<std::complex<double>>& amplitudes,
-                 size_t numQubits, double eps) -> QuantumComputation {
+                 const size_t numQubits, const double eps)
+    -> QuantumComputation {
   QuantumComputation disentangler{numQubits};
   for (size_t i = 0; i < numQubits; ++i) {
     // rotations to disentangle LSB
@@ -222,13 +225,13 @@
       QuantumComputation rzMultiplexer = multiplex(RZ, phis, addLastCnot);
       // append rzMultiplexer to disentangler, but it should only attach on
       // qubits i-numQubits, thus "i" is added to the local qubit indices
-      for (auto& op : rzMultiplexer) {
+      for (const auto& op : rzMultiplexer) {
         for (auto& target : op->getTargets()) {
           target += static_cast<Qubit>(i);
         }
         // as there were some systematic compiler errors with accessing the
         // qubit directly the controls are collected and then newly set
-        qc::Controls newControls;
+        Controls newControls;
         for (const auto& control : op->getControls()) {
           newControls.emplace(control.qubit + static_cast<Qubit>(i));
         }
@@ -243,13 +246,13 @@
       // attach on qubits i-numQubits, thus "i" is added to the local qubit
       // indices
       ryMultiplexer.reverse();
-      for (auto& op : ryMultiplexer) {
+      for (const auto& op : ryMultiplexer) {
         for (auto& target : op->getTargets()) {
           target += static_cast<Qubit>(i);
         }
         // as there were some systematic compiler errors with accessing the
         // qubit directly the controls are collected and then newly set
-        qc::Controls newControls;
+        Controls newControls;
         for (const auto& control : op->getControls()) {
           newControls.emplace(control.qubit + static_cast<Qubit>(i));
         }
@@ -266,9 +269,10 @@
   }
   return disentangler;
 }
+} // namespace
 
 auto createStatePreparationCircuit(
-    std::vector<std::complex<double>>& amplitudes, double eps)
+    std::vector<std::complex<double>>& amplitudes, const double eps)
     -> QuantumComputation {
 
   if (!isNormalized(amplitudes, eps)) {
```

test/algorithms/test_statepreparation.cpp
burgholzer, line 1: Would you mind adding two tests to also test the failure scenarios where the method throws an exception? These should be fairly straightforward.

src/algorithms/StatePreparation.cpp
burgholzer: From looking at the tests that you wrote, I believe it is better and less surprising to make this a const reference and explicitly copy the state in the circuit creation routine. This can be noted in the docstring of the method.
In essence, this does not really limit the scalability of the method so much. It effectively decreases the maximum number of qubits that can be handled by 1 as one needs double the amount of memory for storing the original vector and the copy for the synthesis.

burgholzer (changes requested):
Alright, many thanks for all your work on this.
This is as close to ready as it gets and turned out nicely.
I have a couple of final suggestions left, which you will find in the respective inline comments. After addressing those, this is ready to go in 🚀 

Side note: while reading through the code I couldn't help but notice that it would be very interested to realise the underlying algorithm using decision diagrams as a data structure, which should simplify some of the code and also improve the general performance. Might be a nice follow up project 😌

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: Yes also believe this to be the better option.

Done in 875dc15

src/algorithms/StatePreparation.cpp
M-J-Hochreiter: changes implemented in 156aa96

test/algorithms/test_statepreparation.cpp
M-J-Hochreiter: done in cdaaea2

M-J-Hochreiter:
Implemented all the requested changes. Looks done to me 😄

test/algorithms/test_statepreparation.cpp
```diff
@@ -70,3 +71,23 @@ TEST_P(StatePreparation, StatePreparationCircuitSimulation) {
...
+TEST(StatePreparation, StatePreparationsAmplitudesNotPowerOf2) {
+  const auto amplitudes =
+      std::vector<std::complex<double>>{std::complex<double>{1 / std::sqrt(3)},
+                                        std::complex<double>{1 / std::sqrt(3)},
+                                        std::complex<double>{1 / std::sqrt(3)}};
+
+  // cast to avoid linting nodiscard
+  ASSERT_THROW(static_cast<void>(qc::createStatePreparationCircuit(amplitudes)),
+               std::invalid_argument);
+}
```
burgholzer:
```suggestion
  ASSERT_THROW(std::ignore = qc::createStatePreparationCircuit(amplitudes),
               std::invalid_argument);
}

TEST(StatePreparation, StatePreparationsAmplitudesNotPowerOf2) {
  const auto amplitudes =
      std::vector<std::complex<double>>{std::complex<double>{1 / std::sqrt(3)},
                                        std::complex<double>{1 / std::sqrt(3)},
                                        std::complex<double>{1 / std::sqrt(3)}};

  ASSERT_THROW(std::ignore = qc::createStatePreparationCircuit(amplitudes),
               std::invalid_argument);
}
```
There is a slightly more idiomatic way to ignore nodiscard return values.

burgholzer (approved):
Everything looks good to me now. 🙌🏼 
Many thanks for all your work on this. 🙏🏼  
The contribution turned out quite nicely! ✨ 

I'll apply one last small suggestion and then this can go in 🚀

test/algorithms/test_statepreparation.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "dd/Package.hpp"
+#include "dd/Simulation.hpp"
+#include "ir/QuantumComputation.hpp"
+
+#include <cmath>
+#include <complex>
+#include <cstddef>
+#include <gtest/gtest.h>
+#include <memory>
+#include <stdexcept>
```
burgholzer, line 20:
```suggestion
#include <stdexcept>
#include <utility>
```

test/algorithms/test_statepreparation.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <utility>
```
burgholzer:
```suggestion
#include <tuple>
```
third time's the charm

M-J-Hochreiter:
Thank you for all your valuable Feedback on this!
Glad that even with a bit of delay everything worked out 😄

PR #563, opened 2024-03-13
♻️ Refactor garbage reduction

reb-ddm:
## Description

Until now, `reduceGarbage()` simply summed the entries of garbage qubits in the matrix, in order to ignore the contribution of garbage qubits in the circuit. This behaviour, however, only works for reversible circuis. If, e.g., we have a vector `[sqrt(2)/2, 0, -sqrt(2)/2, 0]`, then summing the weights would lead to an invalid quantum state `[0, 0, 0, 0]`. In order to be sure that probabilities of the result state is the sum of the probabilities of the initial state, we now compute `sqrt(|a|^2 + |b|^2)` when summing two weigths `a` and `b` in the function `reduceGarbage`. In this way, the result of reducing garbage of the previous example `[sqrt(2)/2, 0, -sqrt(2)/2, 0]` would be `[1, 0, 0, 0]`, as expected.

In order to perform this sum in the DD representation, it suffices to compute `sqrt(|a|^2 + |b|^2)` for the respective weights of the two DDs. This is calculated in the function `addMagnitudes()`.

Additionally, a flag `normalizeWeights` was added to `reduceGarbage()`. If it is set to `true`, the function changes all weights in the DD to their magnitude, also for non-garbage qubits. This is used for checking partial equivalence of circuits. For partial equivalence, only the measurement probabilities are considered, so we need to consider only the magnitudes of each entry. 

## Checklist:


burgholzer:
Hey @reb-ddm 👋🏼 

I took the liberty to directly edit the PR a little bit. Most notably, I simplified the code a little bit and changed the `addMagnitude` method so that it uses a compute table that is created on-demand. This should avoid having to unconditionally allocate a substantial amount of memory for a feature that not every user will be using.

With that, I think this PR should be ready to merge once CI passes.
I'll still keep it open for now, while I have a look at the related QCEC PR.

PR #592, opened 2024-04-19
:sparkles: Support for Qiskit's final_layout

### LINKED ISSUE #439: ✨ Support for Qiskit's `final_layout`

When reasoning about quantum circuits there are two degrees of freedom that can always lead to problems/confusion:
An arbitrary permutation of a circuit's qubits might be applied at the very beginning of the circuit (referred to as an "initial layout") or at the end of the circuit (referred to as an "output permutation").
In almost any case, a circuit containing such permutations would still be considered equivalent to a circuit without such permutations.
This is especially relevant for verifying the equivalence of quantum circuits and the results of compilation flows in particular.

For a long time, Qiskit has supported a `_layout` property for its `QuantumCircuit` objects that captured the initial layout applied during compilation. Within the MQT we use this property to infer the `initialLayout` for our `QuantumComputation` objects.
In `0.22.0` (which is now our minimum supported version), Qiskit changed the `_layout` property to a proper `TranspileLayout` class.
`0.23.0` introduced an additional `final_layout` component that (in some confusing fashion) captures the output permutation.
This should allow to also infer the output permutation from Qiskit.
Right now, we use the measurements in a circuit for that purpose, which is rathe prone to errors (e.g., someone not adding measurements to their circuit before compilation).

Starting with `0.45.0.` (which should be the next minor release), the ergonomics of the `TranspileLayout` class have been considerably improved (see https://github.com/Qiskit/qiskit/pull/10835).
That should yield a blueprint for how to interpret the `final_layout` and to infer the `outputPermutation` from Qiskit circuits.

A potential solution would first look if the `final_layout` attribute is present.
If so, it would check whether the new convenient syntax (>=0.45.0) is available.
If not, it replicates the logic to parse the output permutation.
If no final layout is present, the solution should check for measurements in the circuit and infer the output permutation from there (as it currently does).

TeWas:
## Description

This PR adds support for Qiskit's `final_layout` attribute when importing circuits from Qiskit. It interprets the `output_permutation` of an MQT `QuantumComputation` object based on Qiskit's `final_index_layout()`, which maps the original input circuit's qubit indices to the final physical qubit indices after routing. This mapping is the result of applying `routing_permutation()` to the `initial_layout`. Additionally, it fixes the interpretation of the output_permutation in `initializeIOMapping()` to align with Qiskit's final_layout.

Fixes #439

## Checklist:

### DIFF
--- a/src/QuantumComputation.cpp
+++ b/src/QuantumComputation.cpp
@@ -165,8 +165,7 @@ void QuantumComputation::initializeIOMapping() {
           // output permutation was already set before -> permute existing
           // values
           const auto current = outputPermutation.at(qubitidx);
-          if (static_cast<std::size_t>(qubitidx) != bitidx &&
-              static_cast<std::size_t>(current) != bitidx) {
+          if (static_cast<std::size_t>(current) != bitidx) {
             for (auto& p : outputPermutation) {
               if (static_cast<std::size_t>(p.second) == bitidx) {
                 p.second = current;
--- a/src/mqt/core/plugins/qiskit.py
+++ b/src/mqt/core/plugins/qiskit.py
@@ -395,6 +395,11 @@ def _import_layouts(qc: QuantumComputation, circ: QuantumCircuit) -> None:
     if final_layout is None:
         return
 
+    # final_index_layout creates a list of final positions for input circuit qubits
+    final_index_layout = layout.final_index_layout(filter_ancillas=False)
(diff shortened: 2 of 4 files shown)


burgholzer (approved):
Many thanks for this contribution. Nicely done! 🎉 
I just rebased the PR on the current main branch and simplified some of the checks a little bit.

PR #610, opened 2024-05-24
Hybrid Neutral Atom Mapper Functionality

lsschmid:
## Add Hybrid NA Mapper Functionality

This adds two things:
1. Move operation: Abstract operation describing the movement of an atom from one coordinate to another.
2. AodOperation: Describing the movements of the AOD to perform the above Move operation. The dump to qasm returns a format similar to qasm.

## Checklist:


include/mqt-core/operations/OpType.hpp
```diff
@@ -60,7 +60,12 @@ enum OpType : std::uint8_t {
   MultiATrue,
   MultiAFalse,
   // Number of OpTypes
-  OpCount
+  OpCount,
+  // Neutral atom shuttling operations
+  Move,
+  AodActivate,
+  AodDeactivate,
+  AodMove,
```
burgholzer, line 66: These have to go before the `OpCount` enum value because that one is used to track the number of different operations, thus it always has to be last. Could be worth adding a comment on that.

include/mqt-core/operations/Operation.hpp
```diff
@@ -56,6 +56,9 @@ class Operation {
   [[nodiscard]] virtual std::size_t getNcontrols() const {
     return controls.size();
   }
+  [[nodiscard]] std::size_t getNqubits() const {
+    return getNcontrols() + getNtargets();
+  }
```
burgholzer: This method should probably be `virtual` and overridden in `NonUnitaryOperation` (where `controls` don't make sense) and possibly in `CompoundOperation` to make sure this considers all the used qubits there.

include/mqt-core/operations/AodOperation.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include "cmath"
+#include "cstddef"
+#include "cstdint"
+#include "memory"
+#include "operations/Operation.hpp"
+#include "ostream"
+#include "sstream"
+#include "string"
+#include "tuple"
+#include "vector"
```
burgholzer: Please distinguish between system includes (STL, third party libs, etc.) using `#include <...>` and library internal includes using `#include "..."`

```suggestion
#include "Control.hpp"
#include "Definitions.hpp"
#include "OpType.hpp"
#include <algorithm>
#include <cmath>
#include <cstddef>
#include <cstdint>
#include <memory>
#include "operations/Operation.hpp"
#include <ostream>
#include <sstream>
#include <string>
#include <tuple>
#include <vector>
```

(the above suggestion will need a reformatting of the include order, which clang-format does automatically for you, so you can just accept the suggestion and let the pre-commit hook do its job)

src/operations/AodOperation.cpp
```diff
@@ -0,0 +1,112 @@
...
+  name = toString(type);
+}
+
+void AodOperation::dumpOpenQASM(std::ostream& of, const RegisterNames& qreg,
+                                [[maybe_unused]] const RegisterNames& creg,
+                                size_t /*indent*/, bool openQASM3) const {
+  if (openQASM3) {
+    throw std::runtime_error("AOD operations are not supported in OpenQASM3, "
+                             "please use the OpenQASM2 output format.");
+  }
```
burgholzer: Why though? Can't you simply output the same code for OpenQASM 3.0?

src/operations/AodOperation.cpp
```diff
@@ -0,0 +1,112 @@
...
+  of << std::setprecision(std::numeric_limits<fp>::digits10);
+  of << name;
```
burgholzer, line 80: I think you might be missing the proper indent here, see, e.g., the corresponding code in the standard operations class.

```c++
of << std::string(indent * OUTPUT_INDENT_SIZE, ' ');
```

src/operations/AodOperation.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+  of.seekp(-1, std::ios_base::end);
+  of << ";\n";
+}
+
+void AodOperation::invert() {
+  if (type == OpType::AodMove) {
+    for (auto& op : operations) {
+      op.start = op.end;
+      op.end = -op.start;
```
burgholzer: Are you sure this does what you intend? Right now this sets `op.start` and `op.end` to `op.end` and `-op.end`. That does not seem quite right to me.

test/test_operation.cpp
```diff
@@ -135,3 +138,106 @@ TEST(Operation, IsDiagonalGate) {
...
+                              {qc::Dimension::X, qc::Dimension::Y}, {0.0, 1.0},
+                              {1.0, 3.0});
+  std::stringstream ss;
+  qc::RegisterNames qreg;
+  qreg.emplace_back("q", "q[0]");
+  qreg.emplace_back("q", "q[1]");
+  qc::RegisterNames const creg{{"c", "c"}};
+  move.dumpOpenQASM(ss, qreg, creg, 0, false);
+
+  EXPECT_FALSE(ss.str().empty());
```
burgholzer: this doesn't really test anything. Couldn't you just test the actual string that results?
Or, potentially even better, create a circuit with the operation, dump it and compare against a reference.

burgholzer (changes requested):
Finally found the time to go through this PR :relieved:
Overall the changes here look good :+1:
I pushed two minor changes directly and added further smaller comments in the files themselves. Most of these should be pretty minor.

I have one medium-large question and one bigger question I'd like to get out of the way before proceeding here:

First, what about reading the produced QASM circuits back in? Right now, no adaptation on the QASM parser has been performed so far. So the produced QASM files cannot be read anywhere. While not a dealbreaker, it would certainly be nice to support that, wouldn't it?

Second, and probably the bigger question: @ystade recently introduced the `na` subcomponent in mqt-core and mqt-qmap, where he (amongst other things) also defined a ShuttlingOperation (with load, move, and store types). Now this PR implements essentially the same concept, but in a different fashion. While I would be ok with that in the short-term, this duplication is certainly no maintainable solution for the future. How hard do you think it would be to base your changes (especially those in QMAP) on @ystade's infrastructure for neutral atom computations? From the changes here, the QASM dump would be the most important thing to port to the existing `NAComputation` and `NAOperation` classes.
(Expect me asking a similar question in the mqt-qmap PR)

include/mqt-core/operations/OpType.hpp
lsschmid: Changed the order and added a short comment for the future

include/mqt-core/operations/Operation.hpp
lsschmid: Changed it to getUsedQubits().size which should make sense for all Operations I guess.

include/mqt-core/operations/AodOperation.hpp
lsschmid: Didn't know about that, will do in future 👍

src/operations/AodOperation.cpp
lsschmid: Wasn't sure if this is still valid Qasm3. Also the output is not real qasm2 anyway, so I decided to keep it as simple as possible.

burgholzer: Based on the global comment on the QASM import: no QASM parser can currently read the produced files anyway. Might as well enable the special syntax for QASM 2 and QASM 3 😌

src/operations/AodOperation.cpp
lsschmid: Added the indent. Not sure if this is sufficient as I'm not sure how correct qasm actually has to look like.

burgholzer: The indent here is purely for cosmetic purposes. Whitespace does not really play that much of a role in OpenQASM

src/operations/AodOperation.cpp
lsschmid: That's right, fixed that.

test/test_operation.cpp
lsschmid: Added a test to compare to the actually correct string.

lsschmid:
> Finally found the time to go through this PR 😌 Overall the changes here look good 👍 I pushed two minor changes directly and added further smaller comments in the files themselves. Most of these should be pretty minor.
> 
> I have one medium-large question and one bigger question I'd like to get out of the way before proceeding here:
> 
> First, what about reading the produced QASM circuits back in? Right now, no adaptation on the QASM parser has been performed so far. So the produced QASM files cannot be read anywhere. While not a dealbreaker, it would certainly be nice to support that, wouldn't it?
> 
> Second, and probably the bigger question: @ystade recently introduced the `na` subcomponent in mqt-core and mqt-qmap, where he (amongst other things) also defined a ShuttlingOperation (with load, move, and store types). Now this PR implements essentially the same concept, but in a different fashion. While I would be ok with that in the short-term, this duplication is certainly no maintainable solution for the future. How hard do you think it would be to base your changes (especially those in QMAP) on @ystade's infrastructure for neutral atom computations? From the changes here, the QASM dump would be the most important thing to port to the existing `NAComputation` and `NAOperation` classes. (Expect me asking a similar question in the mqt-qmap PR)

Regarding the qasm import:
Back when I wrote the code I also added the correct parsing commands. This took me quite some time as I'm not really familiar with parsing in general. Unfortuantly in the meantime the parser was updated to support qasm3 which broke my version and I could not see an easy way to transfer my code. As for now, the only usecase where the parsing would be interesting is to schedule circuits which have been mapped and saved previously. This is indeed a possible/interesting option but currently not used by us and my knowledge on parsing is not sufficient to do some "quick fix" to that.

Regarding the NA unification:
Obviously, my code already existed before @ystade started with his part. We had some discussion back then on reusing the code but then decided to create two separate versions as a start as we required different things. A unification might probably be possible and would probably make sense for common export/import but currently this is a non-trivial task as we took different approaches to how we handle/represent the AOD moves.

src/operations/AodOperation.cpp
lsschmid: Sure

burgholzer:
> > Finally found the time to go through this PR 😌 Overall the changes here look good 👍 I pushed two minor changes directly and added further smaller comments in the files themselves. Most of these should be pretty minor.
> > I have one medium-large question and one bigger question I'd like to get out of the way before proceeding here:
> > First, what about reading the produced QASM circuits back in? Right now, no adaptation on the QASM parser has been performed so far. So the produced QASM files cannot be read anywhere. While not a dealbreaker, it would certainly be nice to support that, wouldn't it?
> > Second, and probably the bigger question: @ystade recently introduced the `na` subcomponent in mqt-core and mqt-qmap, where he (amongst other things) also defined a ShuttlingOperation (with load, move, and store types). Now this PR implements essentially the same concept, but in a different fashion. While I would be ok with that in the short-term, this duplication is certainly no maintainable solution for the future. How hard do you think it would be to base your changes (especially those in QMAP) on @ystade's infrastructure for neutral atom computations? From the changes here, the QASM dump would be the most important thing to port to the existing `NAComputation` and `NAOperation` classes. (Expect me asking a similar question in the mqt-qmap PR)
> 
> Regarding the qasm import: Back when I wrote the code I also added the correct parsing commands. This took me quite some time as I'm not really familiar with parsing in general. Unfortuantly in the meantime the parser was updated to support qasm3 which broke my version and I could not see an easy way to transfer my code. As for now, the only usecase where the parsing would be interesting is to schedule circuits which have been mapped and saved previously. This is indeed a possible/interesting option but currently not used by us and my knowledge on parsing is not sufficient to do some "quick fix" to that.

Fully understand that point. The new parser is structured quite a bit differently than the old one. Adding a parser for the new operations would definitely make sense in the future. Also triggered by the comment below, I would postpone that implementation to a point where we consolidated on the particular details when it comes to NA Compilation and how to represent the corresponding computations. At that point, this can be a student project. Might be worth tracking that in an issue.
It isn't valid QASM in either version of the standard.
For it to be valid, you would have to add an `opaque` gate declaration for the operation to the top of the file, eliminate the `;` from the parameter list, and you'd have to somehow handle the fact that the parameter list can have arbitrary length.

> Regarding the NA unification: Obviously, my code already existed before @ystade started with his part. We had some discussion back then on reusing the code but then decided to create two separate versions as a start as we required different things. A unification might probably be possible and would probably make sense for common export/import but currently this is a non-trivial task as we took different approaches to how we handle and represent the AOD moves.

It's obvious that this is a non-trivial task. It's still important to think about this early as I would really want to avoid to develop two completely separate strands of features here.
Just for clarification: 
To me, the two AOD representations do not look fundamentally different. You choose to explicitly encode the direction and use floating point numbers for start and end coordinates.
@ystade chose to implicitly store the direction by assuming a 2D plane and using `(x, y)` (integer) coordinates for the start and end coordinates. Functionality-wise, these two are equivalent (at least in 2D and assuming the integer resolution is enough).
Do you really think that adapting your code to that class would be that much of an effort? Most likely, I am simply overlooking things given that I have not spent enough time with the QMAP code yet. So I am really interested in your judgement here.
If this is too much to ask for now (which is totally ok), then I'd at least like to have a clear plan on who has the responsibility to consolidate that in the future and a rough timeframe.

lsschmid:
> Fully understand that point. The new parser is structured quite a bit differently than the old one. Adding a parser for the new operations would definitely make sense in the future. Also triggered by the comment below, I would postpone that implementation to a point where we consolidated on the particular details when it comes to NA Compilation and how to represent the corresponding computations. At that point, this can be a student project. Might be worth tracking that in an issue. It isn't valid QASM in either version of the standard. For it to be valid, you would have to add an `opaque` gate declaration for the operation to the top of the file, eliminate the `;` from the parameter list, and you'd have to somehow handle the fact that the parameter list can have arbitrary length.
>
Yes, exactly; this would probably be the way to go, also for the consolidated version below.


> It's obvious that this is a non-trivial task. It's still important to think about this early as I would really want to avoid to develop two completely separate strands of features here. Just for clarification: To me, the two AOD representations do not look fundamentally different. You choose to explicitly encode the direction and use floating point numbers for start and end coordinates. @ystade chose to implicitly store the direction by assuming a 2D plane and using `(x, y)` (integer) coordinates for the start and end coordinates. Functionality-wise, these two are equivalent (at least in 2D and assuming the integer resolution is enough). Do you really think that adapting your code to that class would be that much of an effort? Most likely, I am simply overlooking things given that I have not spent enough time with the QMAP code yet. So I am really interested in your judgement here. If this is too much to ask for now (which is totally ok), then I'd at least like to have a clear plan on who has the responsibility to consolidate that in the future and a rough timeframe.
>
That's correct. As the basic functionality is quite clear (moving coordinates in 2D), I think both sides could be consolidated in a straightforward fashion conceptional-wise, but with small differences (e.g., I require floats instead of int) . An almost similar problem exists for the NeutralAtomArchitecture Class we used for the Mapping Tasks. For now I would keep the two projects separate and then discuss on an overarching Operation Class + Architecure Class in the following weeks, also in the context of a compiler collection.

burgholzer (approved):
Alright, I think we have a plan then 👍🏼
I'll just continue to merge here for now.

Would you mind creating issues for those two big points with some context, so that we keep track of them?
- First and foremost, the consolidation of the operations between mqt-core operations in general, @ystade's NAOperations, and your new operations here.
- Second, and probably as a follow-up, proper QASM parsing and potentially an improved QASM dump

lsschmid:
I'll write the issues 👍

PR #624, opened 2024-06-04
Generalize concept of removing OpTypes

lsschmid:
## Description

To not only remove Identities but also other Operations I wrote a function that generalizes that concept and allows one to remove arbitrary operation types.

## Checklist:

### DIFF
--- a/include/mqt-core/CircuitOptimizer.hpp
+++ b/include/mqt-core/CircuitOptimizer.hpp
@@ -28,6 +28,9 @@ public:
 
   static void removeIdentities(QuantumComputation& qc);
 
+  static void removeOperation(qc::QuantumComputation& qc,
+                              const std::set<OpType>& opTypes, size_t opSize);
+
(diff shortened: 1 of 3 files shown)


lsschmid:
For some reason some other tests fail, not sure why.

Also, probably I don't need the functionality any more, so feel free to delete the PR

burgholzer:
> For some reason some other tests fail, not sure why.
> 
> Also, probably I don't need the functionality any more, so feel free to delete the PR

Just from a Quick Look: The main reason for the tests failing is that the previous identity removal would also remove (multi-)controlled identities, while your version only removes single-qubit identities without controls.
That should be rather easy to fix though by adding a special case for the identity.

In general, I'd be in favour of getting this merged. It seems like a useful generalisation.

lsschmid:
Now it should work. 
As said, I don't really need it anymore though.

Also: I wrote a comment on the compound operations. Please remove it if it looks fine.

PR #628, opened 2024-06-06
Improvements for Trace Computation

### LINKED ISSUE #336: Improvements for Trace Computation
At the moment, the `partial_trace` (and, hence, the `trace`) function in the package are calculated in a straight-forward fashion. As a result, the performance of the operation scales with the number of paths in the decision diagram as opposed to the number of nodes. Introducing a compute table as for the other decision diagram operations fixes this problem. A possible benchmark to test this is to simply create the matrix of an n-qubit Hadamard transform and compute its trace. This matrix has `n` nodes, but `4^n` paths.

Furthermore, it might be possible to implement an improved `trace` function, that does not just call the partial trace and eliminates every variable, and instead just sums up all the diagonal entries.
It might even be beneficial to compute `trace(...)/2^n` in order to normalise the result to the range `[0, 1]`. Instead of computing `trace(...)` and then dividing by `2^n` (which induces numerical problems), the computed trace can be divided by two in each of the levels in order to continuously normalise it throughout the computation.

TeWas:
## Description

This PR introduces improvements for the trace computation: 

1. Adds a compute table for storing results of eliminated nodes during (partial) trace computation
- this allows the full trace computation to scale linearly with the number of nodes
- however, the partial trace computation still scales with the number of paths in the DD if the bottom qubits are to be eliminated:  non-eliminated nodes cannot be stored in the compute table, as this would prevent their proper elimination in subsequent trace calls
2. Normalizes the result

Fixes #336 

## Checklist:

### DIFF
--- a/include/mqt-core/dd/Package.hpp
+++ b/include/mqt-core/dd/Package.hpp
@@ -2003,7 +2003,10 @@ private:
   /**
    * @brief Computes the normalized (partial) trace using a compute table to
    * store results for eliminated nodes.
-   * @details This optimization allows the full trace
+   * @details At each level, perform lookup and store results in the compute
+   * table only if all lower level qubits are eliminated as well.
+   *
+   * This optimization allows the full trace
    * computation to scale linearly with respect to the number of nodes.
    * However, the partial trace computation still scales with the number of
    * paths in the DD when bottom qubits are to be eliminated.
@@ -2035,11 +2038,18 @@ private:
 
     const auto v = a.p->v;
     if (eliminate[v]) {
-      // Lookup nodes marked for elimination in the compute table: if the trace
-      // has already been computed, return the result
+      // Lookup nodes marked for elimination in the compute table if all
+      // lower level qubits are eliminated as well: if the trace has already
+      // been computed, return the result
       auto& computeTable = getTraceComputeTable<Node>();
-      if (const auto* r = computeTable.lookup(a.p); r != nullptr) {
-        return {r->p, r->w * aWeight};
+      if (std::all_of(
+              eliminate.begin(),
+              eliminate.begin() +
+                  static_cast<std::vector<bool>::difference_type>(level),
+              [](bool e) { return e; })) {
+        if (const auto* r = computeTable.lookup(a.p); r) {
+          return {r->p, r->w * aWeight};
+        }
       }
 
       const auto elims = alreadyEliminated + 1;
@@ -2051,7 +2061,16 @@ private:
       if constexpr (std::is_same_v<Node, mNode>) {
         r.w = r.w / 2.0;
       }
-      computeTable.insert(a.p, {r.p, r.w});
+
+      // Insert result into compute table if all lower level qubits are
+      // eliminated as well
+      if (std::all_of(
+              eliminate.begin(),
+              eliminate.begin() +
+                  static_cast<std::vector<bool>::difference_type>(level),
+              [](bool e) { return e; })) {
+        computeTable.insert(a.p, {r.p, r.w});
+      }
       r.w = r.w * aWeight;
       return r;
     }
--- a/include/mqt-core/dd/Package.hpp
+++ b/include/mqt-core/dd/Package.hpp
@@ -2008,13 +2008,14 @@ private:
    * This optimization allows the full trace
    * computation to scale linearly with respect to the number of nodes.
    * However, the partial trace computation still scales with the number of
-   * paths in the DD when bottom qubits are to be eliminated.
+   * paths to the lowest level in the DD that should be traced out.
    *
-   * Normalization is continuously applied, dividing by two at each level
-   * marked for elimination, thereby ensuring that the result is mapped to the
-   * interval [0,1].
-   * @note Normalization is only applied to matrix nodes, as the trace
-   * of density matrices equals 1 by definition.
+   * For matrices, normalization is continuously applied, dividing by two at
+   * each level marked for elimination, thereby ensuring that the result is
+   * mapped to the interval [0,1] (as opposed to the interval [0,2^N]).
+   *
+   * For density matrices, such normalization is not applied as the trace of
+   * density matrices is always 1 by definition.
    */
   template <class Node>
   CachedEdge<Node> trace(const Edge<Node>& a,
--- a/include/mqt-core/dd/Package.hpp
+++ b/include/mqt-core/dd/Package.hpp
@@ -1962,44 +1962,15 @@ public:
     return {r.p, cn.lookup(r.w)};
   }
 
+  // Normalized trace function
   template <class Node>
   ComplexValue trace(const Edge<Node>& a, const std::size_t numQubits) {
     if (a.isIdentity()) {
-      return a.w * std::pow(2, numQubits);
+      return static_cast<ComplexValue>(a.w);
+      ;
     }
     const auto eliminate = std::vector<bool>(numQubits, true);
     return trace(a, eliminate, numQubits).w;
-  }
-
-  template <class Node>
-  ComplexValue trace2(const Edge<Node>& a, const std::size_t numQubits,
-                      bool normalize = false) {
-    const auto aWeight = static_cast<ComplexValue>(a.w);
-    if (aWeight.approximatelyZero()) {
-      return ComplexValue{0., 0.};
-    }
-
-    if (a.isIdentity()) {
-      if (normalize) {
-        return aWeight;
-      }
-      return aWeight * std::pow(2, numQubits);
-    }
-
-    if (a.isTerminal()) {
-      return ComplexValue(a.w);
-    }
-
-    if (numQubits <= 0) {
-      throw std::invalid_argument("Number of Qubits is incorrect");
-    }
-    auto l = trace2(a.p->e[0], numQubits - 1, normalize);
-    auto r = trace2(a.p->e[3], numQubits - 1, normalize);
-    auto w = aWeight * (l + r);
-    if (normalize) {
-      w = w / 2;
-    }
-    return w;
   }
 
   /**
@@ -2024,7 +1995,6 @@ public:
   }
 
 private:
-  /// TODO: introduce a compute table for the trace?
   template <class Node>
   CachedEdge<Node> trace(const Edge<Node>& a,
                          const std::vector<bool>& eliminate, std::size_t level,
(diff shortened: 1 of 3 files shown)


include/mqt-core/dd/Package.hpp
```diff
@@ -1111,7 +1118,7 @@ template <class Config> class Package {
     auto tmp0 = conjugateTranspose(measZeroDd);
     auto tmp1 = multiply(e, densityFromMatrixEdge(tmp0), false);
     auto tmp2 = multiply(densityFromMatrixEdge(measZeroDd), tmp1, true);
-    auto densityMatrixTrace = trace(tmp2, nrQubits);
+    auto densityMatrixTrace = trace(tmp2, nrQubits) * std::pow(2, nrQubits);
```
burgholzer: Technically, I think we could get rid of this by performing the normalization in the `trace` method based on whether we operate on density matrices or matrices.
For density matrices, it always holds that the trace is 1 (It's a summation of probabilities).
For unitary matrices, the largest trace is the one of the identity, which is $2^n$.

So, if we only divide by two for each level whenever we are dealing with matrices, we can guarantee (and commnicate in the docstring) that the `trace` method will return a value in the interval $[0,1]$.

include/mqt-core/dd/Package.hpp
```diff
@@ -276,10 +279,12 @@ template <class Config> class Package {
       vectorInnerProduct.clear();
       vectorKronecker.clear();
       matrixKronecker.clear();
+      matrixTrace.clear();
       stochasticNoiseOperationCache.clear();
       densityAdd.clear();
       densityDensityMultiplication.clear();
       densityNoise.clear();
+      densityTrace.clear();
```
burgholzer, line 291: I might be mistaken, but I think you do not need to clear these two tables here.
Both tables take only a `Node*` as an input and return a `CachedEdge`, i.e., none of the operands contains a complex number from the respective table. Consequently, it should be fine to leave them as they are if only complex numbers have been collected.

include/mqt-core/dd/Package.hpp
```diff
@@ -260,12 +261,14 @@ template <class Config> class Package {
...
       conjugateMatrixTranspose.clear();
       matrixKronecker.clear();
+      matrixTrace.clear();
       matrixVectorMultiplication.clear();
       matrixMatrixMultiplication.clear();
       stochasticNoiseOperationCache.clear();
       densityAdd.clear();
       densityDensityMultiplication.clear();
       densityNoise.clear();
+      densityTrace.clear();
```
burgholzer: This is technically not really part of this PR, but it is kind of triggered by the change here:
It might make sense to separate the `||` into separate `if`'s here given the fact that no tables operating on density matrices need to be cleared whenever matrix nodes were collected and vice versa.

Would you mind adjusting that (including the top-level comment before the `if`)?

include/mqt-core/dd/Package.hpp
```diff
@@ -1985,16 +2012,17 @@ template <class Config> class Package {
       return CachedEdge<Node>::zero();
     }
 
-    if (std::none_of(eliminate.begin(), eliminate.end(),
-                     [](bool v) { return v; })) {
+    // If a is the identity matrix or there is nothing left to eliminate (e.g. a
+    // is a terminal) then simply return a
```
```suggestion
    // If `a` is the identity matrix or there is nothing left to eliminate,
    // then simply return `a`
```
The terminal itself is considered the identity, but it could very well be that someone requests the partial trace and only wants to trace out a single qubit.

include/mqt-core/dd/Package.hpp
```diff
@@ -1985,16 +2012,17 @@ template <class Config> class Package {
...
+    if (a.isIdentity() || std::none_of(eliminate.begin(), eliminate.end(),
+                                       [](bool v) { return v; })) {
```
burgholzer: Now that I think about it: 
Isn't the check here a bit too pessimistic? In the sense that it always checks the full `eliminated` vector to be none, while it would technically be enough for a stopping condition that none of the qubits from index `0` (very bottom of the DD) up to the current `level` (or `level-1`; needs to be checked) should be eliminated.

Say someone want to trace out only the top-most qubit in a three-qubit system. Then they would pass `eliminate={false, false, true}`, which would never trigger the condition here and would always trigger a recursive computation. Although, after the first level, one could technically stop, because no further levels are eliminated.
It should be possible to realize that by simply exchanging the `eliminate.end()` with the appropriate `eliminate.begin() + x` (where `x` is an index based on the current `level`).
This should drastically improve the partial trace performance.
Maybe you can also add a test for that.

include/mqt-core/dd/Package.hpp
```diff
@@ -2003,6 +2031,8 @@ template <class Config> class Package {
       auto r = add2(trace(a.p->e[0], eliminate, level - 1, elims),
                     trace(a.p->e[3], eliminate, level - 1, elims), v - 1);
 
+      r.w = r.w / 2.0;
```
burgholzer: This line could maybe use a comment stating that this is where we apply the normalization.

Regarding the other comment from above: this is also the place where the normalization would only be applied to matrices and not density matrices.

include/mqt-core/dd/Package.hpp
```diff
@@ -2019,6 +2049,7 @@ template <class Config> class Package {
                                 eliminate.begin(), eliminate.end(), true)) -
                             alreadyEliminated));
     auto r = makeDDNode(adjustedV, edge);
+    // nodes that were not eliminated are not added to the compute table
```
burgholzer: Given that such nodes are not put in the compute table, I am wondering whether one could slightly optimize the compute table lookup at the beginning of this function. In the sense, that no lookup needs to be performed if it is clear from the start that it cannot be successful.

test/dd/test_package.cpp
```diff
@@ -285,21 +285,86 @@ TEST(DDPackageTest, IdentityTrace) {
...
+  auto swapGate2 = dd2->makeTwoQubitGateDD(dd::SWAP_MAT, 0, 1);
   auto ptr = dd->partialTrace(swapGate, {true, false});
-  EXPECT_EQ(ptr.w * ptr.w, 1.);
+  auto fullTrace = dd->trace(ptr, 1);
+  auto fullTraceOriginal = dd2->trace(swapGate2, 2);
+  EXPECT_EQ(dd::RealNumber::val(ptr.w.r), 0.5);
+  // Check that successively tracing out subsystems is the same as computing the
+  // full trace from the beginning
+  EXPECT_EQ(fullTrace.r, fullTraceOriginal.r);
+}
```
burgholzer: Really only a minor comment, but I think you technically do not need to allocate two DD packages here (and two SWAP gates). It should be enough to allocate one package and to create the SWAP DD once.
The rest of the test should "simply work"

test/dd/test_package.cpp
```diff
@@ -285,21 +285,86 @@ TEST(DDPackageTest, IdentityTrace) {
...
+
+TEST(DDPackageTest, PartialTraceKeepInnerQubits) {
+  // Check that the partial trace computation is correct when tracing out the
+  // outer qubits only. This test shows that we should avoid storing
+  // non-eliminated nodes in the compute table, as this would prevent their
+  // proper elimination in subsequent trace calls.
+
+  const std::size_t numQubits = 8;
+  auto dd = std::make_unique<dd::Package<>>(numQubits);
+  auto dd2 = std::make_unique<dd::Package<>>(numQubits);
```
burgholzer: Same as above goes here. You should be able to reduce the number of packages to one.

test/dd/test_package.cpp
```diff
@@ -285,21 +285,86 @@ TEST(DDPackageTest, IdentityTrace) {
...
+    auto hKron = hGate;
+    for (std::size_t i = 0; i < numQubits - 1; ++i) {
+      hKron = dd->kronecker(hKron, hGate, 1);
+    }
+    dd->trace(hKron, numQubits);
+    auto stats = computeTable.getStats();
+    auto lookups = stats.lookups;
+    auto hits = stats.hits;
+    ASSERT_EQ(lookups, 2 * numQubits - 1);
+    ASSERT_EQ(hits, numQubits - 1);
```
burgholzer:
```suggestion
    const auto& stats = computeTable.getStats();
    ASSERT_EQ(stats.lookups, 2 * numQubits - 1);
    ASSERT_EQ(stats.hits, numQubits - 1);
```
Just a tiny bit more compact.

burgholzer (changes requested):
Many thanks for your work on this and for the clear code comments. These are highly appreciated!
I looked through all the changes here and you'll find my comments inline. Some are just minor suggestions that you can quickly incorporate. Some might take a little bit of work, but I think it shouldn't be too bad.

include/mqt-core/dd/Package.hpp
TeWas: Oh yes, indeed. It seems like this is also the case for some other tables, such as vectorKronecker and matrixKronecker, where only the final returned edge weight is stored in cUniqueTable. I will check the others as well!

test/dd/test_package.cpp
TeWas: The issue here is that `partialTrace `and `trace `cannot share the same compute table. Depending on the execution order, one computation will produce incorrect results. For example, if `fullTraceOriginal `is computed first, the partial trace computation will be incorrect because all qubits were eliminated and thus the nodes were stored in the compute table. But then, during the partial trace computation, for the top node, the result from the compute table will be used, even though some lower-level qubits should have been eliminated.

include/mqt-core/dd/Package.hpp
burgholzer: That could very well be the case. Some of these have only recently been refactored to just take a `Node` instead of an `Edge`.
Thanks for looking into it!

test/dd/test_package.cpp
burgholzer: I think you can work around that, by only inserting nodes into the compute table if all nodes below that node are eliminated as well. That disambiguates the two use cases. And while it limits the caching for the partial trace a little, it should allow to use both functions within the same package instance.

In the end, it would be very undesirable if a user would get unexpected results because they called the trace method followed by the partial trace. We should make sure that both of them can work in parallel, even if that means that one of them will perform slightly worse.

include/mqt-core/dd/Package.hpp
```diff
@@ -1985,24 +2026,51 @@ template <class Config> class Package {
...
     if (eliminate[v]) {
+      // Lookup nodes marked for elimination in the compute table if all
+      // lower level qubits are eliminated as well: if the trace has already
+      // been computed, return the result
+      auto& computeTable = getTraceComputeTable<Node>();
+      if (std::all_of(
+              eliminate.begin(),
+              eliminate.begin() +
+                  static_cast<std::vector<bool>::difference_type>(level),
+              [](bool e) { return e; })) {
```
burgholzer: Just a quick side comment (besides the currently failing unittest): I think you can extract this computation to avoid repeating it at the end of this block.
```c++
const auto computeFullTrace = std::all_of(
              eliminate.begin(),
              eliminate.begin() +
                  static_cast<std::vector<bool>::difference_type>(level),
              [](bool e) { return e; });
```
or something similar.

TeWas: Done, thanks!

test/dd/test_package.cpp
TeWas: Yes, so the cases are disambiguated now by checking before the lookup and insertion into the table that the lower-level qubits are eliminated as well.

include/mqt-core/dd/Package.hpp
burgholzer: As discussed offline, I transferred this to a separate issue. See #633

burgholzer (approved):
Many thanks for all your work on this. Especially, for writing thorough tests! ✅ 

I just pushed some finishing touches, which are mostly stylistic/minor. After CI is all green, this is good to go 🚀

PR #629, opened 2024-06-07
🚸 Support for adding tests to NALAC

ystade:
## Description

This PR contains modifications that were necessary for the tests of the Neutral Atom Logical Array Compiler, see cda-tum/mqt-qmap#470.

## Checklist:

### DIFF
--- a/include/mqt-core/operations/CompoundOperation.hpp
+++ b/include/mqt-core/operations/CompoundOperation.hpp
@@ -63,7 +63,7 @@ public:
   std::vector<std::unique_ptr<Operation>>& getOps() { return ops; }
 
   [[nodiscard]] std::set<Qubit>
-  getUsedQubits(const std::function<Qubit(Qubit)>& perm = [](Qubit q) {
+  getUsedQubits(const std::function<Qubit(Qubit)>& perm = [](const Qubit q) {
     return q;
   }) const override;
 
--- a/include/mqt-core/operations/CompoundOperation.hpp
+++ b/include/mqt-core/operations/CompoundOperation.hpp
@@ -62,8 +62,7 @@ public:
 
   std::vector<std::unique_ptr<Operation>>& getOps() { return ops; }
 
-  [[nodiscard]] auto
-  getUsedQubitsPermuted(const std::function<Qubit(const Qubit)>& perm) const
+  [[nodiscard]] auto getUsedQubitsPermuted(const Permutation& perm) const
       -> std::set<Qubit> override;
 
   [[nodiscard]] auto commutesAtQubit(const Operation& other,
--- a/include/mqt-core/operations/CompoundOperation.hpp
+++ b/include/mqt-core/operations/CompoundOperation.hpp
@@ -21,8 +21,8 @@ private:
 public:
   explicit CompoundOperation();
 
-  explicit CompoundOperation(
-      std::vector<std::unique_ptr<Operation>>&& operations);
+  explicit
+  CompoundOperation(std::vector<std::unique_ptr<Operation>>&& operations);
 
   CompoundOperation(const CompoundOperation& co);
 
@@ -62,7 +62,10 @@ public:
 
   std::vector<std::unique_ptr<Operation>>& getOps() { return ops; }
 
-  [[nodiscard]] std::set<Qubit> getUsedQubits() const override;
+  [[nodiscard]] std::set<Qubit>
+  getUsedQubits(const std::function<Qubit(Qubit)>& perm = [](Qubit q) {
+    return q;
+  }) const override;
 
   [[nodiscard]] auto commutesAtQubit(const Operation& other,
                                      const Qubit& qubit) const -> bool override;
--- a/include/mqt-core/operations/CompoundOperation.hpp
+++ b/include/mqt-core/operations/CompoundOperation.hpp
@@ -21,8 +21,8 @@ private:
 public:
   explicit CompoundOperation();
 
-  explicit
-  CompoundOperation(std::vector<std::unique_ptr<Operation>>&& operations);
+  explicit CompoundOperation(
+      std::vector<std::unique_ptr<Operation>>&& operations);
 
   CompoundOperation(const CompoundOperation& co);
 
--- a/include/mqt-core/operations/CompoundOperation.hpp
+++ b/include/mqt-core/operations/CompoundOperation.hpp
@@ -62,10 +62,9 @@ public:
 
   std::vector<std::unique_ptr<Operation>>& getOps() { return ops; }
 
-  [[nodiscard]] std::set<Qubit>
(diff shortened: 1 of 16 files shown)


src/na/NAComputation.cpp
```diff
@@ -9,13 +9,16 @@ auto NAComputation::toString() const -> std::string {
   std::stringstream ss;
   ss << "init at ";
   for (const auto& p : initialPositions) {
-    ss << "(" << p->x << ", " << p->y << ")"
-       << ", ";
+    ss << *p << ", ";
+  }
+  if (ss.tellp() == 8) {
+    ss.seekp(-1, std::ios_base::end);
```
burgholzer, line 15: Might be worth covering by a test given that you explicitly introduced some special casing here.

src/na/operations/NALocalOperation.cpp
```diff
@@ -17,10 +17,14 @@ auto NALocalOperation::toString() const -> std::string {
     ss << ")";
   }
   ss << " at ";
-  for (const auto& p : positions) {
-    ss << *p << ", ";
+  if (positions.empty()) {
+    ss.seekp(-1, std::ios_base::end);
```
burgholzer, line 21: Might be worth covering by a test given that you explicitly introduced some special casing here.

src/operations/Operation.cpp
```diff
@@ -121,6 +121,58 @@ bool Operation::equals(const Operation& op, const Permutation& perm1,
...
+    } else {
+      for (const auto& control : controls) {
+        if (control.type == Control::Type::Pos) {
+          usedQubits1.emplace(perm1.at(control.qubit));
+        } else {
+          negControls1.emplace(perm1.at(control.qubit));
+        }
+      }
+      for (const auto& qubit : targets) {
+        usedQubits1.emplace(perm1.at(qubit));
```
burgholzer: Both of these else branches are uncovered by tests. Would you mind adding some unittests for these cases?

src/operations/Operation.cpp
```diff
@@ -121,6 +121,58 @@ bool Operation::equals(const Operation& op, const Permutation& perm1,
...
+        } else {
+          negControls2.emplace(perm2.at(control.qubit));
+        }
+      }
+      for (const auto& qubit : op.getTargets()) {
+        usedQubits2.emplace(perm2.at(qubit));
+      }
+    }
+
+    return usedQubits1 == usedQubits2 && negControls1 == negControls2;
```
burgholzer: Just an idea at the top of my head:
Quite some code in this function is simply duplicated more-or-less four times. two variants for the two operations and two variants for permutation or no permutation.
Guess it would be possible to simplify this a little bit by introducing further protected member functions in the `Operation` class that cover parts of the code here.
Probably too much to ask for this PR, so feel free to defer to a later point in time. Just wanted to bring it up.

I was also wondering: there is already a `getUsedQubits` method. Could it be worth reusing that here? Maybe in addition to a new `getCointrolType() -> std::vector<qc::Control::Type>` or `getControlPolarity() -> std::vector<bool>` member function.

burgholzer (approved):
LGTM ✅ 
Just a couple of comments on covering changes by tests and one slightly bigger comment on whether we can maybe simplify some of the code.
(Started the review before your last commit, so could be that some comments are obsolete by now)

src/na/NAComputation.cpp
ystade: Did that ;)

src/na/operations/NALocalOperation.cpp
ystade: Did that as well ;)

src/operations/Operation.cpp
ystade: Added a test, will check in the coverage report whether everything is covered.

src/operations/Operation.cpp
ystade: Does https://github.com/cda-tum/mqt-core/blob/f0dfa7115d146ae703d38da653ef41b5a18dfd48/src/operations/Operation.cpp#L124-L171 look better now? It is always a bit difficult to get it right involving the permutation.

burgholzer: That does look better already! 👍🏻
Could someone use a lambda function that is either the identity or applies the permutation and use such a function to construct the respective sets?
One could, for example, augment the `getUsedQubits` function with an optional std::function argument that defaults to the identity.

I believe that could help improve the clarity throughout the whole function here even more.

ystade: Realized that, I think now it is easier when you look a the "files changed" tab.

include/mqt-core/operations/CompoundOperation.hpp
```diff
@@ -62,7 +62,10 @@ class CompoundOperation final : public Operation {
 
   std::vector<std::unique_ptr<Operation>>& getOps() { return ops; }
 
-  [[nodiscard]] std::set<Qubit> getUsedQubits() const override;
+  [[nodiscard]] std::set<Qubit>
+  getUsedQubits(const std::function<Qubit(Qubit)>& perm = [](Qubit q) {
```
burgholzer:
```suggestion
  getUsedQubits(const std::function<Qubit(const Qubit)>& perm = [](const Qubit q) {
```
Just being a bit pedantic here, but I guess the extra `const` doesn't hurt here.
Holds for the other overloads as well.

src/operations/Operation.cpp
```diff
@@ -121,6 +121,46 @@ bool Operation::equals(const Operation& op, const Permutation& perm1,
...
+      if (control.type == Control::Type::Neg) {
+        if (perm2.empty()) {
+          negControls2.emplace(control.qubit);
+        } else {
+          negControls2.emplace(perm2.at(control.qubit));
+        }
+      }
+    }
+
+    return negControls1 == negControls2;
```
burgholzer: Sorry for being so pedantic here, but this can be even simpler.

The second block of code can also make use of such a lambda function.
I'd probably define a `perm` function based on whether the permutation is empty. Then you can reuse that function across the entire function here.

Additionally, these simplifications need not be limited to your newly introduced if block. They can be applied across the whole function.

Something like (untested)
```c++
std::function<Qubit(const Qubit)> permFunc1 = perm1.empty()? [](const Qubit q) { return q; } : [&perm1](const Qubit q) { return perm1.at(q); };
```

burgholzer (changes requested):
Like the overall suggestion and direction. Just some further comments.

ystade:
And something broke due the change in `getUsedQubits` :/

src/operations/Operation.cpp
ystade: I'll change that in the next commit but then I honestly kind of have enough for only wanting to fix the equivalence check of operations.

burgholzer: Fair enough. Sorry for the long detour 😅

burgholzer (approved):
LGTM now ✅

PR #639, opened 2024-07-01
🐛 Support `u` gate when parsing OpenQASM 2

ystade:
## Description

🚧 Should probably have opened an issue but I wanted to directly add the failing test.

### What I did

I downloaded a circuit for AE with 2 qubits compiled hardware-agnostically with qiskit from MQT Bench and inserted this circuit as input to `qc::QuantumComputation::fromQASM()`, which should not throw an error but it does, see the respective test 
https://github.com/cda-tum/mqt-core/blob/f91a0526617dbb68e2ce3589a8994f220bc44347/test/test_operation.cpp#L263-L287

## Checklist:

### DIFF
--- a/include/mqt-core/parsers/qasm3_parser/StdGates.hpp
+++ b/include/mqt-core/parsers/qasm3_parser/StdGates.hpp
@@ -89,6 +89,7 @@ const std::map<std::string, std::shared_ptr<Gate>> STANDARD_GATES = {
     {"id", std::make_shared<StandardGate>(StandardGate({0, 1, 0, qc::I}))},
     {"u2", std::make_shared<StandardGate>(StandardGate({0, 1, 2, qc::U2}))},
     {"u3", std::make_shared<StandardGate>(StandardGate({0, 1, 3, qc::U}))},
+    {"u", std::make_shared<StandardGate>(StandardGate({0, 1, 3, qc::U}))},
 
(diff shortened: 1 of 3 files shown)


ystade:
Also, I am not sure whether the place for the new test is appropriate, feel free to suggest something different.

burgholzer:
This was a pretty quick fix. Although, technically, this is not even a bug on our side, but on the Qiskit side. A well known one in fact.
The Qiskit OpenQASM 2.0 exporter is known to hallucinate the existence of the `u` gate in the standard library, while, in fact, the gate is not part of the standardized `qelib1.inc`, but was only part of Qiskit's extension of said file.

So, technically, our parser is right to reject the circuit since it is not valid OpenQASM 2.0.
However, many people have come to rely on these circuits to be readable and Qiskit itself has implemented a respective workaround for reading the files.

Given that it's a one line change, I think it should be fine to also adopt the workaround here.

burgholzer:
> Also, I am not sure whether the place for the new test is appropriate, feel free to suggest something different.

I moved the test to the OpenQASM parser test suite and extended it slightly.

Also adjusted the labels of this PR. The "bug" label is typically reserved for issues (=bug reports), while the "fix" label is intended for PRs (=bug fixes). And given that this is only a bug fix and not a breaking change of any sort nor a new feature, it qualifies as a `patch` release contribution and does not really warrant a new `minor` version. Hope that makes sense 😌

ystade:
@burgholzer Thanks for taking care of this!

PR #651, opened 2024-07-24
:sparkles: Add `customGate` flag and optimizer functionality to flatten only custom gates

DRovara:
## Description

This pull request adds the `customGate` flag to the `CompoundOperation` class. This flag is set for `CompoundOperations` that were constructed from custom gate calls.

The pull request also provides new functionality to the `flattenOperations` optimizer. Now, a second parameter can be used to indicate that only custom gates should be flattened. It makes sense to put this functionality here as outside of the `CircuitOptimizer` class, certain protected members cannot be accessed anymore.

## Checklist:

### DIFF
--- a/src/parsers/QASM3Parser.cpp
+++ b/src/parsers/QASM3Parser.cpp
@@ -645,7 +645,7 @@ public:
         index++;
       }
 
-      auto op = std::make_unique<qc::CompoundOperation>();
+      auto op = std::make_unique<qc::CompoundOperation>(true);
       for (const auto& nestedGate : compoundGate->body) {
         if (auto barrierStatement =
                 std::dynamic_pointer_cast<BarrierStatement>(nestedGate);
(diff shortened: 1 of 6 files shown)


src/parsers/QASM3Parser.cpp
DRovara, line 1: I'm not 100% sure this is the only place where the flag has to be set. Is Line 799 also invoked for custom gate declarations in some cases, or is it only used for other types of block operations?

burgholzer: I think that should actually be it.

Line 799 is just for the `then` and `else` blocks of `if` statements or the bodies of `for` loops.
Given that your use case right now is limited to OpenQASM 2 anyway, these shouldn't be relevant.

burgholzer (approved):
Many thanks for this nice addition.
I just pushed two commits with a few cleanups.
Once CI passes, this is good to go in 🚀

PR #666, opened 2024-08-09
⚗️ upload to Test PyPI for pushes on `main`

burgholzer:
## Description

This PR updates the CD workflow so that anytime it runs on a push to `main`, it uploads the resulting package to Test PyPI. This allows to battle test the packages before official releases.
This will be especially important in the context of #662.

## Checklist:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,13 +11,19 @@ permissions:
   id-token: write
 
 jobs:
+  # Builds the sdist and wheels on all supported platforms and uploads the resulting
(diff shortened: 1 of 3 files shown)


burgholzer:
Unfortunately, it seems that this is not really feasible.
Test PyPI won't accept packages with local version identifiers and setuptools-scm does not easily allow to disable local versions, e.g., based on the presence of an environment variable.
Maybe sometime in the future.

burgholzer:
Might have found a way to make this possible after all. Investigating further.

PR #751, opened 2024-11-13
🚸 Add Basic Python Bindings for NAComputation

ystade:
## Description

I need the export functionality of the `NAComputation` to a string in the Python bindings of MQT QMap. This PR creates a binding for that on the Python side.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ ninja.version = ">=1.10"
 build-dir = "build/{wheel_tag}/{build_type}"
 
 # Only build the Python bindings target
-build.targets = ["ir"]
+build.targets = ["ir", "na"]
 
 # Only install the Python package component
 install.components = ["mqt-core_Python"]
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -1,2 +1,4 @@
 # add the IR bindings package
 add_subdirectory(ir)
+# add the NA bindings package
+add_subdirectory(na)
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -85,7 +85,7 @@ ninja.version = ">=1.10"
 build-dir = "build/{wheel_tag}/{build_type}"
 
 # Only build the Python bindings target
-build.targets = ["ir"]
+build.targets = ["ir", "na"]
```
burgholzer, line 88: Would it make sense to add an interface target in the `src/python/CMakeLists.txt` file after including all the subdirectories that links all the bindings targets together.
That would let us use a single string here.
Maybe its also fine to leave this as is, given how this will change with #662 again anyway.

burgholzer (changes requested):
This is missing a couple of small things:
- Python stubs similar to https://github.com/cda-tum/mqt-core/tree/main/src/mqt/core/ir (including docstrings)
- A small test on the Python side in the test suite to make sure this works as expected. (we currently do not track coverage of the bindings code, which is unfortunate, but I haven't found time to figure out how to actually do that)

Side note: I'd still be in favour of trying to unify the kinds of computations across the MQT in the future so that this additional bindings module is not even necessary.

Bigger comment: I am almost certain this will not work right away in QMAP and will rather depend on #662 and cda-tum/mqt-qmap#418, which still require some work to get right.
At the moment, you are probably better off to simply expose the `NAComputation` class as part of the QMAP bindings.

burgholzer:
Closing this for now. Let's revisit once that makes more sense.
In the meantime, QMAP seems to be fine without this.

PR #752, opened 2024-11-14
Refactor handling of gate matrices and inverses

### LINKED ISSUE #484: :recycle: Refactor handling of gate matrices and inverses

Joshy-R:
## Description

These commits change how matrices are created from an OpType that reduces the number of switch statements necessary.
The downside of this approach is that some functions can be called with arguments that do not affect the returned value. As the user never calls these functions, I see this as acceptable.

I also removed the CX_MAT and CZ_MAT, as they are only used in tests and do not occur in any other repository.

Fixes #484 

## Checklist:

I have not squashed the commits to make them easier to review, but I can squash them if desired.


include/mqt-core/ir/operations/OpType.hpp
```diff
@@ -8,165 +8,37 @@
 #include <unordered_map>
 
 namespace qc {
+
+enum OpTypeFlags {
+  OP_TYPE_INV = 1,
+  OP_TYPE_DIAG = 2,
+};
```
burgholzer:
```c++
enum class OpTypeFlag: uint8_t {
  NONE = 0b00,
  INV  = 0b01,
  DIAG = 0b10,
};
```
How about the above suggestion? first of all creates a dedicated `NONE` entry that replaces the use of `0`s throughout the code. Uses binary integers to more clearly illustrate the flag character. And makes this an enum class.

However, given that the arithmetic with different enums does seem to upset clang-tidy, it might be worth to simply define these as
```c++
constexpr unit8_t OP_TYPE_FLAG_NONE = 0b00;
constexpr unit8_t OP_TYPE_FLAG_INV  = 0b01;
constexpr unit8_t OP_TYPE_FLAG_DIAG = 0b10;
```

include/mqt-core/ir/operations/OpType.hpp
```diff
@@ -8,165 +8,37 @@
...
+
+constexpr static unsigned N_OP_TYPE_FLAGS_BITS = 2;
```
burgholzer:
```suggestion
constexpr static unsigned NUM_OP_TYPE_FLAG_BITS = 2;
```
Slight naming preference for consistency.

include/mqt-core/ir/operations/OpType.inc
```diff
@@ -0,0 +1,69 @@
+#define OP_TYPE_INV 0b1
+#define OP_TYPE_DIAG 0b10
```
burgholzer: Do these really need to be defined here? Aren't these defined as enums anyway?

include/mqt-core/dd/Operations.hpp
```diff
@@ -22,83 +22,87 @@
...
-    gm = inverse ? V_MAT : VDG_MAT;
-    break;
+    // clang-format off
+  case qc::iSWAP: case qc::iSWAPdg:
+  case qc::Peres: case qc::Peresdg:
+  case qc::S:     case qc::Sdg:
+  case qc::T:     case qc::Tdg:
+  case qc::V:     case qc::Vdg:
+  case qc::SX:    case qc::SXdg:
+  case qc::I: case qc::H: case qc::X: case qc::Y: case qc::Z:
```
burgholzer: This is not yet completely correct and clang-tidy rightfully complains here.
I, H, X, Y, Z are all self inverse. So these need not be XORed.

I am also not a big fan of fall-through switch statements, so I'd rather have an explicit return in each of the blocks.

include/mqt-core/dd/Operations.hpp
```diff
@@ -22,83 +22,87 @@
...
-// single-target Operations
-template <class Config>
-qc::MatrixDD
-getStandardOperationDD(const qc::StandardOperation* op, Package<Config>& dd,
-                       const qc::Controls& controls, const qc::Qubit target,
-                       const bool inverse) {
-  GateMatrix gm;
 
-  const auto type = op->getType();
+inline std::tuple<qc::OpType, fp, fp, fp>
```
burgholzer: I am not yet fully happy with this fixes kind of return type. In principle, gates could have an arbitrary number of parameters. The fact that currently supported gates onlly use up to three parameters is purely artificial and there are examples of gates with 4+ parameters that we might want to support in the future. Do you see any kind of way of supporting that use case here? Either via variadic lists or vectors of parameters or something completely different..

Maybe it would even be better to take this one step further and represent an operation by its number of qubits, its type, and its number of parameters. Then one could probably transform the code here to a very uniform style and consolidate it quite a bit.

Adding even one step further, this might be worth pulling from the DD submodule to the `ir/operations` module as a whole given the general usefulness of getting the (inverse) matrix for a given StandardOperation.

burgholzer (approved):
Hey :wave:

Many thanks for starting to work on this! Great to see someone taking this on!
I really like the solution with the flags for indicating the inverse type of a gate and indicating whether a gate is diagonal! Very elegant.

I added some inline comments that are mostlly minor except one bigger one that might trigger some further refactorings. However, I believe these might be worth it.

One thing on top of all of that: It would be great if you could address the clang-tidy complaints.
Typically these are posted as PR comments, but that feature does not work for PRs from forks. You can still see the warnings either in the check summary here: https://github.com/cda-tum/mqt-core/pull/752/checks or when looking through the "Files Changed" tab on GitHub.

Thanks again and let me know if anything is unclear!

burgholzer (changes requested):
Oops, clicked the wrong button 🙃

include/mqt-core/ir/operations/OpType.hpp
rotmanjanez: `enum class`es are even more strict with conversions and manipulations, it would keep a normal `enum`s and use casts to please clang tidy. Adding a flag `NONE` is indeed nice!

include/mqt-core/ir/operations/OpType.inc
rotmanjanez: Yes, this was an oversight

include/mqt-core/ir/operations/OpType.inc
```diff
@@ -0,0 +1,75 @@
...
+HANDLE_OP_TYPE(32, Teleportation, OpTypeNone, "teleportation")
+
+// Classically-controlled Operation
+HANDLE_OP_TYPE(33, ClassicControlled, OpTypeNone, "classic_controlled")
+
+// Noise operations
+HANDLE_OP_TYPE(34, ATrue, 0, nullptr)
+HANDLE_OP_TYPE(35, AFalse, 0, nullptr)
+HANDLE_OP_TYPE(36, MultiATrue, 0, nullptr)
+HANDLE_OP_TYPE(37, MultiAFalse, 0, nullptr)
```
burgholzer:
```suggestion
HANDLE_OP_TYPE(34, ATrue, 0, "a_true")
HANDLE_OP_TYPE(35, AFalse, 0, "a_false")
HANDLE_OP_TYPE(36, MultiATrue, 0, "multi_a_true")
HANDLE_OP_TYPE(37, MultiAFalse, 0, "multi_a_false")
```

might just give these a proper string representation. then you can get rid of the `nullptr` handling in the string representation function.

include/mqt-core/ir/operations/OpType.hpp
```diff
@@ -9,173 +9,46 @@
...
+#undef HANDLE_OP_TYPE
+#undef LAST_OP_TYPE
+  };
+
+  if (OP_NAMES.find(opType) == OP_NAMES.end() ||
+      OP_NAMES.at(opType) == nullptr) {
     throw std::invalid_argument("Invalid OpType!");
-    // GCOV_EXCL_STOP
   }
+  return OP_NAMES.at(opType);
```
burgholzer: This can be implemented with fewer lookups
```c++
if (const auto it = OP_NAMES.find(opType); it != OP_NAMES.end()) {
  return it->second;
}
throw std::invalid_argument("Invalid OpType!");
```

include/mqt-core/dd/GateMatrixDefinitions.hpp
```diff
@@ -10,31 +10,27 @@
...
 
+extern const std::unordered_map<qc::OpType,
+                                GateMatrix (*const)(const std::vector<fp>&)>
+    MATS_GENERATORS;
+
+extern const std::unordered_map<qc::OpType, TwoQubitGateMatrix (*const)(
+                                                const std::vector<fp>&)>
+    TWO_GATE_MATS_GENERATORS;
+
 inline GateMatrix uMat(const fp lambda, const fp phi, const fp theta) {
```
burgholzer: I would argue that the definitions of these methods for gate matrices could be completely moved to the `.cpp` file.
They shouldn't actually be needed outside of the gate matrix generators. Might as well encapsulate them there.
Any uses of the existing methods can just use the factory methods.

include/mqt-core/dd/GateMatrixDefinitions.hpp
```diff
@@ -10,31 +10,27 @@
...
 constexpr GateMatrix MEAS_ZERO_MAT{1, 0, 0, 0};
 constexpr GateMatrix MEAS_ONE_MAT{0, 0, 0, 1};
 
+extern const std::unordered_map<qc::OpType,
+                                GateMatrix (*const)(const std::vector<fp>&)>
+    MATS_GENERATORS;
+
+extern const std::unordered_map<qc::OpType, TwoQubitGateMatrix (*const)(
+                                                const std::vector<fp>&)>
+    TWO_GATE_MATS_GENERATORS;
```
burgholzer:
```suggestion
    SINGLE_QUBIT_MATRIX_GENERATORS;

extern const std::unordered_map<qc::OpType, TwoQubitGateMatrix (*const)(
                                                const std::vector<fp>&)>
    TWO_QUBIT_GATE_MATRIX_GENERATORS;
```
Let's give these a slightly better name.

include/mqt-core/dd/GateMatrixDefinitions.hpp
```diff
@@ -174,4 +140,15 @@ inline TwoQubitGateMatrix xxPlusYYMat(const fp theta, const fp beta = 0.) {
...
+
+inline GateMatrix opToSingleGateMatrix(qc::OpType t,
+                                       const std::vector<fp>& params = {}) {
+  return MATS_GENERATORS.at(t)(params);
+}
+
+inline TwoQubitGateMatrix
+opToTwoQubitGateMatrix(qc::OpType t, const std::vector<fp>& params = {}) {
+  return TWO_GATE_MATS_GENERATORS.at(t)(params);
+}
```
burgholzer: Also just slightly more precise naming and a small `const` addition.
```suggestion
inline GateMatrix opToSingleQubitGateMatrix(const qc::OpType t,
                                       const std::vector<fp>& params = {}) {
  return MATS_GENERATORS.at(t)(params);
}

inline TwoQubitGateMatrix
opToTwoQubitGateMatrix(const qc::OpType t, const std::vector<fp>& params = {}) {
  return TWO_GATE_MATS_GENERATORS.at(t)(params);
}
```

burgholzer: This is more of a general comment/question:
Given the presence of these functions, are the factory functions actually needed/helpful? 
Don't they simply add an additional level of indirection (if not two, based on the use of lambdas)?
Wouldn't it be enough to simply implement these two methods and have them use an internal switch statement on the OpType?
From experience, the compiler generates pretty efficient code from switch statements that is actually very close to a LUT.

include/mqt-core/dd/Operations.hpp
```diff
@@ -34,83 +34,99 @@
...
+      return {type, {-parameter[0U]}};
+    }
+    return {type, {parameter[0U]}};
+  // inverse by negating the first of two parameter
+  case qc::XXminusYY:
+  case qc::XXplusYY:
+    if (inverse) {
+      return {type, {-parameter[0U], parameter[1U]}};
+    }
+    return {type, {parameter[0U], parameter[1U]}};
```
burgholzer: This comment isn't fully true.

include/mqt-core/dd/Operations.hpp
```diff
@@ -34,83 +34,99 @@
...
+
+  case qc::DCX:
+    if (inverse) {
+      if (target0 == nullptr || target1 == nullptr) {
+        throw qc::QFRException("Invalid target qubits for DCX");
+      }
+
+      // DCX is not self-inverse, but the inverse is just swapping the targets
+      std::swap(*target0, *target1);
+    }
```
burgholzer: Feels a bit strange that this is the only case in all of this function that actually needs the `target0` and `target1` parameters.
That triggered a bit of a thought process.

Maybe a slight optimization for this overall method would be to pass the targets as a `std::vector<Qubit>&` as well. That way, 
- it can contain various entries which makes it more future proof
- only requires 8 Byte for the parameter (instead of 16 Byte for the two pointers)
- directly allows to pipe-through the return value of the `getTargets()` method in the `Operation` class.
- should simplify some of the calls across the library slightly.

This would also allow one to not distinguish between the single-qubit and the two-qubit case all that much and have a single function that takes a vector of targets and handles the distinction internally.
But maybe this is also a stupid idea. It be nice if you could play around a little with this.

include/mqt-core/dd/Operations.hpp
```diff
@@ -119,64 +135,10 @@ qc::MatrixDD
...
-  case qc::XXplusYY:
-    gm = inverse ? xxPlusYYMat(-parameter[0U], parameter[1U])
-                 : xxPlusYYMat(parameter[0U], parameter[1U]);
-    break;
-  default:
-    std::ostringstream oss{};
-    oss << "DD for gate " << op->getName() << " not available!";
-    throw qc::QFRException(oss.str());
-  }
+  auto [type, params] = getOperationParameters(op, &target0, &target1, inverse);
```
burgholzer:
```suggestion
  const auto [type, params] = getOperationParameters(op, &target0, &target1, inverse);
```

include/mqt-core/dd/Operations.hpp
```diff
@@ -34,83 +34,99 @@
...
-  return dd.makeGateDD(gm, controls, target);
+}
+
+// single-target Operations
+template <class Config>
+qc::MatrixDD getStandardOperationDD(const qc::StandardOperation* op,
+                                    Package<Config>& dd,
+                                    const qc::Controls& controls,
+                                    qc::Qubit target, const bool inverse) {
+  auto [type, params] = getOperationParameters(op, &target, nullptr, inverse);
```
burgholzer:
```suggestion
  const auto [type, params] = getOperationParameters(op, &target, nullptr, inverse);
```

src/dd/GateMatrixDefinitions.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <complex>
+#include <unordered_map>
+#include <vector>
+
+namespace dd {
+
+const std::unordered_map<qc::OpType,
+                         GateMatrix (*const)(const std::vector<fp>&)>
+    MATS_GENERATORS{{
+        // clang-format off
```
burgholzer: Given how you purposely disable formatting here, it might be worth to actually format the matrices as matrices instead of linear lists 🤷🏼

test/dd/test_package.cpp
```diff
@@ -50,8 +49,8 @@ TEST(DDPackageTest, TrivialTest) {
   auto dd = std::make_unique<dd::Package<>>(2);
   EXPECT_EQ(dd->qubits(), 2);
 
-  auto xGate = dd->makeGateDD(dd::X_MAT, 0);
-  auto hGate = dd->makeGateDD(dd::H_MAT, 0);
+  auto xGate = dd->makeGateDD(dd::opToSingleGateMatrix(qc::X), 0);
```
burgholzer: Although these are just tests, this might be a nice opportunity to modernise the syntax in some of these.
Instead of directly calling `makeGateDD` here one could construct a `StandardOperation` and call the `getDD` method. Feels slightly less clunky and manual.

burgholzer (changes requested):
Many thanks for the iteration here. This is looking great already and I like where this is heading.
I went through everything in detail again and added my feedback inline. Hope that makes sense. Please let me know if it doesn't!

include/mqt-core/dd/GateMatrixDefinitions.hpp
rotmanjanez: I'm not sure which one is more efficient, because the switch (for the compiler) is non exhaustive, which requires it to generate some checks. But i think the performance differences are minimal, if there are any. But i can rewrite it to use switch statements.

include/mqt-core/dd/Operations.hpp
rotmanjanez: Why not? These are the gates with to parameters that can be inverted by inverting the first parameter? Am i missing something?

burgholzer: That's my bad. I must have read this too quickly. The comments seems just fine 🤨

include/mqt-core/dd/GateMatrixDefinitions.hpp
burgholzer: I believe rewriting to switch statements could reduce the overall complexity of the code at (at most) a negligible runtime overhead. So I'd be in favor of doing that if there are no strong objections.

Joshy-R:
I tried to implement your thoughts as good as possible. One change necessary to transition the test to retrieve the DD via the `getDD` and the `StandardOperation` class was to pass the standard operations as a const reference instead of a pointer. Also, this is closer to the semantics, as the code assumes the pointer is not null, and the code does not use the object's location information. This way, instead of writing `dd::getDD(&(qc::StandardOperation){ 0, qc::RZ, {dd::PI_4}}, dd)`, we can write `dd::getDD(qc::StandardOperation(0, qc::RZ, {dd::PI_4}), dd)`, which is a bit easier to read.

I hope this is a further improvement to the previous draft.

Joshy-R:
I am also trying to understand why one CI build fails, but I don't see how my changes produced this error. Would you happen to have an idea, or would you be able to help me with that?

burgholzer:
> > I am also trying to understand why one CI build fails, but I don't see how my changes produced this error. Would you happen to have an idea, or would you be able to help me with that?
> 
> That failure is unrelated to your changes and either due to an update in pandas or an update in `uv`. I very much suspect it's the latter as I remember reading something about the default resolution strategy changing. I will have a look later.

Turns out it's something else entirely. It's due to `ubuntu-latest` switching from `ubuntu-22.04` to `ubuntu-24.04`. That change comes with a new `gcc` version and that version doesn't seem to be particularly happy with building `pandas` from source. I'll quickly fix this up. `pandas` ships 3.13 wheels as of `v2.2.3`.

Joshy-R:
Okay, thanks. Then, I will rebase onto the new main so that the test passes again. Do you have any further change requests regarding my pull request?

include/mqt-core/dd/Operations.hpp
```diff
@@ -34,149 +35,103 @@
...
+    }
+  } else {
+    // some operations need parameter changes even in the non-inverse case
+    if (type == qc::U) {
+      assert(params.size() == 3U);
+      // shuffle [a, b, c] to [c, b, a]
+      std::reverse(params.begin(), params.end());
+    } else if (type == qc::U2) {
+      params[0U] = params[1U];
+    }
```
burgholzer: Wouldn't it be simpler to eliminate these special cases by refactoring the respective `u2` and `u` matrix methods so that they take their parameters in the opposite order? Sounds like less special handling to me.

Furthermore, the `U2` condition isn't quite right as it does not swap the parameters.

include/mqt-core/dd/Operations.hpp
```diff
@@ -34,149 +35,103 @@
...
-    std::swap(target0, target1);
+qc::MatrixDD getStandardOperationDD(const qc::StandardOperation& op,
+                                    Package<Config>& dd,
+                                    const qc::Controls& controls,
+                                    const std::vector<qc::Qubit>& targets,
+                                    const bool inverse) {
+
+  auto type = op.getType();
+  std::vector<fp> params = op.getParameter();
+  std::vector<qc::Qubit> targetQubits = targets;
```
burgholzer: Sorry for being so picky here, but I believe this could be improved further by a bit.
At the moment, this incurs an extra copy for these variables in the `inverse=false` case that should not really be necessary.
Given that these copies are small, I would be fine with leaving it, but I wanted to bring it up in case you see an elegant solution that avoids that.

One solution that I could think of is to create a separate `GetStandardOperationInverseDD` method that essentially forwards to the `getStandardOperationDD` method after performing the necessary adaptations.
That way, copies should only be required where they are really necessary.
Instead of `break` statements in the respective `case` statements of the new inverse method, it could directly return the respective value from `getStandardOperationDD`, i.e., the self-inverse gates would return without any modifications, and the others would just construct a new StandardOperation with the right parameters and pass it over.

include/mqt-core/dd/Operations.hpp
```diff
@@ -34,149 +35,103 @@
...
+    case qc::U:
+      // shuffle [a, b, c] to [b, c, a]
+      std::rotate(params.begin(), params.begin() + 1, params.end());
+      for (auto& param : params) {
+        param = -param;
+      }
+      break;
+    case qc::U2:
+      params[0U] = -params[1U] + PI;
+      params[1U] = -params[1U] - PI;
```
burgholzer: This is not quite right as it does not preserve the old semantics

burgholzer:
> Okay, thanks. Then, I will rebase onto the new main so that the test passes again. Do you have any further change requests regarding my pull request?

Sounds good.
I am on it ☝️ I might have one or two more requests.

include/mqt-core/dd/Operations.hpp
```diff
@@ -186,17 +141,17 @@ getStandardOperationDD(const qc::StandardOperation* op, Package<Config>& dd,
 // An empty permutation marks the identity permutation.
 
 template <class Config>
-qc::MatrixDD getDD(const qc::Operation* op, Package<Config>& dd,
+qc::MatrixDD getDD(const qc::Operation& op, Package<Config>& dd,
```
burgholzer: Just a comment: I was a bit hesitant about this change at first, as I thought it would cause object slicing, but I convinced myself that this is not the case and this still works as intended. I agree with your assessment that this even makes the intent clearer because `op == nullptr` was never even handled in the first place.

burgholzer: Given the comment above, I am also thinking whether it wouldn't be cleaner to separate the `inverse` function from the `non-inverse` function here.
I think the duplication would be at a minimum and the code would be easier to read overall. 
What do you think?

include/mqt-core/dd/Operations.hpp
```diff
@@ -205,54 +160,53 @@ qc::MatrixDD getDD(const qc::Operation* op, Package<Config>& dd,
...
-    const auto* standardOp = dynamic_cast<const qc::StandardOperation*>(op);
-    const auto& targets = permutation.apply(op->getTargets());
-    const auto& controls = permutation.apply(op->getControls());
+  if (op.isStandardOperation()) {
+    const auto* standardOp = dynamic_cast<const qc::StandardOperation*>(&op);
+    const auto& targets = permutation.apply(op.getTargets());
+    const auto& controls = permutation.apply(op.getControls());
 
     if (qc::isTwoQubitGate(type)) {
       assert(targets.size() == 2);
```
burgholzer: I think this distinction on whether it is a one or two target gate is superfluous based on your changes in previous commits.

src/dd/GateMatrixDefinitions.cpp
```diff
@@ -0,0 +1,223 @@
...
+    return pMat(params[0]);
+  case qc::RX:
+    return rxMat(params[0]);
+  case qc::RY:
+    return ryMat(params[0]);
+  case qc::RZ:
+    return rzMat(params[0]);
+  default:
+    assert(false && "Invalid single-qubit gate type");
+    return {}; // unreachable
```
burgholzer: If you really want to mark this as unreachable, then you can use `qc::unreachable()`.
```suggestion
    assert(false && "Invalid single-qubit gate type");
    qc::unreachable();
```
Although that part of the code is technically reachable as the method is public and can easily be passed a wrong `OpType`.
I'd still be fine with adding the unreachable there (and down below).

test/dd/test_package.cpp
```diff
@@ -137,13 +138,14 @@ TEST(DDPackageTest, BellState) {
...
-  auto h2Gate = dd->makeGateDD(dd::H_MAT, 2);
-  auto swapGate = dd->makeTwoQubitGateDD(dd::SWAP_MAT, qc::Controls{}, 0, 2);
+  auto h0Gate = dd::getDD(qc::StandardOperation(0, qc::H), *dd);
+  auto s0Gate = dd::getDD(qc::StandardOperation(1_pc, 0, qc::S), *dd);
+  auto t0Gate = dd::getDD(qc::StandardOperation(2_pc, 0, qc::T), *dd);
+  auto h1Gate = dd::getDD(qc::StandardOperation(1, qc::H), *dd);
+  auto s1Gate = dd::getDD(qc::StandardOperation(2_pc, 1, qc::S), *dd);
+  auto h2Gate = dd::getDD(qc::StandardOperation(2, qc::H), *dd);
+  auto swapGate = dd->makeTwoQubitGateDD(dd::opToTwoQubitGateMatrix(qc::SWAP),
+                                         qc::Controls{}, 0, 2);
```
burgholzer: You should be able to also replace this with a corresponding StandardOperation call to stay consistent. (same goes for a couple of occurences down below)

test/dd/test_package.cpp
```diff
@@ -1878,68 +1889,6 @@ TEST(DDPackageTest, DDFromSingleElementMatrix) {
...
-            const auto gateDD = dd->makeGateDD(
-                gateMatrix, qc::Controls{{control0, controlType}, control1},
-                target);
-            EXPECT_EQ(controlledGateDD, gateDD);
-          }
-        }
-      }
-    }
-  }
-}
```
burgholzer: I'd be in favor of leaving these tests in, but just locally defining the CX and CZ matrices. These are good sanity checks for the DD generation routines.

burgholzer (changes requested):
Many thanks again for your work on this!
I really like the way this is going and how it simplifies certain aspects in the library.
I feel we are really close to the finish line here. There are a couple of errors that snuck their way into the last couple of commits and some further suggestions for getting the implementation even cleaner and more performant.
I am fairly optimistic that this will be the last round of iteration.

include/mqt-core/dd/Operations.hpp
rotmanjanez: Yes, good point. I think we can solve this without a separate function and just have it inline like this:
```c++
if (not inverse) {
   // return gate directly
}
// construct inverse gate
std::vector<qc::Qubit> targetQubits = targets;
switch {
  // ...
}
// return inverse gate
```
This would have a duplicate call to gate creations, but as this is only two lines, I think this is ok.

include/mqt-core/dd/Operations.hpp
rotmanjanez: I will take a closer look at this

src/dd/GateMatrixDefinitions.cpp
rotmanjanez: If this code can get called by a library user, would it be better to raise an exception?

burgholzer: Most likely, yes.
Could be an `std::invalid_argument` exception.
That way, such errors get a proper error message even in Release mode.

include/mqt-core/dd/Operations.hpp
burgholzer: That works for me 👍🏼 sounds like a nice compromise.

burgholzer:
@Joshy-R gentle ping just in case this might have flown under the radar.

A couple of bits and pieces changed over the last couple of weeks that lead to merge conflicts here. However, these should be easy to resolve (hopefully).

Joshy-R:
No worries, I will take a look at this over the weekend and implement the requested changes.

include/mqt-core/dd/Operations.hpp
rotmanjanez: I don't see the benefit of handling inverse separately here. The amount of code necessary to handle the inverse is minimal and shows the similarity among inverse in non-inverse gates. If you feel strong about this change, i can implement is.

Joshy-R:
I have implemented most of the requested changes (see comment above). macos-14 is failing due to some caches. Retriggering the pipeline will likely fix the issue.

Please let me know if you would like me to make any changes.

burgholzer:
> I have implemented most of the requested changes (see comment above). macos-14 is failing due to some caches. Retriggering the pipeline will likely fix the issue.
> 
> Please let me know if you would like me to make any changes.

Thanks for your work on this. I'll try to have a look at your changes later today or early tomorrow.
I also triggered a re-run of the workflow in question. Have not seen that error in the past 🙃

include/mqt-core/dd/Operations.hpp
burgholzer: Looking at the code, I am fine with the current solution. Many thanks 🙏🏼

burgholzer (approved):
Many thanks for another round of changes.
I just went through the PR and made some final adjustments.
I also added the `extensive-cpp-ci` label, which kicks off a whole set of CI runs under different OSes and compilers, and checks whether everything still works.
Depending on how this goes, this is ready to go in 🚀 

Many thanks again for your contribution. Great work. 🙌🏼

burgholzer:
Just pushed another small commit (https://github.com/cda-tum/mqt-core/pull/752/commits/c595265e588076718c2c500b0d50792cc137485d), where the main aim was to get some coverage for the lines that were reported uncovered. This has led me to an intersting observation on this PR:
I just compared Python binary wheel sizes from before this PR, before https://github.com/cda-tum/mqt-core/pull/752/commits/c595265e588076718c2c500b0d50792cc137485d, and with https://github.com/cda-tum/mqt-core/pull/752/commits/c595265e588076718c2c500b0d50792cc137485d.

| lib | [v3.0.0b3](https://pypi.org/project/mqt-core/3.0.0b3/) | before https://github.com/cda-tum/mqt-core/pull/752/commits/c595265e588076718c2c500b0d50792cc137485d (diff to [v3.0.0b3](https://pypi.org/project/mqt-core/3.0.0b3/)) | after https://github.com/cda-tum/mqt-core/pull/752/commits/c595265e588076718c2c500b0d50792cc137485d (diff to [v3.0.0b3](https://pypi.org/project/mqt-core/3.0.0b3/)) |
| :---------------- | ----: | ----: | ----: |
| libmqt-core-algorithms.so | 109464 | 80688 (-26%) | 80688 (-26%) |
| libmqt-core-circuit-optimizer.so | 150608 | 146536 (-2.7%) | 134208 (-10.9%) |
| libmqt-core-dd.so | 562592 | 525664 (-6.6%) | 525664 (-6.6%) |
| libmqt-core-ds.so | 64104 | 60024 (-6.4%) | 59952 (-6.5%) |
| libmqt-core-ir.so | 1212752 | 1139120 (-6.1%) | 1130928 (-6.7%) |
| libmqt-core-na.so | 64000 | 47616 (-25.6%) | 39256 (-38.7%) |
| libmqt-core-zx.so | 310912 | 306808 (-1.3%) | 298568 (-4%) |
| ir.cpython | 524528 | 499952 (-4.7%) | 495856 (-5.5%) |
| Overall | 2998960 | 2806408 (-6.4%) | 2765120 (-7.8%) |

which is quite impressive for such a considerably small change ⚡

burgholzer:
Our coverage collection certainly does not like const static variables in functions.
I suspect that even at the lowest optimization level, which we use for coverage collection, this is optimized in some form where the coverage data cannot be mapped to the respective lines. I guess it is what it is and we are going to have to live with that.

PR #846, opened 2025-03-03
🎨 Refactor `NAComputation` with concrete base classes for every operation and ouput new `.naviz` format

ystade:
## Description

The tool [`mqt-naviz`](github.com/cda-tum/mqt-naviz) purposefully introduces a new textual format that allows to execute operations at a specific time stamp. The current output format of NA computation is not really compatible even though `mqt-naviz` supports a legacy feature to read in this old format with some caveats.

This issue triggered a wider refactoring of the `NAComputation`. Its entire structure is refactored. Most importantly, concrete Operation have their own class representation inheriting from some super class. At the end, all operations inherit from the class `Op`. As part of the refactoring, there is a straight forward way to test whether an operation is of some particular type and then it can be casted to this type.

## Checklist:


include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,102 @@
...
   }
-  template <class T, class... Args> auto emplaceBack(Args&&... args) -> void {
-    static_assert(std::is_base_of_v<NAOperation, T>,
-                  "T must be a subclass of NAOperation.");
-    operations.emplace_back(std::make_unique<T>(std::forward<Args>(args)...));
+  [[nodiscard]] auto
+  getLocationOfAtomAfterOperation(const std::unique_ptr<Atom>& atom,
+                                  const std::unique_ptr<Op>& op) const
+      -> Location {
+    return getLocationOfAtomAfterOperation(atom.get(), op.get());
```
burgholzer: This is more of a general interface design remark. It applies to various functions and methods throughout the PR. I'll annotate this here and leave the remaining places uncommented.

This method does neither change ownership nor does it modify any of its parameters. Furthermore, it is only reasonable to apply this method for non-null arguments. As such, these parameters should simply be const references to the underlying classes instead of unique pointers.
Calls where you currently do `ptr->get()` would be replaced by `*ptr`.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,102 @@
...
   }
-  auto clear(const bool clearInitialPositions = true) -> void {
-    operations.clear();
-    if (clearInitialPositions) {
-      initialPositions.clear();
-    }
+  [[nodiscard]] auto
+  getLocationOfAtomAfterOperation(const std::unique_ptr<Atom>& atom,
+                                  const Op* op) const -> Location {
+    return getLocationOfAtomAfterOperation(atom.get(), op);
```
burgholzer: I'd argue that these additional functions are redundant and just create boilerplate code. I'd personally prefer to have just the version taking const reference arguments.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,102 @@
...
-    initialPositions.emplace_back(std::move(p));
+  [[nodiscard]] auto getLocationOfAtomAfterOperation(const Atom* atom,
+                                                     const Op* op) const
+      -> Location;
+  auto emplaceBackAtom(std::string name) -> const Atom* {
+    return atoms.emplace_back(std::make_unique<Atom>(std::move(name))).get();
+  }
+  auto emplaceBackZone(std::string name) -> const Zone* {
+    return zones.emplace_back(std::make_unique<Zone>(std::move(name))).get();
+  }
```
burgholzer: Would it make sense for these to return references instead of pointers given how the return values will never be null.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,102 @@
...
+  auto emplaceInitialLocation(const Atom* atom, const Location& loc) -> void {
+    initialLocations.emplace(atom, loc);
+  }
```
burgholzer: For consistency reasons, this should probably also return a const reference.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,102 @@
...
+    return std::vector<std::unique_ptr<Op>>::emplace_back(
+               std::make_unique<T>(std::forward<T>(op)))
+        .get();
+  }
+  template <class T, typename... Args>
+  auto emplaceBack(Args&&... args) -> const Op* {
+    return std::vector<std::unique_ptr<Op>>::emplace_back(
+               std::make_unique<T>(std::forward<Args>(args)...))
+        .get();
   }
```
burgholzer: Are these really necessary?
They seem to be covered by existing vector functionality.

include/mqt-core/na/NAUtils.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <utility>
+
+namespace na {
+/**
+ * @brief Checks whether a gate is global.
+ * @details A StandardOperation is global if it acts on all qubits.
+ * A CompoundOperation is global if all its sub-operations are
+ * StandardOperations of the same type with the same parameters acting on all
+ * qubits. The latter is what a QASM line like `ry(π) q;` is translated to in
+ * MQT-core. All other operations are not global.
```
burgholzer: It's either "MQT Core" or "mqt-core". Holds for all docstrings and references.

include/mqt-core/na/NAUtils.hpp
```diff
@@ -0,0 +1,68 @@
...
+ */
+[[nodiscard]] inline auto isGlobal(const qc::Operation& op,
```
burgholzer: Why is this not a method of the Operation class that is overridden in the subclasses? Defaults to false and to the respective branch code for the respective classes.

include/mqt-core/na/NAUtils.hpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+template <> struct std::hash<std::pair<std::size_t, std::size_t>> {
+  std::size_t
+  operator()(const std::pair<std::size_t, std::size_t>& p) const noexcept {
+    const std::size_t h1 = std::hash<std::size_t>{}(p.first);
+    const std::size_t h2 = std::hash<std::size_t>{}(p.second);
+    return qc::combineHash(h1, h2);
+  }
+};
```
burgholzer: Given the above comment, I think these could be moved to the place where they are actually required.
As you probably know by now, I am not a big fan of "utils" files.

include/mqt-core/na/entities/Atom.hpp
```diff
@@ -0,0 +1,51 @@
...
+public:
+  Atom() = default;
+  explicit Atom(std::string name) : name(std::move(name)) {}
+  Atom(const Atom& atom) = default;
+  Atom(Atom&& atom) noexcept = default;
+  Atom& operator=(const Atom& atom) = default;
+  Atom& operator=(Atom&& atom) noexcept = default;
+  ~Atom() = default;
+  [[nodiscard]] auto getName() const -> std::string { return name; }
+  [[nodiscard]] auto toString() const -> std::string { return name; }
```
burgholzer: This method kind of seems redundant. I'd have a slight tendency to remove it and just keep the `getName` method around. Or do you envision a future where the Atom class will get more members that need to be part of the string representation?

include/mqt-core/na/entities/Atom.hpp
```diff
@@ -0,0 +1,51 @@
...
+    return !(*this == other);
+  }
+};
+} // namespace na
+template <> struct std::hash<const na::Atom*> {
+  [[nodiscard]] auto operator()(const na::Atom* atom) const noexcept
+      -> std::size_t {
+    return hash<std::string>{}(atom->getName());
+  }
+}; // namespace std
```
burgholzer: You should not actually need to provide a hash for a pointer type. The regular hash should be good enough. Do you see a compelling reason for keeping this?

include/mqt-core/na/entities/Location.hpp
```diff
@@ -0,0 +1,111 @@
...
+  explicit Location(const std::pair<double, double>& p)
+      : Location(p.first, p.second) {};
+  explicit Location(const std::pair<std::int32_t, std::int32_t>& p)
+      : Location(p.first, p.second) {};
+  explicit Location(const std::pair<std::int64_t, std::int64_t>& p)
+      : Location(p.first, p.second) {};
+  explicit Location(const std::pair<std::uint32_t, std::uint32_t>& p)
+      : Location(p.first, p.second) {};
+  explicit Location(const std::pair<std::uint64_t, std::uint64_t>& p)
+      : Location(p.first, p.second) {};
```
burgholzer: C++ already defines all kinds of casting rules. Do we really need all of these constructors here? Isn't one constructor taking a pair of doubles and one taking two doubles entirely sufficient?
Not a dealbreaker per-se, so we can also leave this in if you prefer them. However, it kind of makes me question whether there truly are so many diverse cases for specifying a location or if it's just the consuming code that should be streamlined to simply rely on using `double`.

Edit: technically, this is such a simple POD class that it probably wouldn't even need explicit constructors.

include/mqt-core/na/entities/Location.hpp
```diff
@@ -0,0 +1,111 @@
...
+  Location() = default;
+  Location(const Location& loc) = default;
+  Location(Location&& loc) noexcept = default;
+  Location& operator=(const Location& loc) = default;
+  Location& operator=(Location&& loc) noexcept = default;
+  ~Location() = default;
```
burgholzer: These can all be removed as they are defaulted, well..., by default.

include/mqt-core/na/entities/Location.hpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+  Location operator-(const Location&& loc) const {
+    return {x - loc.x, y - loc.y};
+  }
+  Location operator+(const Location& loc) const {
+    return {x + loc.x, y + loc.y};
+  }
+  Location operator+(const Location&& loc) const {
+    return {x + loc.x, y + loc.y};
+  }
```
burgholzer: The rvalue overloads here are redundant. The compiler will happily pass rvalue references to a const reference parameter.
```suggestion
  Location operator-(const Location& loc) const {
    return {x - loc.x, y - loc.y};
  }
  Location operator+(const Location& loc) const {
    return {x + loc.x, y + loc.y};
  }
```
I am also pretty sure that these methods should be marked `[[nodiscard]]`

include/mqt-core/na/entities/Location.hpp
```diff
@@ -0,0 +1,111 @@
...
+  [[nodiscard]] auto length() const -> double {
+    return std::sqrt((x * x) + (y * y));
```
burgholzer:
```suggestion
    return std::hypot(x, y);
```
slightly more efficient and numerically stable.

include/mqt-core/na/entities/Location.hpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+  [[nodiscard]] auto operator>(const Location& other) const -> bool {
+    return x > other.x || (x == other.x && y > other.y);
+  }
+  [[nodiscard]] auto operator<=(const Location& other) const -> bool {
+    return x <= other.x || (x == other.x && y <= other.y);
+  }
+  [[nodiscard]] auto operator>=(const Location& other) const -> bool {
+    return x >= other.x || (x == other.x && y >= other.y);
+  }
```
burgholzer: You can avoid some repetition here, which also eliminates some risk for potential errors.
```suggestion
  [[nodiscard]] auto operator<(const Location& other) const -> bool {
    return x < other.x || (x == other.x && y < other.y);
  }
  [[nodiscard]] auto operator>(const Location& other) const -> bool {
    return other < *this;
  }
  [[nodiscard]] auto operator<=(const Location& other) const -> bool {
    return !(*this < other);
  }
  [[nodiscard]] auto operator>=(const Location& other) const -> bool {
    return !(other < *this);
  }
```

include/mqt-core/na/entities/Location.hpp
```diff
@@ -0,0 +1,111 @@
...
+    return (*this - loc).length();
+  }
+  [[maybe_unused]] [[nodiscard]] auto
+  getManhattanDistanceX(const Location& loc) const -> double {
+    return std::abs(x - loc.x);
+  }
+  [[maybe_unused]] [[nodiscard]] auto
+  getManhattanDistanceY(const Location& loc) const -> double {
+    return std::abs(y - loc.y);
+  }
```
burgholzer: Why are these marked as `[[maybe_unused]]`? If they are important somewhere else (e.g., QMAP), I'd recommend at least testing them here.

include/mqt-core/na/entities/Zone.hpp
```diff
@@ -0,0 +1,43 @@
...
+  std::string name;
+
+public:
+  Zone() = default;
+  explicit Zone(std::string name) : name(std::move(name)) {}
+  Zone(const Zone& zone) = default;
+  Zone(Zone&& zone) noexcept = default;
+  Zone& operator=(const Zone& zone) = default;
+  Zone& operator=(Zone&& zone) noexcept = default;
+  ~Zone() = default;
```
burgholzer:
```suggestion
```

include/mqt-core/na/entities/Atom.hpp
```diff
@@ -0,0 +1,51 @@
...
+  std::string name;
+
+public:
+  Atom() = default;
+  explicit Atom(std::string name) : name(std::move(name)) {}
+  Atom(const Atom& atom) = default;
+  Atom(Atom&& atom) noexcept = default;
+  Atom& operator=(const Atom& atom) = default;
+  Atom& operator=(Atom&& atom) noexcept = default;
+  ~Atom() = default;
```
burgholzer: If all of these are defaulted, you might as well remove them, as they will be auto generated by the compiler anyway.
```suggestion
```

include/mqt-core/na/entities/Zone.hpp
```diff
@@ -0,0 +1,43 @@
...
+  [[nodiscard]] auto getName() const -> std::string { return name; }
+  [[nodiscard]] auto toString() const -> std::string { return name; }
```
burgholzer: Same argument as for the `Atom` class

include/mqt-core/na/operations/Op.hpp
```diff
@@ -0,0 +1,34 @@
...
+  }
+  template <class T> [[nodiscard]] auto is() const -> bool {
+    return dynamic_cast<const T*>(this) != nullptr;
+  }
+  template <class T> [[nodiscard]] auto as() -> T& {
+    return dynamic_cast<T&>(*this);
+  }
+  template <class T> [[nodiscard]] auto as() const -> const T& {
+    return dynamic_cast<const T&>(*this);
+  }
```
burgholzer: I like these. Very elegant. Should probably add that to the regular `qc::Operation` as well.
Although the sequence of checking if an operation and subsequently casting it now requires two dynamic casts instead of just one, if one manually does 
```c++
if(auto* derivedOp = dynamic_cast<DerivedOp*>(op); op!=nullptr) {
  ... do something with derivedOp
}
```
instead of 
```c++
if (op.is<DerivedOp>()) {
  auto& derivedOp = op.as<DerivedOp>();
  ... do something with derivedOp
}
```
I quickly played with that in compiler explorer and it seems that modern compilers are clever enough to optimise the above pattern to a single dynamic cast. See https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGISRqkrgAyeAyYAHI%2BAEaYxCBmAOykAA6oCoRODB7evv6BaRmOAqHhUSyx8Um2mPbFDEIETMQEOT5%2BATV1WY3NBKWRMXEJyQpNLW15nWN9A%2BWVIwCUtqhexMjsHMgGCgoA1BZMSnsmiVYaAIIpXtG0eMggJpd7B0eYEIsnAMwAIntYVCYXloBBMX3OFxeADc8C0vGI9gA/Q5KD7fP4AoEgsEQl4mACsVkJDAweAUomI6AJP2pexhcIRQKIe2AmAIAEkfmi0AwxnsALRg7B7MIEdF7DQ4p6QvYETAsFIGOXfNzbI77AAqQpOhIJFhJ%2BHJzSp%2BJppr2TNQIoU3IEfMFX2F0VQnhOZ2lLxexDZawY/wAnswWPcAPqiMZg1V2sUagBUQogBAQZM%2BYDAYL%2BDGBtBSBGIUuebpphblCqVmBVat2ey1jp1RP1pKNlOptMtFptnwdwq1ZgAbG7cZ7vQRfQGg6Hw6Cvm5e32E7GkymCzLTsWZaXFUxlZGq5rtXq9QayRSTWa/u2jrbeWLu3seXy54OPcOfcQ/ehA6xJ0dp1GbzW5jzo6ECLsmCiLCueKJOua4rtKe57D8cR4FCmDoHsNDMLQIB7Nctz3K8xynEOLyiiKVLgtK%2BF3A8L7IcQqHoRAD5iuRDCfLheDoBAHHPjBeKFoexJNqerbmu2rIclynysQK2rkagaHEIxWDPoWXpvn63FQUW0pwVRlzSu2LBMGEaJ3uRJEvi8ckopgsZ7NEbzKHm4rhAA7khKFoVSZxcAZpEioIewjuKkqGTKZFUHsEDOUornEN2ZKRgxTGUdgHyfNZGmerZ0YWl4RBAf8PnoeK8UqHm3ZHKlZUZR8ul5ZpYoZqVjG%2BQAdFJnKNZFeVwblWD0MqlWJU1oVaZN04QmuHDLDhHD4rwfgcFopCoJwbjWNYewKKs6wVuYXw8KQBCaPNywANYgPigSLZIK0XRtnC8AoICBOda3zaQcCwDAiAoKgCp0HE5CUGgIP0PEwBSGYfB0HKxDvXFz3RGEzT%2BmjGPEP6ADy0TaJgDicKdkNsIIeMMLQWPfaQWDRF4wBuGItDvdwvBYKZRjiHT%2BDeg4THs%2BtmCqMTRWbOtoq1M9dzRMQmMeFgpO8HmeAsCrpDKc6SjIdzwB3EYF3LFQBjAAoABqeCYB5eMpIwmv8IIIhiOwUgyIIigqOodO6Fw%2BiGMYO2WPoeDRO9kDLKguZZOz/J418ApjJRPymJY1hcIkvBKXEqmYBHHxdMT9QuAw7ieO0eghGEgwVMM/uFJkAiTH4DfpE3DBzEM8T%2B3Yxc9OMrQV3kve1P3Ai9C0Xd1z3tiDy3egzFPNfzPXyz7WsGwSAtnDLaQq3rZtHB7KoAAcfb8n2kgssgyB7FInVmLFuCECQJxmCdiyq8b123fdnCPVIBrRI%2BJOoAE4QGZ0SF8DQkgwFX3xF8fez0j5vQ%2BmdY2v0AYQCQKsAg1wCDgwgOhV%2Bqk9BO2EKIcQ7sKFezUM9P2pAPIKxSCrHeS0np0yPnjIqBC9ioBimfC%2BV8b53wfk/CAHgoZxHfp/b%2B31FjLAQJgJgWB4iFwerwDWZhJCdQ0Do0%2BZh%2BxgLul8fEkgr7IK4a9Ww6CvpaEUaQG6d19CcC%2BJww%2BNj7GXVcRwMwHjs5eJ/lrOIGRnCSCAA

include/mqt-core/na/operations/LocalOp.hpp
```diff
@@ -0,0 +1,47 @@
...
+  std::vector<qc::fp> params;
+  std::vector<const Atom*> atoms;
+
+  LocalOp(std::vector<qc::fp> params, std::vector<const Atom*> atoms)
+      : params(std::move(params)), atoms(std::move(atoms)) {}
+  explicit LocalOp(std::vector<const Atom*>& atoms)
+      : LocalOp({}, std::move(atoms)) {}
+  explicit LocalOp(std::vector<qc::fp> params, const Atom* atom)
+      : LocalOp(std::move(params), std::vector{atom}) {}
+  explicit LocalOp(const Atom* atom) : LocalOp({}, atom) {}
```
burgholzer: Not a must, obviously, but by switching around the parameters and the atoms, some of these constructors can be simplified.
```suggestion
  std::vector<const Atom*> atoms;
  std::vector<qc::fp> params;

  explicit LocalOp(std::vector<const Atom*> a, std::vector<qc::fp> p = {})
      : atoms(std::move(a)), params(std::move(p)) {}
  explicit LocalOp(const Atom* atom, std::vector<qc::fp> p = {})
      : LocalOp(std::vector{atom}, std::move(p)) {}
```
Note also the rename of the constructor parameters to avoid shadowing (when you look at the build output, gcc will typically scream at you for that).

include/mqt-core/na/operations/LocalOp.hpp
```diff
@@ -0,0 +1,47 @@
...
+public:
+  LocalOp() = delete;
+  [[nodiscard]] auto getAtoms() -> decltype(atoms)& { return atoms; }
+  [[nodiscard]] auto getAtoms() const -> const decltype(atoms)& {
+    return atoms;
+  }
+  [[nodiscard]] auto getParams() -> decltype(params)& { return params; }
+  [[nodiscard]] auto getParams() const -> const decltype(params)& {
+    return params;
+  }
```
burgholzer: Not saying you have to do this, but the signatures of these functions can be simplified a little, which saves some typing. Compilers are clever enough to deduce the right types here.
```suggestion
  [[nodiscard]] auto getAtoms() -> auto& { return atoms; }
  [[nodiscard]] auto getAtoms() const -> const auto& {
    return atoms;
  }
  [[nodiscard]] auto getParams() -> auto& { return params; }
  [[nodiscard]] auto getParams() const -> const auto& {
    return params;
  }
```

include/mqt-core/na/operations/LocalRZOp.hpp
```diff
@@ -0,0 +1,31 @@
...
+namespace na {
+class LocalRZOp final : public LocalOp {
+public:
+  LocalRZOp(qc::fp angle, const Atom* atom) : LocalOp({angle}, atom) {
+    name = "rz";
+  }
+  LocalRZOp(qc::fp angle, std::vector<const Atom*> atom)
+      : LocalOp({angle}, std::move(atom)) {
+    name = "rz";
+  }
```
burgholzer: truly minor
```suggestion
  LocalRZOp(const qc::fp angle, const Atom* atom) : LocalOp({angle}, atom) {
    name = "rz";
  }
  LocalRZOp(const qc::fp angle, std::vector<const Atom*> atom)
      : LocalOp({angle}, std::move(atom)) {
    name = "rz";
  }
```

include/mqt-core/na/operations/GlobalOp.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+namespace na {
+class GlobalOp : public Op {
+protected:
+  std::string name;
+  std::vector<qc::fp> params;
+  const Zone* zone;
+  GlobalOp(std::vector<qc::fp> params, const Zone* zone)
+      : params(std::move(params)), zone(zone) {}
+  explicit GlobalOp(const Zone* zone) : GlobalOp({}, zone) {}
```
burgholzer: Similar to the `LocalOp` comments, these could probably read
```c++
  explicit GlobalOp(const Zone* z, std::vector<qc::fp> p = {})
      : zone(z), params(std::move(p)) {}
```

I was also wondering, whether it should actually read
```suggestion
  explicit GlobalOp(const Zone& z, std::vector<qc::fp> p = {})
      : zone(&z), params(std::move(p)) {}
```
given how I assume the zone must not be null in order for the operation to make sense.

include/mqt-core/na/operations/GlobalOp.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+public:
+  GlobalOp() = delete;
+  [[nodiscard]] auto getParams() -> decltype(params)& { return params; }
+  [[nodiscard]] auto getParams() const -> const decltype(params)& {
+    return params;
+  }
+  [[nodiscard]] auto getZone() -> const Zone*& { return zone; }
+  [[nodiscard]] auto getZone() const -> const Zone* { return zone; }
```
burgholzer: Similar comments to the `LocalOp` class apply here.
Although for the `Zone` one, both methods really don't make a difference in practice. A reference will be the same size as the actual pointer (8 byte). And using it, is just the same. Given how it's a const pointer, it cannot be mutated anyway. Might as well skip one of the methods here.

Alternatively, you could return a `const Zone&` here by dereferencing the pointer which, I assume, must not be null anyway. I think I'd prefer that solution as it makes using this a little more convenient (need not check the result for being null, can use `.` instead of `->`)

include/mqt-core/na/operations/LocalOp.hpp
burgholzer: A small follow-up on this triggered by one of the comments in the `GlobalOp` class:
At least for the single atom constructor, I would assume this can also be
```c++
explicit LocalOp(const Atom& atom, std::vector<qc::fp> p = {})
      : LocalOp(std::vector{&atom}, std::move(p)) {}
```
as I would argue that `atom` must not be null anyway.
Something like that is a bit harder to enforce for the `vector` constructor as I am pretty sure that `std::vector<const Atom&>` doesn't really work well (because a reference to an Atom cannot be default constructed).

I'd still opt for changing the single-atom constructor.

include/mqt-core/na/operations/ShuttlingOp.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <utility>
+#include <vector>
+
+namespace na {
+class ShuttlingOp : public Op {
+protected:
+  std::vector<const Atom*> atoms;
+  explicit ShuttlingOp(std::vector<const Atom*> atoms)
+      : atoms(std::move(atoms)) {}
```
burgholzer: Why does this not have a single-atom constructor? might be worth adding one for consistency.

include/mqt-core/na/operations/ShuttlingOp.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+public:
```
burgholzer, line 37: In the other classes, you explicitly deleted the default constructor. I'd argue this makes sense here as well.

include/mqt-core/na/operations/ShuttlingOp.hpp
```diff
@@ -0,0 +1,35 @@
...
+  [[nodiscard]] auto getAtoms() -> decltype(atoms)& { return atoms; }
+  [[nodiscard]] auto getAtoms() const -> const decltype(atoms)& {
+    return atoms;
+  }
```
burgholzer:
```suggestion
  [[nodiscard]] auto getAtoms() -> auto& { return atoms; }
  [[nodiscard]] auto getAtoms() const -> const auto& {
    return atoms;
  }
```

include/mqt-core/na/operations/LoadOp.hpp
```diff
@@ -0,0 +1,58 @@
...
+                  std::vector<Location> targetLocations)
+      : ShuttlingOp(std::move(atoms)),
+        targetLocations(std::move(targetLocations)) {
+    if (this->atoms.size() != this->targetLocations->size()) {
+      throw std::invalid_argument(
+          "Number of atoms and target locations must be equal.");
+    }
+  }
+  explicit LoadOp(std::vector<const Atom*> atoms)
+      : ShuttlingOp(std::move(atoms)) {}
```
burgholzer:
```suggestion
  explicit LoadOp(std::vector<const Atom*> atms,
                  std::vector<Location> locations)
      : ShuttlingOp(std::move(atms)),
        targetLocations(std::move(locations)) {
    if (atoms.size() != targetLocations->size()) {
      throw std::invalid_argument(
          "Number of atoms and target locations must be equal.");
    }
  }
  explicit LoadOp(std::vector<const Atom*> atms)
      : ShuttlingOp(std::move(atms)) {}
```
Would be great to avoid the shadowing compiler warnings. I am open for bike shedding on the abbreviation policy. We could also simply add an underscore at the end of parameters or class members instead of crippling the names.

include/mqt-core/na/operations/ShuttlingOp.hpp
```diff
@@ -0,0 +1,35 @@
...
+  [[nodiscard]] virtual auto getTargetLocations() -> std::vector<Location>& = 0;
```
burgholzer: Does it make sense to get a mutable reference for the target locations or would it be better to leave these as a read-only member that is set at construction?

include/mqt-core/na/operations/StoreOp.hpp
```diff
@@ -0,0 +1,58 @@
...
+                   std::vector<Location> targetLocations)
+      : ShuttlingOp(std::move(atoms)),
+        targetLocations(std::move(targetLocations)) {
+    if (this->atoms.size() != this->targetLocations->size()) {
+      throw std::invalid_argument(
+          "Number of atoms and target locations must be equal.");
+    }
+  }
+  explicit StoreOp(std::vector<const Atom*> atoms)
+      : ShuttlingOp(std::move(atoms)) {}
```
burgholzer: Similar comments as for the `LoadOp` apply here.

include/mqt-core/na/operations/MoveOp.hpp
```diff
@@ -0,0 +1,47 @@
...
+public:
+  explicit MoveOp(std::vector<const Atom*> atoms,
+                  std::vector<Location> targetLocations)
+      : ShuttlingOp(std::move(atoms)),
+        targetLocations(std::move(targetLocations)) {
+    if (this->atoms.size() != this->targetLocations.size()) {
+      throw std::invalid_argument(
+          "Number of atoms and target locations must be equal.");
+    }
+  }
```
burgholzer: I guess it would also make sense to have single-atom, single-location constructors for the individual shuttling operations.
These could take a `const Atom& atom` and a `const Location& location`.
Or `const std::optional<Location>& = std::nullopt` in case of the load and store operations.

src/na/operations/LocalOp.cpp
```diff
@@ -0,0 +1,44 @@
...
+      if (!params.empty()) {
+        for (const auto& p : params) {
+          ss << p << " ";
+        }
+      }
+      ss << *atom << "\n";
+    }
+    ss << "]";
+  }
+  return ss.str();
```
burgholzer: This might be a lot of personal preference, but I am a big fan of reducing indentation in code by using early returns. IMHO it makes code more readable.
If you agree or have no major objections, this could be rewritten to:
```suggestion
  if (atoms.size() == 1) {
    if (!params.empty()) {
      for (const auto& p : params) {
        ss << " " << p;
      }
    }
    ss << " " << *(atoms.front());
    return ss.str();
  }
  ss << " [\n";
  for (const auto* const atom : atoms) {
    ss << "    ";
    if (!params.empty()) {
      for (const auto& p : params) {
        ss << p << " ";
      }
    }
    ss << *atom << "\n";
  }
  ss << "]";
  return ss.str();
```

src/na/operations/MoveOp.cpp
```diff
@@ -0,0 +1,31 @@
...
+  ss << "@+ move";
+  if (atoms.size() == 1) {
+    ss << " " << targetLocations.front() << " " << *(atoms.front());
+  } else {
+    ss << " [\n";
+    for (std::size_t i = 0; i < atoms.size(); ++i) {
+      ss << "    " << targetLocations[i] << " " << *(atoms[i]) << "\n";
+    }
+    ss << "]";
+  }
```
burgholzer: Depending on your preference, this could use an early return as well. This actually holds for all of these print functions. So I am not going to annotate this further.

test/na/test_na_utils.cpp
burgholzer, line 1: Based on the comment in the header file, this test can be moved to some other test file. Probably either the QASM importer tests or the QFR functionality.
At some point, we really need to split up the test suite a little more.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,102 @@
...
 #include <memory>
 #include <ostream>
 #include <string>
+#include <unordered_map>
+#include <utility>
 #include <vector>
 
 namespace na {
-class NAComputation {
+class NAComputation final : protected std::vector<std::unique_ptr<Op>> {
```
burgholzer: As much as I like the reduction in the amount of code that needs to be written by simply deriving from `std::vector`, this is considered a bad practice in modern C++.

I'll just paste an answer from Claude on this here. you can also simply google "c++ inheriting from std::vector" and read through all the forum posts.

While the usage here seems to be mostly fine, it also seems like a footgun we might want to avoid from the start.

---

Based on the conversation history and the code you've shared, I'll address your new question about STL class inheritance.

Inheriting from STL containers like `std::vector` is generally considered bad practice in modern C++. There are several key reasons for this:

1. **Non-virtual destructors** - Most STL containers don't have virtual destructors, which can cause undefined behavior when derived classes are deleted through base class pointers.

2. **Inadvertent slicing** - Container objects might be copied or moved in ways that could slice derived class data.

3. **Invariant violations** - Your derived class might introduce functionality that breaks the container's invariants or expected behavior.

4. **Composition over inheritance** - The standard approach is to use composition (containing an STL container as a member) rather than inheritance.

Instead of:
```cpp
class MyContainer : public std::vector<int> { /* ... */ };
```

Prefer:
```cpp
class MyContainer {
private:
    std::vector<int> data;
public:
    // Expose only the operations you want
};
```

This composition approach provides better encapsulation and prevents unintended behaviors. If needed, you can selectively expose container functionality through wrapper methods or by using public inheritance from an interface class that defines the desired behavior.

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
 #include <sstream>
 #include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
 
 namespace na {
+auto NAComputation::getLocationOfAtomAfterOperation(const Atom* atom,
+                                                    const Op* op) const
```
burgholzer: This function should take both arguments as const references since they must not be null.

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
-    ss << *p << ", ";
-  }
-  if (ss.tellp() == 8) {
-    ss.seekp(-1, std::ios_base::end);
-  } else {
-    ss.seekp(-2, std::ios_base::end);
+  std::vector<std::pair<const Atom*, Location>> initialLocationsAsc(
+      initialLocations.begin(), initialLocations.end());
+  std::sort(initialLocationsAsc.begin(), initialLocationsAsc.end(),
+            [](const auto& a, const auto& b) { return a.second < b.second; });
```
burgholzer: You could simply make this an `std::map<Location, const Atom*>` which would automatically be sorted correctly.

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
+        //===-----------------------------------------------------------------//
+        // Load Operations
+        //-----------------------------------------------------------------===//
+        if (std::any_of(opAtoms.begin(), opAtoms.end(),
+                        [&currentlyShuttling](const auto* atom) {
+                          return currentlyShuttling.find(atom) !=
+                                 currentlyShuttling.end();
+                        })) {
+          std::cout << "Error in op number " << counter
+                    << " (atom already loaded)\n";
```
burgholzer: Is printing to `std::cout` really the best way to handle this here?
Wouldn't it be cleaner to have an `std::string errorMessage` member in the class and store the respective error message there? In addition with a Getter and some convenience function for printing errors.

It is not a really good practice to pollute the standard output.
It would be slightly better if these were reported to `std::cerr`, but still not really ideal.

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
+          return false;
+        }
+        std::for_each(opAtoms.begin(), opAtoms.end(),
+                      [&currentlyShuttling](const auto* atom) {
+                        currentlyShuttling.insert(atom);
+                      });
```
burgholzer: Can't help myself and have to annotate it. For me, this is so much harder to read than a simple
```suggestion
        for (const auto* atom: opAtoms) {
          currentlyShuttling.emplace(atom);
        }   
```
It is also shorter and less verbose.

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
-          if (s1->y > s2->y && e1->y <= e2->y) {
-            std::cout << "Error in op number " << counter
-                      << " (row order not preserved)\n";
-            return false;
+      //===----------------------------------------------------------------===//
+      // All Shuttling Operations that move atoms
+      //===----------------------------------------------------------------===//
+      if ((op->is<LoadOp>() && op->as<LoadOp>().hasTargetLocations()) ||
+          (op->is<StoreOp>() && op->as<StoreOp>().hasTargetLocations()) ||
+          op->is<MoveOp>()) {
```
burgholzer: Couldn't you simplify this by adding a 
```c++
[[nodiscard]] virtual auto hasTargetLocations() const -> bool {
  return true;
}
```
method to the `ShuttlingOp` class and replacing the code here with
```c++
if(shuttlingOp.hasTargetLocations()) {
```

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
+                        << " (column order not preserved)\n";
+              return false;
+            }
+            // Exp.:
+            //   o
+            // o |
+            // | |
+            // v |
+            // o v
+            //   o
```
burgholzer: Really like these examples! Great documentation

src/na/NAComputation.cpp
```diff
@@ -9,100 +9,227 @@
...
+          currentLocations[opAtoms[i]] = targetLocations[i];
+        }
+      }
+      if (shuttlingOp.is<StoreOp>()) {
+        //===-----------------------------------------------------------------//
+        // Store Operations
+        //-----------------------------------------------------------------===//
+        std::for_each(opAtoms.begin(), opAtoms.end(), [&](const auto* atom) {
+          currentlyShuttling.erase(atom);
+        });
```
burgholzer: regular range-based for loop?

test/na/test_nacomputation.cpp
```diff
@@ -8,134 +8,168 @@
...
+  // row order not preserved
+  qc.emplaceBack<StoreOp>(std::vector{atom2, atom1},
+                          std::vector{Location{0, 1}, Location{2, 2}});
+  EXPECT_FALSE(qc.validate());
+  qc.clear();
+  qc.emplaceBack<LoadOp>(std::vector{atom1});
+  qc.emplaceBack<StoreOp>(std::vector{atom1});
+  qc.emplaceBack<StoreOp>(std::vector{atom1});
+  EXPECT_FALSE(qc.validate());
+}
```
burgholzer: This would be much cleaner as a test fixture and individual unit tests, which would also allow running them separately.

test/na/test_nacomputation.cpp
```diff
@@ -8,134 +8,168 @@
...
+  const auto* const atom3 = qc.emplaceBackAtom("atom3");
+  qc.emplaceInitialLocation(atom0, 0, 0);
+  qc.emplaceInitialLocation(atom1, 1, 0);
+  qc.emplaceInitialLocation(atom2, 0, 2);
+  qc.emplaceInitialLocation(atom3, 1, 2);
+  qc.emplaceBack<LoadOp>(std::vector{atom0, atom1},
+                         std::vector{Location{0, 1}, Location{1, 1}});
+  EXPECT_TRUE(qc.validate());
+  // atom already loaded
+  qc.emplaceBack<LoadOp>(std::vector{atom0}, std::vector{Location{0, 1}});
```
burgholzer: This does not have to be now, but would it make sense to add convenience methods to the computation class for applying these operations? For example
```c++
qc.load(atom, location)
qc.load(atoms, locations)
qc.move(atom, location)
qc.move(atoms, locations)
qc.store(atom, location)
qc.store(atoms, locations)
qc.localRZ(atom, theta)
qc.localRZ(atoms, theta)
qc.globalRY(zone, theta)
qc.globalCZ(zone)
```
I believe that already covers all currently possible operations and might make using the class more convenient.

Maybe also something like
```c++
const auto& atom = qc.initializeAtom(name, x, y)
```
makes sense.

burgholzer (changes requested):
Many thanks @ystade for creating this PR.
These are really some great improvements to the usability of the NA components of MQT Core.
As this kind of sets a precedent for changes we might also want to apply to the `qc::Operation` and `qc::QuantumComputation` hierarchy. I went through the code with some extra scrutiny.
You will find most of that in the inline comments.
Some of these are really up for discussion, but I think it is worth having that discussion.

I have two more general points:
- Please update the PR title to better reflect the changes in this PR. This is fundamentally refactoring how we handle NA computations and not simply updating an output format. Probably also worth to update the PR description
- Since a couple PRs back, we also have C++ documentation as part of our docs. As such, it would be really great to also have proper C++ documentation for all newly added C++ code. In times of LLM-assistance, I hope this should be moderately easy.

Many thanks for your work on this 🙏🏼

include/mqt-core/na/NAComputation.hpp
ystade: I agree for most of the other functions but here I think it is different. Here, I really need the address of the operation as it e.g. a RZ operation may occur multiple times but here I refer to a special one. So what I would need to do here to follow your reasoning would be to get the address of the value passed by reference. This also feels odd. What do you suggest?

include/mqt-core/na/NAComputation.hpp
ystade: They are just for convenience to skip the creation of the unique pointer.

include/mqt-core/na/NAComputation.hpp
burgholzer: Doing something like
```c++
const auto& ref = ...;
const auto* addr = &ref;
```
Is a pretty common and standard pattern.

include/mqt-core/na/entities/Location.hpp
ystade: I do not know. They came from another PR. I will add tests for them.

include/mqt-core/na/NAComputation.hpp
burgholzer: I think by just inheriting from the vector, you would have access to the respective method already. But given how that is bound to change, this is very similar to https://github.com/cda-tum/mqt-core/blob/e1e4515dbb4345a4d109733cb0efe7a6ef4385f6/include/mqt-core/ir/QuantumComputation.hpp#L562-L564
I just thought that you would get this for free by inheriting from std::vector

include/mqt-core/na/NAUtils.hpp
ystade: First, it was in this location before and I didn't think of moving it. Second, I find it nice to have it in the NA part as it is specific to NA and does not serve a general purpose.

burgholzer: Considering that we (hopefully) will merge the NA computation and the OG QuantumComputation at some point, a trait-like `isGlobal` still makes sense to me.
Even though this function might currently only be used in NA parts of the code base, it is essentially a basic feature of the MQT Core IR.

test/na/test_nacomputation.cpp
ystade: I keep that for later as this representation is bound to change when we transition to a unifrom representation perhaps based on MLIR.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -9,94 +9,168 @@
...
+  /// arguments and returns a reference to the newly created location.
+  /// @param atom The atom to set the initial location for.
+  /// @param loc The parameters for the location of the atom.
+  /// @return A reference to the newly created location.
+  template <typename... Args>
+  auto emplaceInitialLocation(const Atom& atom, Args&&... loc)
+      -> const Location& {
+    return initialLocations_
+        .emplace(&atom,
+                 Location{static_cast<double>(std::forward<Args>(loc))...})
```
ystade, line 135: Following your suggestion, @burgholzer I removed the plethora of constructors for the Location. Now, I need this cast here. Is there a better solution or is that the correct way here anyways?

burgholzer: I believe this is the correct way. This will scream whenever the thing being forwarded cannot be cast to a double, which is exactly as desired.

test/na/test_nacomputation.cpp
burgholzer: Alright. Then let's just keep this in mind as it would definitely benefit the usability.

ystade:
@burgholzer  The tests should run through now and all your feedback should be integrated to the best of my knowledge.

burgholzer:
Great! Many thanks for addressing this so quickly 🙏🏼 
I will have a look later today.

ystade:
@burgholzer There is (at least) one issue remaining. I used the convention to append a `_` to members to avoid name shadowing as suggsted. This conflicts with the naming convention. What do we want to do about it?

burgholzer:
> @burgholzer There is (at least) one issue remaining. I used the convention to append a `_` to members to avoid name shadowing as suggsted. This conflicts with the naming convention. What do we want to do about it?

Hm. Probably the cleanest solution would be to adjust the clang-tidy config so that this is not a violation anymore.
In order to not create dozens of warnings for the remainder of MQT Core, one could use a regex where the trailing underscore is optional.

burgholzer:
> Apparently, there is a specific key for that option. See the changes from the last commit.

I am pretty sure that is not what we want, as this enforces the trailing underscore. And I am also pretty sure you do not have the time resources to go through all of MQT Core and change that.

burgholzer (approved):
Many thanks for working through all the comments.
This is a really clean PR. I like that very much 😌 
I just put on some finishing touches in the last couple of commits. Specifically:
- removed two redundant files that were left over from before the last iteration
- added `@file` docstrings to all new files. (important for getting proper doxygen output)
- completed some missing docstrings
- shortened some existing ones where a one-line brief was sufficient
- adjusted the spacing in headers so that the files don't look so cluttered.
- applied some left-over suggestions in the NA computation.

Once CI is green, I will go ahead and merge this.
I'll also create a new 3.0 Beta release so that it becomes easier to directly rely on this in QMAP.

PR #862, opened 2025-03-14
✨ Add a couple of NA Operations and `convertToLocal` Function

ystade:
## Description

Even though neutral atom quantum computers usually only support `CZ`, `RZ`, `RX`, and `RY` gates where some of them must be global on a zone, we need more gates (Clifford Gates) for a project, where the `NAComputation` is used as an exchange format and not supposed to be executed on the hardware. Hence, this PR adds the missing operations to the infrastructure.

Additionally, this project will only handle local gates. Therefore, we add a function to the `NAComputation` that converts all global gates to respective local gates. To facilitate this, we need to know the extent of the zone to evaluate whether an atom is affected by a global gate or not. Additionally, for entangling gates, like `cz`, we also need to know the `radbergRadius` to evaluate which atoms interact with each other. All necessary changes are added to the corresponding classes.

## Checklist:

### DIFF
--- a/include/mqt-core/na/NAComputation.hpp
+++ b/include/mqt-core/na/NAComputation.hpp
@@ -106,10 +106,12 @@ public:
 
   /// Emplaces a new zone with the given name and returns a reference to the
   /// newly created zone.
-  /// @param name The name of the zone.
+  /// @param args The name of the zone and, optionally, the extent of the zone.
   /// @return A reference to the newly created zone.
-  auto emplaceBackZone(std::string name) -> const Zone& {
-    return *zones_.emplace_back(std::make_unique<Zone>(std::move(name)));
+  template <typename... Args>
+  auto emplaceBackZone(Args&&... args) -> const Zone& {
+    return *zones_.emplace_back(
+        std::make_unique<Zone>(std::forward<Args>(args)...));
   }
 
   /// Emplaces a new initial location for the given atom with the given location
@@ -178,5 +180,11 @@ public:
   /// @returns a pair of a Boolean indicating whether the NAComputation is valid
   /// and a string containing the error message if the NAComputation is invalid.
   [[nodiscard]] auto validate() const -> std::pair<bool, std::string>;
+
+  /// Calculates the position of each atom at every stage and replaces global
+  /// gates by local gates with the atoms that are affected by the global gate.
+  /// @param rydbergRange The range of the Rydberg interaction.
+  /// @note There is no way back after this function has been called.
+  auto convertToLocalGates(double rydbergRange) -> void;
 };
 } // namespace na
--- a/include/mqt-core/na/entities/Zone.hpp
+++ b/include/mqt-core/na/entities/Zone.hpp
@@ -13,7 +13,9 @@
 
 #pragma once
 
+#include <optional>
 #include <ostream>
+#include <stdexcept>
 #include <string>
 #include <utility>
 
@@ -23,16 +25,50 @@ namespace na {
 /// To maintain the uniqueness of zones, the name of the zone should be unique
 /// for this zone.
 class Zone final {
+public:
+  /// A simple struct to represent the extent of a zone.
+  struct Extent {
+    double minX = 0.0;
+    double minY = 0.0;
+    double maxX = 0.0;
+    double maxY = 0.0;
+  };
+
+private:
   /// The identifier of the zone.
   std::string name_;
+  /// The extent of the zone.
+  std::optional<Extent> extent_;
 
 public:
   /// Creates a new zone with the given name.
   /// @param name The name of the zone.
   explicit Zone(std::string name) : name_(std::move(name)) {}
 
+  /// Creates a new zone with the given name.
+  /// @param name The name of the zone.
+  /// @param extent The extent of the zone.
+  Zone(std::string name, const Extent& extent)
+      : name_(std::move(name)), extent_(extent) {}
+
   /// Returns the name of the zone.
   [[nodiscard]] auto getName() const -> std::string { return name_; }
+
+  /// Returns the extent of the zone.
+  [[nodiscard]] const std::optional<Extent>& getExtent() const {
+    return extent_;
+  }
+
+  /// Sets the extent of the zone.
+  void setExtent(const Extent& extent) { extent_ = extent; }
+
+  [[nodiscard]] auto contains(const Location& location) const -> bool {
+    if (!extent_) {
+      throw std::runtime_error("Zone's extent is not set.");
+    }
+    return location.x >= extent_->minX && location.y >= extent_->minY &&
+           location.x <= extent_->maxX && location.y <= extent_->maxY;
+  }
 
   /// Prints the zone to the given output stream.
   /// @param os The output stream to print the zone to.
--- a/include/mqt-core/na/operations/GlobalCZOp.hpp
+++ b/include/mqt-core/na/operations/GlobalCZOp.hpp
@@ -13,8 +13,19 @@
 
 #pragma once
 
+#include "LocalCZOp.hpp"
+#include "LocalOp.hpp"
+#include "na/entities/Atom.hpp"
+#include "na/entities/Location.hpp"
 #include "na/entities/Zone.hpp"
 #include "na/operations/GlobalOp.hpp"
+
+#include <array>
+#include <iterator>
+#include <map>
+#include <memory>
+#include <unordered_map>
+#include <vector>
 
 namespace na {
 /// Represents a global CZ operation in the NAComputation.
@@ -23,5 +34,34 @@ public:
   /// Creates a new CZ operation in the given zone.
   /// @param zone The zone the operation is applied to.
   explicit GlobalCZOp(const Zone& zone) : GlobalOp(zone, {}) { name_ = "cz"; }
+
+  /// Returns a local representation of the operation.
+  /// @param atomsLocations The locations of the atoms.
+  /// @param rydbergRange The range of the Rydberg interaction.
+  [[nodiscard]] auto
+  toLocal(const std::unordered_map<const Atom*, Location>& atomsLocations,
+          const double rydbergRange) const
+      -> std::unique_ptr<LocalOp> override {
+    // use a map here with the location as keys to ensure a deterministic order
+    // of the atoms
+    std::map<Location, const Atom*> affectedAtoms;
+    for (const auto& [atom, loc] : atomsLocations) {
+      if (zone_->contains(loc)) {
+        affectedAtoms.emplace(loc, atom);
+      }
+    }
+    std::vector<std::array<const Atom*, 2>> atomPairs;
+    for (auto it1 = affectedAtoms.cbegin(); it1 != affectedAtoms.cend();
+         ++it1) {
+      const auto& [loc1, atom1] = *it1;
+      for (auto it2 = std::next(it1); it2 != affectedAtoms.cend(); ++it2) {
+        const auto& [loc2, atom2] = *it2;
+        if ((loc1 - loc2).length() <= rydbergRange) {
+          atomPairs.emplace_back(std::array{atom1, atom2});
+        }
+      }
+    }
+    return std::make_unique<LocalCZOp>(atomPairs);
+  }
 };
 } // namespace na
--- a/include/mqt-core/na/operations/GlobalOp.hpp
+++ b/include/mqt-core/na/operations/GlobalOp.hpp
@@ -15,10 +15,15 @@
 #pragma once
 
 #include "Definitions.hpp"
+#include "LocalOp.hpp"
+#include "na/entities/Atom.hpp"
+#include "na/entities/Location.hpp"
 #include "na/entities/Zone.hpp"
 #include "na/operations/Op.hpp"
 
+#include <memory>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
@@ -53,5 +58,12 @@ public:
 
   /// Returns a string representation of the operation.
   [[nodiscard]] auto toString() const -> std::string override;
+
+  /// Returns a local representation of the operation.
+  /// @param atomsLocations The locations of the atoms.
+  /// @param rydbergRange The range of the Rydberg interaction.
+  [[nodiscard]] virtual auto
+  toLocal(const std::unordered_map<const Atom*, Location>& atomsLocations,
+          double rydbergRange) const -> std::unique_ptr<LocalOp> = 0;
 };
 } // namespace na
--- a/include/mqt-core/na/operations/GlobalRYOp.hpp
+++ b/include/mqt-core/na/operations/GlobalRYOp.hpp
@@ -14,8 +14,17 @@
 #pragma once
 
 #include "Definitions.hpp"
+#include "LocalOp.hpp"
+#include "LocalRYOp.hpp"
+#include "na/entities/Atom.hpp"
+#include "na/entities/Location.hpp"
 #include "na/entities/Zone.hpp"
 #include "na/operations/GlobalOp.hpp"
+
+#include <map>
+#include <memory>
+#include <unordered_map>
+#include <vector>
 
 namespace na {
 /// Represents a global RY operation in the NAComputation.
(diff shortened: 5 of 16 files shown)


include/mqt-core/na/entities/Zone.hpp
```diff
@@ -23,17 +27,51 @@ namespace na {
...
   /// Returns the name of the zone.
   [[nodiscard]] auto getName() const -> std::string { return name_; }
 
+  /// Returns the extent of the zone.
+  [[nodiscard]] const std::optional<Extent>& getExtent() const {
+    return extent_;
+  }
+
+  /// Sets the extent of the zone.
+  void setExtent(const Extent& extent) { extent_ = extent; }
```
burgholzer, line 65: trailing return types? (surprised clang-tidy did not complain)

include/mqt-core/na/entities/Zone.hpp
```diff
@@ -23,17 +27,51 @@ namespace na {
...
+
+  [[nodiscard]] auto contains(const Location& location) const -> bool {
+    if (!extent_) {
+      throw std::runtime_error("Zone's extent is not set.");
+    }
+    return location.x >= extent_->minX && location.y >= extent_->minY &&
+           location.x <= extent_->maxX && location.y <= extent_->maxY;
```
burgholzer, line 72: Maybe it is just me, but I always like these to be ordered as inequalities that could be read left to right just as you would write them in a math book.
```suggestion
    return extent_->minX <= location.x && location.x <= extent_->maxX &&
           extent_->minY <= location.y && location.y <= extent_->maxY;
```

include/mqt-core/na/entities/Zone.hpp
```diff
@@ -23,17 +27,51 @@ namespace na {
...
+    double minY = 0.0;
+    double maxX = 0.0;
+    double maxY = 0.0;
+  };
+
+private:
   /// The identifier of the zone.
   std::string name_;
+  /// The extent of the zone.
+  std::optional<Extent> extent_;
```
burgholzer, line 43: Is the extent of a zone truly optional? Every zone must have a certain extent.
You could be initializing the values with the min and max double values if you want a reasonable default.

include/mqt-core/na/operations/GlobalOp.hpp
```diff
@@ -53,5 +58,12 @@ class GlobalOp : public Op {
 
   /// Returns a string representation of the operation.
   [[nodiscard]] auto toString() const -> std::string override;
+
+  /// Returns a local representation of the operation.
+  /// @param atomsLocations The locations of the atoms.
+  /// @param rydbergRange The range of the Rydberg interaction.
```
burgholzer: I'd argue the common name for that is "rydbergRadius".

include/mqt-core/na/operations/GlobalRYOp.hpp
```diff
@@ -27,5 +36,26 @@ class GlobalRYOp final : public GlobalOp {
...
     name_ = "ry";
   }
+
+  /// Returns a local representation of the operation.
+  /// @param atomsLocations The locations of the atoms.
+  [[nodiscard]] auto
+  toLocal(const std::unordered_map<const Atom*, Location>& atomsLocations,
+          const double /* unused */) const
+      -> std::unique_ptr<LocalOp> override {
+    // make sure atoms are added always in the same order based on their
```
burgholzer:
```suggestion
    // make sure atoms are always added in the same order based on their
```

include/mqt-core/na/operations/GlobalRYOp.hpp
```diff
@@ -27,5 +36,26 @@ class GlobalRYOp final : public GlobalOp {
   GlobalRYOp(const Zone& zone, qc::fp angle) : GlobalOp(zone, {angle}) {
     name_ = "ry";
   }
+
+  /// Returns a local representation of the operation.
+  /// @param atomsLocations The locations of the atoms.
+  [[nodiscard]] auto
+  toLocal(const std::unordered_map<const Atom*, Location>& atomsLocations,
+          const double /* unused */) const
+      -> std::unique_ptr<LocalOp> override {
```
burgholzer, line 46: Any reason for not moving these functions to a source file?

include/mqt-core/na/operations/GlobalRYOp.hpp
burgholzer: probably also wrong in th other comments 🙃

include/mqt-core/na/operations/LocalCZOp.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <array>
+#include <string>
+#include <vector>
+
+namespace na {
+/// Represents a local RZ operation in the NAComputation.
+class LocalCZOp final : public LocalOp {
+public:
+  /// Creates a new RZ operation with the given atoms and angle.
```
burgholzer:
```suggestion
  /// Creates a new RZ operation with the given atoms.
```
CZ has no angle

burgholzer: please also fix in the remainder of this file.

include/mqt-core/na/operations/LocalHOp.hpp
```diff
@@ -0,0 +1,36 @@
...
+  /// Creates a new RZ operation with the given atoms and angle.
+  /// @param atom The atoms the operation is applied to.
+  explicit LocalHOp(const std::vector<const Atom*>& atom) : LocalOp(atom, {}) {
+    name_ = "h";
+  }
+
+  /// Creates a new RZ operation with the given atom and angle.
+  /// @param atom The atom the operation is applied to.
+  explicit LocalHOp(const Atom& atom) : LocalHOp({&atom}) {}
+};
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/operations/LocalRXOp.hpp
```diff
@@ -0,0 +1,40 @@
...
+  /// @param angle The angle of the operation.
+  LocalRXOp(const std::vector<const Atom*>& atom, const qc::fp angle)
+      : LocalOp(atom, {angle}) {
+    name_ = "rx";
+  }
+
+  /// Creates a new RZ operation with the given atom and angle.
+  /// @param atom The atom the operation is applied to.
+  /// @param angle The angle of the operation.
+  LocalRXOp(const Atom& atom, const qc::fp angle) : LocalRXOp({&atom}, angle) {}
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/operations/LocalRYOp.hpp
```diff
@@ -0,0 +1,40 @@
...
+  /// @param atom The atoms the operation is applied to.
+  /// @param angle The angle of the operation.
+  LocalRYOp(const std::vector<const Atom*>& atom, const qc::fp angle)
+      : LocalOp(atom, {angle}) {
+    name_ = "ry";
+  }
+
+  /// Creates a new RZ operation with the given atom and angle.
+  /// @param atom The atom the operation is applied to.
+  /// @param angle The angle of the operation.
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/operations/LocalCZOp.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright (c) 2025 Chair for Design Automation, TUM
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+/** @file
+ * @brief Defines a class for representing local RZ operations.
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/operations/GlobalCZOp.hpp
```diff
@@ -13,15 +13,55 @@
...
+        affectedAtoms.emplace(loc, atom);
+      }
+    }
+    std::vector<std::array<const Atom*, 2>> atomPairs;
+    for (auto it1 = affectedAtoms.cbegin(); it1 != affectedAtoms.cend();
+         ++it1) {
+      const auto& [loc1, atom1] = *it1;
+      for (auto it2 = std::next(it1); it2 != affectedAtoms.cend(); ++it2) {
+        const auto& [loc2, atom2] = *it2;
+        if ((loc1 - loc2).length() <= rydbergRange) {
```
burgholzer: The `Location` class might benefit from a `distance` method if that is not already implemented.

include/mqt-core/na/operations/LocalXOp.hpp
```diff
@@ -0,0 +1,36 @@
...
+public:
+  /// Creates a new RZ operation with the given atoms and angle.
+  /// @param atom The atoms the operation is applied to.
+  explicit LocalXOp(const std::vector<const Atom*>& atom) : LocalOp(atom, {}) {
+    name_ = "x";
+  }
+
+  /// Creates a new RZ operation with the given atom and angle.
+  /// @param atom The atom the operation is applied to.
+  explicit LocalXOp(const Atom& atom) : LocalXOp({&atom}) {}
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/operations/LocalYOp.hpp
```diff
@@ -0,0 +1,36 @@
...
+  /// Creates a new RZ operation with the given atoms and angle.
+  /// @param atom The atoms the operation is applied to.
+  explicit LocalYOp(const std::vector<const Atom*>& atom) : LocalOp(atom, {}) {
+    name_ = "y";
+  }
+
+  /// Creates a new RZ operation with the given atom and angle.
+  /// @param atom The atom the operation is applied to.
+  explicit LocalYOp(const Atom& atom) : LocalYOp({&atom}) {}
+};
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/operations/LocalZOp.hpp
```diff
@@ -0,0 +1,36 @@
...
+public:
+  /// Creates a new RZ operation with the given atoms and angle.
+  /// @param atom The atoms the operation is applied to.
+  explicit LocalZOp(const std::vector<const Atom*>& atom) : LocalOp(atom, {}) {
+    name_ = "z";
+  }
+
+  /// Creates a new RZ operation with the given atom and angle.
+  /// @param atom The atom the operation is applied to.
+  explicit LocalZOp(const Atom& atom) : LocalZOp({&atom}) {}
```
burgholzer: C&P mistake in the docstrings.

include/mqt-core/na/NAComputation.hpp
```diff
@@ -178,5 +180,11 @@ class NAComputation final {
   /// @returns a pair of a Boolean indicating whether the NAComputation is valid
   /// and a string containing the error message if the NAComputation is invalid.
   [[nodiscard]] auto validate() const -> std::pair<bool, std::string>;
+
+  /// Calculates the position of each atom at every stage and replaces global
+  /// gates by local gates with the atoms that are affected by the global gate.
+  /// @param rydbergRange The range of the Rydberg interaction.
```
burgholzer: Similar to the other comment: "Rydberg Radius"?

test/na/test_nacomputation.cpp
```diff
@@ -39,6 +48,13 @@ TEST(NAComputation, Zone) {
   ss << zone;
   EXPECT_EQ(ss.str(), "zone");
 }
+TEST(NAComputation, ZonesExtent) {
+  const auto zone = Zone("zone", {0, 0, 2, 2});
+  EXPECT_TRUE(zone.contains({1., 1.}));
```
burgholzer, line 53: Could be nice to have a convenience function that simply takes two doubles instead of a location for `contains`

burgholzer (changes requested):
Hey @ystade 👋🏼 

As requested, I took a look through the changes here.
I can't really help it, but I feel this addition to the code is rather rushed and makes some not-so-lightweight decisions.
A lot of the code comments are wrong because of copy-paste mistakes. You'll find some highlighted in the inline comments.

Furthermore, this PR decides to turn "local" operations, which have exclusively been single-atom operations, into vectors of potentially multi-atom operations.
This does not quite feel right, as it feels like it is mixing concepts from zoned-based NA architectures with non-zoned ones.
It is also quite a fundamental change of the model of thinking about the operations.
I wouldn't want us to make such a change lightly.
It also feels slightly weird that there are now local and global versions of certain gates, but not for others. For example, this PR adds local RX and RY, but it does not add GlobalRX and GlobalRZ.

If this was all done for the greater good and with a master plan in mind, let's move this along. I won't block this after the inline comments have been addressed.
However, if this was indeed rushed and not fully thought through, I'd rather take a step back and reconsider the options here.

ystade:
Thanks for the quick review, and I'm sorry for the many copy and paste errors. That's the danger of this class hierarchy with separate classes for each operation—as much as I like it.

### Local Operations
I see the `NAComputation` as a representation of general NA computations. With that said, the `NAComputation`does indeed mix concepts from both architectures. For the non-zoned ones, local gates on multiple qubits do exist. However, they are still local because they only execute *one* gate not gates on all atoms within a zone. There is no particular reason for the missing `GlobalRX` and `GlobalRY`—they are just not needed yet for any tool using the `NAComputation`.

### Merge with QuantumComputation
Apart from this PR, we started the discussion on whether it makes sense to merge `NAComputation` into the existing `QuantumComputation`. As it stands right now, I would still argue that it makes sense to keep the concepts separate even though I see the duplication of code here. There are a few fundamental differences, that I want to list in the following. Most of the points are a consequence of that the `QuantumComputation` was primarily designed with SC and qiskit in mind.
- SC does not have the concept of zoned (global) operations. Hence, they are missing in the `QuantumComputation`. Even though this is not a deal breaker to add those, it may add additional checks to existing tools, whether a given `QuantumComputation` can be processed or not.
- The same holds for all shuttling operations. Again, even though it is not a deal breaker to add those to the `QuantumComputation` (which actually exist somewhere already created by Ludwig), this causes some changes in the infrastructure. For example, the standard output format of a `QuantumComputation` is QASM. However, QASM does not support shuttling operations. Me personally, I would be quite hesitant to add some custom syntax to QASM for shuttling operations as seems not the right way to be to augment something that is tailored to SC for NA. This is mainly related to the next point.
- As a consequence of the above, atoms (qubits) do not have a fixed position. Even though it is still possible to index the existing atoms from 0 to n as one does for SC, at least some operations need coordinates as input. I have the feeling that this does not fit very well into the QASM format. One reason for this might be, that we definitely want to move multiple atoms at the same time. QASM does not offer a good way to express the scheduling of operations, e.g., to express that a group of operations is executed simultaneously.
- The main purpose of #846 was to make `NAComputation` output the new format that [mqt-naviz](https::/github.com/cda-tum/mqt-naviz) understands. I would very much like to keep this feature. This means when merging both representations, `QuantumComputation` requires a new output method for the `.naviz` format.

Overall, I think everything is possible and there is no fundamental obstacle of merging both representations. It just generates some work to merge the different concepts. Additionally, one must accept that one merges the concepts of computations meant for SC and for NA.

burgholzer:
First of all: I extracted the discussion about a potential merger of the `QuantumComputation` and the `NAComputation` into a separate thread here #875. Since this is a bigger topic in general, I'd like to keep that discussion separate from the PR here, which we should be integrating rather quickly, as you pointed out.

As for the PR here: I am fine with moving this along the way it is right now and deferring further and bigger changes to a point where #875 has been thoroughly discussed.
However, I can't really help it and I am really unsure whether the changes to the `LocalOp` class here really make that much sense. In my opinion, a global operation should be broken down into multiple local operations that each represent a single atomic operation. If some kind of grouping is desirable, then a concept similar to the `CompoundOperation` from the `QuantumComputation` hierarchy seems suitable. This could also be flattened on demand.
Quite some of these concepts already exist in the `QuantumComputation` hierarchy (including all kinds of AOD operations).

Anyway, at the very least, the docstrings here need to be updated and I would propose to add the two other global gates just for completeness sake.

ystade:
Closing this for now as not planned to be continued.

PR #879, opened 2025-03-19
🚨 Run clang-tidy on MLIR subproject, switch to `LLVM@20`, and update to `CMake >= 3.24`

burgholzer:
## Description

This pull request enables running `clang-tidy` for the MLIR subproject and updates the project to use LLVM version 20. This enhances code quality checks and aligns the project with the latest LLVM updates. It is also the same version currently shipped with CLion.

In addition, this PR raises the minimum supported CMake version to 3.24, which brings in several improvements:
- Better FetchContent functionality (no more backwords-compatible workarounds)
- Support for file sets which can be used to automatically install header files associated with a target.
- [CMAKE_VERIFY_INTERFACE_HEADER_SETS](https://cmake.org/cmake/help/latest/variable/CMAKE_VERIFY_INTERFACE_HEADER_SETS.html#variable:CMAKE_VERIFY_INTERFACE_HEADER_SETS) to enable build rules that verify all headers in header sets can be used on their own.

## Checklist:


burgholzer:
@flowerthrower @ystade just venting my frustration here. Of course the LLVM set up would be making problems with clang-tidy.. it's not like that is actually part of the same monorepository.. 
I'll play around with this a little more and try to make it not complain, but at some point I am just going to put `// IWYU pragma: ignore` etc. in the respective parts and merge this to move it forward.

burgholzer:
Alright, after the revelation that the project needs to be built in order for the MLIR files in the build directory to be generated, this seems to work now. I guess we will see in the upcoming MLIR PRs.
(I am still wondering if there is a way to more cleverly include files in various places so that the `// IWYU pragma: keep` statements are not necessary when the includes seem to be absolutely necessary for template instantiations.)

I am disregarding the patch coverage here, as these are mostly code paths that should never be exercised and merely exist out of caution.

PR #880, opened 2025-03-19
MLIR Conversion(s)

flowerthrower:
## Description

This PR originates from [this mqt-mlir PR](https://github.com/cda-tum/mqt-mlir/pull/8) and introduces conversions between Catalysts quantum dialect and the MQTOpt dialect.

## Checklist:


burgholzer:
Now that #879 is merged, this needs to be rebased

CMakeLists.txt
```diff
@@ -67,10 +67,12 @@ get_mqt_core_version()
 
 project(
   mqt-core
-  LANGUAGES CXX
+  LANGUAGES CXX C
   VERSION ${MQT_CORE_VERSION}
   DESCRIPTION "MQT Core - The Backbone of the Munich Quantum Toolkit")
 
+# to prevent `Targets Defined: nlohmann_json` error, but should eventually go in mlir/CMakeLists.txt
+find_package(Catalyst REQUIRED CONFIG)
```
burgholzer: Just putting this comment here so that this does not get overlooked.
At the moment, the whole CI pipeline fails because of this.

cmake/ExternalDependencies.cmake
```diff
@@ -36,7 +36,7 @@ set(JSON_SystemInclude
     CACHE INTERNAL "Treat the library headers like system headers")
 cmake_dependent_option(JSON_Install "Install nlohmann_json library" ON "MQT_CORE_INSTALL" OFF)
 FetchContent_Declare(nlohmann_json URL ${JSON_URL} FIND_PACKAGE_ARGS ${JSON_VERSION})
-list(APPEND FETCH_PACKAGES nlohmann_json)
+# list(APPEND FETCH_PACKAGES nlohmann_json)
```
burgholzer: This should also be reverted.
We definitely use the nlohmann_json package and it needs to be available.
If a line has to change, then it's the one a couple lines above where we force the install when we install mqt-core.
However, I'd also expect that something in catalyst has to change for all of this to work properly.
Maybe we should consider vendoring nlohmann_json as part of mqt-core.
Would solve a couple of issues.

include/mqt-core/dd/statistics/MemoryManagerStatistics.hpp
burgholzer, line 1: Should be reverted before merging. Also holds for all other occurrences.

src/dd/CMakeLists.txt
burgholzer, line 1: This should also be reverted. This target is bound to be defined.

mlir/CMakeLists.txt
```diff
@@ -10,12 +10,17 @@ set(MQT_MLIR_INCLUDE_BUILD_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
 # MLIR must be installed on the system
 find_package(MLIR REQUIRED CONFIG)
+find_package(Catalyst REQUIRED CONFIG)
```
burgholzer: Can we scope this a little tighter?
I am a bit hesitant to add a permanent dependency on catalyst for all MLIR activities, which, at some point, might mean all of mqt-core.
Especially since I very carefully removed the dependency on Qiskit a while back and turned that into an optional dependency.

It should be possible to only depend on catalyst whenever one wants to build the Catalyst conversion passes and wants to enable that functionality.
I am fairly sure there are some common best practices for that out there.

mlir/include/mlir/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2025 Chair for Design Automation, TUM
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+add_subdirectory(MQTOptToQuantum)
+add_subdirectory(QuantumToMQTOpt)
```
burgholzer: I find it irritating that the MQT dialect has a prefix but the Catalyst dialect does not have one.
If we, in the future, add further conversion passes, I believe this will quickly become confusing.
What about adding a "Catalyst" prefix in all relevant naming?

mlir/include/mlir/Conversion/MQTOptToQuantum/MQTOptToQuantum.h
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2025 Chair for Design Automation, TUM
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#ifndef LIB_CONVERSION_MQTOPTTOQUANTUM_MQTOPTTOQUANTUM_H_
+#define LIB_CONVERSION_MQTOPTTOQUANTUM_MQTOPTTOQUANTUM_H_
```
burgholzer: We typically use 
```
#pragma: once
```

Instead of header guards.
I'd strive for consistency here.

(Comment holds for all header files that we have control over)

mlir/include/mlir/Conversion/MQTOptToQuantum/MQTOptToQuantum.td
```diff
@@ -0,0 +1,33 @@
...
+//
+// Licensed under the MIT License
+
+//===-- Passes.td - Arith pass definition file --------*- tablegen -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
burgholzer: This seems to be copy pasted and does not belong here. Please also check in other files.

mlir/include/mlir/Conversion/MQTOptToQuantum/MQTOptToQuantum.td
```diff
@@ -0,0 +1,33 @@
...
+#ifndef LIB_CONVERSION_MQTOPTTOQUANTUM
+#define LIB_CONVERSION_MQTOPTTOQUANTUM
+
+include "mlir/Pass/PassBase.td"
+
+def MQTOptToQuantum : Pass<"mqtopt-to-quantum"> {
+  let summary = "Lower `mqtopt` to `quantum` dialect.";
+
+  let description = [{
+    This pass lowers the `mqtopt` to `quantum` dialect.
```
burgholzer: It could be worth extending this description slightly to cover more than the summary is already covering.
Same holds for the other conversion pass.

mlir/lib/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2025 Chair for Design Automation, TUM
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+# Navigate up five levels using relative path notation
+get_filename_component(CATALYST_BASE_DIR "${MLIR_DIR}/../../../../.." ABSOLUTE)
```
burgholzer: This feels wrong on so many levels. There has to be a better way to get catalyst set up properly.

mlir/lib/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# Define the MLIR include directory directly
+set(MLIRQUANTUM_INCLUDE_DIR "${CATALYST_BASE_DIR}/mlir/include")
+
+# Global settings for compiler flags
+add_compile_options(-fexceptions)
```
burgholzer: I have seen this before in our MLIR code base. Are you sure this is not redundant here?

mlir/lib/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+# Add include directory for MLIRQuantum headers
+include_directories(${MLIRQUANTUM_INCLUDE_DIR})
+
+# Optionally, if using TableGen-generated headers, include build/include (skip if not depending on
+# generated .inc files)
+include_directories("${CATALYST_BASE_DIR}/build/include")
```
burgholzer: This also doesn't quite feel right.
Eventually, this should work as a plugin for an installed version of catalyst. There is no way that we have access to a build directory at that point.
Do you currently need this line?

mlir/lib/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# Navigate up five levels using relative path notation
+get_filename_component(CATALYST_BASE_DIR "${MLIR_DIR}/../../../../.." ABSOLUTE)
+# Define the MLIR include directory directly
+set(MLIRQUANTUM_INCLUDE_DIR "${CATALYST_BASE_DIR}/mlir/include")
+
+# Global settings for compiler flags
+add_compile_options(-fexceptions)
+
+# Add include directory for MLIRQuantum headers
+include_directories(${MLIRQUANTUM_INCLUDE_DIR})
```
burgholzer: I'd really hope that at some point we can eliminate the global includes.
We explicitly know the targets that these headers are needed in. So we might as well add them on a target basis.

mlir/lib/Conversion/CMakeLists.txt
burgholzer, line 1: Can we maybe nest this one level deeper in some Catalyst specific directory? Might help with isolation as well.

mlir/lib/Conversion/MQTOptToQuantum/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2025 Chair for Design Automation, TUM
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+file(GLOB CONVERSION_SOURCES *.cpp)
+
+set(LIBRARIES ${conversion_libs} MLIRIR MLIRPass MQT::CoreIR ${MLIRQuantum_LIB})
```
burgholzer: Are MLIRIR and MLIRPass explicitly needed here? What is in `conversion_libs`?

mlir/tools/quantum-opt/CMakeLists.txt
```diff
@@ -15,7 +15,10 @@ set(LIBS
     ${transform_libs}
     ${passes_libs}
     MLIROptLib
-    MLIRMQTOptTransforms)
+    MLIRMQTOptTransforms
+    MQTOptToQuantum
+    QuantumToMQTOpt
+    MLIRQuantum)
```
burgholzer: I think this has to be slightly adjusted for the fact that Catalyst might not always be available and will become an optional dependency. So also these link libraries should be optional.

mlir/tools/quantum-opt/quantum-opt.cpp
```diff
@@ -7,8 +7,18 @@
...
+#include "mlir/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.h"
+#include "mlir/Dialect/Func/Extensions/AllExtensions.h"
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h"
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h" // IWYU pragma: keep
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h"
 #include "mlir/Dialect/MQTOpt/Transforms/Passes.h" // IWYU pragma: keep
+#include "mlir/IR/DialectRegistry.h"
+#include "mlir/InitAllDialects.h"
+#include "mlir/InitAllPasses.h"
+#include "mlir/Tools/mlir-opt/MlirOptMain.h"
```
burgholzer: Are all of these includes necessary? Seems a bit overboard. Especially since some, such as the DialectLibrary, are already included below.
I am fairly sure clang-tidy will scream at you once this builds 🙃

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
```diff
@@ -0,0 +1,476 @@
...
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/OperationSupport.h"
+#include "mlir/IR/TypeRange.h"
+#include "mlir/IR/Value.h"
+#include "mlir/IR/ValueRange.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/raw_ostream.h"
```
burgholzer: Every include that is not from our own library should use `<...>` for the includes.

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+    // Get the result of the new operation, which represents the qubit register
+    auto trgtQreg = mqtoptOp->getResult(0);
+
+    // Collect the users of the original operation to update their operands
+    std::vector<mlir::Operation*> users(op->getUsers().begin(),
+                                        op->getUsers().end());
+
+    // Iterate over the users in reverse order
+    for (auto* user : llvm::reverse(users)) {
```
burgholzer: Can't you just refactor this to a loop that starts with
```
const auto& users = op->getUsers();
for(auto it=users.rbegin()); it != users.rend(); ++it) { ...
```
Then there is no need to construct a temporary object at all.
Similar reasoning can be applied to other parts of the code.

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
```diff
@@ -0,0 +1,476 @@
...
+    });
+  }
+};
+
+struct ConvertQuantumAlloc
+    : public OpConversionPattern<catalyst::quantum::AllocOp> {
+
+  ConvertQuantumAlloc(const TypeConverter& typeConverter, MLIRContext* context)
+      : OpConversionPattern<catalyst::quantum::AllocOp>(typeConverter,
+                                                        context) {}
```
burgholzer: Maybe I am stupid, and it does not really Safe much, but I think you can inherit this constructor from
The base class with a single
`using ...` where the dots are the constructor method name.
Also holds for all other classes here.

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+  LogicalResult
+  matchAndRewrite(catalyst::quantum::DeallocOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter& rewriter) const override {
+
+    // Extract operand(s)
+    auto qregValue = adaptor.getQreg();
+
+    // Prepare the result type(s)
+    auto resultTypes = ::mlir::TypeRange({});
```
burgholzer: Seems redundant. Can't this simply be inlined below with `{}` for the function parameter?

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+    auto staticParams = ::mlir::DenseF64ArrayAttr();
+    auto paramsMask = ::mlir::DenseBoolArrayAttr();
+
+    if (gateName == "CNOT" || gateName == "CY" || gateName == "CZ" ||
+        gateName == "CRX" || gateName == "CRY" || gateName == "CRZ" ||
+        gateName == "ControlledPhaseShift") {
+
+      assert(inQubitsVec.size() == 2 && "Expected 1 control + 1 target qubit");
+      inCtrlQubitsVec.push_back(inQubitsVec[0]);
```
burgholzer: Prefer to use `emplace_back` over `push_back` in most cases.

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
```diff
@@ -0,0 +1,476 @@
...
+    populateCallOpTypeConversionPattern(patterns, typeConverter);
+    target.addDynamicallyLegalOp<func::CallOp>(
+        [&](func::CallOp op) { return typeConverter.isLegal(op); });
+
+    populateBranchOpInterfaceTypeConversionPattern(patterns, typeConverter);
+    target.markUnknownOpDynamicallyLegal([&](Operation* op) {
+      return isNotBranchOpInterfaceOrReturnLikeOp(op) ||
+             isLegalForBranchOpInterfaceTypeConversionPattern(op,
+                                                              typeConverter) ||
+             isLegalForReturnOpTypeConversionPattern(op, typeConverter);
```
burgholzer: To be fair, I have no idea what these lines are doing. Maybe an indication that this could use some more documentation. (Or that I need sleep 😜)

mlir/lib/Conversion/MQTOptToQuantum/MQTOptToQuantum.cpp
```diff
@@ -0,0 +1,534 @@
...
+    llvm::SmallVector<mlir::Value> i1Values;
+    if (!inCtrlQubits.empty()) {
+      mlir::Type i1Type = rewriter.getI1Type();
+      auto constTrue = rewriter.create<mlir::arith::ConstantOp>(
+          op.getLoc(), i1Type, rewriter.getBoolAttr(true));
+      for (size_t i = 0; i < inCtrlQubits.size(); ++i) {
+        i1Values.push_back(constTrue.getResult());
+      }
+    }
+    mlir::ValueRange inCtrlQubitsValues(i1Values);
```
burgholzer: This indeed feels ugly. There should be better solutions for that, right?

mlir/lib/Conversion/MQTOptToQuantum/MQTOptToQuantum.cpp
```diff
@@ -0,0 +1,534 @@
...
+    patterns.add<ConvertMQTOptSimpleGate<::mqt::ir::opt::RYOp>>(typeConverter,
+                                                                context);
+    patterns.add<ConvertMQTOptSimpleGate<::mqt::ir::opt::RZOp>>(typeConverter,
+                                                                context);
+
+    patterns.add<ConvertMQTOptSimpleGate<::mqt::ir::opt::HOp>>(typeConverter,
+                                                               context);
+    patterns.add<ConvertMQTOptSimpleGate<::mqt::ir::opt::SWAPOp>>(typeConverter,
+                                                                  context);
+    patterns.add<ConvertMQTOptSimpleGate<::mqt::ir::opt::POp>>(typeConverter,
```
burgholzer: What about operations that MQT supports but catalyst does not?
There are quite some gates that could be supported. S, T, SX, their inverses all have equivalents in terms of rotation gates. Potentially even more would make sense.

burgholzer (changes requested):
Hey @flowerthrower 👋🏻 

Many thanks for all your work on this.
I just managed to do a first very rough pass over everything in here. Take everything with a grain of salt please. The review was mostly written on my phone while navigating through the city of Hannover 😅
As you probably expected, most of my comments focus on the general setup of the catalyst connection and the structure of how we neatly integrate that into the project.
The actual functionality of this PR is already 95% fine. I only had a couple minor code comments.

I hope most of the comments make sense. Let me know if some of them don't.
It will probably take us a bit to get this integration as nice as we want it.
We should reach a point though, where CI succeeds again on this branch 😌

mlir/lib/Conversion/QuantumToMQTOpt/QuantumToMQTOpt.cpp
flowerthrower: It seems like these reverse iterators are only available on standard containers (like std::vector, std::list, etc.) or if the range type implements them explicitly, which `llvm::iterator_range` unfortunately does not.

mlir/lib/Conversion/MQTOptToQuantum/MQTOptToQuantum.cpp
flowerthrower: Perhaps @BertiFlorea can add a few more (incl. tests) once this is merged?

BertiFlorea: Noted. 👍

mlir/CMakeLists.txt
```diff
@@ -26,6 +26,20 @@ string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
 list(GET MLIR_VERSION_COMPONENTS 0 MLIR_VERSION_MAJOR)
 add_compile_definitions(MLIR_VERSION_MAJOR=${MLIR_VERSION_MAJOR})
 
+string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
+list(GET MLIR_VERSION_COMPONENTS 0 MLIR_VERSION_MAJOR)
+add_compile_definitions(MLIR_VERSION_MAJOR=${MLIR_VERSION_MAJOR})
+
```
burgholzer: Merge artefact
```suggestion
```

mlir/CMakeLists.txt
```diff
@@ -26,6 +26,20 @@ string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
...
+option(ENABLE_CATALYST "Enable Catalyst conversion passes" OFF)
+if(ENABLE_CATALYST)
+  if(NOT MLIR_VERSION_MAJOR EQUAL 19)
+    message(FATAL_ERROR "Catalyst currently requires MLIR 19.x")
+  endif()
```
burgholzer: I am still trying to understand if this is a hard requirement.
I'd like to see this in action in CI and see how/if/when it fails.

mlir/CMakeLists.txt
```diff
@@ -26,6 +26,20 @@ string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
...
+  find_package(Catalyst REQUIRED CONFIG)
+  list(APPEND CMAKE_MODULE_PATH "${CATALYST_CMAKE_DIR}")
+  set(MLIR_INCLUDE_DIRS "${MLIR_INCLUDE_DIRS};${CATALYST_INCLUDE_DIRS}")
```
burgholzer:
```suggestion
  list(APPEND MLIR_INCLUDE_DIRS "${CATALYST_INCLUDE_DIRS}")
```
Might as well use consistent syntax

mlir/CMakeLists.txt
```diff
@@ -26,6 +26,20 @@ string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
 list(GET MLIR_VERSION_COMPONENTS 0 MLIR_VERSION_MAJOR)
 add_compile_definitions(MLIR_VERSION_MAJOR=${MLIR_VERSION_MAJOR})
 
+string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
+list(GET MLIR_VERSION_COMPONENTS 0 MLIR_VERSION_MAJOR)
+add_compile_definitions(MLIR_VERSION_MAJOR=${MLIR_VERSION_MAJOR})
+
+option(ENABLE_CATALYST "Enable Catalyst conversion passes" OFF)
```
burgholzer: I think this option deserves a more specific name as it seems rather generic and MQT-independent at the moment. Maybe following the other options and calling this `ENABLE_MQT_CORE_MLIR_CATALYST_PLUGIN`?

mlir/include/mlir/Conversion/Catalyst/CatalystQuantumToMQTOpt/CatalystQuantumToMQTOpt.h
```diff
@@ -0,0 +1,23 @@
...
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#pragma once
+
+#include "mlir/Pass/Pass.h" // from @llvm-project
```
burgholzer: Since this is not a local include, this should be
```suggestion
#include <mlir/Pass/Pass.h> // from @llvm-project
```

mlir/include/mlir/Conversion/Catalyst/CatalystQuantumToMQTOpt/CatalystQuantumToMQTOpt.h
burgholzer, line 1: Is it best practice to have the directory and the header/source files have the same name? Seems a bit redundant to me.

mlir/include/mlir/Conversion/Catalyst/CatalystQuantumToMQTOpt/CatalystQuantumToMQTOpt.td
```diff
@@ -0,0 +1,32 @@
...
+// All rights reserved.
+//
+// SPDX-License-Identifier: MIT
+//
+// Licensed under the MIT License
+
+include "mlir/Pass/PassBase.td"
+
+def CatalystQuantumToMQTOpt : Pass<"quantum-to-mqtopt"> {
+  let summary = "Lower `quantum` to `mqtopt` dialect.";
```
burgholzer: Nitpick, but technically this is not really a "lowering" right? it is a conversion between two equally powerful Quantum dialects on the same abstraction level.

mlir/include/mlir/Conversion/Catalyst/CatalystQuantumToMQTOpt/CatalystQuantumToMQTOpt.td
```diff
@@ -0,0 +1,32 @@
...
+// Copyright (c) 2025 Munich Quantum Software Company GmbH
+// All rights reserved.
+//
+// SPDX-License-Identifier: MIT
+//
+// Licensed under the MIT License
+
+include "mlir/Pass/PassBase.td"
+
+def CatalystQuantumToMQTOpt : Pass<"quantum-to-mqtopt"> {
```
burgholzer: Another nitpick: In most places the code explicitly says `CatalystQuantum` in one form or another. Here, the pass name does not include `catalyst`. It probably should for consistency.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+// All rights reserved.
+//
+// SPDX-License-Identifier: MIT
+//
+// Licensed under the MIT License
+
+include "mlir/Pass/PassBase.td"
+
+def MQTOptToCatalystQuantum : Pass<"mqtopt-to-quantum"> {
+  let summary = "Convert `mqtopt` to `quantum` dialect.";
```
burgholzer: I like this better than the version in the other conversion pass that I commented on. However, this one is missing the `Catalyst` reference in the summary.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+
+  let description = [{
+    This pass converts the `mqtopt` to Catalyst's `quantum` dialect.
+    The following operations are currently NOT converted (and instead marked legal):
+    - IOp
+    - GPhaseOp
```
burgholzer: Catalyst has a `GlobalPhaseOp`: https://github.com/PennyLaneAI/catalyst/blob/e89b83d393be158be499a95ce02b1312e7e89bbd/mlir/include/Quantum/IR/QuantumOps.td#L484

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+
+include "mlir/Pass/PassBase.td"
+
+def MQTOptToCatalystQuantum : Pass<"mqtopt-to-quantum"> {
+  let summary = "Convert `mqtopt` to `quantum` dialect.";
+
+  let description = [{
+    This pass converts the `mqtopt` to Catalyst's `quantum` dialect.
+    The following operations are currently NOT converted (and instead marked legal):
+    - IOp
```
burgholzer: The identity gate is mapped to the Catalyst name "Identity". So this is definitely supported for conversion

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+    - GPhaseOp
+    - BarrierOp
+    - SOp, SdgOp
+    - TOp, TdgOp
```
burgholzer: Both `S` and `T` are supported as gates in Catalyst. So are there inverses via the respective modifier.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+    - VOp, VdgOp
+    - UOp, U2Op
+    - SXOp, SXdgOp
+    - iSWAPOp, iSWAPdgOp
```
burgholzer: `ISWAP` and, consequentially, its inverse are also directly supported in Catalyst.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+    - PeresOp, PeresdgOp
+    - DCXOp
+    - ECROp
+    - RXXOp, RYYOp, RZZOp, RZXOp
```
burgholzer: According to https://github.com/PennyLaneAI/catalyst/blob/e89b83d393be158be499a95ce02b1312e7e89bbd/frontend/test/pytest/test_operations.py#L114-L124, at least RXX, RYY and RZZ are also supported.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+    This pass converts the `mqtopt` to Catalyst's `quantum` dialect.
+    The following operations are currently NOT converted (and instead marked legal):
+    - IOp
+    - GPhaseOp
+    - BarrierOp
+    - SOp, SdgOp
+    - TOp, TdgOp
+    - VOp, VdgOp
+    - UOp, U2Op
+    - SXOp, SXdgOp
```
burgholzer, line 18: This as well as the `ECR` gate are declared as not supported here https://github.com/PennyLaneAI/catalyst/blob/e89b83d393be158be499a95ce02b1312e7e89bbd/frontend/test/pytest/test_operations.py#L84-L85
It would be interesting to know, what is missing in order to enable the corresponding support.
If it turns out to be easy, it might be worth adding.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+    - iSWAPOp, iSWAPdgOp
+    - PeresOp, PeresdgOp
+    - DCXOp
```
burgholzer: Just to add another comment here: One could probably add support for these gates by specifying them as a unitary matrix, right?

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+
+def MQTOptToCatalystQuantum : Pass<"mqtopt-to-quantum"> {
+  let summary = "Convert `mqtopt` to `quantum` dialect.";
+
+  let description = [{
+    This pass converts the `mqtopt` to Catalyst's `quantum` dialect.
+    The following operations are currently NOT converted (and instead marked legal):
+    - IOp
+    - GPhaseOp
+    - BarrierOp
```
burgholzer: I am not quite sure we want to keep barriers around if we are translating between dialects. Ideally, we would want the conversion to be a true conversion in the sense that the whole circuit is translated.

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
```diff
@@ -0,0 +1,36 @@
...
+    - TOp, TdgOp
+    - VOp, VdgOp
+    - UOp, U2Op
+    - SXOp, SXdgOp
+    - iSWAPOp, iSWAPdgOp
+    - PeresOp, PeresdgOp
+    - DCXOp
+    - ECROp
+    - RXXOp, RYYOp, RZZOp, RZXOp
+    - XXminusYY, XXplusYY
```
burgholzer: One final comment here: For some of these gates, a single gate could be replaced by a decomposed version of the gate only using gates supported by the other dialect.
No one says that a roundtrip between dialects must be lossless.

mlir/include/mlir/Dialect/MQTOpt/IR/MQTOptDialect.h
```diff
@@ -10,7 +10,9 @@
 
 #pragma once
 
+#include <llvm/ADT/SmallVector.h>
 #include <mlir/Bytecode/BytecodeOpInterface.h>
+#include <mlir/Interfaces/SideEffectInterfaces.h>
```
burgholzer: These did not seem to be necessary before, why are they now?

mlir/lib/Dialect/MQTDyn/Transforms/ConstantFolding.cpp
```diff
@@ -18,6 +18,7 @@
 namespace mqt::ir::dyn {
 
 #define GEN_PASS_DEF_CONSTANTFOLDING
+#include "mlir/Dialect/Common/Compat.h"
```
burgholzer: This did not seem necessary before. Why now?
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInverses.cpp
```diff
@@ -18,8 +18,8 @@
 namespace mqt::ir::opt {
 
 #define GEN_PASS_DEF_CANCELCONSECUTIVEINVERSES
+#include "mlir/Dialect/Common/Compat.h"
```
burgholzer: This did not seem necessary before. Why now?
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPass.cpp
```diff
@@ -18,8 +18,8 @@
 namespace mqt::ir::opt {
 
 #define GEN_PASS_DEF_QUANTUMSINKPASS
+#include "mlir/Dialect/Common/Compat.h"
```
burgholzer: This did not seem necessary before. Why now?
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInverses.cpp
```diff
@@ -18,8 +18,8 @@
...
 #include "mlir/Dialect/MQTOpt/Transforms/Passes.h.inc"
-
```
burgholzer: stray whitespace change

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPass.cpp
```diff
@@ -18,8 +18,8 @@
...
 #include "mlir/Dialect/MQTOpt/Transforms/Passes.h.inc"
-
```
burgholzer: stray whitespace change

mlir/tools/quantum-opt/CMakeLists.txt
```diff
@@ -19,7 +19,18 @@ set(LIBS
     MLIRMQTOptTransforms
     MLIRMQTDynTransforms)
 
+option(ENABLE_CATALYST "Enable Catalyst conversion passes" ON)
```
burgholzer: Why is this option here again? there should only be one top-level option for enabling catalyst support.

mlir/tools/quantum-opt/quantum-opt.cpp
```diff
@@ -8,6 +8,14 @@
  * Licensed under the MIT License
  */
 
+// At the top of quantum-opt.cpp:
```
burgholzer: Seems like a redundant comment
```suggestion
```

mlir/test/Conversion/mqtopt.mlir
burgholzer, line 1: I know this will probably be painfull to write, but can we please
- split these tests up so that they are more fine-granular and test more specific things
- explicitly test all supported operations as well as non-supported operations and their conversion.

https://github.com/munich-quantum-toolkit/core/blob/main/mlir/test/Dialect/MQTOpt/Transforms/consecutive-inverses.mlir is a good blueprint for more extensive tests at the moment.

If this is enough of a pain point to find a better way to test the infrastructure, then so be it.
We are adding quite some functionality here, so it should also be tested appropriately.
Most likely, we should be collecting coverage metrics for the MLIR part of the project. (this will have to wait for another PR though)

burgholzer (changes requested):
Thanks @flowerthrower for your tireless work on this!
I still haven't managed to fully review the actual conversion passes, but made another detailed pass on the infrastructure surrounding the conversions.

Most of the feedback is inline already.
The main reason for clang-tidy to report somewhat bogus results is that it simply cannot find the catalyst headers, so it gets confused.
This also points to one of the things, that is _not_ an inline comment: We really want to be testing this integration in CI with an appropriate amount of tests.
The CI workflow should be set up to install Catalyst and successfully build the project. Otherwise, the confidence in the code working as expected will not be very high over time.
At some point, we want to distribute this integration to third parties, where this has to be seamless.

Ideally, we would add an optional python dependency group for catalyst, similar to qiskit (https://github.com/munich-quantum-toolkit/core/blob/79cb99f53440f951d7ba6e163474dd4ae7063b08/pyproject.toml#L49-L51) and would use Python whenever available to detect the installed version of Catalyst or to point `find_package` to the right direction. We do something pretty similar for mqt-core and pybind11 in most of our projects: https://github.com/cda-tum/mqt-qcec/blob/aa833928fcd47ee132075e90a301aa0277e7f476/cmake/ExternalDependencies.cmake#L6-L33

mlir/include/mlir/Conversion/Catalyst/CatalystQuantumToMQTOpt/CatalystQuantumToMQTOpt.h
flowerthrower: I agree, but that seems to be the case 🤷🏻‍♂️

mlir/CMakeLists.txt
```diff
@@ -26,6 +25,13 @@ string(REPLACE "." ";" MLIR_VERSION_COMPONENTS ${MLIR_VERSION})
 list(GET MLIR_VERSION_COMPONENTS 0 MLIR_VERSION_MAJOR)
```
ystade, line 26: Just because I stumbe over it right now: Is this gymnastic with the version string necessary or can the `VERSION_LESS` operator of cmake be used directly? That might make the code cleaner here.

burgholzer: Seems reasonable to use `VERSION_LESS` 👍🏼

.github/workflows/ci_catalyst.yml
```diff
@@ -0,0 +1,122 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # set up ccache for faster C++ builds
+      - name: Setup ccache
+        uses: Chocobo1/setup-ccache-action@v1
+        with:
+          prepend_symlinks_to_path: false
+          override_cache_key: c++-tests-mlir
```
burgholzer: fairly sure you need to use a different cache key here, otherwise there will be cache contention between the jobs.
```suggestion
          override_cache_key: c++-tests-mlir-catalyst
```
If we extend the matrix of CI workflows above, the cache key should also be updated.

.github/workflows/ci_catalyst.yml
```diff
@@ -0,0 +1,122 @@
...
+          echo "LLVM_DIR=$(brew --prefix llvm@19)/lib/cmake/llvm" >> $GITHUB_ENV
+          echo "$(brew --prefix llvm@19)/bin" >> $GITHUB_PATH
+
+      # clone PennyLane Catalyst (from patched branch) - excl. submodules
+      - name: Clone PennyLane Catalyst
+        run: |
+          git clone --branch mqt_integration --depth=1 https://github.com/flowerthrower/catalyst.git catalyst
+
+      - name: Set macOS deployment target for linker
+        run: echo "MACOSX_DEPLOYMENT_TARGET=14.0" >> $GITHUB_ENV
```
burgholzer, line 79: This should not be necessary as you are running on macOS 14 already.

burgholzer, line 79: This should not be necessary as you are running on macOS 14 already.

.github/workflows/ci_catalyst.yml
```diff
@@ -0,0 +1,122 @@
...
+            -DCMAKE_BUILD_TYPE=Release \
+            -DMLIR_DIR=$MLIR_DIR \
+            -DLLVM_DIR=$LLVM_DIR \
+            -DCATALYST_BUILD_TOOLS=OFF \
+            -DEnzyme_DIR="none"
+
+      # build MLIRQuantum
+      - name: Build MLIRQuantum dialect
+        run: |
+          ninja -C catalyst/mlir/build MLIRQuantum
```
burgholzer:
```suggestion
          cmake --build catalyst/mlir/build --target MLIRQuantum
```

.github/workflows/ci_catalyst.yml
```diff
@@ -0,0 +1,122 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  cpp-test-catalyst:
+    name: 🇨‌ Test MLIR with LLVM@${{ matrix.llvm-version }}
+    runs-on: macos-14
+    strategy:
+      matrix:
+        llvm-version: [19]
```
burgholzer, line 25: Would be interesting to see this work on `macos-13` (x86 runners) as well as `llvm-version: 20`.
The current build pipeline seems to pass, which is great already!

Unrelated to this PR in particular, but what could be a nice byproduct in a separate PR is that this already represents a functional macOS-14 CI pipeline for MLIR, which would be part of #925.

cmake/ExternalDependencies.cmake
```diff
@@ -73,3 +73,23 @@ endif()
...
+  message(STATUS "Found Catalyst package: ${Python_Catalyst_DIR}")
+
+  # TODO: once the Catalyst Python package provides the necessary files set(Catalyst_DIR
+  # "${Python_Catalyst_DIR}/mlir/build/lib/cmake/catalyst")
+  if(Catalyst_DIR)
+    list(APPEND CMAKE_PREFIX_PATH "${Catalyst_DIR}")
+    find_package(Catalyst REQUIRED)
+  else()
+    message(STATUS "Could not detect Catalyst CMake directory")
+  endif()
```
burgholzer: This needs to fail hard if Catalyst cannot be found.
```suggestion
  find_package(Catalyst REQUIRED)
```

cmake/ExternalDependencies.cmake
```diff
@@ -73,3 +73,23 @@ endif()
...
+  execute_process(
+    COMMAND "${Python_EXECUTABLE}" -c
+            "import catalyst, os; print(os.path.dirname(catalyst.__file__))"
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    OUTPUT_VARIABLE Python_Catalyst_DIR
+    ERROR_QUIET)
+  message(STATUS "Found Catalyst package: ${Python_Catalyst_DIR}")
+
+  # TODO: once the Catalyst Python package provides the necessary files set(Catalyst_DIR
+  # "${Python_Catalyst_DIR}/mlir/build/lib/cmake/catalyst")
```
burgholzer, line 88: Ideally, there would be no need to do something like `set(Catalyst_DIR "${Python_Catalyst_DIR}/...`, but `Catalyst_DIR` would already be a result of the `execute_process` call.
Either catalyst provides an entry point for that (similar to pybind, nanobind, or mqt-core), or we parse this somewhat from output similar to the existing command here by a simple Python script.

cmake/ExternalDependencies.cmake
burgholzer: Once a compatible version of Catalyst is released, this will probably also require a version specifier to indicate the minimum version required.

mlir/CMakeLists.txt
```diff
@@ -12,9 +12,8 @@ set(MQT_MLIR_MIN_VERSION 19.0)
 
 # MLIR must be installed on the system
 find_package(MLIR REQUIRED CONFIG)
-if(MLIR_VERSION VERSION_LESS MQT_MLIR_MIN_VERSION)
-  message(FATAL_ERROR "MLIR version must be at least ${MQT_MLIR_MIN_VERSION}")
-endif()
+# if(MLIR_VERSION VERSION_LESS MQT_MLIR_MIN_VERSION) message(FATAL_ERROR "MLIR version must be at
+# least ${MQT_MLIR_MIN_VERSION} but found ${MLIR_VERSION}") endif()
```
burgholzer: Why was this commented out?
```suggestion
if(MLIR_VERSION VERSION_LESS MQT_MLIR_MIN_VERSION)
  message(FATAL_ERROR "MLIR version must be at least ${MQT_MLIR_MIN_VERSION}")
endif()
```

burgholzer (commented):
Just a couple of comments because the PR notifications where in my inbox 😌

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
BertiFlorea: Thanks for the reference, added in 2163bcc57605271c7f70d026cb2e529b52d94811 👍🏼

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
BertiFlorea: Added in 86e31d23fbb3dcfaac942d09949e85b2e8c41859 👍🏼

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
BertiFlorea: Added in 8a0e9cde2454b3d81eed029a82b8152325e40aef 👍🏼 
Support for inverses added in 0ce0025714a69659ad2c512decd4066b04e5d5f1 👍🏼

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
BertiFlorea: Added in 8a0e9cde2454b3d81eed029a82b8152325e40aef 👍🏼

mlir/include/mlir/Conversion/Catalyst/MQTOptToCatalystQuantum/MQTOptToCatalystQuantum.td
BertiFlorea: Support for inverse added in 0ce0025714a69659ad2c512decd4066b04e5d5f1 👍🏼

PR #885, opened 2025-03-24
✨ Collect clifford blocks

jannikpflieger:
## Description

Added the ability to separate the collection of Clifford and non Clifford Blocks by adding a new function parameter. For that a isCliffordCheck was added as well. The isClifford check is very rough but includes all common Clifford gates. However, it does not include rotation gates that with the right angle could be considered Clifford. It does not include custom gates that could be Clifford as well.

The separation is done by finalizing a block when we iterate over a non-Clifford Block. When we want to collect Clifford and non Clifford gates separately then the function handles non-Clifford gates as "cannot be processed" and the canProcess variable is set to false. This ensures that we still have large CliffordBlocks but we have only single non CliffordBlocks. Essentially separating the non-CliffordBlocks from all Clifford and other non-CliffordBlocks.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add Clifford block collection pass to `CircuitOptimizer` module ([#885]) ([**jannikpflieger**], [**@burgholzer**])
 - ✨ Add `isControlled()` method to the `UnitaryInterface` MLIR class ([#1157]) ([**@taminob**], [**@burgholzer**])
 - 📝 Integrate generated MLIR documentation ([#1147]) ([**@denialhaag**], [**@burgholzer**])
 - ✨ Add `IfElseOperation` to C++ library and Python package to support Qiskit's `IfElseOp` ([#1117]) ([**@denialhaag**], [**@burgholzer**], [**@lavanya-m-k**])
@@ -221,6 +222,7 @@ _📚 Refer to the [GitHub Release Notes](https://github.com/munich-quantum-tool
 [#893]: https://github.com/munich-quantum-toolkit/core/pull/893
 [#892]: https://github.com/munich-quantum-toolkit/core/pull/892
 [#886]: https://github.com/munich-quantum-toolkit/core/pull/886
+[#885]: https://github.com/munich-quantum-toolkit/core/pull/885
 [#883]: https://github.com/munich-quantum-toolkit/core/pull/883
 [#882]: https://github.com/munich-quantum-toolkit/core/pull/882
 [#879]: https://github.com/munich-quantum-toolkit/core/pull/879
@@ -270,6 +272,7 @@ _📚 Refer to the [GitHub Release Notes](https://github.com/munich-quantum-tool
 [**@li-mingbao**]: https://github.com/li-mingbao
 [**@lavanya-m-k**]: https://github.com/lavanya-m-k
 [**@taminob**]: https://github.com/taminob
+[**@jannikpflieger**]: https://github.com/jannikpflieger
 
 <!-- General links -->
 


include/mqt-core/circuit_optimizer/CircuitOptimizer.hpp
```diff
@@ -90,7 +90,8 @@ class CircuitOptimizer {
    * @param qc the quantum circuit
    * @param maxBlockSize the maximum size of a block
    */
-  static void collectBlocks(QuantumComputation& qc, std::size_t maxBlockSize);
+  static void collectBlocks(QuantumComputation& qc, std::size_t maxBlockSize,
+                            bool collectCliffords);
```
burgholzer:
```suggestion
                            bool onlyCollectCliffords = false);
```
Maybe that name would be a little more suitable?
Also, this parameter is missing a description in the corresponding docstring.
Finally, this should default to false, otherwise this would be a breaking change.

include/mqt-core/ir/operations/Operation.hpp
```diff
@@ -154,6 +154,33 @@ class Operation {
...
+    case SXdg:
+    case DCX:
+    case SWAP:
+    case iSWAP:
+    case ECR:
+      return true;
+    default:
+      return false;
+    }
+  }
```
burgholzer: This is not yet correct. As it stands, this would, e.g., classify a controlled-S gate as Clifford.
While the aspiration of this function is not to yield absolutely no false-negatives, it must not yield false positives.

- you also do not need the `this->` in the member function.
- the function here should actually simply return `false` and an overload of the method should be implemented in the `StandardOperation` class that contains this implementation. The `CompoundOperation` class will also need an overload that checks if every operation in the compound operation is Clifford. Other similar functions are already implemented in the respective classes and serve as a blueprint.
- while we are at it, would you mind adding a docstring to any method that you add to the code base?
- you can tweak the code slightly to improve the checks. Identity gates as well as barriers are supported with arbitrary numbers of qubits. thus, you should check for these first. Then maybe check for the gates that allow a single control (X, Y, Z) and then for the gates that must have no controls (the rest from the list above).

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1432,12 +1433,22 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
...
     }
 
+    // check if the operation is a Clifford operation
+    if (collectCliffords && op->isClifford()) {
+      isClifford = true;
+    }
+
+    if (collectCliffords && !isClifford) {
+      canProcess = false;
+    }
```
burgholzer:
```suggestion
    // check if the operation can be processed
    if (!op->isUnitary() || (collectCliffords && !op->isClifford()) {
      canProcess = false;
    }
```
much simpler

test/circuit_optimizer/test_collect_blocks.cpp
burgholzer, line 1: All the changes in this file where 
```
qc::CircuitOptimizer::collectBlocks(..., ..., false);
```
Should be reverted to remain unchanged by this PR based on the default argument for the pass.

test/circuit_optimizer/test_collect_blocks.cpp
```diff
@@ -16,15 +16,31 @@
...
 TEST(CollectBlocks, emptyCircuit) {
   QuantumComputation qc(1);
-  qc::CircuitOptimizer::collectBlocks(qc, 1);
+  qc::CircuitOptimizer::collectBlocks(qc, 1, false);
+  EXPECT_EQ(qc.getNindividualOps(), 0);
+}
+
+TEST(CollectBlocks, emptyCliffordCircuit) {
+  QuantumComputation qc(1);
+  qc::CircuitOptimizer::collectBlocks(qc, 1, true);
```
burgholzer: This is a pretty useless test. The new flag has no influence at all on empty circuits.

test/circuit_optimizer/test_collect_blocks.cpp
```diff
@@ -16,15 +16,31 @@
...
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 1);
+  EXPECT_TRUE(qc.front()->isStandardOperation());
+}
+
+TEST(CollectBlocks, singleCliffordGate) {
+  QuantumComputation qc(1);
+  qc.t(0);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectBlocks(qc, 1, true);
```
burgholzer: The T gate is not a Clifford gate. This seems to be a non-sense test. Please adjust to make this a reasonable test.

test/circuit_optimizer/test_collect_blocks.cpp
```diff
@@ -37,20 +53,63 @@ TEST(CollectBlocks, collectMultipleSingleQubitGates) {
...
+  qc.h(1);
+  qc.t(0);
+  qc.t(1);
+  qc.x(0);
+  qc.x(1);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectBlocks(qc, 1, true);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 6);
+}
```
burgholzer: The test name implies that this tests some kind of collection. However, the actual test is set up to not change a single thing. Please revise the test so that it is actually useful.

test/circuit_optimizer/test_collect_blocks.cpp
```diff
@@ -37,20 +53,63 @@ TEST(CollectBlocks, collectMultipleSingleQubitGates) {
...
+  qc.h(1);
+  qc.t(0);
+  qc.x(0);
+  qc.x(1);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectBlocks(qc, 2, true);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 4);
+  EXPECT_TRUE(qc.front()->isCompoundOperation());
+}
```
burgholzer: This test does not seem quite right.
Looking at the circuit, I would have imagined there to be 3 blocks in the resulting circuit (H(0), H(1), X(1)), (T(0)), and (X(0)).
Please check the implementation again. and devise reasonable tests.
Instead of simply asserting the number of operations, assert what the actual operations are to make sure that the pass is doing what it is supposed to do.

test/circuit_optimizer/test_collect_blocks.cpp
```diff
@@ -37,20 +53,63 @@ TEST(CollectBlocks, collectMultipleSingleQubitGates) {
...
+  qc.cx(0, 1);
+  qc.t(0);
+  qc.x(0);
+  qc.x(1);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectBlocks(qc, 2, true);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 4);
+  EXPECT_TRUE(qc.front()->isCompoundOperation());
+}
```
burgholzer: A couple more tests are in order. Tests should cover at least every Clifford gate once and every non-Clifford gate once.

test/ir/test_operation.cpp
```diff
@@ -147,6 +148,44 @@ TEST(Operation, IsIndividualGate) {
...
+  EXPECT_TRUE(sxdg.isClifford());
+  const qc::StandardOperation dcx(0, 1, qc::DCX);
+  EXPECT_TRUE(dcx.isClifford());
+  const qc::StandardOperation swap(0, 1, qc::SWAP);
+  EXPECT_TRUE(swap.isClifford());
+  const qc::StandardOperation iswap(0, 1, qc::iSWAP);
+  EXPECT_TRUE(iswap.isClifford());
+  const qc::StandardOperation ecr(0, 1, qc::ECR);
+  EXPECT_TRUE(ecr.isClifford());
+}
```
burgholzer: These are not really conclusive (see comment where the function is defined).

burgholzer (changes requested):
Thanks for this initial PR 🙏🏼 
I just went through everything and left a couple of inline comments. This still needs a bit of work before it is ready to be merged.
Don't hesitate to ask if you should have any questions.

burgholzer:
@jannikpflieger any updates on this? it's been three weeks since the review above 😌

jannikpflieger:
@burgholzer still doing it, since I was on vacation + exam, but I'm on it! I'm back since yesterday evening :D

burgholzer:
Hey @jannikpflieger 👋🏻

What's the status here? 🙂

jannikpflieger:
Hey @burgholzer,
Yes so the correction I did doesn't pass my tests. I was hoping that someone could take a look at that bu I think I need to update the branch anyway.

burgholzer:
> Hey @burgholzer,
> 
> Yes so the correction I did doesn't pass my tests. I was hoping that someone could take a look at that bu I think I need to update the branch anyway.
> 
> 

I can take a look a little later today if you'd like.
Although I am just on my phone, so it's only gonna be a couple of comments probably.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1434,9 +1435,10 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
     bool canProcess = true;
     bool makesTooBig = false;
 
-    // check if the operation can be processed
     if (!op->isUnitary()) {
       canProcess = false;
+    } else if (onlyCollectCliffords && !op->isClifford()) {
+      continue;
```
burgholzer: This is definitely not working as expected yet, which some improved tests would have shown. I recommend to improve the testing significantly to ensure that this is working as expected.
Non-Clifford operations need to finalize a block for a certain qubit. so that the next block comes _after_ the non-Clifford gate(s).

test/circuit_optimizer/test_collect_blocks.cpp
burgholzer, line 1: I fixed up a couple of things in the other parts of the PR, but these tests here are still broken. They do not assert the right behaviour/structure of the resulting circuit and, what makes it worse, the implementation currently does not produce correct results, as can be seen fro example from the `nonCliffordOnAll` test results

```console
i:   0   1
1:  sx   |
2:   | sxd
3:   t   |
4:   |   t
5:   x   |
6:   |   x
o:   0   1


i:   0   1
1:--------
 :   | sxd
 :   |   x
 ---------
2:--------
 :  sx   |
 :   x   |
 ---------
3:   |   t
4:   t   |
o:   0   1
```
which clearly shows an incorrect grouping and ordering of the operations.

burgholzer (changes requested):
Hey @jannikpflieger 👋🏼 
I took the liberty to fix up some small parts of this PR to move this forward.
As you'll see by the comments, the implementation is not yet correct and will need additional work that I cannot do for you. This all starts with proper testing of the desired feature to ensure that it truly works as expected.

include/mqt-core/circuit_optimizer/CircuitOptimizer.hpp
```diff
@@ -93,6 +93,19 @@ class CircuitOptimizer {
...
 
+  /**
+   * @brief Collects all Clifford blocks in the circuit.
+   * @details The circuit is traversed and all operations that are part of a
+   * Clifford block are collected into a compound operation with a certain
+   * maximum Block Size. All non-Clifford Blocks remain as single blocks. The
+   * compound operation is then appended to the circuit. Light optimizations are
+   * applied to the blocks, such as removing identity gates.
+   * @param qc the quantum circuit
+   * @param maxBlockSize the maximum size of a block
```
burgholzer: Just some minor rewording.
```suggestion
   * @brief Collects all Clifford blocks in the circuit.
   * @details The circuit is traversed and all operations that are part of a
   * Clifford block are collected into a compound operation with a certain
   * maximum block size. All non-Clifford operations remain untouched. 
   * Light optimizations are applied to the blocks, such as removing identity gates.
   * @param qc the quantum circuit
   * @param maxBlockSize the maximum size of a block
```

src/circuit_optimizer/CircuitOptimizer.cpp
burgholzer, line 1: Please try to revert the unrelated changes in this file.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1572,6 +1571,152 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
...
+    bool isCliff;
+    std::vector<Qubit> qubits;
+  };
+  std::vector<OperationData> opData;
+  opData.reserve(qc.size());
+  for (auto opIt = qc.begin(); opIt != qc.end(); ++opIt) {
+    const std::set<Qubit> used = opIt->get()->getUsedQubits();
+    opData.push_back(OperationData{
+        opIt, opIt->get()->isClifford(), {used.begin(), used.end()}});
+  }
```
burgholzer: This just feels like a wasteful copy of information. The QuantumComputation is already a vector of operations that contains all the necessary information and that can be accessed by an index.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,141 @@
...
+  qc.rx(0.1, 0);
+  qc.x(0);
+  qc.y(1);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectCliffordBlocks(qc, 2);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 3);
+  EXPECT_TRUE(qc.front()->isCompoundOperation());
+  EXPECT_TRUE(qc.back()->isCompoundOperation());
+  EXPECT_EQ(dynamic_cast<qc::CompoundOperation&>(*qc.front()).size(), 3);
```
burgholzer: This cannot be right. The last Y gate should be part of the first block, which means that this should only result in a compound operation with 4 operations.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
burgholzer, line 1: Almost none of these tests assert the full structure of the resulting circuit.
This makes it harder than necessary to judge the correctness of the transformations.
I would argue that tests should contain a full replica of the expected circuit and compare against that using the circuit equality method.

Furthermore, none of these tests are particularly "unit" tests in the sense that they are minimal to demonstrate that a particular behaviour works as expected.
Try to think of minimal scenarios that should work as well as minimal scenarios for things that should not work.

Lastly, I do not think these tests cover all relevant aspects of the block consolidation. Just an example of what is definitely missing: Clifford operations that would form a three-qubit block, but only running collection with two-qubits maximum and checking that the result looks as expected.

Please take that into consideration and refactor the test setup appropriately.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1572,6 +1571,152 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
...
+      break; // don't need to look into other blocks
+    }
+
+    // If we didn't merge into any block, we need to start a new one
+    // and it is for sure Clifford because we already checkked non Clifford
+    if (!didMerge) {
+      blocks.push_back(
+          Block{{i}, true, {opD.qubits.begin(), opD.qubits.end()}});
+    }
+  }
```
burgholzer: From a complexity point of view, this seems overly complex and inefficient.
Based on one of the tests, I also suspect that this is not quite correct yet.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1572,6 +1571,152 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
...
+    const int lastK = block.idx.back();
+    *opData[static_cast<std::size_t>(lastK)].opIt = std::move(compound);
+
+    // Erase the rest, backwards to keep iterators valid
+    for (int opD = static_cast<int>(block.idx.size()) - 2; opD >= 0; --opD) {
+      qc.erase(opData[static_cast<std::size_t>(
+                          block.idx[static_cast<std::size_t>(opD)])]
+                   .opIt);
+    }
+  }
```
burgholzer: As this whole function is merely reshuffling and/or reordering/regrouping gates, there should be no need for any `clone` or other copies of operations. This should be totally achievable with moves only.

jannikpflieger:
@burgholzer thanks for the comment, I get working on it ASAP.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
jannikpflieger: For that exact  test, I wanted Y not part of the first block. the idea of this test was that we have H0, S1 and CX 01 in one block as there are three gates but block size is limited to two. 

I agree if we ran the circuit with maxBlockSize 3 then it should behave as you wrote. I can try executing it with larger block size, and see if it behaves as expected.

burgholzer: The max block size is in terms of qubits, not in terms of gates.

jannikpflieger: Okay, this clarifies then a lot. Thanks! I'll adjust it. Then I agree that the other tests are faulty in a similar manner and do not show the intended behaviour.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,377 @@
...
+}
+
+TEST(CliffordBlocks, singleGate) {
+  QuantumComputation qc(1);
+  qc.sx(0);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectCliffordBlocks(qc, 1);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 1);
+}
```
burgholzer: Would be great to assert what that single gate is. Is it a compound operation or a single-qubit gate? I'd very much prefer it to be the single-qubit gate as a compound operation only containing one gate does not make too much sense.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,377 @@
...
+  QuantumComputation qc2(2);
+  qc2.h(0);
+  qc2.cx(0, 1);
+  qc2.x(1);
+
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectCliffordBlocks(qc, 1);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 3);
+  EXPECT_TRUE(qc == qc2);
```
burgholzer:
```suggestion
  EXPECT_EQ(qc, qc2);
```
also applies to other tests.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,377 @@
...
+TEST(CliffordBlocks, largerGatethenBlock) {
+  QuantumComputation qc(2);
+  qc.h(0);
+  qc.cx(0, 1);
+  qc.x(1);
+
+  QuantumComputation qc2(2);
+  qc2.h(0);
+  qc2.cx(0, 1);
+  qc2.x(1);
```
burgholzer: It might make sense to give this a more "speaking" variable name. Like `expected`.
Holds for the remaining tests.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,377 @@
...
+  QuantumComputation qc2(3);
+  QuantumComputation op1(3);
+  QuantumComputation op2(3);
+  op1.s(1);
+  op1.h(0);
+  op1.cx(0, 1);
+  qc2.emplace_back(op1.asCompoundOperation());
+  qc2.rx(0.1, 0);
+  op2.cx(1, 2);
+  op2.y(1);
```
burgholzer: These tests are written in a way that is quite hard to judge for correctness. The difference between `op2` and `qc2` is visually very hard to make out. A little more structure would be beneficial here.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,377 @@
...
+  EXPECT_TRUE(qc2 == qc);
+}
+
+TEST(CliffordBlocks, handleCompoundOperation2) {
+  QuantumComputation qc(1);
+  QuantumComputation op(1);
+  qc.h(0);
+  op.t(0);
+  op.z(0);
+  qc.emplace_back(op.asCompoundOperation());
```
burgholzer: What happens if you extend this test as follows:
```suggestion
  qc.h(0);
  op.t(0);
  op.z(0);
  op.x(0);
  qc.emplace_back(op.asCompoundOperation());
```

test/circuit_optimizer/test_collect_clifford_blocks.cpp
```diff
@@ -0,0 +1,377 @@
...
+  QuantumComputation qc(1);
+  qc.h(0);
+  qc.barrier(0);
+  qc.h(0);
+  std::cout << qc << "\n";
+  qc::CircuitOptimizer::collectCliffordBlocks(qc, 1);
+  std::cout << qc << "\n";
+  EXPECT_EQ(qc.size(), 3);
+  EXPECT_TRUE(qc.back()->isStandardOperation());
+}
```
burgholzer: this could be a little more assertive on the exact circuit similar to the other tests.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1625,6 +1625,241 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
...
+        blockQubits, [this](const Qubit q) { return blocked.contains(q); });
+  }
+
+  /**
+   * @brief Check if adding qubits used by gate would exceed maxBlockSize
+   * @param used Qubits used by the gate
+   * @param maxBlockSize Maximum allowed block size
+   */
+  [[nodiscard]] bool
+  checkMaxBlockSize(const std::set<Qubit>& used,
```
burgholzer: The name of this function could be improved to reflect what it actually means to return true or false.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -1625,6 +1625,241 @@ void CircuitOptimizer::collectBlocks(QuantumComputation& qc,
...
+   * @param movePosition Whether to move the position of the block
+   * @param step Current logical step in the quantum computation
+   */
+  void addOp(std::unique_ptr<Operation>& op, const std::set<Qubit>& used,
+             QuantumComputation& qc, QuantumComputation::iterator& it,
+             const bool movePosition, const std::size_t step) {
+    operations->emplace_back(std::move(op));
+
+    if (movePosition) {
+      // we move block into the right position with idenites because they are
```
burgholzer:
```suggestion
      // we move block into the right position with identities because they are
```

test/circuit_optimizer/test_collect_clifford_blocks.cpp
jannikpflieger: Thanks for the comment. 
Because t,z and x are one CompoundOperation the whole operations is interpreted as a non-clifford operation. 
So we would get exactly this back. The Hadamard in Block 1 and then the non-Clifford Block consisting of T,Z and X in one non-Clifford Block. I tested it, here is the result:

i:   0
1:   h
2:----
 :   t
 :   z
 :   x
 : -----
o:   0

Do you have another outcome in mind?
In the code I generally don't break compound operations up into their single operations. So an output like this:

i:   0
1:   h
2:  t
3:----
 :   z
 :   x
 : -----
o:   0

is not considered right now.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
jannikpflieger: Just a clarifying question, as we compare the two circuits not only size but all the other assertion are a bit redundant too right? 
Such as is qc.front() a Compound operation for example. Should I delete them as well?

test/circuit_optimizer/test_collect_clifford_blocks.cpp
burgholzer: I suppose this is fine. If I would have expected anything then I would have expected a nested compound operation that contains the z and x gate as part of the original compound operation, but this could be a potential optimization for the future.

test/circuit_optimizer/test_collect_clifford_blocks.cpp
burgholzer: Yeah. Technically they are redundant. they might help in noticing a particular error, but given how the circuits are printed anyway as part of every test, this should also not be necessary. So feel free to delete the redundant checks in all the tests.

PR #886, opened 2025-03-24
♻️ Eliminate templates from `dd::Package` class

burgholzer:
## Description

This PR removes the `Config` template from the DD package, which makes the DD package a non-templated class ✨ 
The compile time information was only used anymore in the `ComputeTables` for determining the number of buckets.
Given how we want to extend these tables in the future anyway, there is no hope of keeping the hash maps as compiled-time sized constructs. Thus, this PR replaces the respective data structures with `std::vector`.
By removing the `Config` template argument, a lot of code can now be easily moved to source files, which should drastically improve compile times and compiler diagnostics. It should also help with binary size.
This required some changes to the way special configurations are handled, which is now no longer handled via inheritance.

@Joshy-R For awareness. I hope this does not interfere too much with what you are currently working on. It is just something that I wanted to get in before the official v3 release of mqt-core as it's a considerable breaking change and will require changes from literally everyone using the DD package.

## Checklist:


Joshy-R:
Thank you, this was something i was too afraid to do yet. When will v3 be reeased? Some changes I would like to propose would also be breaking changes.

Also what changes do you consider to be breaking ones? Is just the python interface considered to be stable or the entire package file with the c++ interface?

burgholzer:
> Thank you, this was something i was too afraid to do yet. When will v3 be reeased? Some changes I would like to propose would also be breaking changes.

It's something I wanted to do for quite a while. These compile-time constructs were only limiting the things we could explore. I am glad to see them gone.

V3 should have probably released a while ago 😅 the Changelog keeps getting longer and longer. Typically, I'd be in no hurry, but we currently have an ongoing journal submission for MQT Core that is just waiting for a tagged release of MQT Core. So I will probably release very soon.

> Also what changes do you consider to be breaking ones? Is just the python interface considered to be stable or the entire package file with the c++ interface?

Generally speaking, we are not following semantic versioning very strictly.
Typically we try to only make breaking changes (C++ or Python) in minor releases and reserve patch releases for bug fixes and other smaller updates.
If the Python interface remains as stable as possible, that's definitely desirable.
Given how the C++ interface is mostly used by our own MQT tools (so far, and at least that I am aware of), making more drastic changes is not as bad. It just requires us to update the consuming packages.

I'd not worry too much about your changes being breaking ones. We can always include them in v3.1 (even if that might be released rather soon after 3.0; releases are cheap 😌)

PR #892, opened 2025-03-31
Add first optimization passes to MQTOpt

DRovara:
## Description

This pull request adds the first optimization passes to MQTOpt.

All new passes can be called using the `quantum-opt` tool.

## `--cancel-consecutive-inverses`

This pass finds consecutive applications of inverses of gates and cancels them.

_Original_:
```mlir
%q_1 = mqtopt.x() %q_0 : !mqtopt.Qubit
%q_2 = mqtopt.z() %q_1 : !mqtopt.Qubit
%q_3 = mqtopt.z() %q_2 : !mqtopt.Qubit
```
_After pass_:
```mlir
%q_1 = mqtopt.x() %q_0 : !mqtopt.Qubit
```

## --quantum-sink

This pass sinks (single-qubit*) quantum gates into deeper branches.

_Original_:
```mlir
%q_1 = mqtopt.x() %q_0 : !mqtopt.Qubit
cf.cond_br %i, ^then(%q_1 : !mqtopt.Qubit), ^else(%q_1 : !mqtopt.Qubit)

^then(%q_1then : !mqtopt.Qubit):
  cf.br ^continue(%q_1then : !mqtopt.Qubit)

^else(%q_1else : !mqtopt.Qubit):
  %q_2 = mqtopt.x() % q_1else : !mqtopt.Qubit
  cf.br ^continue(%q_2 : !mqtopt.Qubit)

^continue(%q_3 : !mqtopt.Qubit):
  ...
```

_After pass_:
```mlir
cf.cond_br %i, ^then(%q_0 : !mqtopt.Qubit), ^else(%q_0 : !mqtopt.Qubit)

^then(%q_0then : !mqtopt.Qubit):
  %q_1then = mqtopt.x() %q_0then : !mqtopt.Qubit
  cf.br ^continue(%q_1then : !mqtopt.Qubit)

^else(%q_0else : !mqtopt.Qubit):
  %q_1else = mqtopt.x() %q_0else : !mqtopt.Qubit
  %q_2 = mqtopt.x() %q_1else : !mqtopt.Qubit
  cf.br ^continue(%q_2 : !mqtopt.Qubit)

^continue(%q_3 : !mqtopt.Qubit):
  ...
```

Note that, due to linear types and SSA, branch "calls" and "returns" will always need to pass the variables needed by a block.

*currently, this pass only works for single-qubit gates. Multi-qubit gates might have multiple users and therefore, the logic has to be adapted for them.

## Checklist:


mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveSelfInverse.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <mlir/Transforms/GreedyPatternRewriteDriver.h>
+#include <utility>
+
+namespace mqt::ir::opt {
+
+#define GEN_PASS_DEF_CANCELCONSECUTIVESELFINVERSE
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h.inc"
+
+/**
+ * @brief This pass attempty to cancel consecutive self-inverse operations.
```
burgholzer:
```suggestion
 * @brief This pass attempts to cancel consecutive self-inverse operations.
```

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveSelfInversePattern.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <mlir/IR/MLIRContext.h>
+#include <mlir/IR/PatternMatch.h>
+#include <mlir/IR/ValueRange.h>
+#include <mlir/Support/LLVM.h>
+#include <mlir/Support/LogicalResult.h>
+
+namespace mqt::ir::opt {
+
+/**
+ * @brief This pattern attempty to cancel consecutive self-inverse operations.
```
burgholzer:
```suggestion
 * @brief This pattern attempts to cancel consecutive self-inverse operations.
```

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveSelfInversePattern.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+  [[nodiscard]] static bool
+  areUsersUnique(mlir::ResultRange::user_range users) {
+    return std::none_of(users.begin(), users.end(),
+                        [&](auto* user) { return user != *users.begin(); });
+  }
+
+  mlir::LogicalResult match(UnitaryInterface op) const override {
+    if (!(mlir::isa<XOp>(op) || mlir::isa<YOp>(op) || mlir::isa<ZOp>(op) ||
+          mlir::isa<HOp>(op) || mlir::isa<IOp>(op) || mlir::isa<SWAPOp>(op))) {
```
burgholzer: It would be fairly easy to extend this to not only work on self-inverse gates, but to really cancel gates that are the inverse of another.
Easy examples include:
- T, Tdg
- S, Sdg
- Sx, Sxdg

The CircuitOptimizer supports some of these already in the `singleQubitGateFusionPass` https://github.com/cda-tum/mqt-core/blob/d71fa7c1ba73af053327c6b9900529c0ffb7e176/src/circuit_optimizer/CircuitOptimizer.cpp#L178-L271

Naturally, it would also be kind of straight forward to extend this to rotation gates.
We have rules for how to generate the inverses of all supported MQT gates within the DD package: https://github.com/cda-tum/mqt-core/blob/d71fa7c1ba73af053327c6b9900529c0ffb7e176/src/dd/Operations.cpp#L75-L134

I am fine with merging the pass as it is for now. But I believe we should be quickly expanding it for further use cases. Maybe worth to turn some of the above into a proper issue description.

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPass.cpp
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @brief This pass attempty to sink quantum operations into the block where
+ * their results are used.
+ */
+struct QuantumSinkPass final : impl::QuantumSinkPassBase<QuantumSinkPass> {
+
+  void runOnOperation() override {
+    // Get the current operation being operated on.
+    mlir::Operation* op = getOperation();
+    mlir::MLIRContext* ctx = &getContext();
```
burgholzer:
```suggestion
    auto* op = getOperation();
    auto* ctx = &getContext();
```
Not totally necessary, but this could save 1 or 2 import in the file and the functions themselves already convey the type being returned. (this also holds for the other pass)

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPushPattern.cpp
```diff
@@ -0,0 +1,346 @@
...
+        depth = result + 1;
+      } else {
+        depth = std::min(depth, result + 1);
+      }
+    }
+    return depth;
+  }
+
+  /**
+   * @brief Out of a set of users, returns the closes one.
```
burgholzer:
```suggestion
   * @brief Out of a set of users, returns the closest one.
```

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkShiftPattern.cpp
```diff
@@ -0,0 +1,157 @@
...
+    return std::all_of(
+               users.begin(), users.end(),
+               [&](auto* user) { return user->getBlock() != op->getBlock(); })
+               ? mlir::success()
+               : mlir::failure();
+  }
+
+  /**
+   * @brief Repdoes not precede the block.laces all uses of a given operation
+   * with the results of a clone,
```
burgholzer: Something seems to be messed up with this docstring.

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
```diff
@@ -0,0 +1,47 @@
...
+    // CHECK: %[[Q12_1:.*]]:2 = mqtopt.x() %[[Q1_0]] ctrl %[[Q2_0]] : !mqtopt.Qubit, !mqtopt.Qubit
+    // CHECK: %[[Q12_2:.*]]:2 = mqtopt.x() %[[Q12_1]]#1 ctrl %[[Q12_1]]#0 : !mqtopt.Qubit, !mqtopt.Qubit
+
+    // --------------------- Check for operations that should be canceled -----------------------------------
+    // CHECK-NOT: %[[ANY:.*]], %[[ANY:.*]] = mqtopt.x() %[[ANY:.*]] ctrl %[[Q0_0]] : !mqtopt.Qubit, !mqtopt.Qubit
+
+    %q1_1, %q0_1 = mqtopt.x() %q1_0 ctrl %q0_0 : !mqtopt.Qubit, !mqtopt.Qubit
+    %q1_2, %q0_2 = mqtopt.x() %q1_1 ctrl %q0_1 : !mqtopt.Qubit, !mqtopt.Qubit
+    %q1_3, %q2_1 = mqtopt.x() %q1_2 ctrl %q2_0 : !mqtopt.Qubit, !mqtopt.Qubit
+    %q2_2, %q1_4 = mqtopt.x() %q2_1 ctrl %q1_3 : !mqtopt.Qubit, !mqtopt.Qubit
```
burgholzer: It would be good to check if the optimization pass also handles controlled gates with different polarity correctly. Say, one CNOT which is positively controlled followed by one which is negatively controlled.

In addition, it would probably be nice to check that multi-controlled X gates with different number of controls are not cancelled.

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
burgholzer, line 1: Side note: at least from a reviewer perspective it seems laborious to construct these tests compared to how similar tests are written for the CircuitOptimizer (https://github.com/cda-tum/mqt-core/blob/main/test/circuit_optimizer/test_single_qubit_gate_fusion.cpp).
Is this a concern? Typically, we should strive to write many tests for the respective functionality to ensure this is working. DO we have a strategy for making that easier or is the current setup just fine?

burgholzer, line 1: As far as I remember, there is a way to combine multiple tests into one file with MLIR.
I believe it might make sense to create one test file per optimization pass and have several (small) test cases in each of them.

burgholzer (changes requested):
Hey @DRovara 👋🏼 

Many thanks for this PR. This is looking great!
Most of the comments are just minor things to fix up.
2 or 3 are more so on how we want to handle certain things and how we want to expand them in the future.

One last thing: I still believe that there also might be value in doing the reverse of the QuantumSink pass you created. Pulling out common operations at the beginning or the end of branches reduces the code within the branches themselves.
For `IfElse` statements that might not be that important, but for loop constructs like `for` or `while`, this could save significant resources. I believe this typically goes under the name of "hoisting" (see, e.g., https://compileroptimizations.com/category/hoisting.htm)
Could also be a nice follow-up.

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPass.cpp
DRovara: Sounds good to me :+1:. Just for clarification though, `op` has to be `auto`, not `auto*`, becase the inferred type is `mlir::ModuleOp`, not `mlir::Operation*`, and `ModuleOp` is already implicitly defined as a pointer.

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
DRovara: Absolutely agreed. Writing these tests is a pain. I spent more time fixing typos in the expected test results than I spent fixing bugs that were uncovered by these tests. Also, they end up only testing the string representation of the output which is obviously also limited.

Long-term I think we should search for a more efficient way to do this.
LIT tests allow you to run any verification program like
```
// RUN: quantum-opt %s --mqt-core-round-trip | <your-program> %s
```
So maybe we can find something that works better  than FileCheck or we write our own program for that.
Of course, that would still be just a test of the string output.

Alternatively, I believe there is nothing that stops us from using these `gtests` also on the MLIR data structures.


In any case, I think this is a issue for the future, not for  this PR

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPass.cpp
burgholzer: I am pretty sure that clang-tidy will recommend you to use `auto*` for `op` in this context, but I am happy to be proven wrong 🙃 
Naturally, `auto` works just fine and the inferred type will be `mlir::Operation*` (or whatever `getOperation()` returns). Clang-tidy just likes to make it explicit when something is a pointer.

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
burgholzer: > In any case, I think this is a issue for the future, not for this PR

Most definitely. Maybe something worth opening an issue on.

> Alternatively, I believe there is nothing that stops us from using these gtests also on the MLIR data structures.

When I read through the Catalyst source code, I saw that MLIR/LLVM apparently has some integration with `googletest` already (https://github.com/PennyLaneAI/catalyst/blob/9f1c8f0b84906c2f8e65c899f5f305f449064585/mlir/CMakeLists.txt#L100-L145). Might be worth exploring that in the near future.

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveSelfInversePattern.cpp
DRovara: I have adapted the pass to now also cancel the other inverse gates like `S` and `Sdg`.

However, I feel like merging rotation gates would make more sense as its own Pass (or at least its own Pattern) to keep these patterns as atomic as possible.

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
DRovara: Done :+1:

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
DRovara: Good idea. Apparently there are multiple ways to achieve that, butmaybe the best (from my limited research) seems to be using the `split-input-file` method. This splits the test file into slices based on `// -----` markers. Each of these markers can then have a description that says what the test does etc.

Advantages:
- Simple and straightforward to extend
- Clean
Disadvantage:
- When a test fails, it will always only tell us the name of the failed test file. Using the `-v` option also gives us the location in the code where the test failed, but we will never get a definitive output that says "slice X failed".

An alternative would be using `CHECK-PREFIXES`. This involves adding a label to every `CHECK` instruction based on the corresponding test case. This way, we could have more information about what test failed exactly, but the maintenance overhead is a lot higher this way.

The LLVM testing guidelines suggest using these prefixes for e.g. testing different platforms, not to identify different test cases. ChatGPT even says using prefixes to differentiate test cases is bad practice.

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveSelfInversePattern.cpp
DRovara: Here is the issue that proposes that feature #898

mlir/test/Dialect/MQTOpt/Transforms/consecutive-self-inverse-2q.mlir
DRovara: #899 now summarizes the current testing situation/possible solutions.

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInversesPattern.cpp
```diff
@@ -0,0 +1,132 @@
...
+        {"peres", "peresdg"},
+    };
+
+    const auto aName = a->getName().stripDialect().str();
+    const auto bName = b->getName().stripDialect().str();
+    return (INVERSE_PAIRS.find(aName) != INVERSE_PAIRS.end() &&
+            INVERSE_PAIRS.at(aName) == bName) ||
+           (INVERSE_PAIRS.find(bName) != INVERSE_PAIRS.end() &&
+            INVERSE_PAIRS.at(bName) == aName);
+  }
```
burgholzer: I believe this could be slightly improved. Although this is more of a nitpick than anything.

The arguments to the function could be const references because they must not be null for this function to work.
Additionally, calling `find` and then `at` technically incurs two lookups. I am not sure all compilers are clever enough to reuse the previous result. At least I would not trust it.
So what I would typically do is to t store the iterator and if it's not the end iterator, dereference it and do the comparison with the dereferenced iterator.

This is really not critical for performance here, but I believe it is worth to adopt some good practices on interface design and being mindful of the complexity of certain STL calls.

Edit: After reading through all of the PR, I believe the interface point holds for quite some of the functions here. Most of the functions that take pointers only work if these pointers are non-null because they are dereferenced unchecked. At that point, I'd personally make these (const) references instead.
(MLIR is not the biggest fan of const; so it might just have to be regular references)

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPushPattern.cpp
```diff
@@ -0,0 +1,348 @@
...
+   * @param branchOp The branch operation that caused the critical edge.
+   * @param rewriter The pattern rewriter to use.
+   * @return The new block that was created.
+   */
+  static mlir::Block* breakCriticalEdge(mlir::Block* oldTarget,
+                                        mlir::Operation* branchOp,
+                                        mlir::PatternRewriter& rewriter) {
+    auto* newBlock = rewriter.createBlock(oldTarget->getParent());
+    std::vector<mlir::Value> newBlockOutputs;
+    for (auto arg : oldTarget->getArguments()) {
```
burgholzer: Another small C++ nit.
You could pre-allocate the required size for the vector before the loop by pulling out the `getArguments` call.
I am actually surprised that clang-tidy didn't suggest this.

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPushPattern.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+/**
+ * @brief Populates the given pattern set with the
+ * `QuantumSinkPushPattern`.
+ *
+ * @param patterns The pattern set to populate.
+ */
+void populateQuantumSinkPushPatterns(mlir::RewritePatternSet& patterns) {
+  patterns.add<QuantumSinkPushPattern>(patterns.getContext());
+}
```
burgholzer, line 350: Just out of genuine interest:
These functions feel so repetitive. Isn't this something that can be autogenerated via tablegen?

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkShiftPattern.cpp
```diff
@@ -0,0 +1,157 @@
...
+   * @return A vector of all earliest users of the given operation.
+   */
+  [[nodiscard]] std::vector<mlir::Operation*>
+  getEarliestUsers(mlir::ResultRange::user_range users) const {
+    std::vector<mlir::Operation*> earliestUsers;
+    std::copy_if(
+        users.begin(), users.end(), std::back_inserter(earliestUsers),
+        [&](auto* user) {
+          return std::none_of(users.begin(), users.end(), [&](auto* other) {
+            std::set<mlir::Block*> visited;
```
burgholzer: I am pretty sure that you want to use an `unordered_set` (hash set) here for the visited set because you do not seem to need a sorted container and you need a fast membership check.
Maybe worth to check in the other functions and files as well.

burgholzer (changes requested):
Thanks for the additional round of changes here. The test structure looks much cleaner now and I also like the additionally supported inverse cancellations.

What's left are some rather nitpicky comments on C++ specifics. None of these are truly performance critical for the code at hand. However, I thought I'd still share them with you because they could become important for other scenarios.

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInversesPattern.cpp
DRovara: From what I have read, LLVM/MLIR suggests to not use const variables for operations, values etc.
In particular, none of the methods implemented by these classes are defined as const. Therefore (correct me if I'm wrong), I believe making these arguments const references would not work because we use non-const methods on them, e.g. in Line 54, right?

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPushPattern.cpp
DRovara: To be honest, I'm not even sure if it even is best practice to implement them like this. I believe we initially decided on this with @ystade, do you have any opinion on it?

Edit: I just checked, and I think the reason was, that we can define these populate methods in the Pass' header file. This way, we don't require additional includes in the pass implementation.

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInversesPattern.cpp
burgholzer: Yeah. But references (without the const) would work and would still imply that the arguments must be non-null.

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPushPattern.cpp
burgholzer: Some quick googling resulted in 
- https://mlir.llvm.org/docs/Tutorials/QuickstartRewrites/#register-the-pattern

and 
- https://mlir.llvm.org/docs/DeclarativeRewrites/

The latter describes how whole rewrites can be written in Tablegen. But I don't fully understand if that would solve any of our problems.

DRovara: Yes, I do remember that we discussed this before. We decided against using Tablegen for patterns, because a majority of the pattern implementation would just be C++ code anyways that we would then have to wrap in a `NativeCodeCall` (or something similar) instruction in tablegen.

We decided it makes more sense to write C++ code in cpp files so that's why we don't use tablegen for rewrite patterns.

burgholzer: That makes sense! Thanks for checking again. Let's just stick with what we currently have. It's not bad in terms of boilerplate code. It's just noticeable.

mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInversesPattern.cpp
DRovara: Alright, I replaced the pointers with references now. Honestly, my intuition told me it would be less pretty/consistent to just have a mix of pointers and references throughout the whole file, but I'm also no C++ expert and I get the idea of using references to indicate that it should be something non-null, so I'm also fine with it this way.

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPushPattern.cpp
```diff
@@ -0,0 +1,352 @@
...
+   * precedes it.
+   *
+   * @param toCheck The block to check.
+   * @param op The operation to find the distance to.
+   * @return The distance between the block and the operation or -1 if the
+   * operation is not in a predecessor block.
+   */
+  int getDepth(mlir::Block& toCheck, const UnitaryInterface& op) const {
+    auto& originalBlock = *op->getBlock();
+    if (&toCheck == &originalBlock) {
```
burgholzer:
```suggestion
    auto* originalBlock = op->getBlock();
    if (&toCheck == originalBlock) {
```
Just slightly more idiomatic

burgholzer (approved):
Thanks again. This is looking great in my opinion.
Feel free to ignore the one last comment. It's just a very small and insignificant change.
Feel free to either incorporate the change or to simply resolve the comment. 
In any case, you can merge the PR 🙌

PR #893, opened 2025-04-01
Clean Up mqt-core round-trip pass

DRovara:
## Description

The pass used a method that was deprecated under the newest MLIR/LLVM standards.

Also, the test case was very bare-bones. This pull requests updates the test case to actually test the functionality of the pass.

## Checklist:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -15,7 +15,7 @@ jobs:
   # workflow is triggered from a PR or a release, respectively.
   python-packaging:
     name: 🐍 Packaging
-    uses: cda-tum/mqt-workflows/.github/workflows/reusable-python-packaging.yml@v1.7
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-python-packaging.yml@v1.7
 
   # Downloads the previously generated artifacts and deploys to PyPI on published releases.
   deploy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,13 +14,13 @@ concurrency:
 jobs:
   change-detection:
     name: 🔍 Change
-    uses: cda-tum/mqt-workflows/.github/workflows/reusable-change-detection.yml@v1.7
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-change-detection.yml@v1.7
 
   cpp-tests:
     name: 🇨‌ Test
     needs: change-detection
     if: fromJSON(needs.change-detection.outputs.run-cpp-tests)
-    uses: cda-tum/mqt-workflows/.github/workflows/reusable-cpp-ci.yml@v1.7
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-ci.yml@v1.7
     with:
       # Runs to enable
       enable-ubuntu2404-gcc-release: true
@@ -68,7 +68,7 @@ jobs:
     name: 🇨‌ Test (Extensive)
     needs: change-detection
     if: fromJSON(needs.change-detection.outputs.run-cpp-tests) && github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'extensive-cpp-ci')
-    uses: cda-tum/mqt-workflows/.github/workflows/reusable-cpp-ci.yml@v1.7
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-ci.yml@v1.7
     with:
       # Runs to enable
       enable-ubuntu2404-gcc-debug: true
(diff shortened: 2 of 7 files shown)


burgholzer (changes requested):
I was aware of this before. Unfortunately, the non-deprecated method does not exist yet in the LLVM@19 series, which makes the build fail on that version.
I am still a bit hesitant to require LLVM@20 for the project here.
Do you think that is just overly cautious or would you agree that we should be sticking with LLVM@19 as a minimum version?
Maybe also something where @ystade or @flowerthrower can weigh in.

The test changes are obviously fine and much appreciated.

DRovara:
> I was aware of this before. Unfortunately, the non-deprecated method does not exist yet in the LLVM@19 series, which makes the build fail on that version. I am still a bit hesitant to require LLVM@20 for the project here. Do you think that is just overly cautious or would you agree that we should be sticking with LLVM@19 as a minimum version? Maybe also something where @ystade or @flowerthrower can weigh in.
> 
> The test changes are obviously fine and much appreciated.

Yeah I noticed that issue, too.
Considering a new major LLVM version comes out every few months, I'm not quite sure how we could scale that efficiently in the long-term. Chances are, if a person does not have the latest version of LLVM installed on their machine, it might just as well be 18.x, 17.x, 16.x and so on. 
So considering there is a breaking change for us between 19 and 20, it doesn't seem so unreasonable to me not to support LLVM-19 anymore.

burgholzer:
> > I was aware of this before. Unfortunately, the non-deprecated method does not exist yet in the LLVM@19 series, which makes the build fail on that version. I am still a bit hesitant to require LLVM@20 for the project here. Do you think that is just overly cautious or would you agree that we should be sticking with LLVM@19 as a minimum version? Maybe also something where @ystade or @flowerthrower can weigh in.
> > The test changes are obviously fine and much appreciated.
> 
> Yeah I noticed that issue, too. Considering a new major LLVM version comes out every few months, I'm not quite sure how we could scale that efficiently in the long-term. Chances are, if a person does not have the latest version of LLVM installed on their machine, it might just as well be 18.x, 17.x, 16.x and so on. So considering there is a breaking change for us between 19 and 20, it doesn't seem so unreasonable to me not to support LLVM-19 anymore.

I generally agree with your statement. Although, technically, this is not a breaking change. It is a deprecation warning. The function is still available in LLVM@20. It just emits a warning.
And if I understand it correctly, it's roughly a new major version per year, which is an okayish release cadence to adapt to in my opinion.

I just checked and it seems that at least Homebrew now also defaults to llvm@20 when doing `brew install llvm`. For linux, there is the separate download script that we also use in CI, so that also works.
Chocolatey, one of the package mangers for Windows also defaults to llvm@20.
So I believe I'd actually be fine to drop the LLVM 19 support here.

What we should do though in my opinion is to make the version requirement explicit in the sense that we already catch mismatches during the CMake configure.
That should be done in a separate PR though.

@DRovara would you mind setting that up?

DRovara:
> > > I was aware of this before. Unfortunately, the non-deprecated method does not exist yet in the LLVM@19 series, which makes the build fail on that version. I am still a bit hesitant to require LLVM@20 for the project here. Do you think that is just overly cautious or would you agree that we should be sticking with LLVM@19 as a minimum version? Maybe also something where @ystade or @flowerthrower can weigh in.
> > > The test changes are obviously fine and much appreciated.
> > 
> > 
> > Yeah I noticed that issue, too. Considering a new major LLVM version comes out every few months, I'm not quite sure how we could scale that efficiently in the long-term. Chances are, if a person does not have the latest version of LLVM installed on their machine, it might just as well be 18.x, 17.x, 16.x and so on. So considering there is a breaking change for us between 19 and 20, it doesn't seem so unreasonable to me not to support LLVM-19 anymore.
> 
> I generally agree with your statement. Although, technically, this is not a breaking change. It is a deprecation warning. The function is still available in LLVM@20. It just emits a warning. And if I understand it correctly, it's roughly a new major version per year, which is an okayish release cadence to adapt to in my opinion.
> 
> I just checked and it seems that at least Homebrew now also defaults to llvm@20 when doing `brew install llvm`. For linux, there is the separate download script that we also use in CI, so that also works. Chocolatey, one of the package mangers for Windows also defaults to llvm@20. So I believe I'd actually be fine to drop the LLVM 19 support here.
> 
> What we should do though in my opinion is to make the version requirement explicit in the sense that we already catch mismatches during the CMake configure. That should be done in a separate PR though.
> 
> @DRovara would you mind setting that up?

Sure I can do that. I reverted the change for this PR so that it still uses `applyPatternsAndFoldGreedily` for now, and we can then update that call (and all other such calls for different passes) in that PR.

burgholzer (approved):
LGTM 👍🏼

burgholzer:
#894 should fix the standard workflows not starting. Should hopefully not take too long to get the CI through for that PR.

PR #897, opened 2025-04-02
⬆️ Require C++20

burgholzer:
## Description

This PR updates the required C++ standard to C++20.
The standard has been around for a while, and it is time for us to move on from C++17.
Some libraries that we want to interface with in the near future, such as CUDA-Q, use C++20.
This means we should, at the very least, ensure that we can build in a C++20 environment.
However, C++20 also has some appealing features that we may want to adopt.

The CI in this PR will prove that all tested configurations allow building the project.
The PR itself will only contain the minimal fixes to please clang-tidy.
Other features, such as 
- [Three-way comparison](https://en.cppreference.com/w/cpp/language/operator_comparison.html#Three-way_comparison) operator `<=>` and [`operator==() = default`](https://en.cppreference.com/w/cpp/language/default_comparisons.html)
- [Designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html#Designated_initializers)
- Init-statements and initializers in [range-for](https://en.cppreference.com/w/cpp/language/range-for.html)
- New [attributes](https://en.cppreference.com/w/cpp/language/attributes.html): \[\[[no_unique_address](https://en.cppreference.com/w/cpp/language/attributes/no_unique_address.html)\]\], \[\[[likely](https://en.cppreference.com/w/cpp/language/attributes/likely.html)\]\], \[\[[unlikely](https://en.cppreference.com/w/cpp/language/attributes/likely.html)\]\]
- [Formatting library](https://en.cppreference.com/w/cpp/utility/format.html)
- constexpr for [\<algorithm\>](https://en.cppreference.com/w/cpp/header/algorithm.html), [\<utility\>](https://en.cppreference.com/w/cpp/header/utility.html), [\<complex\>](https://en.cppreference.com/w/cpp/header/complex.html)
- [std::ssize](https://en.cppreference.com/w/cpp/iterator/size.html)
- [Ranges](https://en.cppreference.com/w/cpp/ranges.html)
- [Mathematical constants](https://en.cppreference.com/w/cpp/numeric/constants.html) in [\<numbers\>](https://en.cppreference.com/w/cpp/header/numbers.html)

will be added in subsequent PRs. It might be worth to move the above list to a separate issue, once this PR is merged.

## Checklist:


burgholzer:
Alright. I do not have the time to look into these diverse errors at the moment.
This will have to wait for v3.1+.
If anyone wants to pick this up, feel free to.

burgholzer:
Okay. After all the CI updates, this is really looking promising now.
The entire test CI should now run successfully.

Next step: Resolve all the clang-tidy warnings. They are mostly for either using `contains`, designated initializers, or range-based versions of certain constructs.
Overall, these should be fairly easy to address but will probably take a while.

Besides that, we should also test the CD pipeline to ensure that the compilers available in the cibuildwheel environments are sufficient for compiling. Additionally, we might have to raise the `MACOSX_DEPLOYMENT_TARGET`. I am not 100% which version is required for C++20 support, but one should be able to find that.

@denialhaag could you take a stab at this? and potentially also turn the PR description into an appropriate issue for a follow up to this PR?

(I am also not quite sure why the release drafter suddenly started failing sporadically)

denialhaag:
> @denialhaag could you take a stab at this? and potentially also turn the PR description into an appropriate issue for a follow up to this PR?

Yes, very happy to!
 
> (I am also not quite sure why the release drafter suddenly started failing sporadically)

I also started noticing that. So far, it has only been happening in Core. 🤔

denialhaag:
I have removed the labels for extensive testing for now; I will add them back later.

.github/workflows/ci.yml
```diff
@@ -154,9 +160,10 @@ jobs:
       cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON -DBUILD_MQT_CORE_BINDINGS=ON
       clang-version: 20
       build-project: true
-      files-changed-only: true
+      files-changed-only: false
```
denialhaag: Set this back to `true` before merging. 

```suggestion
      files-changed-only: true
```

denialhaag:
@burgholzer, I'm quite confident that the remaining linter issues are false positives.

1. I could not use `std::ranges::reverse_view` because my compiler does not support it.
2. I tried quite a bit and also tried to find more information online, but it looks like the MLIR iterators do not work with `std::ranges`.

If you agree, I would ignore the remaining linter issues, and then this should be ready to go in.

UPGRADING.md
```diff
@@ -4,6 +4,8 @@ This document describes breaking changes and how to upgrade. For a complete list
 
 ## [Unreleased]
 
+With this release, MQT Core requires C++20.
```
burgholzer:
```suggestion
With this release, the minimum required C++ version has been raised from C++17 to C++20. All compilers available by default on the systems we test with support all relevant features of the standard. Some frameworks we plan to integrate with even require C++20 by now.
```
Just a little bit more information here. I'd propose to keep it at that and extend the paragraph once we actually use more of the C++20 features.

bindings/dd/register_matrix_dds.cpp
```diff
@@ -20,7 +20,7 @@
 
 #include <pybind11/buffer_info.h>
 #include <pybind11/cast.h>
-#include <pybind11/numpy.h>
+#include <pybind11/numpy.h> // NOLINT(misc-include-cleaner)
```
burgholzer, line 23: Do we actually need that header? (Also holds for the other file)

bindings/ir/operations/register_compound_operation.cpp
```diff
@@ -42,7 +42,7 @@ void registerCompoundOperation(const py::module& m) {
     if (i < 0) {
       i += static_cast<DiffType>(size);
     }
-    if (i < 0 || static_cast<SizeType>(i) >= size) {
+    if (i < 0 || std::cmp_greater_equal(static_cast<SizeType>(i), size)) {
```
burgholzer: I think one of the main advantages of the std::cmp functions is that the static cast is actually not really necessary anymore. Could you try without the explicit casts and see if clang-tidy complains?

mlir/lib/Conversion/MQTDynToMQTOpt/MQTDynToMQTOpt.cpp
```diff
@@ -215,8 +215,10 @@ struct ConvertMQTDynExtract final : OpConversionPattern<dyn::ExtractOp> {
 
     // add an entry to the qubitDataMap to store the indices and the register
     // for the insertOperation
-    qubitDataMap->try_emplace(
-        dynQubit, QubitData{dynQreg, op.getIndex(), op.getIndexAttrAttr()});
+    qubitDataMap->try_emplace(dynQubit,
+                              QubitData{.qReg = dynQreg,
+                                        .index = op.getIndex(),
+                                        .indexAttr = op.getIndexAttrAttr()});
```
burgholzer:
```suggestion
                              dynQreg,
                                      op.getIndex(),
                                        op.getIndexAttrAttr()});
```
Does that actually work? (Modulo whitespace)
The whole point of the emplace methods is that the object is not constructed beforehand.
This comment might apply to multiple places.

src/na/NAComputation.cpp
```diff
@@ -99,11 +98,10 @@ auto NAComputation::validate() const -> std::pair<bool, std::string> {
...
         // Load Operations
         //-----------------------------------------------------------------===//
-        if (std::any_of(opAtoms.begin(), opAtoms.end(),
-                        [&currentlyShuttling](const auto* atom) {
-                          return currentlyShuttling.find(atom) !=
-                                 currentlyShuttling.end();
-                        })) {
+        if (std::ranges::any_of(opAtoms, [&currentlyShuttling](
+                                             const auto* atom) {
+              return currentlyShuttling.find(atom) != currentlyShuttling.end();
```
burgholzer: Shouldn't this be a "contains" check as well?
Might also hold for other places.

src/zx/FunctionalityConstruction.cpp
```diff
@@ -41,8 +42,8 @@ bool FunctionalityConstruction::checkSwap(const op_it& it, const op_it& end,
       const auto tar2 = p.at(op2->getTargets().front());
       const auto ctrl1 = p.at((*op1->getControls().begin()).qubit);
       const auto ctrl2 = p.at((*op2->getControls().begin()).qubit);
-      return ctrl == static_cast<Qubit>(tar1) && tar1 == ctrl2 &&
-             target == static_cast<Qubit>(ctrl1) && ctrl1 == tar2;
+      return std::cmp_equal(ctrl, static_cast<Qubit>(tar1)) && tar1 == ctrl2 &&
```
burgholzer: Static casts necessary?

src/zx/FunctionalityConstruction.cpp
```diff
@@ -624,7 +625,7 @@ ZXDiagram FunctionalityConstruction::buildFunctionality(
 }
 bool FunctionalityConstruction::transformableToZX(
     const qc::QuantumComputation* qc) {
-  return std::all_of(qc->cbegin(), qc->cend(), [&](const auto& op) {
+  return std::ranges::all_of(qc->cbegin(), qc->cend(), [&](const auto& op) {
```
burgholzer:
```suggestion
  return std::ranges::all_of(qc, [&](const auto& op) {
```
This might be even simpler. Probably also applies to other places. Might be worth to search through all occurrences.

src/zx/ZXDiagram.cpp
```diff
@@ -424,6 +426,6 @@ ZXDiagram::getConnectedSet(const std::vector<Vertex>& s,
 }
 
 bool ZXDiagram::isIn(const Vertex& v, const std::vector<Vertex>& vertices) {
-  return std::find(vertices.begin(), vertices.end(), v) != vertices.end();
+  return std::ranges::find(vertices, v) != vertices.end();
```
burgholzer, line 429: I don't know if this is any faster, but it seems a little more idiomatic to use `std::ranges::contains` here and in similar cases.

test/zx/test_simplify.cpp
```diff
@@ -32,7 +33,8 @@ ZXDiagram makeIdentityDiagram(const std::size_t nqubits,
   }
 
   // add identity spiders
-  for (Qubit qubit = 0; static_cast<std::size_t>(qubit) < nqubits; ++qubit) {
+  for (Qubit qubit = 0; std::cmp_less(static_cast<std::size_t>(qubit), nqubits);
```
burgholzer: Static cast necessary?

burgholzer:
As for the clang-tidy warnings:
- did you include the `ranges` header when trying to use it in the IR library?
- are there maybe any MLIR/LLVM specific implementations of the ranges concepts that could be used instead? Some that might be enabled now because the standard has been raised. Might be worth checking.

On a separate note: I was also thinking whether we need to provide an extra setting for clang-format to indicate that we now use C++20. if I remember correctly, the config file defaults to c++17, which is why we don't explicitly specify the standard.
The same question would actually be interesting for the clang-tidy config file, but I don't think that there is something to configure on that.

Thanks for checking 🙏

src/zx/ZXDiagram.cpp
denialhaag: Unfortunately, `std::ranges::contains` seems to exist only for C++23 and later (see [here](https://en.cppreference.com/w/cpp/algorithm/ranges/contains)).

test/zx/test_simplify.cpp
denialhaag: The `static_cast` was indeed not necessary. I removed it. 👍

src/zx/FunctionalityConstruction.cpp
denialhaag: The `static_cast` was indeed not necessary. I removed it. 👍

bindings/ir/operations/register_compound_operation.cpp
denialhaag: The `static_cast` was indeed not necessary. I removed it. 👍

bindings/dd/register_matrix_dds.cpp
denialhaag: Yes, in those two files it is necessary. I tried removing it, but the Python tests failed. The dependency seems to be calls to `format` such as [here](https://github.com/munich-quantum-toolkit/core/blob/main/bindings/dd/register_matrix_dds.cpp#L80).

src/zx/FunctionalityConstruction.cpp
denialhaag: This does not work because `'const qc::QuantumComputation *&' does not satisfy 'input_range'`. If I understand this correctly, changes to `QuantumComputation` class would be necessary to make this work. Maybe this is something that can be done as part of #1069. 🤔 

In general, I tried using the `ranges` functions without `begin` and `end` first, so it should also be necessary in the other places, but I'll have another look.

bindings/dd/register_matrix_dds.cpp
burgholzer: Ah. That makes sense 👍🏻

src/zx/FunctionalityConstruction.cpp
burgholzer: I believe it's just a typo in my suggestion above. If you replace `qc` with `*qc` it might as well work already.
Otherwise, it's of course fine to defer this to a follow-up PR.

src/zx/ZXDiagram.cpp
burgholzer: Ah, damn. I overlooked that. Feel free to resolve this comment then.

mlir/lib/Conversion/MQTDynToMQTOpt/MQTDynToMQTOpt.cpp
denialhaag: That did work! I checked if there are other instances of this, but it seemed to be the only one. 👍 

Sorry for not just applying the suggestion. 😬

src/na/NAComputation.cpp
denialhaag: Thanks for catching that! There were quite a few more places where `contains` could be used. Now it should be used everywhere. 👍

src/zx/FunctionalityConstruction.cpp
denialhaag: You are right; with `*qc` it does work! Thanks! 

There were also a few more instances where we can avoid `begin` and `end` now. 👍

denialhaag (approved):

.clang-format
```diff
@@ -1,3 +1,4 @@
 BasedOnStyle: LLVM
 IncludeBlocks: Regroup
 PointerAlignment: Left
+Standard: c++20
```
denialhaag, line 4:
```suggestion
Standard: c++20

```

PR #900, opened 2025-04-03
:sparkles: Initial implementation of the `mqtdyn` Dialect

DRovara:
## Description

This pull request migrates the incomplete implementation of the `mqtdyn` Dialect from the internal repository and finishes the implementation.

The initial implementation was done by @ystade.

It also contains one initial transformation pass, `constant-folding`, to set up the transformation infrastructure.

---

Before merging, we should consider the following question:
There is quite a lot of shared code (e.g., consider MQTDynTraits.h and MQTOptTraits.h). Are we okay with that or do we want to reuse some code? In case we want more reuse, we probably need to implement some data structures that are shared over the different dialects; the examples above both use their respective dialect's `UnitaryInterface` implementation. Alternatively, maybe C++ templates also do the trick.

## Checklist:


burgholzer:
> Before merging, we should consider the following question: There is quite a lot of shared code (e.g., consider MQTDynTraits.h and MQTOptTraits.h). Are we okay with that or do we want to reuse some code? In case we want more reuse, we probably need to implement some data structures that are shared over the different dialects; the examples above both use their respective dialect's `UnitaryInterface` implementation. Alternatively, maybe C++ templates also do the trick.

I haven't really looked through the code in detail, but in general I would argue that we are currently at a point where we should really try to find the cleanest possible design for what we plan to do. Right now, the code base is still small enough and has no dependents so that we can (more or less) easily make big breaking changes. With every PR that grows the MLIR part, this becomes harder and harder.
That being said, I would very much like to avoid redundancy wherever possible.
If we already need to duplicate quite some code while the code base is small, this is not really a good indication for the general design at the moment. So let's strive for making this as clean as possible.

If I am not mistaken, traits and interfaces are not necessarily bound to a particular dialect, are they? So we could be sharing (elements of) them across dialects?
Any other particular suggestions for keeping it simple here?

mlir/include/mlir/Dialect/Common/IR/CommonTraits.h
```diff
@@ -77,4 +79,75 @@ template <size_t N> class ParameterArity {
...
+    auto measureOp = mlir::cast<ConcreteOp>(op);
+    if (measureOp.getInQubits().size() != measureOp.getOutBits().size()) {
+      return measureOp->emitOpError()
+             << "number of input qubits (" << measureOp.getInQubits().size()
+             << ") " << "and output bits (" << measureOp.getOutBits().size()
+             << ") must be the same";
+    }
+    return mlir::success();
+  }
+};
```
burgholzer, line 111: Couldn't one generalize this trait to not only apply for measurements but more generally? (Or does this still depend on the refactor we talked about offline a couple days back on how controls are represented?)

mlir/include/mlir/Dialect/Common/IR/CommonTraits.h
```diff
@@ -77,4 +79,75 @@ template <size_t N> class ParameterArity {
...
+
+template <typename ConcreteOp>
+class HasOptionalSizeEitherOrTrait
```
burgholzer: The naming confuses me a little here. Wouldn't this be better as something like `UniquelyDefinedSizeTrait` or `UniqueSizeDefinitionTrait` or `ValidSizeDefinitionTrait` or any of the above without the "Definition".

Same holds for at least the trait below.

mlir/include/mlir/Dialect/Common/IR/CommonTraits.h
```diff
@@ -77,4 +79,75 @@ template <size_t N> class ParameterArity {
...
+    auto castOp = mlir::cast<ConcreteOp>(op);
+    const auto hasAttr = op->hasAttr("size_attr");
+    const auto hasOperand = castOp.getSize() != nullptr;
+    if (!hasAttr && !hasOperand) {
+      return op->emitOpError() << "expected an operand or attribute for size";
+    }
+    if (hasAttr && hasOperand) {
+      return op->emitOpError()
+             << "expected either an operand or attribute for size";
+    }
```
burgholzer: These can be combined to a single XOR with an error message that only one may be specified and what the actual provided values where. Makes this a little more precise.
Also holds for other traits of this kind.

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynInterfaces.td
```diff
@@ -0,0 +1,88 @@
...
+            /*methodName=*/  "getAllInQubits",
+            /*args=*/        (ins),
+            /*methodBody=*/  [{}],
+            /*defaultImpl=*/ [{
+                std::vector<mlir::Value> operands{};
+                const auto& inQubits = $_op.getInQubits();
+                operands.insert(operands.end(), inQubits.begin(), inQubits.end());
+                const auto& controls = $_op.getCtrlQubits();
+                operands.insert(operands.end(), controls.begin(), controls.end());
+                return operands;
```
burgholzer: I don't know if it makes much of a difference for the generated code, but these default implementations could be slightly improved to first query both control vectors, then pre-allocating the right size for the resulting vector, and only then emplacing all of them into the rightly sized vector.

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynOps.td
```diff
@@ -0,0 +1,212 @@
...
+//===----------------------------------------------------------------------===//
+
+def MQTDynDialect : Dialect {
+    // The dialect name used in the MLIR file to prefix operations.
+    let name = "mqtdyn";
+
+    let summary = "The MQT dynamic qubit allocation (mqtdyn) dialect.";
+
+    let description = [{
+        // todo: add description
```
burgholzer: Todo 😉

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynOps.td
```diff
@@ -0,0 +1,212 @@
...
+class MQTDynType<string name, string typeMnemonic, list<Trait> traits = []>
+        : TypeDef<MQTDynDialect, name, traits> {
+    let mnemonic = typeMnemonic;
+}
+
+def QubitType : MQTDynType<"Qubit", "Qubit"> {
+    let summary = "A qubit reference.";
+}
+
+def QuregType : MQTDynType<"QubitRegister", "QubitRegister"> {
```
burgholzer, line 64: This is a minor nit and maybe just a personal one, but I stumbled upon this multiple times now, so I might as well say it: "qureg" is fairly uncommon. The typical short version for a qubit register or a quantum register is "qreg".
I'd prefer, if we would stick with that.

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynOps.td
```diff
@@ -0,0 +1,212 @@
...
+    let assemblyFormat = [{
+        `(` $params ( `static` $static_params^ )? ( `mask` $params_mask^ )? `)` attr-dict $in_qubits ( `ctrl` $pos_ctrl_qubits^ )? ( `nctrl` $neg_ctrl_qubits^ )?
+    }];
+}
+
+def XOp : UnitaryOp<"x", [OneTarget, NoParameter]> {
+    let summary = "X operation";
+
+    let description = [{
+        // todo: add description
```
burgholzer: A couple todos here and there throughout this file.
I don't want to postpone the PR here from being merged for too long. So I am fine with leaving these in and tackling them in follow up. We should just make sure that this follow up happens.

mlir/include/mlir/Dialect/MQTDyn/Transforms/Passes.h
```diff
@@ -0,0 +1,35 @@
...
+class RewritePatternSet;
+
+} // namespace mlir
+
+namespace mqt::ir::dyn {
+
+#define GEN_PASS_DECL
+#include "mlir/Dialect/MQTDyn/Transforms/Passes.h.inc" // IWYU pragma: export
+
+void populateFoldExtractQubitPatterns(mlir::RewritePatternSet& patterns);
```
burgholzer: Could we improve the naming of this method slightly? Seems a bit confusing to me.
Maybe by adding a "Constant..." prefix.

mlir/include/mlir/Dialect/MQTDyn/Transforms/Passes.td
```diff
@@ -0,0 +1,24 @@
...
+
+def ConstantFolding : Pass<"constant-folding", "mlir::ModuleOp"> {
+  let summary = "This pass performs constant folding for several operands that might also be added as static arguments";
+  let description = [{
+    This pass transforms constants that are passed to mqtdyn operations as operands into static arguments.
+
+    It is currently compatible with:
+    - mqtdyn.extractQubit
+  }];
+}
```
burgholzer, line 22: This is probably more a question due to my lack of knowledge but maybe you can help me out: couldn't this also be seen as some kind of canonicalization of operations (and implemented as such)? Or can it not because canonicalization merely operates on the information of a single operation?

mlir/include/mlir/Dialect/MQTOpt/IR/MQTOptOps.td
```diff
@@ -537,7 +506,7 @@ def XXplusYY : UnitaryOp<"xxplusyy", [TwoTarget, TwoParameters]> {
     }];
 }
 
-def MeasureOp : GateOp<"measure", [SameVariadicResultSize]> {
+def MeasureOp : GateOp<"measure", [SameVariadicResultSize, MatchingMeasureInOuts]> {
```
burgholzer, line 509: Does the trait here actually add any value compared to the existing trait?

burgholzer (changes requested):
Hey @DRovara 👋🏻 
Thanks for all your work on this. This is looking great.
All the feedback is truly minor and should be easy to address.
Looking forward to merging this once everything is addressed 🙂
And to try to reduce the duplication a little more across these files.

mlir/include/mlir/Dialect/Common/IR/CommonTraits.h
DRovara: Unfortunately, I don't really see a way to solve this. In MLIR, the only way you can access operand "groups" is by their names. If we could use something like `op->getOperand(0)` to access the first *group* of operands, then we could turn this into a parametric trait and have a generic solution.
However, `op->getOperand(0)` will just return the first operand. That means that if the input is a variadic set of arguments, then this will just return the first of these arguments. Therefore, we cannot access the `variadic.size()` method this way.
The only solution is to access the variadic inputs/outputs using the `MeasureOp`'s `.getInQubits()` etc. and that will only work with measurements.

mlir/include/mlir/Dialect/Common/IR/CommonTraits.h
DRovara: Fair point. I have now adapted the names :+1:

mlir/include/mlir/Dialect/Common/IR/CommonTraits.h
burgholzer: Thanks for the explanation. That (unfortunately) makes sense.
Guess that's how it is then 😌

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynOps.td
DRovara: Good point.
However, I suggest we handle that in @ystade's planned update on the dialect definitions / code deduplication.

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynOps.td
DRovara: No problem, this was just part of the code that was not yet fully copied over from the MQTOptOps.td definitions. Resolving those todos was rather straightforward.

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynOps.td
burgholzer: Sounds reasonable to me. @ystade can you keep track of that?

mlir/include/mlir/Dialect/MQTOpt/IR/MQTOptOps.td
DRovara: Yup. `SameVariadicResultSize` only matches *within* ins/outs. So if we have two variadic outs, it will make sure that they have the same sizes. It will not make sure that a variadic ins group also has the same size - that's what this new trait does.

burgholzer: Thanks for the clarification! Makes sense

mlir/include/mlir/Dialect/MQTDyn/Transforms/Passes.td
DRovara: That's actually a good point and honestly I didn't really consider the question before.

You're right, this is definitely a canonicalization pattern. That being said, the `cancelConsecutiveInverses` pattern from the MQTOpt dialect can also be considered a canonicalization (local scope, (largely) keeps semantics intact and constant time complexity). So it seems this is a question that can be extended to further parts of the code base.

Also, the idea here was just to have *some* transformation pass so that the infrastructure is set up already.

My suggestion (feel free to suggest something else) would be: We keep this as is for now and set up an issue that concerns rethinking transformations vs. canonicalizations for the full MLIR code base for a later time.

burgholzer: That seems like a fair suggestion.
The pass here definitely fulfills its role for this PR.
Would you mind creating the respective issue?

mlir/include/mlir/Dialect/MQTDyn/IR/MQTDynInterfaces.td
```diff
@@ -0,0 +1,90 @@
...
+            /*args=*/        (ins),
+            /*methodBody=*/  [{}],
+            /*defaultImpl=*/ [{
+                std::vector<mlir::Value> operands{};
+                operands.reserve($_op.getInQubits().size() + $_op.getPosCtrlQubits().size() + $_op.getNegCtrlQubits().size());
+                const auto& inQubits = $_op.getInQubits();
+                operands.insert(operands.end(), inQubits.begin(), inQubits.end());
+                const auto& controls = $_op.getCtrlQubits();
+                operands.insert(operands.end(), controls.begin(), controls.end());
+                return operands;
```
burgholzer:
```suggestion
                const auto& posCtrlQubits = $_op.getPosCtrlQubits();
                const auto& negCtrlQubits = $_op.getNegCtrlQubits();
                std::vector<mlir::Value> controls{};
                controls.reserve(posCtrlQubits.size() + negCtrlQubits.size());
                controls.insert(controls.end(), posCtrlQubits.begin(), posCtrlQubits.end());
                controls.insert(controls.end(), negCtrlQubits.begin(), negCtrlQubits.end());
                return controls;
            }]>,
        InterfaceMethod<
            /*desc=*/        "Returns all input qubits of the operation incl. all controlling qubits.",
            /*returnType=*/  "std::vector<mlir::Value>",
            /*methodName=*/  "getAllInQubits",
            /*args=*/        (ins),
            /*methodBody=*/  [{}],
            /*defaultImpl=*/ [{
                const auto& inQubits = $_op.getInQubits();
                const auto& controls = $_op.getCtrlQubits();
                std::vector<mlir::Value> operands{};
                operands.reserve(inQubits.size() + controls.size());
                operands.insert(operands.end(), inQubits.begin(), inQubits.end());
                operands.insert(operands.end(), controls.begin(), controls.end());
                return operands;
```
While I am fairly sure that the compiler is clever enough to avoid calling the getter twice, it never hurts to be sure.

burgholzer: Would you mind applying a similar suggestion to the Opt dialect? GitHub won't let me comment on enough lines to do it as part of the review.

burgholzer (changes requested):
This looks great. Just one more minor nit.

burgholzer (approved):
Great 👍🏻 thanks for also adding the Changelog entry 🙂

Issue #644, opened 2024-07-04
✨ Mark and Sweep Garbage Collection for DD Package


burgholzer:
### What's the problem this feature will solve?

As off now, our decision diagram package uses manual reference counting for garbage collection, i.e., every DD `Node` contains a 32bit field tracking a reference count.
The reference count is increased and decreases before and after manipulating decision diagrams.

For example, in classical simulation, the reference count of the initial state (DD) is increased. Then, the quantum operations of the computation are applied one after the other while always increasing the reference count of the new state and decreasing the reference count of the old state.

After each adjustment, the garbage collection routine is called, which checks all the unique unique tables for whether they have reached their (configurable) garbage collection limit.
If one of them reaches their limit, the real collection is started and the respective compute tables are cleared.

On the positive side, this kind of reference counting always allows one to determine the amount of "dead" nodes in the unique tables, i.e., nodes that have a reference count of 0.
If only a single active DD is maintained at a time, this directly gives the size of that decision diagram in `O(1)` complexity. It also allows to directly judge the percentage of active vs dead nodes in the unique tables, which can be used as the basis for a decision to grow the unique table or to conduct garbage collection. However, the latter scenario is hypothetical at the moment, as the tables can currently not be resized.

On the negative side, 
- the reference count has a negative impact on the memory footprint of each DD node. Space savings in each node can be extremely valuable and drastically increase performance.
- the reference count has a negative runtime impact in the sense that constantly increasing and decreasing it consumes a measurable amount of time that would be better spent in actual computations most of the time
- the way we currently update the reference count during simulation is not very favourable to keeping compute table entries around. Since we never increase the reference count of the quantum operation DDs when performing the simulation, any matrix node will always be considered "dead". So even if we were to adopt something like #335, it would always collect the whole MxV multiplication table if there are too many matrix nodes.

In essence, keeping track of the reference count doesn't really get us much, while it leads to several performance throttles.

### Describe the solution you'd like

The literature on Binary Decision Diagrams (BDDs) has established two kinds of garbage collection schemes: reference counting, as we adopted it, or a mark-and-sweep approach, where there is no dedicated reference count field in each node and no continuous reference counting.
Instead, at periodic intervals, the population of the unique tables is checked (our garbage collection check), and if a table contains too many nodes, all active DDs are traversed once in a depth-first fashion where every node is _marked_. (This mark can probably be optimized/hidden in one of the pointer bits of the node)
[...]

PR #908, opened 2025-04-06
✨ Migrate and Refactor Approximation Functionality from DDSim Package

MatthiasReumann:
## Description
This PR adds DD approximation functionality. Most of the functions originate from the DDSim Package. However, instead of simply copying and gluing the packages together, I re-implemented and refactored most of it. 

## Checklist:


src/dd/Simulation.cpp
```diff
@@ -210,37 +249,42 @@ std::map<std::string, std::size_t> sample(const qc::QuantumComputation& qc,
...
+  return out;
 }
+template VectorDD simulate(const qc::QuantumComputation& qc, const VectorDD& in,
+                           Package& dd, const Approximation<None>& approx);
+template VectorDD simulate(const qc::QuantumComputation& qc, const VectorDD& in,
+                           Package& dd,
+                           const Approximation<FidelityDriven>& approx);
+template VectorDD simulate(const qc::QuantumComputation& qc, const VectorDD& in,
+                           Package& dd,
+                           const Approximation<MemoryDriven>& approx);
```
burgholzer: Just a very quick side comment, because this is still early in its development and now's probably the time to make more drastic changes:
I am not the biggest fan of using templates here; especially with the explicit instantiations that become necessary due to code being moved to the source file.
Over the past couple of months, we worked hard to reduce the number of templated classes throughout the DD package as they have become quite a pain.

Couldn't we solve this somehow differently? In principle, any approximation functions takes a DD and produces a DD given certain parameters. Do these functions need state? For certain application strategies probably yes. For certain others, not really.
Maybe we can make something out of that.

src/dd/Simulation.cpp
```diff
@@ -210,37 +249,42 @@ std::map<std::string, std::size_t> sample(const qc::QuantumComputation& qc,
...
-    auto oldW = e.w;
-    // adjust for global phase
-    const auto globalPhase = ComplexValue{std::polar(1.0, qc.getGlobalPhase())};
-    e.w = dd.cn.lookup(e.w * globalPhase);
-    // adjust reference count
-    dd.cn.incRef(e.w);
-    dd.cn.decRef(oldW);
+  if (fp phase = qc.getGlobalPhase(); hasGlobalPhase(phase)) {
+    applyGlobalPhase(phase, out, dd);
   }
```
burgholzer: I like changes like these. These could go into a separate PR though 😌

include/mqt-core/dd/Simulation.hpp
```diff
@@ -42,13 +43,17 @@ namespace dd {
...
  * std::size_t, std::size_t).
  *
+ * @tparam stgy The approximation strategy applied.
  * @param qc The quantum computation to simulate
  * @param in The input state to simulate. Represented as a vector DD.
  * @param dd The DD package to use for the simulation
+ * @param approx The object describing the approximation strategy.
  * @return A vector DD representing the output state of the simulation
  */
+template <const ApproximationStrategy stgy = None>
```
burgholzer:
```suggestion
template <const ApproximationStrategy strategy = None>
```
No need to be cryptic with the names here in general. I'd much prefer to spell these out.

include/mqt-core/dd/Approximation.hpp
burgholzer, line 1: This is really more of a general thought, but do you think it might be possible to avoid having to traverse the full DD twice to perform approximation? Specifically, does one really need to compute all of the node contributions first before starting the approximation? If that can be avoided somehow, this could save quite a bit of memory.
Just a thought 🤷🏼

burgholzer (commented):
Just a couple of thoughts because I skimmed over the PR 😌 
I also approved the CI so that the corresponding checks should run now.

src/dd/Simulation.cpp
MatthiasReumann: Opened a new PR: see #915

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    const Approx& ap = approximate(edge, amplitude * edge.w, budget, dd);
+    if (ap.edge.isTerminal() || ap.contrib > budget) {
+      edges[i] = ap.edge;
+      sum += ap.contrib;
+    } else {
+      edges[i] = vEdge::zero();
+      budget -= ap.contrib;
+    }
+  }
```
burgholzer: If I interpret this correctly, then this is doing a depth-first traversal of the DD. This would lead to bottom-most nodes being eliminated first.
I would argue that a breadth-first traversal would be more beneficial as it would potentially eliminate large subtrees early on.

I would have probably expected this to loop over the edges, check whether any of them has a contribution that is suitable for approximation. If yes, greedily eliminate the edge and update the budget.
Then recurse down on any non-eliminated successor until the budget is exhausted.
After the recursion, if something has been eliminated, the edge weights can be renormalized.

There are two tricky parts that are not yet addressed:
- you want to avoid iterating over the same node multiple times. Once a node has been processed, it is done. Right now, the implementation traverses all paths in the decision diagram, which most of the time will be exponentially many no matter the node count. Most likely, there needs to be some kind of "visited" set for flagging nodes. The `size` method does something like that. Some of the main methods in the package use a Compute Table for that. Yet another option would be a mark and sweep approach as detailed in one of the open issues in this Repo.
- the reference counts are currently not accounted for after approximation. They would need to be decreased for any node that is being eliminated and updated for any node changing as a result of that. The pattern for that can be observed across the code base: store the old edge in a temporary, compute the new edge, increase the ref count of the new edge, decrease the ref count of the old edge.

I am just on my phone right now, so excuse the rather short answer, but I hope this does make sense.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "dd/ComplexNumbers.hpp"
+#include "dd/Node.hpp"
+#include "dd/Package.hpp"
+
+namespace dd {
+namespace {
+struct Approx {
+  vEdge edge;
+  double contrib;
+};
```
burgholzer: Do you actually need that struct?
The contribution is always gonna be the squared magnitude of the corresponding edge weight, right?

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+Approx approximate(const vEdge& curr, const ComplexValue& amplitude,
```
burgholzer: This could probably use a shortcut for when the budget is completely exhausted and there can't be any more approximation.

burgholzer (changes requested):
Thanks for your work on this. Now that the implementation is much smaller, it is also much easier to reason about.
You'll find some comments on the implementation in the file comments.

I hope these make sense 😌

src/dd/Approximation.cpp
MatthiasReumann: No, not really. Could easily be a `std::pair` also. 

Yes, `contrib` equals the squared magnitude.

src/dd/Approximation.cpp
MatthiasReumann: **Thanks** @burgholzer ! Your comments always nudge me towards better solutions. 

I've just realized that node contributions can be calculated much more conveniently (and intuitively) in a breadth-first manner. This is also the way the [old DDSim implementation](https://github.com/cda-tum/mqt-ddsim/blob/2f7f7d673bcd8651bf98954558d347c6f3cd00bd/src/Simulator.cpp#L343) does it. 

Here's an interesting (potentially faulty) thought: Given the contributions of a layer (which sum up to one) isn't it a knapsack problem to find the optimal set of nodes to delete? Wouldn't that be an _optimal_ DD approximation scheme for the given budget? As a side note, the current implementation traverses the tree (breadth-first) layer by layer.

Regarding reference counts: Does calling `decRef` decrease the reference counts of all children nodes? The documentation says it only does so when the reference count of the to be deleted node "hits zero". Does that mean to truly delete a node I must recursively traverse all children and call `decRef` 🤔?

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+            probs[&edge] += parent * mag2(edge.w);
+          }
+        }
+      }
+    }
+  }
+
+  auto& mm = dd.getMemoryManager<vNode>();
+  state = vEdge::normalize(state.p, state.p->e, mm, dd.cn);
+  state.w = dd.cn.lookup(state.w / std::sqrt(mag2(state.w)));
```
MatthiasReumann: I am not sure if this is the correct way to normalize a DD.

include/mqt-core/dd/Approximation.hpp
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @brief Approximate the input state to a given final fidelity.
+ *
+ * @details Traverses the decision diagram layer by layer in a breadth-first
+ * manner and eliminates edges greedily until the budget is exhausted.
+ *
+ * @param state The DD to approximate.
+ * @param fidelity The desired final fidelity after approximation.
+ * @param dd The DD package to use for the simulation
+ */
```
burgholzer: This could potentially use a small addition stating that the fidelity between the original and the resulting state is guaranteed to be _at least_ `@p fidelity`.

I believe this is important to note because it very well might be that the true fidelity is larger.

Just a thought that I haven't fully thought through: would it make sense for this method to return the actual fidelity of the approximation? Is that even accurate/can be tracked throughout? Assume `@p fidelity == 0.1` and the routine cuts off to nodes with a contribution of `0.05` and `0.025`. Is the fidelity between the resulting and the original state then `1.0 - 0.05 - 0.025`?

include/mqt-core/dd/Simulation.hpp
```diff
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include "dd/Approximation.hpp"
```
burgholzer: Let's maybe not touch the `Simulation.[hpp|cpp]` files in this PR at all and just focus on implementing and refining the actual approximation method.
We can tackle its application in the simulation routine in a follow-up PR.
```suggestion
```

src/dd/Simulation.cpp
```diff
@@ -10,9 +10,11 @@
 
 #include "dd/Simulation.hpp"
 
+#include "dd/Node.hpp"
 #include "dd/Operations.hpp"
 #include "dd/Package.hpp"
 #include "ir/Definitions.hpp"
+#include "ir/Permutation.hpp"
```
burgholzer:
```suggestion
#include "dd/Operations.hpp"
#include "dd/Package.hpp"
#include "ir/Definitions.hpp"
```
As per the previous comment.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,106 @@
...
+  approximate(state, fidelity, dd);
+
+  auto norm = getNorm(state.getVector());
+
+  // no nodes deleted. must be the same.
+  EXPECT_EQ(state.size(), 3);
+  // norm must be one.
+  EXPECT_NEAR(norm.real(), 1., 1e-6);
+  EXPECT_NEAR(norm.imag(), 0., 1e-6);
+}
```
burgholzer:
```suggestion
TEST(ApproximationTest, KeepAll) {
  constexpr std::size_t nq = 1;
  constexpr double fidelity = 1;
  Package dd(nq);

  qc::QuantumComputation qc(nq);
  qc.x(0);

  auto state = simulate(qc, dd.makeZeroState(nq), dd);
  approximate(state, fidelity, dd);

  auto norm = getNorm(state.getVector());

  // no nodes deleted. must be the same.
  EXPECT_EQ(state.size(), 2);
  // norm must be one.
  EXPECT_NEAR(norm.real(), 1., 1e-6);
  EXPECT_NEAR(norm.imag(), 0., 1e-6);
}
```
While it does not make a huge difference, I always like to keep the tests as small and focused as possible.
A single-qubit circuit should be sufficient for this basic test.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,106 @@
...
+  auto norm = getNorm(state.getVector());
+
+  // has correct number of nodes.
+  EXPECT_EQ(state.size(), 3);
+  // correct edge is deleted.
+  EXPECT_EQ(state.p->e[1], vEdge::zero());
+  // norm must be one.
+  EXPECT_NEAR(norm.real(), 1., 1e-6);
+  EXPECT_NEAR(norm.imag(), 0., 1e-6);
+}
```
burgholzer: I think it would make sense to add one more test on a single-qubit that uses an `ry` gate similar to here, where the assertion basically is that the resulting DD must be equal to the DD of the `|1>` state.

More generally for the tests here, I think it would be good to also assert the actual state resulting from the approximation being correct.
Given how small these circuits are, that should be fairly trivial.

Last, but not least, a code comment that displays the original state before approximation would probably make is even easier to follow the tests.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+  constexpr auto mag2 = ComplexNumbers::mag2;
+
+  double budget = 1 - fidelity;
+
+  std::unordered_map<const vEdge*, double> probs{{&state, mag2(state.w)}};
+  std::deque<vEdge*> q{&state};
+  while (!q.empty() && budget > 0) {
+    const std::vector<vEdge*> layer(q.begin(), q.end());
+
+    q.clear();
```
burgholzer: I think you could be avoiding this conversion to a vector if you would
- directly iterate over the `deque` (which could probably be a single-linked list as well as you never really need the backwards iteration)
- in every iteration of the while loop, construct a temporary single-linked list that you push to
- at the end of the iteration, `std::move` the temporary list to `q`.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+      if (parent <= budget) {
+        dd.decRef(*lEdge);
+        *lEdge = vEdge::zero();
+        budget -= parent;
+      } else {
+        for (auto& edge : node->e) {
+          if (!edge.isTerminal() && !edge.w.exactlyZero()) {
+            if (probs.find(&edge) == probs.end()) {
+              q.push_back(&edge);
```
burgholzer:
```suggestion
              q.emplace_back(&edge);
```
Does not make a huge difference here, but in general it safes a copy.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+      const double parent = probs[lEdge];
+
+      if (parent <= budget) {
+        dd.decRef(*lEdge);
+        *lEdge = vEdge::zero();
+        budget -= parent;
+      } else {
+        for (auto& edge : node->e) {
+          if (!edge.isTerminal() && !edge.w.exactlyZero()) {
+            if (probs.find(&edge) == probs.end()) {
```
burgholzer: Just a thought: could it happen that one reaches the same `edge`, but with a different parent `contribution`, which would effectively change the approximation threshold? Probably not, right?
That could only happen if we were using `vNode`s instead of `vEdge`s I suppose.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+  std::unordered_map<const vEdge*, double> probs{{&state, mag2(state.w)}};
+  std::deque<vEdge*> q{&state};
+  while (!q.empty() && budget > 0) {
+    const std::vector<vEdge*> layer(q.begin(), q.end());
+
+    q.clear();
+    for (vEdge* lEdge : layer) {
+      vNode* node = lEdge->p;
+      const double parent = probs[lEdge];
```
burgholzer: Would `contribution` be a more suitable name here?

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+      if (parent <= budget) {
+        dd.decRef(*lEdge);
+        *lEdge = vEdge::zero();
```
burgholzer: Unfortunately, I don't think it is as easy as this.
First of all, you are decreasing the reference count here, without increasing it elsewhere, which means that something is lost along the way. Skip the dereferencing for now. I'll add a separate comment somewhere else to address this.
Anyhow, it would be great to add a test that catches this kind of failure so that it does not happen again. For this, you could:
- create a state to be approximated
- approximate the state
- call `decRef` on the resulting state
- call `dd->garbageCollect` with the `force` flag set to on.
- assert that the vector unique table is completely empty since one would expect that all nodes have been collected.

I am fairly certain that your current implementation will trigger a segfault on the `decRef` call because it tries to decrease the refcount of a node that has already reached zero.

---

Furthermore, you are setting a particular `edge` to zero. Assuming that the edge was not zero before (which will always be the case in approximation), this makes its parent node unnormalised. However, you do not have a reference to the parent node available here, so there is no way you could be performing the normalization.
As it stands, this is modifying the successors of a `vNode` in-place. This affectively changes their hash and, by that, would also mean that they need to be placed in a different unique table bucket.
As you see it's complicated.

It is not enough to simply traverse the DD downward and remove edges. There needs to be an upwards traversal as well that updates all affected nodes.
You can basically see this in the old DDSIM implementation (specifically, the `removeNodes` function): https://github.com/cda-tum/mqt-ddsim/blob/00c599505255273f3be3bdf3489f0b33e1e5b3ad/src/Simulator.cpp#L305-L329
The interesting part for you is probably: https://github.com/cda-tum/mqt-ddsim/blob/00c599505255273f3be3bdf3489f0b33e1e5b3ad/src/Simulator.cpp#L322-L326
which constructs a new pair of edges and calls the `makeDDNode` method. That is the kind of pattern you are looking for in your implementation as well.

It would be great to craft a test that would surface this issue so that it becomes easier to judge when it no longer occurs.
One way to do this is to
- build a state that can be approximated and has a known approximation
- approximate the state
- build the approximated state from scratch
- assert that the two DDs are the same, i.e., they point to the same node and have the same edge weight.

You current implementation should fail this test.

---

One major difference between the old DDSIM implementation and the new one here, is that the new implementation tries to delete edges rather than nodes. I have not fully thought this through, but I think this should actually make things easier

src/dd/Approximation.cpp
burgholzer: If you take care of the main comment in this review, then these lines here should not be needed.
What you will probably want to do instead, is to correct the reference counts here.
In other major DD operations, reference counts are only ever updated after the operation has been applied. The typical pattern is
- increase ref count of new state
- decrease ref count of old state
- set old state to new state
- call garbage collection

you can kind of see this here: https://github.com/munich-quantum-toolkit/core/blob/72e5e2c993b82367966b7b5f6cbafd63cad653a4/src/dd/Package.cpp#L894-L900

---

The way reference counting works (to also answer one of the open questions from the previous review) is that:
- on increase, if the ref count of the node is non-zero, it will simply be increased by one. If it is zero (i.e., this is a new node), the ref count of the node will be set to 1 and `incRef` will be called on all children of the current node.
- on decrease, if the ref count will be decreased by 1. If it does not hit 0, the operation ends. If it hits zero, `decRef` is recursively called on all children.

---

To keep the reference count after approximation correct, you probably want to replicate the above pattern: increase the new, decrease the old, garbageCollect.

At least this is the easiest solution. The alternative would be to update the reference count on the fly, which would safe some traversals of the DD, but is way harder to get right.
Given how we are currently thinking about dropping this kind of reference counting entirely (see https://github.com/munich-quantum-toolkit/core/issues/644), I would not bother trying.

burgholzer (changes requested):
Thanks for you continued work on this! 🙏🏼 
The review here turned out a bit longer than expected. ups.
I hope it provides you with enough context to understand, why the current solution is not yet working correctly and also how you could be adding tests that assert it works correctly.
I think generally, writing more dedicated tests that assert the right functionality could really be key here.
Let me know if you have any further questions!

src/dd/Approximation.cpp
burgholzer: I think the knapsack analogy is accurate. I am not 100% sure what the actual notion of optimality would eventually be, but it would be some kind of optimal solution.
Might be worth keeping in mind for a follow-up to the greedy strategy being implemented here.

As for the ref-count: see the other comment :o)

include/mqt-core/dd/Approximation.hpp
MatthiasReumann: Very good point. I've changed the description accordingly. 

---

> [...] the routine cuts off to nodes with a contribution of 0.05 and 0.025. Is the fidelity between the resulting and the original state then 1.0 - 0.05 - 0.025?

Yes, should be so. 

Let $|\psi\rangle$ be the state to be approximated. Use the Projector $P=\sum_{i\in{}I}|i\rangle\langle{}i|$ with a given set $I$ which (effectively) truncates amplitudes in a state vector. Thus, the approximated state is:

$$
|\psi_{I}\rangle = \frac{P_{I}|\psi\rangle}{||\psi\rangle||_{l2}}.
$$

Note that
1. $P_{I} = P_{I}^{\dagger}$[^1]
2. $P_{I}|\psi_{I}\rangle = |\psi_{I}\rangle$[^2]
3. $\Vert\psi_{I}\rangle\Vert_{2} = \sqrt{F(|\psi\rangle, |\psi^{I}\rangle)}$[^2]

Define Fidelity as $F(|\psi\rangle, |\phi\rangle) = |\langle\psi{}|\phi\rangle|^2$.

Then, if I'm not mistaken:

equation-3



Consequently, the fidelity between the source and approximated state is exactly the sum of contributions of the remaining amplitudes (dot product; also note that the renormalization is missing), or `1 - {contribution of removed amplitudes}`.

---

In theory, yes, we could return the fidelity. The old DD-Sim implementation does so also. I guess the question (which you probably can answer best) is, what would the "enduser" do with this information? As stated in the original paper, one use case would be to calculate the final fidelity by multiplying the individual ones. Could be useful for the `simulate` function with approximation. What do you say?

[^1]: (Nielson Chuang, Quantum Computation and Quantum Information P.70)
[^2]: (Hillmich, As Accurate as Needed, as Efficient as Possible:
Approximations in DD-based Quantum Circuit Simulation P.5)

burgholzer: Thanks for the derivation! That looks totally reasonable 👍🏼 
I think overall, it is just a quantity that is easy to compute and _might_ help for certain scenarios.
In general, repeated approximations might benefit from more accurate estimates of the available budget from previous approximations.
So I do not have a very clear use case in mind, but I believe there might be some.

src/dd/Approximation.cpp
MatthiasReumann: Yes that could happen! That's why we add the contributions. This shouldn't be a problem since we calculate all the contributions for the next layer based on the current layer contributions. (Essentially: We only check the threshold once all partial contributions have been added.)

The following DD for $(H\otimes{}H)|00\rangle$ shows an example where we reach the same edge with a different parent contribution.

Screenshot 2025-04-29 at 09 27 12


Please let me know if I'm mistaken.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+///-----------------------------------------------------------------------------
+///                      \n simulate with approximation \n
+///-----------------------------------------------------------------------------
+
+TEST(ApproximationTest, OneQubitKeepAllBudgetZero) {
+
+  // Test: If the budget is 0, no approximation will be applied.
+  //
+  // |state⟩ = 0.866|0⟩ + 0.5|1⟩
```
burgholzer: That's not the state created by this test circuit😉

I suspect you might want to change the circuit.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,239 @@
...
+  constexpr double fidelity = 1;
+
+  auto dd = std::make_unique<dd::Package>(nq);
+
+  qc::QuantumComputation qc(nq);
+  qc.x(0);
+
+  auto state = simulate(qc, dd->makeZeroState(nq), *dd);
+  auto p = approximate(state, fidelity, *dd);
+  auto approx = p.first;
```
burgholzer: Probably just a matter of taste, but this could look a little cleaner using structured bindings and an additional assertion that the returned fidelity is 1.
Something like
```
const auto& [approx, fid] = approximate(state, fidelity, *dd);
EXPECT_EQ(fid, fidelity);
```

Similar reasoning can be applied to the other tests.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,239 @@
...
+  const CVec expected{{0}, {1}};
+  EXPECT_EQ(approx.getVector(), expected);
+  EXPECT_EQ(approx.size(), 2);
+}
+
+TEST(ApproximationTest, OneQubitKeepAllBudgetTooSmall) {
+
+  // Test: If the budget is too small, no approximation will be applied.
+  //
+  // |state⟩ = 0.866|0⟩ + 0.5|1⟩
```
burgholzer: Also not the state being prepared in the test.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,239 @@
...
+  qc::QuantumComputation qc(nq);
+  qc.ry(qc::PI / 3, 0);
+
+  auto state = simulate(qc, dd->makeZeroState(nq), *dd);
+  auto p = approximate(state, fidelity, *dd);
+
+  dd->decRef(p.first);
+  dd->garbageCollect(true);
+
+  EXPECT_EQ(dd->vUniqueTable.getNumEntries(), 0);
```
burgholzer: I'd argue that this could be combined with the previous test.
The description comment is still helpful and could also be copied over.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,239 @@
...
+  qc::QuantumComputation qc(nq);
+  qc.h(0);
+  qc.h(1);
+  qc.ry(qc::PI / 3, 0);
+
+  qc::QuantumComputation qcRef(nq);
+  qcRef.h(0);
+  qcRef.x(1);
+  qcRef.cx(0, 1);
+  qcRef.cx(1, 0);
```
burgholzer: Maybe I am stupid, but the resulting state is a product state, so single-qubit operations are completely sufficient to construct the reference state.
Specifically,
```suggestion
  qcRef.x(0);
  qcRef.h(1);
```
Should just work.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+TEST(ApproximationTest, ThreeQubitRemoveNodeWithChildren) {
+
+  // Test: Remove node that has a subtree attached (i.e. has children).
+  //
+  // |state⟩ = 0+0.866j|000⟩ + 0-0.096j|100⟩ + 0+0.166j|101⟩ + 0+0.231j|110⟩
+  //           + 0-0.4j|111⟩
+  //
+  // Eliminate parent of |1xx⟩ with contribution ~ 0.25.
+  //     → |approx⟩ = i|000⟩
```
burgholzer, line 277: I really really like these test descriptions! If only all tests in all projects were written like that.

If I could wish for one thing, which probably is not really feasibly or too much effort, it would be an actual visualization of the respective decision  diagrams in ASCII art to illustrate the approximation. That would be the cherry on top in all these tests. 🍒

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace dd {
+namespace {
+/**
+ * @brief Recursively rebuild @p state. Exclude edges contained in @p exclude.
+ * @return Rebuilt VectorDD.
+ */
+VectorDD rebuild(const VectorDD& state,
+                 const std::forward_list<const vEdge*>& exclude, Package& dd) {
+  const auto it = std::find(exclude.begin(), exclude.end(), &state);
```
burgholzer: This has a complexity of O(n), where `n` is the length of the list.
Since this kind of membership check is executed for every node in the decision diagrams, this turns out to be quite costly.
If you swap the `std::forward_list` for an `unordered_set` the membership check complexity is reduced to O(1). That should yield a reasonable speedup.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  if (state.isTerminal()) {
+    return state;
+  }
+
+  const std::array<vEdge, RADIX> edges{rebuild(state.p->e[0], exclude, dd),
+                                       rebuild(state.p->e[1], exclude, dd)};
+
+  VectorDD edge = dd.makeDDNode(state.p->v, edges);
+  edge.w = dd.cn.lookup(edge.w * state.w);
```
burgholzer: This is really deep in the micro optimization domain, but I think this could benefit from an extra check, whether the multiplication and the lookup are actually necessary. Something like (untested):
```
if (state.w.exactlyOne()) {
  return edge;
}
edge.w = dd.cn.lookup(edge.w * state.w);
```

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,100 @@
...
+                next.begin(), next.end(),
+                [&eChild](const auto& p) { return p.first == eChild; });
+            if (it == next.end()) {
+              next.emplace_front(eChild, childContribution);
+            } else {
+              (*it).second += childContribution;
+            }
+          }
+        }
+      }
```
burgholzer: This might just be me and my personal programming preference, so there is definitely no need for you to change that, but just so that I have said it.

I am a big fan of reducing the levels of nesting in code.
I would probably place a `continue` statement in the first if block and de-nest the subsequent else if, by flipping the if condition and continuing if the condition is not met.
That reduces at least one level of indirection, if not two.
At times, this can really help to make the code a little cleaner and seam less complex.
The `exactlyZero` check could also be flipped to de-nest the subsequent code.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        for (const vEdge& eChildRef : n->e) {
+          const vEdge* eChild = &eChildRef;
+
+          if (!eChild->w.exactlyZero()) {
+            const double childContribution =
+                contribution * ComplexNumbers::mag2(eChild->w);
+            const auto it = std::find_if(
+                next.begin(), next.end(),
+                [&eChild](const auto& p) { return p.first == eChild; });
```
burgholzer: Thinking about the complexity here, I believe this could also benefit from using an unordered_set.
Iteration speed is similar. Lookup is way faster.
Should be a no brainer.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,100 @@
...
+          }
+        }
+      }
+    }
+
+    curr = std::move(next);
+  }
+
+  VectorDD approx = rebuild(state, exclude, dd);
+  approx.w = dd.cn.lookup(approx.w / std::sqrt(ComplexNumbers::mag2(approx.w)));
```
burgholzer: I was about to write that this computation is likely to be redundant because you could simply set the top weight to one, but that would discard the global phase.

What could work though:
Before starting the approximation, safe the top edge weight of the incoming DD.
Then set the weight to exactly one and start the approximation procedure.
At the end, set `approx.w` to the initial edge weight.

I am fairly certain that the top edge weight is not affected by approximation, so the above simplification should be valid.

include/mqt-core/dd/Approximation.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace dd {
+
+/**
+ * @brief Approximate the @p state based on fidelity. The fidelity of the
+ * approximated state will be at least @p fidelity.
+ *
+ * @details Traverses the decision diagram layer by layer in a breadth-first
+ * manner (iterative deepening algorithm) and eliminates edges greedily until
+ * the budget (1 - @p fidelity) is exhausted.
```
burgholzer, line 35: Once all the details of the implementation are clarified, I believe it would be helpful to add a brief description of the complexity of the approximation operation to the docstring here.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <forward_list>
+#include <utility>
+
+namespace dd {
+namespace {
+/**
+ * @brief Recursively rebuild @p state. Exclude edges contained in @p exclude.
+ * @return Rebuilt VectorDD.
+ */
+VectorDD rebuild(const VectorDD& state,
```
burgholzer: At the moment, this method traverses the full DD down to the terminals. And it does so without checking whether a certain successor has already been processed. As such, this function will necessarily scale exponentially in the number of qubits since it will visit every possible path through the decision diagram (which, in a decision diagram without zero terminals, are always exponentially many).
This is undesirable and seems unnecessary.

Furthermore, with this function, there is a second top-down traversal of the decision diagram in the approximation routine. One for collecting the nodes to be eliminated, and another one for eliminating the nodes themselves. I am fairly certain that there is a way to directly eliminate the nodes as part of the main descent through the decision diagram.
This might not be as apparent because the main approximation loop is implemented iteratively instead of recursively (like this function is).
However, I think this can be made to work even in the iterative case.
This would also avoid one of the drawbacks here that the current method really rebuilds the entire DD; even when there is nothing to eliminate.
Just as an example for a potential optimization (which will probably be obsolete if this two-step pass is avoided), if I look at the set of edges to be excluded an none of them has a variable index smaller than X, then I do not need to descend further than to level X+1 because nothing will change below that level.

burgholzer (changes requested):
I really really like where this is going. Thanks for your continued work on this.
The structure of the approximation has already become way cleaner and clearer.
As you will see from the comments, there are a couple of complexity aspects that are still suboptimal and where I believe improvements make sense.
Most of these should be fairly easy to incorporate. The hardest one will probably be avoiding the two-stage approximation traversal. But, at the same time, it's also the most drastic performance blocker.

I would be tempted to say that some tests that track the complexity of the approximation would be in order, but I know from experience that these are near impossible to write. Or they require some kind of additional tracking, which results in an overhead memory as well as runtime wise.
Some kind of middle ground could be to run the tests of the test suite through a profile and check the call counts for certain critical functions. This would probably reveal the exponential complexity of the current rebuild function.
Maybe these thoughts trigger an idea on your side for how to further improve the tests (which are already really great!)

burgholzer:
Ah, and don't be worried about the failing Python tests in the CI. These are unrelated to your changes and will be more or less automatically resolved soon 😌

test/dd/test_approximations.cpp
MatthiasReumann: thanks & sorry🤦🏼

test/dd/test_approximations.cpp
MatthiasReumann: You are absolutely right.

test/dd/test_approximations.cpp
MatthiasReumann: 🍒 it is.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,390 @@
...
+      const std::size_t numEdges = (state.size() * 2) + 1;
+
+      const auto t1 = std::chrono::high_resolution_clock::now();
+      approximate(state, fidelity, *dd);
+      const auto t2 = std::chrono::high_resolution_clock::now();
+      const std::chrono::duration<double, std::micro> runtime = t2 - t1;
+
+      // Runtime should be sub-linear in terms of edges.
+      // 500μs is an "acceptable" difference.
+      EXPECT_TRUE(runtime.count() - static_cast<double>(numEdges) < 500);
```
burgholzer: In principle, I very much like this test. In practice, it is rather prone to become flaky as it very much depends on the actual hardware the test is executed on.
If I perform this on bad enough hardware, it is bound to fail.
Would it make sense to introduce some kind of statistics tracking that would allow us to assert this condition more robustly (even if that costs a little extra memory and runtime)?
Something like the number of total edges processed in the queue.
Based on how the rest of this review goes (this was the first comment), further/other statistics might make sense.

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,390 @@
...
+      // Each state has two outgoing edges + root edge.
+      const std::size_t numEdges = (state.size() * 2) + 1;
+
+      const auto t1 = std::chrono::high_resolution_clock::now();
+      approximate(state, fidelity, *dd);
+      const auto t2 = std::chrono::high_resolution_clock::now();
+      const std::chrono::duration<double, std::micro> runtime = t2 - t1;
+
+      // Runtime should be sub-linear in terms of edges.
+      // 500μs is an "acceptable" difference.
```
burgholzer:
```suggestion
      // 500us is an "acceptable" difference.
```
I'd rather like to avoid using unicode symbols in code (documentation). While it is not really important here, we had cases where this would mess up the documentation PDF build (yay, latex can be great...).

test/dd/test_approximations.cpp
```diff
@@ -0,0 +1,390 @@
...
+  const std::array<vEdge, RADIX> edges{
+      buildExponentialDD(v + 1, height, dd, dis, gen),
+      buildExponentialDD(v + 1, height, dd, dis, gen),
+  };
+
+  vEdge eNew = dd.makeDDNode(v, edges);
+  eNew.w = dd.cn.lookup(static_cast<fp>(dis(gen)));
+
+  return eNew;
+}
```
burgholzer: Written like this, I am pretty sure that the resulting DD is not properly normalized.
The top-edge weight of the final DD is a random number in `[0, 1]`.

I also think that the last lookup (after the recursive calls) is not really necessary since the originally created edge weights are already random.

Finally, the variable order is mixed up as the current implementation will have index 0 at the top and index n-1 at the bottom of the DD if I am not completely mistaken.

There are multiple ways to accomplish what you want to be doing here a little more straight forward I believe.

The first thing that comes to mind is to use two functions. One that basically does
```c++
if (v == static_cast<Qubit>(0)) {
  auto e = vEdge::one();
  e.w = dd.cn.lookup(static_cast<fp>(dis(gen)));
  return e;
}

const std::array<vEdge, RADIX> edges{
  buildExponentialDD(v-1, dd, dis, gen),
  buildExponentialDD(v-1, dd, dis, gen),
};
```
and one top-level function that essentially does
```c++
auto e = buildExponentialDD(height, dd, dis, gen);
e.w = dd::Complex::one();
```

An alternative would be to simply generate a random (valid) state vector and use the `from_vector` method to get the resulting DD. Given how the DD is supposedly exponentially large anyway, the exponentially large vector won't really hurt.

Third alternative, but that is probably something for a different PR, would be to turn this into a slightly more pragmatic solution that can be customised to create DDs of arbitrary shape and can be made part of the main library. I started something like that in the past, but never really followed through and continued to work on it. See https://github.com/munich-quantum-toolkit/core/commit/159bbf173be0993ecdd3ef440ef082f0f38a555a if you are interested

src/dd/Approximation.cpp
burgholzer, line 1: First things first: this new implementation seems to work and that is a great achievement already. The solution is also improving with every iteration, which is great to see.

While the previous solution did too little, I believe this one does a little too much in terms of bookkeeping and data structures it allocates and manages. Especially the child-to-parent map will eventually be as large as the original DD (which might be exponentially large). However, I currently do not really have the capacity to come up with a full proposal on what to change. In the end, it would be great to not scale with `|E|` (number of edges in the diagram), but rather with `|V|` (number of nodes in the diagram), or some combination thereof.
It would probably be nice to also include this as a metric in the benchmarks to be run.
Generally, it would probably be interesting to see benchmarks on non-worst-case decision diagrams in order to judge the real world performance of the approximation method.

One potential avenue for improvement is to somehow make use of the `flags` member of the `NodeBase` class to eliminate the `PathFlags` variable and, instead, using the node internal flags to mark edges for removal.

I have looked a little bit into the literature on (classical) BDD approximation, which yielded https://mcmil.net/pubs/DAC98.pdf, where the authors use a three-step procedure of (analyse) - (mark) - (rebuild).
Potentially also interesting: https://par.nsf.gov/servlets/purl/10465240
Maybe you could quickly skim these as well to see if there are any synergies to be exploited here.

burgholzer (changes requested):
Many thanks for yet another great iteration on the code here 🙏🏼 
Really fun to think about this.
And I have to say again: I really like the testing. This is how tests should be written 🙌🏼 

I did not have enough time to really think every aspect of the current implementation through, but something is telling me that this is not yet 100% it. I hope the rather long comment helps to point you in some directions that can (1) improve the overall implementation even further and (2) lead to even more comprehensive benchmarking of the method to better understand its complexity.

Keep up the good work 🙏🏼

include/mqt-core/dd/Approximation.hpp
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param state The DD to approximate.
+ * @param fidelity The desired minimum fidelity after approximation.
+ * @param dd The DD package to use for the approximation.
+ * @param meta An optional struct which will be filled with meta information.
+ * @return The fidelity between the source and the approximated state.
+ */
+double approximate(VectorDD& state, double fidelity, Package& dd,
+                   struct ApproxMeta* meta = nullptr);
```
burgholzer: How about making this a `std::optional<ApproxMeta>&`?
Seems like the more C++-like solution to me.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  double budget = 1 - fidelity;
+  Qubit min{std::numeric_limits<Qubit>::max()};
+
+  TerminalList candidates{};
+
+  ApproxMeta rmeta{};
+
+  while (budget > 0) {
+    Contributions c; // Stores contributions of the next layer.
```
burgholzer: Just a thought: could it make sense to eliminate this map and, instead, directly store the respective information in the `next` layer?
Would eliminate the duplication of information. But probably slower, because adding contributions to existing entries is O(layer length) instead of O(1).
Maybe still worth thinking about.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+      curr.pop();
+      rmeta.nodesVisited++;
+
+      // If possible, flag a node for deletion and decrease the budget.
+      // If necessary, reset the lowest qubit number effected.
+      if (budget >= n.contribution) {
+        n.ptr->flags = FLAG_DELETE;
+        budget -= n.contribution;
+        min = std::min(min, n.ptr->v);
+        removeAbove(candidates, budget);
```
burgholzer: Instead of constantly updating this list, it should be sufficient to do this once at the end of the main loop, right?

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+      }
+
+      // Compute the contributions of the next layer.
+      for (std::size_t i = 0; i < RADIX; ++i) {
+        const vEdge& eRef = n.ptr->e[i];
+        const double contribution =
+            n.contribution * ComplexNumbers::mag2(eRef.w);
+
+        if (eRef.isTerminal()) {        // Don't add terminals to the queue.
+          if (budget >= contribution) { // They (potentially) can be deleted.
```
burgholzer: If I am reading this correctly, it is adding all zero terminals to the list. (they have zero contribution)
But eliminating zero terminals does not make any sense.
I suppose the second if statement here should start with a check that the weight is not exactly zero.

If I remember correctly, there might even be a `isNonZeroTerminal` check that might be helpful here.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    Layer next{}; // Prioritize nodes for next iteration.
+    for (auto& [n, contribution] : c) {
+      next.emplace(n, contribution, budget);
+    }
+
+    if (next.empty()) { // Break early. Avoid std::move.
+      break;
+    }
```
burgholzer: Could break even earlier than that before `next` is being constructed. Doesn't safe much, but I think the intent becomes even clearer.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+  // If a node is flagged, reset the flag and return a zero edge.
+  if (n->flags == FLAG_DELETE) {
+    n->flags = 0U;
+    return vEdge::zero();
+  }
+
+  // If a node has been visited once, return the already rebuilt node
+  // and set the edge weight accordingly.
+  if (l.find(n) != l.end()) {
+    vEdge eR = l.at(n);
```
burgholzer: While I am quite sure that the compiler generates optimized code for this already, this is still technically conducting two lookups. The typical pattern for this is
```suggestion
  if (auto it = l.find(n); it != l.end()) {
    vEdge eR = it->second;
```

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+  while (budget > 0) {
+    Contributions c; // Stores contributions of the next layer.
+    while (!curr.empty()) {
+      const LayerNode n = curr.top();
+      curr.pop();
+      rmeta.nodesVisited++;
+
+      // If possible, flag a node for deletion and decrease the budget.
+      // If necessary, reset the lowest qubit number effected.
+      if (budget >= n.contribution) {
```
burgholzer: Maybe that's just my brain being wired wrongly, but somehow the order in these comparisons triggers me.
I'd find it more "logical"/natural for all of these to read
If (contribution <= budget)

Maybe you agree.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+  // Lastly, check if any terminals can be deleted.
+  while (!candidates.empty()) {
+    const auto [n, flag, contribution] = candidates.front();
+    candidates.pop_front();
+    if (budget >= contribution) {
+      n->flags = FLAG_DELETE + flag;
+      budget -= contribution;
+      min = std::min(min, n->v);
+    }
+  }
```
burgholzer: There is no real need to manipulate the list here. It can simply be traversed in a read-only fashion from beginning to end (or until the budget is exhausted).

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+    eR.w = curr.w;
+    return eR;
+  }
+
+  // Otherwise traverse down to rebuild each non-terminal edge.
+  std::array<vEdge, RADIX> edges{};
+  for (std::size_t i = 0; i < RADIX; ++i) {
+    const vEdge& eRef = n->e[i];
+
+    if (eRef.isTerminal()) {
```
burgholzer, line 187: Could probably also use a shortcut for when this is a zero terminal.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+      edges[i] = eRef;
+      continue;
+    }
+
+    edges[i] = sweep(eRef, min, l, dd);
+  }
+
+  // Rebuild node and set its edge weight accordingly.
+  vEdge eR = dd.makeDDNode(n->v, edges);
+  eR.w = curr.w;
```
burgholzer, line 211: This seems to work, but I am not quite sure I understand why at the moment.
Couldn't the makeDDNode function already return a non-one edge weight?
Why is the current edge weight involved here at all?
Probably just me sitting on the phone and overlooking something 😄

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,246 @@
...
+  l[n] = eR;
```
burgholzer, line 212: This lookup table of replacements might grow quite big, right?
Eventually, it will contain all nodes of the entire DD that need to be replaced.

We might want to optimize this somehow.
What I am thinking at the moment:
Assuming that the reference tracking for the DD is accurate and that there is only one active DD in the Dd package, the reference count of a node indicates by how many nodes it is reached.
One could manually decrement the counter for each replacement being performed (while increasing the reference count of the replacement). If the reference count of the replaced node hits 0, it can be removed from the lookup map.
Specifically, nodes with a reference count of 1 don't even need to be added to the replacement map. This would also get rid of the reference counting after approximation I suppose.

Since it is not known, whether we really want to keep the reference count in the nodes in the future (or completely replace them with mark and sweep), it could be worth thinking about whether this counting information can be incorporated into the Mark phase (using the flags for example).

This is clearly on the high end of optimizations. I'd be totally fine with just going for the reference counting solution here.
Actually, I'd be fine with the existing solution and the above could be tackled in a follow-up.

src/dd/Approximation.cpp
MatthiasReumann: Funnily enough, I think we already had such a solution some time ago 🤔 I think we dropped it because of the O(L) runtime - which makes sense in my opinion.

src/dd/Approximation.cpp
MatthiasReumann: You are absolutely right. Thanks for pointing out - totally missed that.

src/dd/Approximation.cpp
burgholzer: Hahah. That could very well be. Then let's stick with this for now. Enough shenanigans 😉

include/mqt-core/dd/Approximation.hpp
MatthiasReumann: You know, I think I spent like 15 Minutes thinking about using `nullptr` vs `std::optional`. 

In the end I decided to use the nullptr because using a reference like that it isn't possible to use`std::nullopt` as default parameter. I feel like this is a deal breaker regarding API design.

src/dd/Approximation.cpp
MatthiasReumann: > [...] and that there is only one active DD in the Dd package

Is this something we can assume on a general basis? 

---
I really like the idea of using the `flags` for that. I think this should be possible. 
It is probably reasonable to do this in a follow-up 👍🏻

include/mqt-core/dd/Approximation.hpp
burgholzer: Ah. Damn it. That would only work for a const reference.
In that case, the current solution is perfectly fine. 👍🏻

src/dd/Approximation.cpp
burgholzer: In almost all cases, where the approximation would be used, I would say so.
One could conceive situations where this is not true. For example during equivalence checking (done in our QCEc library), where two circuits are simulated simultaneously and where the resulting states are compared after the simulations.
However, even in these cases, the reference count is an upper bound to how many nodes in the current DD might be referring to the node. That is, some of the counters might just hit zero as part of the approximation and, hence, will remain in the lookup map.

With the dedicated Marking method for counting, this would be avoided because one would only count the nodes from the DD to be approximated.

Let's safe this for a follow-up though.

include/mqt-core/dd/Approximation.hpp
MatthiasReumann: What's your opinion on returning an `std::optional<ApproximationMeta>` and only returning it if a flag is set to true, e.g.: `includeMeta=true`, otherwise `std::nullopt`? In some sense the fidelity returned is meta information too, right?

burgholzer: How about taking it one step further and returning an `ApproximationMetadata` object (i think id slightly prefer the spelled out name) that has some std::optional members that are only populated if a flag is set?

MatthiasReumann: Sounds good. However, now if I think about it, the meta data is always calculated. Should we just drop the `std::optional` altogether? If only the fidelity is required one can always do

```cpp
const auto& [fidelity, ...others] = approximate(...)
```

burgholzer: Fine with me 👍🏻

src/dd/Approximation.cpp
MatthiasReumann: The way I think about the recursive function `sweep` is that it _rebuilds_ the node attached to `curr`. I also think about each node being a decision going left with probability `p` or right with `1-p`, respectively. 

If we were to rebuild a DD without any deletion, we would be assigning the current edge weight to the rebuilt node to keep those probabilities intact.

Now, if we delete a node by "setting" its ingoing edge to `vEdge::zero()` the parent node's `makeDDNode` call will normalize the node to have probabilities `1` for the edge (or decision) kept and `0` for the deleted one. The ingoing edge of the parent node will now have the norm as a weight. However, this normalization factor should not affect the probabilities of the decision on the layer above. Hence, again we simply assign `eR.w = curr.w` and eliminate the normalization factor. In other words, we "forget" that there was a probability `p < 1` that we could visit this edge - now it's a one-way street. This pattern repeats until the root edge.

Hope this makes somewhat sense 🙈

burgholzer: Yeah, thanks for spelling it out. After reading this, it makes sense.

Maybe this is something to add as a code comment in some way. That might safe future me or others to scratch their head again 😌

include/mqt-core/dd/Approximation.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * approximated state will be at least @p fidelity.
+ *
+ * @details Traverses the decision diagram layer by layer in a breadth-first
+ * manner (iterative deepening algorithm) and eliminates edges greedily until
+ * the budget (1 - @p fidelity) is exhausted.
+ *
+ * @param state The DD to approximate.
+ * @param fidelity The desired minimum fidelity after approximation.
+ * @param dd The DD package to use for the approximation.
+ * @param meta An optional struct which will be filled with meta information.
```
burgholzer:
```suggestion
```
This is no longer a parameter of the method.

include/mqt-core/dd/Approximation.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * @return An ApproximationMetadata object.
```
burgholzer:
```suggestion
 * @return Metadata about the approximation.
```

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,235 @@
...
+using Lookup = std::unordered_map<const vNode*, vEdge>;
+/// @brief A node, a flag indicating left or right, and its contribution.
+using Terminal = std::tuple<vNode*, uint16_t, double>;
+
+/**
+ * @brief Search and mark nodes for deletion until the budget 1 - @p fidelity is
+ * exhausted.
+ * @details Uses a prioritized iterative-deepening search.
+ * Iterating layer by layer ensures that each node is only visited once.
+ */
```
burgholzer: Tiny nitpick, but the docstring here does not describe the function parameters and the return value.

src/dd/Approximation.cpp
```diff
@@ -0,0 +1,235 @@
...
+  Layer curr{};
+  curr.emplace(state.p);
+
+  std::forward_list<Terminal> candidates{};
+
+  ApproximationMetadata meta{fidelity, 0, std::numeric_limits<Qubit>::max()};
+
+  double budget = 1 - fidelity;
+  while (budget > 0) {
+    Contributions c; // Stores contributions of the next layer.
```
burgholzer: Just a thought: might it be more performant to allocate this hash map once and then to simply clear it in each iteration of the whole loop?
That might safe some memory allocations and resizing operations on subsequent iterations.

burgholzer (changes requested):
This is really great!
Some final remarks and slight touch ups, then this is ready to go in 🙌

burgholzer (approved):
Great job on this PR! Let's get this in 🚀

burgholzer:
Ah, of course I forgot one thing: could you please add a Changelog entry for this PR in the unreleased `### Added` section of the top-level CHANGELOG.md file?

burgholzer (approved):
Nice 🚀

PR #915, opened 2025-04-08
🎨 Add Helper Functions For `simulate` And `sample`

MatthiasReumann:
## Description

This pull request adds an handful of helper functions for the `simulate` and `sample` functions in an unnamed namespace. 

## Checklist:

- [X] The pull request only contains commits that are related to it.
- [X] I have added appropriate tests and documentation.
- [X] The pull request introduces no new warnings and follows the project's style guidelines.

### DIFF
--- a/include/mqt-core/dd/Operations.hpp
+++ b/include/mqt-core/dd/Operations.hpp
@@ -205,6 +205,33 @@ applyClassicControlledOperation(const qc::ClassicControlledOperation& op,
                                 const qc::Permutation& permutation = {});
 
 /**
+ * @brief Check whether @p op is virtually executable.
+ *
+ * @param op The operation in question.
+ * @return Boolean that indicates whether @p op is virtually executable.
+ */
+bool isExecutableVirtually(const qc::Operation& op) noexcept;
+
+/**
+ * @brief Apply virtual operation @p op by updating @p permutation.
+ *
+ * @param op The virtual operation to apply.
+ * @param permutation The to be updated permutation.
+ */
+void applyVirtualOperation(const qc::Operation& op,
+                           qc::Permutation& permutation) noexcept;
+
+/**
+ * @brief Apply global phase to a given DD.
+ *
+ * @param in The input DD
+ * @param phase The phase to apply
+ * @param dd The DD package to use
+ * @return The output DD
+ */
+VectorDD applyGlobalPhase(VectorDD& in, const fp& phase, Package& dd);
+
+/**
  * @brief Change the permutation of a given DD.
  *
  * @details This function changes the permutation of the given DD @p on from
--- a/include/mqt-core/dd/Operations.hpp
+++ b/include/mqt-core/dd/Operations.hpp
@@ -208,15 +208,15 @@ applyClassicControlledOperation(const qc::ClassicControlledOperation& op,
  * @brief Check whether @p op is virtually executable.
  *
  * @param op The operation in question.
- * @return Boolean that indicates whether @p op is virtually executable.
+ * @return Whether @p op is virtually executable.
  */
 bool isExecutableVirtually(const qc::Operation& op) noexcept;
 
 /**
- * @brief Apply virtual operation @p op by updating @p permutation.
+ * @brief Apply virtual operation @p op.
  *
  * @param op The virtual operation to apply.
- * @param permutation The to be updated permutation.
+ * @param permutation If suitable, the to be updated permutation.
  */
 void applyVirtualOperation(const qc::Operation& op,
                            qc::Permutation& permutation) noexcept;
--- a/src/dd/Simulation.cpp
+++ b/src/dd/Simulation.cpp
@@ -17,11 +17,10 @@
 #include "ir/operations/ClassicControlledOperation.hpp"
 #include "ir/operations/NonUnitaryOperation.hpp"
 #include "ir/operations/OpType.hpp"
+#include "ir/operations/Operation.hpp"
 
 #include <algorithm>
 #include <array>
-#include <cmath>
-#include <complex>
 #include <cstddef>
 #include <map>
 #include <memory>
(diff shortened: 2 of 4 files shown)


src/dd/Simulation.cpp
```diff
@@ -31,6 +31,42 @@
...
+ */
+void virtualSwap(const qc::Operation& op, qc::Permutation& perm) noexcept {
+  const auto& targets = op.getTargets();
+  std::swap(perm.at(targets[0U]), perm.at(targets[1U]));
+}
+
+/**
+ * @brief Returns `true` if the circuit has a global phase.
+ */
+bool hasGlobalPhase(const fp& phase) noexcept { return std::abs(phase) > 0; }
```
burgholzer: Wouldn't it be more natural for this function to take a circuit as a const reference? Same holds for the function below.

Thinking about it, the function here could actually become a public method of the QuantumComputation class, couldn't it?

src/dd/Simulation.cpp
```diff
@@ -31,6 +31,42 @@
...
+  return op.getType() == qc::SWAP && !op.isControlled();
+}
+
+/**
+ * @brief Virtually SWAP by permuting the layout.
+ */
+void virtualSwap(const qc::Operation& op, qc::Permutation& perm) noexcept {
+  const auto& targets = op.getTargets();
+  std::swap(perm.at(targets[0U]), perm.at(targets[1U]));
+}
```
burgholzer: One could slightly generalize this so it might have some future use.
What about calling the first function something like `isExecutableVirtually(const Operation& op)`
Which could subsume all operations that can be applied virtually. The ones I can think about:
- SWAP operations (as in the current implementation)
- Identity operations (no matter how many controls or targets)
- Barriers (no matter how many targets)

Then the second function would be `applyVirtualOperation` or `executeVirtualOperation` (apply probably fits better with the existing terminology).

Does that make sense?

src/dd/Simulation.cpp
burgholzer, line 1: The docstrings for the methods here are kind of incomplete. Would you mind adding parameter and return value descriptions as well?

src/dd/Simulation.cpp
```diff
@@ -212,34 +246,25 @@ std::map<std::string, std::size_t> sample(const qc::QuantumComputation& qc,
 
 VectorDD simulate(const qc::QuantumComputation& qc, const VectorDD& in,
                   Package& dd) {
-  auto permutation = qc.initialLayout;
-  auto e = in;
+  qc::Permutation permutation = qc.initialLayout;
+  dd::VectorDD out = in;
```
burgholzer: Any particular reason for not keeping these as `auto`? (I like the change to refer to this as `out` instead of the generic `e`)

src/dd/Simulation.cpp
```diff
@@ -212,34 +246,25 @@ std::map<std::string, std::size_t> sample(const qc::QuantumComputation& qc,
...
-  if (std::abs(qc.getGlobalPhase()) > 0) {
-    // create a temporary copy for reference counting
-    auto oldW = e.w;
-    // adjust for global phase
-    const auto globalPhase = ComplexValue{std::polar(1.0, qc.getGlobalPhase())};
-    e.w = dd.cn.lookup(e.w * globalPhase);
-    // adjust reference count
-    dd.cn.incRef(e.w);
-    dd.cn.decRef(oldW);
+  if (fp phase = qc.getGlobalPhase(); hasGlobalPhase(phase)) {
```
burgholzer:
```suggestion
  if (const fp phase = qc.getGlobalPhase(); hasGlobalPhase(phase)) {
```
But this is subject to change anyway based on the comments above 🙂

burgholzer (changes requested):
Many thanks for this nice small PR.
This is perfect for establishing some best practices and playing through the whole PR workflow.

I have a couple of minor comments which you will find inline. I hope they make sense. 🙂 feel free to ping me whenever you are done with implementing the respective changes

burgholzer:
Ah, btw, if you want to see the longer output from the failed check you can click on the workflow run and see the bit message as part of the summary at the very bottom. Specifically, here: https://github.com/munich-quantum-toolkit/core/actions/runs/14326926906/attempts/2#summary-40167853871

src/dd/Simulation.cpp
MatthiasReumann: You are absolutely right. Does also fit well with the `getGlobalPhase` method.

src/dd/Simulation.cpp
MatthiasReumann: Nope. Should have changed that back to `auto`.

This is actually something I wanted to ask you eventually (or your opinion on this topic)!:
The `auto` keyword is used quite extensively in this project. Do you generally prefer `auto` over naming the exact type? In my head the rule of thumb is: use `auto` for lambdas, iterators, and types with long names. However, `qc::Permutation` and `dd::VectorDD` aren't really that many characters. Moreover, when using `auto` any good idea will output the actual type right next to `auto` (so it doesn't really save screen space). Nonetheless, it's very, very, convenient to use `auto`.

src/dd/Simulation.cpp
MatthiasReumann: I've added a method `hasGlobalPhase` to the `QuantumComputation` class and moved the `applyGlobalPhase` function to `Operations.{hpp,cpp}`. However, I kept the `const fp& phase` parameter for the latter. I would argue that one could use this function also without having a `qc` object at hand.

src/dd/Simulation.cpp
MatthiasReumann: Are the Identities and Barriers simply "no-ops", or do I make a grave mistake here?

src/dd/Simulation.cpp
burgholzer: Very good point. I like that 👍🏻

src/dd/Simulation.cpp
burgholzer: Yeah, both of these can be considered no-ops for simulation purposes.

src/dd/Simulation.cpp
burgholzer: Just my personal opinion:
The biggest argument is convenience. As you say, IDEs are clever enough by now to figure out the right type and display it appropriately. It saves you from having to think what the actual type should be and what you need to type. That creates a much smoother development experience in my opinion.
I'd generally just avoid `auto` wherever it might add to the clarity of the code to spell out the type.
Or in cases where you actually want a base class pointer but the function returns a pointer from a derived class.
Otherwise I'd almost always go for the auto variant.
The only real downside I can see is when you are not in an IDE and don't have access to all the automatic type hints. Such as when reviewing code on GitHub.

include/mqt-core/dd/Operations.hpp
```diff
@@ -204,6 +204,33 @@ applyClassicControlledOperation(const qc::ClassicControlledOperation& op,
                                 const std::vector<bool>& measurements,
                                 const qc::Permutation& permutation = {});
 
+/**
+ * @brief Check whether @p op is virtually executable.
+ *
+ * @param op The operation in question.
+ * @return Boolean that indicates whether @p op is virtually executable.
```
burgholzer:
```suggestion
 * @return Whether @p op is virtually executable.
```
Just keeping it brief

include/mqt-core/dd/Operations.hpp
```diff
@@ -204,6 +204,33 @@ applyClassicControlledOperation(const qc::ClassicControlledOperation& op,
...
+ */
+bool isExecutableVirtually(const qc::Operation& op) noexcept;
+
+/**
+ * @brief Apply virtual operation @p op by updating @p permutation.
```
burgholzer: Nitpick: Not every virtual operation may update the permutation. Could be worth rewording slightly.

src/dd/Simulation.cpp
MatthiasReumann: okay, thanks!

burgholzer (approved):
This is looking great. I just have two really minor last comments. Then this can go in 🙂

burgholzer (approved):
Nice 👍🏻 
From now on out, you also shouldn't need approval from a maintainer anymore for CI to start.

MatthiasReumann:
> Nice 👍🏻 From now on out, you also shouldn't need approval from a maintainer anymore for CI to start.

Merci!

PR #932, opened 2025-04-22
🐛 [MLIR] Fix bug in MLIR round-trip

flowerthrower:
## Description

Fix bugs that only occur on macOS as a result of undefined behavior.

## Checklist:


burgholzer:
@flowerthrower What's the status here?

mlir/lib/Dialect/MQTOpt/Transforms/FromQuantumComputationPattern.cpp
```diff
@@ -62,6 +62,34 @@ struct FromQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+        reg.getResult(), // The register to insert into
+        qubit, nullptr,  // The qubit to insert
+        rewriter.getI64IntegerAttr(static_cast<std::int64_t>(index)) // Index
+    );
+    return insert;
+  }
+
+  // rewriter.create<::mqt::ir::opt::InsertOp>(
+  //         op.getLoc(), resultType, inQregValue, qubitValue, idxValue,
+  //         idxIntegerAttr);
```
burgholzer: The commented-out code here can be removed
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -125,34 +134,60 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
    * @param op The operation to convert.
    * @param currentQubitVariables The list of previously defined qubit
    * variables.
+   *
+   * return if the operation was successfully handled
```
burgholzer: This is not a proper doxygen comment.
```suggestion
   * @returns if the operation was successfully handled
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -125,34 +134,60 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+      for (size_t i = 0; i < ctrlInsIndices.size(); i++) {
+        currentQubitVariables[ctrlInsIndices[i]] = outs[i + 1];
+      }
+      currentQubitVariables[targetIndex] = outs[0];
+
+      // Add the operation to the QuantumComputation.
+      auto operation = qc::StandardOperation(
+          qc::Controls{ctrlInsIndices.cbegin(), ctrlInsIndices.cend()},
+          targetIndex, opType);
+      circuit.push_back(operation);
```
burgholzer:
```suggestion
      circuit.emplace_back<StandardOperation(qc::Controls{ctrlInsIndices.cbegin(), ctrlInsIndices.cend()},
          targetIndex, opType);
```
This should work and is more efficient as it constructs the operation inplace.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -125,34 +134,60 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
     const auto ctrlIns = op.getCtrlQubits();
     const auto outs = op.getOutQubits();
 
-    // Get the qubit index of every control qubit.
-    std::vector<size_t> ctrlInsIndices(ctrlIns.size());
-    std::transform(ctrlIns.begin(), ctrlIns.end(), ctrlInsIndices.begin(),
-                   [&currentQubitVariables](const mlir::Value val) {
-                     return findQubitIndex(val, currentQubitVariables);
-                   });
+    try {
```
burgholzer, line 186: This `try` block is scoped very broadly. Could you please narrow it down to the actual call(s) that might fail?

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -125,34 +134,60 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+    } catch (const std::runtime_error& e) {
+      if (e.what() ==
+          std::string(
+              "Qubit was not found in list of previously defined qubits")) {
+        // Try again later when all qubits are available
+        return false;
+      }
+      {
+        throw; // Rethrow the exception if it's not the expected one.
+      }
```
burgholzer: Is there any particular reason for this being in its own block?
```suggestion
      throw; // Rethrow the exception if it's not the expected one.
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -125,34 +134,60 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+
+      // Add the operation to the QuantumComputation.
+      auto operation = qc::StandardOperation(
+          qc::Controls{ctrlInsIndices.cbegin(), ctrlInsIndices.cend()},
+          targetIndex, opType);
+      circuit.push_back(operation);
+    } catch (const std::runtime_error& e) {
+      if (e.what() ==
+          std::string(
+              "Qubit was not found in list of previously defined qubits")) {
```
burgholzer: I am pretty sure you do not need the explicit string conversion here and might as well do something like
```suggestion
          "Qubit was not found in list of previously defined qubits") {
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -266,52 +316,55 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
     circuit.addQubitRegister(numQubits, regName);
     circuit.addClassicalRegister(numQubits);
 
-    std::unordered_set<mlir::Operation*> visited{};
+    std::set<mlir::Operation*> visited{};
+    std::set<mlir::Operation*> mqtUsers{};
```
burgholzer: I am fairly certain that these should both be `unordered_set` as you want to use them as hash maps with quick membership checks and you do not need these to be sorted.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -266,52 +316,55 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
 
-    // Visit all operations in the AST using Breadth-First Search.
-    while (!toVisit.empty()) {
-      mlir::Operation* current = *toVisit.begin();
-      toVisit.erase(toVisit.begin());
-      if (visited.find(current) != visited.end()) {
+    mlir::Operation* current = op;
+    while (current != nullptr) {
+      llvm::outs() << current->getName().getStringRef() << "\n";
+
```
burgholzer: This seems like a print out from debugging?
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -321,26 +374,28 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
-    // NOLINTNEXTLINE(bugprone-nondeterministic-pointer-iteration-order)
-    for (auto* operation : visited) {
-      if (operation->getDialect()->getNamespace() != DIALECT_NAME_MQTOPT) {
-        updateMQTOptInputs(*operation, rewriter, newAlloc.getQureg(),
-                           measureCount);
-      }
+    for (auto* operation : llvm::make_early_inc_range(mqtUsers)) {
+      mqtUsers.erase(operation); // safe deletion as op will be erased next
+      updateMQTOptInputs(*operation, rewriter, newAlloc.getQureg());
     }
```
burgholzer: This seems like a fairly cryptic and LLVM-specific way of doing this (also holds below).
There is a more idiomatic way to do this base don using iterators.
Any `erase` operation from an STL container in C++ will return an updated iterator that can be used for iterating further.
See https://en.cppreference.com/w/cpp/container/vector/erase for an example of how this is done.
Let's rather use that.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -321,26 +374,28 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+    while (!visited.empty()) {
+      // Enable updates of `visited` set in the loop.
+      for (auto* operation : llvm::make_early_inc_range(visited)) {
+        if (operation->getDialect()->getNamespace() == DIALECT_NAME_MQTOPT) {
+          if (operation->getUsers().empty()) {
+            visited.erase(operation);
+            rewriter.eraseOp(
+                operation); // deletes all definitions of defining OPs
+          }
+        }
```
burgholzer: This seems like a confusing way to write this removal here.
The while loop has no break conditions, so the `visited` container` must eventually be empty for this function to return.
Yet, the inner loop has two special if conditions that check for some kind of property before deciding to remove an operation.
Maybe I am overlooking things, but isn't there a clearer way to write this?

burgholzer (changes requested):
Hey @flowerthrower 

I took the liberty to review this real quick in order to move this along a little quicker.
You can find detailed comments inline.

One note for the future: it would be great if PRs like these would really only contain the fixes for the bug that has been observed. This PR as it stands clearly also adds in other changes that make it harder to judge what has actually been fixed.

burgholzer:
Also: could you please add a respective changelog entry for this PR?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- 🐛 Fixed a bug on macOS, caused by accessing an invalidated iterator after erasure in a loop ([#932](https://github.com/munich-quantum-toolkit/core/pull/932)) ([**@flowerthrower**](https://github.com/flowerthrower))
```
```suggestion
- 🐛 Fix bug in MLIR roundtrip passes caused by accessing an invalidated iterator after erasure in a loop ([#932](https://github.com/munich-quantum-toolkit/core/pull/932)) ([**@flowerthrower**](https://github.com/flowerthrower))
```

See https://common-changelog.org/#241-change for some guidelines on how to write changelog entries.
The previous entry provided no context for people reading the changelog on what kind of bug was fixed.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -16,15 +16,18 @@
 #include "mlir/Dialect/MQTOpt/Transforms/Passes.h"
 
 #include <algorithm>
-#include <cstddef>
```
burgholzer, line 19: This include is actually needed.
Under macOS it is sometimes greyed out because the standard library linked by default includes the respective types through a different path.
It is still needed.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -76,12 +79,17 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
       size_t qubitArrayIndex = 0;
-      if (auto opResult =
-              llvm::dyn_cast<mlir::OpResult>(currentQubitVariables[i])) {
-        qubitArrayIndex = opResult.getResultNumber();
+      if (currentQubitVariables[i] != nullptr) {
+        auto opResult =
+            llvm::dyn_cast<mlir::OpResult>(currentQubitVariables[i]);
+        qubitArrayIndex =
+            opResult
+                .getResultNumber(); // e.g. is 1 if it comes from an extract op
```
burgholzer: The formatting here irritates me
```suggestion
        // e.g. is 1 if it comes from an extract op
        qubitArrayIndex = opResult.getResultNumber(); 
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -76,12 +79,17 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
       } else {
+        // Will be caught, so further ops can be collected until the qubit is
```
burgholzer: > Will be caught ...

By whom and where? 
Hard to follow the logic here without that information.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -125,34 +133,56 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+    try {
+      std::transform(ctrlIns.begin(), ctrlIns.end(), ctrlInsIndices.begin(),
+                     [&currentQubitVariables](const mlir::Value val) {
+                       return findQubitIndex(val, currentQubitVariables);
+                     });
+      // Get the qubit index of the target qubit.
+      targetIndex = findQubitIndex(in, currentQubitVariables);
+    } catch (const std::runtime_error& e) {
+      if (std::string_view(e.what()) ==
+          "Qubit was not found in list of previously defined qubits") {
```
burgholzer: Absolutely no need to use a string_view here. Both `e.what()` and the string constant are plain C strings.
You can simply use `strcmp` from `<cstring>` here as
```
if (strcmp(e.what(), "...") == 0)
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -321,26 +367,39 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
-    for (auto* operation : visited) {
-      if (operation->getDialect()->getNamespace() != DIALECT_NAME_MQTOPT) {
-        updateMQTOptInputs(*operation, rewriter, newAlloc.getQureg(),
-                           measureCount);
-      }
+    for (auto it = mqtUsers.begin(); it != mqtUsers.end(); /* no increment */) {
+      auto* operation = *it;
+      it = mqtUsers.erase(it); // erase returns the next valid iterator
+      updateMQTOptInputs(*operation, rewriter, newAlloc.getQureg());
     }
```
burgholzer: Maybe I am overlooking something, but why is this actually deleting anything from the `mqtUsers` vector? The vector is not used again afterwards and, currently, all its elements are deleted eventually.
You could simply replace this with 
```suggestion
    for (auto& operation: mqtUsers) {
      updateMQTOptInputs(*operation, rewriter, newAlloc.getQureg());
    }
```
and it would accomplish the same thing while being more efficient.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -321,26 +367,39 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+      progress = false;
+
+      for (auto it = visited.begin(); it != visited.end(); /* no increment */) {
+        mlir::Operation* operation = *it;
+
+        // Skip if not part of the mqtopt dialect
+        if (operation->getDialect()->getNamespace() != DIALECT_NAME_MQTOPT) {
+          ++it;
+          continue;
+        }
```
burgholzer: Given how you repeatedly process the visited set until convergence and non-mqt operations will always hit this condition in every single loop, wouldn't it be better to either:
- remove them from the visited set upon first encounter in this loop (certainly possible)
- not add them to the visited set in the first place (unsure whether that is possible)

burgholzer (changes requested):
I think this needs one more iteration, then it should be ready.
I hope you are not annoyed by the level of detail, but I think it is important to learn these subtleties and to pay attention to them.

mlir/lib/Dialect/MQTOpt/Transforms/FromQuantumComputationPattern.cpp
```diff
@@ -214,6 +247,15 @@ struct FromQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
       }
     }
 
+    // Now insert all the qubits back into the registers they were extracted
+    // from.
+    for (size_t i = 0; i < numQubits; i++) {
+      auto insertOp = createRegisterInsert(
+          newAlloc, currentRegister, currentQubitVariables[i], i, rewriter);
+      // Keep track of qubit register access
+      currentRegister = insertOp.getOutQureg();
```
burgholzer:
```suggestion
      currentRegister = insertOp.getOutQreg();
```
Since #933 will now merge before this PR, I think this line has to change after the merge.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -62,33 +68,41 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
    * variables.
    *
    * @return The index of the qubit in the list of previously defined qubit
-   * variables.
+   * variables. Throws an exception if the qubit is not found.
```
burgholzer: Doxygen has a dedicated syntax for documenting the exceptions a function may throw.
It's also always a good idea to denote the particular type of exception so that people know what they should be catching.

```suggestion
   * variables. 
   *
   * @throws `std::runtime_error` if the qubit is not found.
```

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -321,26 +394,32 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
 
     // Update the inputs of all non-mqtopt operations that use mqtopt operations
     // as inputs, as these will be deleted later.
-    // NOLINTNEXTLINE(bugprone-nondeterministic-pointer-iteration-order)
-    for (auto* operation : visited) {
-      if (operation->getDialect()->getNamespace() != DIALECT_NAME_MQTOPT) {
-        updateMQTOptInputs(*operation, rewriter, newAlloc.getQureg(),
-                           measureCount);
-      }
+    // NOLINTNEXTLINE(readability-qualified-auto)
```
burgholzer:
```suggestion
```
Is that ignore actually ignoring something? does not seem necessary to me

burgholzer (changes requested):
Thanks for all your work on this! This is 99% ready to merge. It will just need some fixing up once #933 merges and two small little details, then we are good to go 🚀

burgholzer (approved):
Alright. This should be ready to go in now 🎉 many thanks!

Issue #899, opened 2025-04-02
🧪 MLIR - Advanced Testing for Transformation passes


DRovara:
Currently, we use LLVM's LIT framework to test transformation passes.
For each pass, we have a single test file that is split into multiple slices for individual test cases.
These test cases use `FileCheck` and require assertions to be written manually which is quite painful.
Therefore, it might be a good idea to have a more sophisticated testing setup.

## LLVM Testing Infrastructure Guidelines

A good standard for how to set up testing might be the LLVM Testing Infrastructure Guide (https://llvm.org/docs/TestingGuide.html). They make the following suggestions:
- Put related tests into a single file (as we already do)
- Use check-prefixes to test different settings, e.g. different architectures  that might require different behaviour (we currently do not really have a use case for this)

The LLVM framework also distinguishes between three classes of tests:
- Regression Tests (the "main" tests in the form of small LIT tests)
- "Whole Program Tests" (LIT tests with bigger programs, in a different directory)
- Unit Tests (tests that use gtest)
According to their guidelines, Unit Tests should only be used to test internal data structures etc., not the functionality of the passes themselves.

The LLVM framework also has additional tools that help them with LIT tests:

1) They have their own scripts that automatically generate assertions (it's considered best practice to almost exclusively use auto-generated checks)
2) They support custom operations that can be used inside the tests (not sure if we have a use case for this currently, but it's best to keep it in mind)
3) They suggest expanding LIT configurations (https://github.com/munich-quantum-toolkit/core/blob/4e08f11fb89491ff61e352a20ce1172aec4c6d9e/mlir/test/lit.cfg.py) to best help for the specific tests

While it is not necessary to follow all of these suggestions, it might be helpful to implement at least one of them, as writing new tests currently is quite a lot of work.

## Deviating from the LLVM Guidelines

It is not strictly necessary to follow these guidelines. We might also implement more gtests instead of having the large number of LIT tests (see https://github.com/PennyLaneAI/catalyst/blob/9f1c8f0b84906c2f8e65c899f5f305f449064585/mlir/CMakeLists.txt#L100-L145 for example)

However, any types of improvements would be quite welcome for the MLIR testing infrastructure.

burgholzer:
Since the dialect redesign has been merged, lit tests have been entirely replaced by googletest-based test suites.
Hence, I would consider this resolved.

PR #948, opened 2025-05-05
⚗️ Testing functionality of MQTOpt value dialect

BertiFlorea:
## Description

As described in #899, we want to improve the testing setup and coverage across the MLIR part of the Core library. The purpose of this PR is to have a structured approach to testing all functionalities of the MQTOpt dialect. In other words, we want to make sure that each operation behaves as expected.

This PR is only a contribution to #899, not a fix.

## Checklist:


mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
 
-// RUN: quantum-opt %s | FileCheck %s
+// RUN: quantum-opt %s -split-input-file | FileCheck %s
 
+// -----
+// This test checks if a qubit register is allocated correctly
+module {
+    // CHECK-LABEL: func.func @testAllocOp
+    func.func @testAllocOp() -> (!mqtopt.QubitRegister) {
+        // CHECK: %[[Reg_0:.*]] = "mqtopt.allocQubitRegister"
```
DRovara: I think in this case, we should also include the attributes to the check string, just because we are testing the AllocOp for the first time here and the attribute is important for it to work

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
 // Licensed under the MIT License
 
-// RUN: quantum-opt %s | FileCheck %s
+// RUN: quantum-opt %s -split-input-file | FileCheck %s
 
+// -----
+// This test checks if a qubit register is allocated correctly
+module {
+    // CHECK-LABEL: func.func @testAllocOp
+    func.func @testAllocOp() -> (!mqtopt.QubitRegister) {
```
DRovara: Since the `AllocOp` has two variants (one with an operand, one with static attributes), we should test both of those (probably the best solution is to have two different test cases for that.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
 //
 // Licensed under the MIT License
 
-// RUN: quantum-opt %s | FileCheck %s
+// RUN: quantum-opt %s -split-input-file | FileCheck %s
 
+// -----
+// This test checks if a qubit register is allocated correctly
```
DRovara: Just a minor thing, but this test does not "check if a qubit register is allocated correctly", it checks if the `AllocOp` is parsed and handled correctly. Same goes with other operations (e.g. the `DeallocOp` test checks if the `DeallocOp` is read correctly, it does not actually check if the register is de-alloced.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        return
+    }
+}
+
+// -----
+// This test checks if a qubit is extracted correctly from a register
+module {
+    // CHECK-LABEL: func.func @testExtractOp
+    func.func @testExtractOp() -> (!mqtopt.QubitRegister, !mqtopt.Qubit) {
```
DRovara: Same thing here, we should also add a second `ExtractOp` test that uses an `index` operand instead of the attribute.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        return %reg_1, %q_0 : !mqtopt.QubitRegister, !mqtopt.Qubit
+    }
+}
+
+// -----
+// This test checks if a qubit is inserted back into its register correctly
+module {
+    // CHECK-LABEL: func.func @testInsertOp
+    func.func @testInsertOp() -> (!mqtopt.QubitRegister) {
```
DRovara: Same thing with the attribute/operand tests

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        return %q1_1, %q0_1 : !mqtopt.Qubit, !mqtopt.Qubit
+    }
+}
+
+// -----
+// This test checks if a negative CX gate is applied correctly
+module {
+    // CHECK-LABEL: func.func @testCXOp
+    func.func @testCXOp() -> (!mqtopt.Qubit, !mqtopt.Qubit) {
```
DRovara: This function should probably have a different name than the one above. Don't forget to also update the `CHECK-LABEL` if you rename the function.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+        // CHECK: %[[Reg_3:.*]], %[[Q2_0:.*]] = "mqtopt.extractQubit"(%[[Reg_2]]) <{index_attr = 2 : i64}>
+        %reg_3, %q2_0 = "mqtopt.extractQubit"(%reg_2) <{index_attr = 2 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+
+        //===------------------------------------------------------------------===//
+        // q_0: ──■──
+        //      ┌─┴─┐
+        // q_1: ┤ X ├
+        //      └─┬─┘
+        // q_2: ──■──
+        //===----------------------------------------------------------------===//
```
DRovara: I like these circuit previews, they help visualize the circuit.
Minor nitpick, feel free to ignore it: the variables of these qubits are not called `q_0`, `q_1`, and `q_2`, they are called either `q0, q1, q2` or, with the index suffix, `q0_0, q1_0, q2_0`.
Maybe it becomes even more helpful if we use the actual variable names in the figure? No high priority though

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        // CHECK: %[[Q1_1:.*]], %[[Q02_2:.*]]:2 = mqtopt.x() %[[Q1_0]] ctrl %[[Q0_0]], %[[Q2_0]] : !mqtopt.Qubit ctrl !mqtopt.Qubit, !mqtopt.Qubit
+        %q1_1, %q02_1:2 = mqtopt.x() %q1_0 ctrl %q0_0, %q2_0 : !mqtopt.Qubit ctrl !mqtopt.Qubit, !mqtopt.Qubit
+
+        // ==========================  Check that there are no further MCX operations ==============================
+        // CHECK-NOT: mqtopt.x() [[ANY:.*]] ctrl %[[ANY:.*]]:2
```
DRovara: If possible, I would love to have multiple test cases that test the two formats
`%q1, q%2, %q3 = mqtopt.x() ...`
AND
`%q:3 = mqtopt.x() ...`
That helps as an example for readers and makes sure that both formats are always working correctly.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+        //===------------------------------------------------------------------===//
+        // q_0: ──■──
+        //      ┌─┴─┐
+        // q_1: ┤ X ├
+        //      └─┬─┘
+        // q_2: ──■──
+        //===----------------------------------------------------------------===//
+
+        // CHECK: %[[Q1_1:.*]], %[[Q02_2:.*]]:2 = mqtopt.x() %[[Q1_0]] nctrl %[[Q0_0]], %[[Q2_0]] : !mqtopt.Qubit nctrl !mqtopt.Qubit, !mqtopt.Qubit
+        %q1_1, %q02_1:2 = mqtopt.x() %q1_0 nctrl %q0_0, %q2_0 : !mqtopt.Qubit nctrl !mqtopt.Qubit, !mqtopt.Qubit
```
DRovara: Maybe I'm missing something, but does this format actually work?
Specifically the
`: !mqtopt.Qubit nctrl !mqtopt.Qubit, !mqtopt.Qubit`
part - is this allowed? Same with the test above.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        return %q02_1#0, %q1_1, %q02_1#1 : !mqtopt.Qubit, !mqtopt.Qubit, !mqtopt.Qubit
+    }
+}
+
+// -----
+// This test checks if a negative MCX gate is applied correctly
+module {
+    // CHECK-LABEL: func.func @testMCXOp
+    func.func @testMCXOp() -> (!mqtopt.Qubit, !mqtopt.Qubit, !mqtopt.Qubit) {
```
DRovara: This test also should have a different name than the one above

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+    }
+}
+
+// -----
+// This test checks if an identity gate is applied to a single qubit correctly
+module {
+    // CHECK-LABEL: func.func @testIOp
+    func.func @testIOp() -> (!mqtopt.Qubit) {
+        // CHECK: %[[Reg_0:.*]] = "mqtopt.allocQubitRegister"
+        %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 2 : i64}> : () -> !mqtopt.QubitRegister
```
DRovara: For the sake of simplicity: As this test case only uses one qubit, should the size just be 1 here?

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        // CHECK: %[[Reg_1:.*]], %[[Q_0:.*]] = "mqtopt.extractQubit"(%[[Reg_0]]) <{index_attr = 0 : i64}>
+        %reg_1, %q_0 = "mqtopt.extractQubit"(%reg_0) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+
+        // ==========================  Check that there are no further extractions ==============================
+        // CHECK-NOT: "mqtopt.extractQubit"([[ANY:.*]])
+
+        return %reg_1, %q_0 : !mqtopt.QubitRegister, !mqtopt.Qubit
```
DRovara: I'm not quite sure about the return values of these test cases:
Is there a reason why the different tests all have different return values? Intuitively I would think that having the same (i.e. none) return values for all test cases would be the simplest. Am I missing something here?

DRovara:
Thanks a lot @BertiFlorea for the effort!

Other than my comments above, I still have a couple of thoughts that I'll add below. In any case, this is already a good start and I'm looking forward to seeing tests for the remaining parametrized gates as well!

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        // CHECK: %[[Q1_1:.*]], %[[Q0_1:.*]] = mqtopt.x() %[[Q1_0]] ctrl %[[Q0_0]] : !mqtopt.Qubit ctrl !mqtopt.Qubit
+        %q1_1, %q0_1 = mqtopt.x() %q1_0 ctrl %q0_0 : !mqtopt.Qubit ctrl !mqtopt.Qubit
+
+        // ==========================  Check that there are no further CX operations ==============================
+        // CHECK-NOT: %[[ANY:.*]], %[[ANY:.*]] = mqtopt.x() %[[ANY:.*]] ctrl %[[ANY:.*]]
+
+        return %q1_1, %q0_1 : !mqtopt.Qubit, !mqtopt.Qubit
+    }
+}
```
DRovara: Thinking about it, I feel like we don't actually need individual test cases for each single-qubit unitary gate. If anything, all the boilerplate required for them (alloc, extract, etc.) just adds unnecessary complexity.
If there was some way to have something like parametrized test cases, and the parameters are the individual 1q-unitary gates, that would be awesome. But most likely, that's not possible. Alternatively, it might be best to just collect them all in one single test function.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+
+        // CHECK: %[[Q1_1:.*]], %[[Q02_2:.*]]:2 = mqtopt.x() %[[Q1_0]] nctrl %[[Q0_0]], %[[Q2_0]] : !mqtopt.Qubit nctrl !mqtopt.Qubit, !mqtopt.Qubit
+        %q1_1, %q02_1:2 = mqtopt.x() %q1_0 nctrl %q0_0, %q2_0 : !mqtopt.Qubit nctrl !mqtopt.Qubit, !mqtopt.Qubit
+
+        // ==========================  Check that there are no further MCX operations ==============================
+        // CHECK-NOT: mqtopt.x() [[ANY:.*]] nctrl %[[ANY:.*]]:2
+
+        return %q02_1#0, %q1_1, %q02_1#1 : !mqtopt.Qubit, !mqtopt.Qubit, !mqtopt.Qubit
+    }
+}
```
DRovara: I'd actually like some more test cases for controlled gates. First of all, we should definitely exhaust all the possible combinations of `ctrls`/`nctrls`:
- [x] 1 target, 1 ctrl
- [x] 1 target, 1 nctrls
- [x] 1 target, 2 ctrls
- [x] 1 target, 2 nctrls
- [x] 1 target, 1 ctrl, 1 nctrl
- [ ] 1 target, 1 nctrl, 1 ctrl (is this order allowed?)  
- [ ] are multiple targets allowed? If yes, that would be nice, too.

Additionally, currently the test all use CX gates. It's fine to generally use just CX gates, but what would be interesting (again, not sure if supported):
- [x] controlled swap (positive controls, negative controls, mixed...)
- [x] controlled parametrized gates
- [ ] controlled gphase gate (if this is supported)

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,8 +6,637 @@
...
+        // ==========================  Check that there are no further ECR operations ==============================
+        // CHECK-NOT: mqtopt.ecr() [[ANY:.*]], [[ANY:.*]]
+
+        return %q1_1, %q0_1 : !mqtopt.Qubit, !mqtopt.Qubit
+    }
+}
+
+// -----
 module {
     // CHECK-LABEL: func @bell_state()
```
DRovara: Of course, this huge test case should no longer exist later.
However, it could be interesting to still have one or two "larger" test cases, e.g. a bell state circuit.

DRovara:
I know this is easier said than done, but I would also like to add negative test cases, i.e. test cases where we expect an error. In particular, this should include all situations that are currently enforced by interface/trait verifiers, like:
- mismatch between input qubits and output qubits of a controlled operation
- invalid formats of controlled operations
- invalid formats of parametrized operations
- measurements with mismatch of in-qubits vs out-qubits
- measurements with mismatch of out-qubits vs out-bits
- alloc/insert/extract operations that define input/size operands AND attributes
- alloc/insert/extract operations that do not define input/size operands or attributes

DRovara (changes requested):
Thanks a lot @BertiFlorea for the effort!

I have added some comments on things I would change, but feel free to further discuss the topics if you don't agree with me. In any case, this is already a good start and I'm looking forward to seeing tests for the remaining parametrized gates as well!

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
DRovara: If this is allowed,  then we definitely should have one test that uses this format, and one that simply uses the format `: !mqtopt.Qubit, !mqtopt.Qubit`

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: I assumed that if I did not insert a return statement and use the values defined inside the function, some instructions would be removed due to dead code elimination (DCE). I assume I was wrong, but to understand it better, does DCE happen only when applying passes?

I will update the tests accordingly and remove the return statements.

DRovara: Good point. I think DCE will be performed if we don't do anything with the values.

However, somehow I'm still not sure what I think about the returns. The alternative approach would be to insert the qubits into the register and then deallocate the register. That would also prevent DCE. It would lead to more boilerplate code but the tests would become more readable.

Also, including the whole qubit allocation and deallocation process in test cases helps with future-proofing: Maybe sooner or later we *require* dealloc to be called, and then all these tests would suddenly fail

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: Sadly, the last two cases from the first list are not possible. The first would require the assembly format of the `UnitaryOp` to be modified, and the second is forbidden by the trait imposed for each unitary gate.

I am still failing in testing the `gphase` gate correctly (still trying to figure out a way to do it correctly).

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: I have compressed all the single-qubit tests into one. In my opinion, this is more elegant and easier to read. However, it is possible to do a unique test that can be parametrised, but I still need some time to figure out how to do it correctly.

DRovara: I'd say that the parametrisation has the lowest priority at the moment. In fact, that might even be fitting for a second PR later on. So you don't necessarily have to focus on that for now.

BertiFlorea:
Once again, @DRovara, many thanks for the elaborate feedback!
I managed to address almost all the parts in the review. There are still a couple of TODOs, and they are the following:
- Add negative tests (I've researched a little bit how to do this best, I will try to implement this in the next couple of days)
- Add insertion of qubits into registers and deallocate them to prevent DCE (dead code elimination)
- Test the `gphase` gate

To prevent a lot of repetition in our code, I would propose a single test for two target qubit gates (swap, iswap, ECR, etc.). What do you think?

DRovara:
> To prevent a lot of repetition in our code, I would propose a single test for two target qubit gates (swap, iswap, ECR, etc.). What do you think?

That sounds reasonable to me. I believe they are all defined through the same trait, right? In that case, it makes sense to join them into one test.

That does remind me though: Do we have tests for e.g. controlled swap gates? Maybe that would be nice to have too, just to be on the safer side.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+
+        return
+    }
+}
+
+// -----
+// This test checks if the AllocOp is parsed and handled correctly using a dynamic operand
+module {
+    // CHECK-LABEL: func.func @testAllocOp
+    func.func @testAllocOp() {
```
DRovara: The test cases should have different names, e.g. `testAllocOpOperand` and `testAllocOpAttribute`.

While it still works the way it is right now due to file-split, having different names for test cases helps find error locations when a specific test fails.

(Also, just as a reminder, don't forget to update the `CHECK-LABEL` if you update the function name)

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+
+        return
+    }
+}
+
+// -----
+// This test checks if the ExtractOp is parsed and handled correctly using a dynamic operand
+module {
+    // CHECK-LABEL: func.func @testExtractOp
+    func.func @testExtractOp() {
```
DRovara: Once again here, the tests should have different names.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+        // CHECK: %[[Reg_2:.*]] = "mqtopt.insertQubit"(%[[Reg_1]], %[[Q_0]], %[[Index]])  : (!mqtopt.QubitRegister, !mqtopt.Qubit, i64) -> !mqtopt.QubitRegister
+        %reg_2 = "mqtopt.insertQubit"(%reg_1, %q_0, %index) : (!mqtopt.QubitRegister, !mqtopt.Qubit, i64) -> !mqtopt.QubitRegister
+
+        // ==========================  Check that there are no further insertions ==============================
+        // CHECK-NOT: "mqtopt.insertQubit"([[ANY:.*]], [[ANY:.*]])
+
+        return
+    }
+}
+
```
DRovara: Here we test a sort of mix-and-match where we alloc a register using a static attribute and extract a qubit using an operand.
That's a good thing to test, but since it doesn't cost us anything, I would add other combinations, too. Especially `alloc` with operand and `extract` with attribute could be interesting for the future.

That's just my intuition, though. Maybe we also decide that we don't need to test all these combinations.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+    // CHECK-LABEL: func.func @testMeasureOp
+    func.func @testMeasureOp() {
+        // CHECK: %[[Reg_0:.*]] = "mqtopt.allocQubitRegister"
+        %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 2 : i64}> : () -> !mqtopt.QubitRegister
+
+        // CHECK: %[[Reg_1:.*]], %[[Q_0:.*]] = "mqtopt.extractQubit"(%[[Reg_0]]) <{index_attr = 0 : i64}>
+        %reg_1, %q_0 = "mqtopt.extractQubit"(%reg_0) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+
+        // CHECK: %[[Q_1:.*]], [[M0_0:.*]] = "mqtopt.measure"(%[[Q_0]])
+        %q_1, %m0_0 = "mqtopt.measure"(%q_0) : (!mqtopt.Qubit) -> (!mqtopt.Qubit, i1)
```
DRovara: Do we also support measuring multiple qubits at once?
If so, we should test that. If not, just ignore this comment.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+
+        // ==========================  Check that there are no further measurements ==============================
+        // CHECK-NOT: "mqtopt.measure"([[ANY:.*]])
+
+        return
+    }
+}
+
+// -----
+// This test checks if single qubit  gate is parsed and handled correctly
```
DRovara: There's a typo in this comment.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
-        %out_qubits_2, %out_bits_3 = "mqtopt.measure"(%6) : (!mqtopt.Qubit) -> (!mqtopt.Qubit, i1)
-        %8 = "mqtopt.insertQubit"(%out_qreg, %out_qubits) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
-        %9 = "mqtopt.insertQubit"(%out_qreg_0, %out_qubits_2) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
-        "mqtopt.deallocQubitRegister"(%8) : (!mqtopt.QubitRegister) -> ()
-        "mqtopt.deallocQubitRegister"(%9) : (!mqtopt.QubitRegister) -> ()
-        return %out_bits, %out_bits_3 : i1, i1
+// RUN: quantum-opt %s -split-input-file | FileCheck %s
+
+// -----
+// This test checks if the AllocOp is parsed and handled correctly using a static attribute
```
DRovara: Just a minor thing, but doc-type comments such as the description of test cases should always end with periods (".") That makes them cleaner for documentation purposes.
This relates to all the test case descriptions in this document.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+        // CHECK-NOT: mqtopt.y() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.z() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.s() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.sdg() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.t() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.tdg() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.v() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.vdg() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.sx() [[ANY:.*]]
+        // CHECK-NOT: mqtopt.sxdg() [[ANY:.*]]
```
DRovara: Optional, but all of these lines could be replaced by a single regex, like

```
// CHECK-NOT: mqtopt.[[ANY:.*]]() [[ANY:.*]] : !mqtopt.Qubit
```

That would make extending the test case in the future easier.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+        %q_5 = mqtopt.ry(%c0_f64) %q_4 : !mqtopt.Qubit
+        %q_6 = mqtopt.rz(%c0_f64) %q_5 : !mqtopt.Qubit
+
+        // ==========================  Check that there are no further single qubit rotation operations ==============================
+        // CHECK-NOT: mqtopt.u(%[[ANY:.*]], %[[ANY:.*]], %[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.u2(%[[ANY:.*]], %[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.p(%[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.rx(%[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.ry(%[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.rz(%[[ANY:.*]]) %[[ANY:.*]]
```
DRovara: Same thing here, we could have a single test string like

```
// CHECK-NOT: mqtopt.[[ANY:.*]]([[ANY:.*]]) %[[ANY:.*]] : !mqtopt.Qubit
```

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+    }
+}
+
+// -----
+// This test checks if controlled parameterized single qubit gates are parsed and handled correctly
+module {
+    // CHECK-LABEL: func.func @testControlledSingleQubitRotationOp
+    func.func @testControlledSingleQubitRotationOp() {
+        // CHECK: %[[Reg_0:.*]] = "mqtopt.allocQubitRegister"
+        %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 1 : i64}> : () -> !mqtopt.QubitRegister
```
DRovara: This needs to be `size_attr = 2` instead.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+        %q0_5, %q1_5 = mqtopt.ry(%c0_f64) %q0_4 ctrl %q1_4 : !mqtopt.Qubit ctrl !mqtopt.Qubit
+        %q0_6, %q1_6 = mqtopt.rz(%c0_f64) %q0_5 ctrl %q1_5 : !mqtopt.Qubit ctrl !mqtopt.Qubit
+
+        // ==========================  Check that there are no further single qubit rotation operations ==============================
+        // CHECK-NOT: mqtopt.u(%[[ANY:.*]], %[[ANY:.*]], %[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.u2(%[[ANY:.*]], %[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.p(%[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.rx(%[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.ry(%[[ANY:.*]]) %[[ANY:.*]]
+        // CHECK-NOT: mqtopt.rz(%[[ANY:.*]]) %[[ANY:.*]]
```
DRovara: This can also be combined into one check string.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+        // ==========================  Check that there are no further MCX operations ==============================
+        // CHECK-NOT: mqtopt.x() [[ANY:.*]] ctrl %[[ANY:.*]] nctrl %[[ANY:.*]]
+
+        return
+    }
+}
+
+
+// -----
+// This test checks if an SWAP gate is parsed and handled correctly
```
DRovara: checks if ~an~ a SWAP gate

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,853 @@
...
+}
+
+// -----
+// This test checks if a mixed controlled SWAP gate is parsed and handled correctly
+module {
+    // CHECK-LABEL: func.func @testMixedControlledSWAPOp
+    func.func @testMixedControlledSWAPOp() {
+        // CHECK: %[[Reg_0:.*]] = "mqtopt.allocQubitRegister"
+        %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 4 : i64}> : () -> !mqtopt.QubitRegister
+
```
DRovara: Would be cool to have a circuit illustration here too, similar to above.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,750 @@
...
+        return
+    }
+}
+
+// -----
+// This test checks if parameterized multiple qubit gates are parsed and handled correctly
+module {
+    // CHECK-LABEL: func.func @testMultipleQubitRotationOp
+    func.func @testMultipleQubitRotationOp() {
```
DRovara: What's the difference between this test and the one above?

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,750 @@
...
+        %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 2 : i64}> : () -> !mqtopt.QubitRegister
+
+        // CHECK: %[[Reg_1:.*]], %[[Q0_0:.*]] = "mqtopt.extractQubit"(%[[Reg_0]]) <{index_attr = 0 : i64}>
+        %reg_1, %q0_0 = "mqtopt.extractQubit"(%reg_0) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+
+        // CHECK: %[[Reg_2:.*]], %[[Q1_0:.*]] = "mqtopt.extractQubit"(%[[Reg_0]]) <{index_attr = 1 : i64}>
+        %reg_2, %q1_0 = "mqtopt.extractQubit"(%reg_0) <{index_attr = 1 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+
+        // CHECK: %[[Q0_1:.*]] = mqtopt.x() %[[Q0_0]] : !mqtopt.Qubit
+        %q0_1 = mqtopt.x() %q0_0 : !mqtopt.Qubit
```
DRovara: For a bell state, this should be a `h` gate, not an `x` gate.

DRovara (changes requested):
Thanks a lot @BertiFlorea for the next revision!

I like the changes that you introduced, but I still have some minor concerns, please see my comments regarding that.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: This test should be called `@testControlled...` since the difference to the one above is that it uses the controlled version of the methods tested above.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: I've added a separate test which verifies the remaining case where all the resources in our dialect use an operand. The case where all the resources use a static attribute is already covered in the insert case for static attributes.

I would argue, for the moment, not to add all the other combinations because if one command fails in some way, it would already fail in the other tests for `alloc` or `extract`.

(Fine by me either way)

DRovara: Sure, sounds good to me. At the end of the day, these types of test cases only really matter if we have something like canonicalisation passes anyways, so we can just wait until we have the need for them.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: As far as I understand, we cannot have other ways of writing this instruction due to the traits we use and how our dialect defines unitary operations. I tested many combinations, including the ones suggested in the following comment. However, they either fail due to one of the traits or because the assembly printer doesn't allow it.

The assembly printer (the printer responsible for taking IR and printing textual output) compresses the target and control qubits in different lists because of the `AttrSizedResultSegments` trait.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
BertiFlorea: Addressed in the comment above.
Note: Indeed, here was a typo in the check command in `%[[Q02_2:.*]]:2` which should have been `%[[Q02_1:.*]]:2`.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,1093 @@
...
+    }
+}
+
+// -----
+// This test checks if the ExtractOp is parsed and handled correctly using a static attribute.
+module {
+    // CHECK-LABEL: func.func @testExtractOpAttribute
+    func.func @testExtractOpAttribute() {
+        // CHECK: %[[Reg_0:.*]] = "mqtopt.allocQubitRegister"
+        %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 2 : i64}> : () -> !mqtopt.QubitRegister
```
DRovara: Just a small thing, but I think we should use a size of 1 if we only need one qubit for the test (same in later tests).

I can imagine a canonicalisation in the future that reduces the number of allocated qubits to the smallest required number, so in that case, this test would suddenly fail.

Not super important, but I guess there's no hard in changing it.

mlir/test/Dialect/MQTOpt/IR/bell_state_value_dialect.mlir
```diff
@@ -6,28 +6,1093 @@
...
+        //===------------------------------------------------------------------===//
+        //       ┌──────┐
+        // q0_0: ┤ SWAP ├ q0_1
+        //       └───┬──┘
+        //       ┌───┴──┐
+        // q1_0: ┤ SWAP ├ q1_1
+        //       └───┬──┘
+        // q2_0: ────■─── q2_1
+        //           |
+        // q3_0: ────■─── q3_1
```
DRovara: I think the correct way to represent the SWAP would be to have just one big block that says SWAP over the two qubits, rather than the two small SWAP blocks.

DRovara (approved):
Thanks a lot @BertiFlorea for the work! I'm super happy with how all of this turned out, and the negative tests look great as well. I just have two minor changes left I would suggest.

Also, I believe it's time to rename the test file. How about something like `dialect_features.mlir`?

After that, I believe this PR is ready to merge.

burgholzer (approved):
This is great 🙌🏼 Many thanks @BertiFlorea for the work on this and @DRovara for diligently reviewing the PR. 🙏🏼 

I just have two really highly-level questions:
- Is the explicit `return` statement in all these tests necessary? It kind of seems wasteful.
- The creation of register, extraction of qubits, as well as the insertion and deallocation of registers is creating quite some boilerplate code for the tests. Is there any way to abstract this into a utility method or something like that? This would reduce the size of the test file considerably and would avoid having to repeat the same code pattern over and over again.

None of the above is a dealbreaker for getting this merged. I was just wondering if there is some room for improvement here.

BertiFlorea:
Thanks for the questions/suggestions @burgholzer and @DRovara!

> * Is the explicit return statement in all these tests necessary? It kind of seems wasteful.

Regarding the above, I don't have a precise resource to point at, but as far as I understood from reading, you would need to have a terminator at the end of each block. Additionally, `func.func` carries the trait
`ImplicitTerminator<func::ReturnOp>` which means that the verifier checks that the last op in every block is a `func.return` (so not having a return statement would result in wrong IR, therefore an error). I've also checked how tests are written in Catalyst and all types of tests have a return statement, so the explanation from before seems to hold. However, I agree with you, it is wasteful.

> * they advocate very much for only checking what should actually be checked. So if I interpret this right, then they would not check the register creation, the extraction, insertion and deallocation for every single test, but would only check them in their respective tests and focus on just the essential checks for a particular unit test in the remainder. Opinions on that?

Indeed, I also agree that we can remove the checks which are not related to the functionality of a given test. Should I do that before we merge it?

burgholzer:
> Thanks for the questions/suggestions @burgholzer and @DRovara!
> 
> > * Is the explicit return statement in all these tests necessary? It kind of seems wasteful.
> 
> Regarding the above, I don't have a precise resource to point at, but as far as I understood from reading, you would need to have a terminator at the end of each block. Additionally, `func.func` carries the trait `ImplicitTerminator<func::ReturnOp>` which means that the verifier checks that the last op in every block is a `func.return` (so not having a return statement would result in wrong IR, therefore an error). I've also checked how tests are written in Catalyst and all types of tests have a return statement, so the explanation from before seems to hold. However, I agree with you, it is wasteful.
> 
> > * they advocate very much for only checking what should actually be checked. So if I interpret this right, then they would not check the register creation, the extraction, insertion and deallocation for every single test, but would only check them in their respective tests and focus on just the essential checks for a particular unit test in the remainder. Opinions on that?
> 
> Indeed, I also agree that we can remove the checks which are not related to the functionality of a given test. Should I do that before we merge it?

Just had a brief discussion with Copilot about this. See https://copilot.microsoft.com/shares/N9T4QvPEsMifWzBmqwEBQ

It seems to agree that the `return` statement is necessary.
It also suggests some potential solutions for tackling the duplication. However, I feel like none of these are directly applicable.

I'd say let's do one last refinement on the CHECK statements and their grouping here and merge this.

I'll also briefly request a review from Copilot. Sometimes it provides helpful comments.
Feel free to ignore them if they do not seem helpful.

DRovara:
> * they typically collect all the check statements and do not interleave them with the actual code. That kind of looks nicer to me. Any opinions on that?

Yes, I was aware of that. But afaik, they don't strongly push towards that or the alternative. In any case, I don't have a strong preference, but here are some points:
- The "interleaving" style we use right now works because the input and output are roughly a 1-to-1 mapping. Once this mapping is not so clear anymore, collecting the checks at the bottom is definitely easier.
- On the other hand, at least to me, the blocked checks in the end give the impression that that block should, on its own, represent a full program. Once we start only testing slices of a program, that may introduce confusion.
- Just a minor point, but intuitively I feel like automated test generation tools have an easier time with collecting blocks of checks at the end of the program, rather than interleaving them, but that's just a hunch.

So there are arguments for and against both styles. I am happy with either alternative, to be honest. It's largely just a matter of preference.

> * they advocate very much for only checking what should actually be checked. So if I interpret this right, then they would not check the register creation, the extraction, insertion and deallocation for every single test, but would only check them in their respective tests and focus on just the essential checks for a particular unit test in the remainder. Opinions on that?

I didn't know about that, but it absolutely makes sense. However, that also introduces a new question:
The check string that tests e.g. the `extractOp` introduces an identifier for the extracted variable that is later used to reference it in the following operations. At the same time, that check string also uses identifiers that were introduced by previous check strings. So the alternatives are, more or less:
1) Use checks for all setup code so that we can access those identifiers and make sure the following operations use the correct qubits.
2) Just don't use an identifier for the first tested operations and instead accect `[[ANY:.*]]`

Approach 2) seems fine as well for this test file, but as soon as optimisation passes are also involved, it won't work anymore, so at least that's something to keep in mind.

burgholzer:
> > * they typically collect all the check statements and do not interleave them with the actual code. That kind of looks nicer to me. Any opinions on that?
> 
> 
> 
> Yes, I was aware of that. But afaik, they don't strongly push towards that or the alternative. In any case, I don't have a strong preference, but here are some points:
> 
> - The "interleaving" style we use right now works because the input and output are roughly a 1-to-1 mapping. Once this mapping is not so clear anymore, collecting the checks at the bottom is definitely easier.
> 
> - On the other hand, at least to me, the blocked checks in the end give the impression that that block should, on its own, represent a full program. Once we start only testing slices of a program, that may introduce confusion.
> 
> - Just a minor point, but intuitively I feel like automated test generation tools have an easier time with collecting blocks of checks at the end of the program, rather than interleaving them, but that's just a hunch.
> 
> 
> 
> So there are arguments for and against both styles. I am happy with either alternative, to be honest. It's largely just a matter of preference.

Yeah. I see that there is no clear winner, but maybe a slight tendency towards collecting blocks of checks. If there are no objections, I'd go with that.

> 
> > * they advocate very much for only checking what should actually be checked. So if I interpret this right, then they would not check the register creation, the extraction, insertion and deallocation for every single test, but would only check them in their respective tests and focus on just the essential checks for a particular unit test in the remainder. Opinions on that?
> 
> 
> 
> I didn't know about that, but it absolutely makes sense. However, that also introduces a new question:
> 
> The check string that tests e.g. the `extractOp` introduces an identifier for the extracted variable that is later used to reference it in the following operations. At the same time, that check string also uses identifiers that were introduced by previous check strings. So the alternatives are, more or less:
> 
> 1) Use checks for all setup code so that we can access those identifiers and make sure the following operations use the correct qubits.
> 
> 2) Just don't use an identifier for the first tested operations and instead accect `[[ANY:.*]]`
> 
> 
> 
> Approach 2) seems fine as well for this test file, but as soon as optimisation passes are also involved, it won't work anymore, so at least that's something to keep in mind.
> 

Hm. Good point. I think this is also part of what they advocate for on the page when they say tests should only test what they really intend to test. So I think it would be fair to not fix the identifiers in these tests because they are not essential to the actual test. Maybe I am overlooking things though.

BertiFlorea:
I've grouped the check statements similarly to how they are presented in the testing guide. Please let me know if there are any necessary changes left!

DRovara:
Yeah I'd also say we follow the testing guide.

That means we remove all check strings that are not directly related to the tested operations (e.g. all the allocs and deallocs)

BertiFlorea:
Done ✅ 
Many thanks for your feedback! 😃

mlir/test/Dialect/MQTOpt/IR/dialect_features.mlir
```diff
@@ -0,0 +1,772 @@
...
+
+// -----
+// This test checks if a mixed controlled SWAP gate is parsed and handled correctly.
+module {
+    // CHECK-LABEL: func.func @testMixedControlledSWAPOp
+    func.func @testMixedControlledSWAPOp() {
+        // CHECK: %[[Reg_1:.*]], %[[Q0_0:.*]] = "mqtopt.extractQubit"(%[[Reg_0]]) <{index_attr = 0 : i64}>
+        // CHECK: %[[Reg_2:.*]], %[[Q1_0:.*]] = "mqtopt.extractQubit"(%[[Reg_1]]) <{index_attr = 1 : i64}>
+        // CHECK: %[[Reg_3:.*]], %[[Q2_0:.*]] = "mqtopt.extractQubit"(%[[Reg_2]]) <{index_attr = 2 : i64}>
+        // CHECK: %[[Reg_4:.*]], %[[Q3_0:.*]] = "mqtopt.extractQubit"(%[[Reg_3]]) <{index_attr = 3 : i64}>
```
burgholzer:
```suggestion
```
Seems like a leftover.

burgholzer (approved):
Many thanks! I believe this is great and ready to go in!
I'll apply the one suggestion where there were still leftover checks and will then merge the PR.
That should allow us to tackle #960 next

Issue #925, opened 2025-04-14
👷🏼 MLIR - Set up proper cross-plattform CI


burgholzer:
### What's the problem this feature will solve?

Right now, our MLIR CI pipeline (https://github.com/munich-quantum-toolkit/core/blob/main/.github/workflows/ci_mlir.yml) is very barebones, as it only runs on Ubuntu and simply executes the tests.
Our typical CI (https://github.com/munich-quantum-toolkit/core/blob/main/.github/workflows/ci.yml with workflows defined here: https://github.com/munich-quantum-toolkit/workflows/tree/main/.github/workflows) is much more elaborate and reliable.

In the short time working on MLIR, this has already bitten us a couple of times, where Ubuntu would just be slightly more forgiving than macOS in certain aspects, and the code that would run just fine in CI would fail on macOS.
Furthermore, we currently do not track coverage for the MLIR part of the codebase, which creates a lot of uncertainty on how well the code base is tested.

### Describe the solution you'd like

The goal of this issue is to set up a more complete CI pipeline for the MLIR subcomponent of MQT Core.
This involves (at least) setting up runners on 
- `macos-13` (x86, intel),
- `macos-14` (arm64),
- `windows-latest`.

Ideally, we could "just" tweak the existing CI workflows and cover all scenarios.
The main challenge in all of this is to get the MLIR/LLVM dependency installed with an appropriate version and, hopefully, without building it from scratch on CI runners.
From experience, an additional challenge to the installation itself can be the struggle to have the CMake build find the installed library on all the different operating systems.

In addition to the above and similar to our regular CI, a coverage collection run should be set up. I'd expect this to need some reading on best practices for that in the MLIR ecosystem. This will also need adjustments to the codecov setup (https://github.com/munich-quantum-toolkit/core/blob/main/.github/codecov.yml), where a new "MLIR" flag probably makes sense.

mhmohona:
Hello @burgholzer, thank you for your kind review on #993. I have addressed the feedback, please have a look.

PR #952, opened 2025-05-14
:construction_worker: add windows CI for MLIR

DRovara:
## Description

This PR adds MLIR CI tests for windows runners.
This is achieved by building MLIR from source so the first execution should take a bit longer.
Afterwards, the cache will be used to store the built versions of MLIR.

This PR contributes to #925

**Note:** To install a specific version, the runner clones the tag `llvmorg-{version}.1.0`. It seems that the first release of a new major `llvm-project` version is typically called `{version}.1.0` so this should keep working, but there is no guarantee.

## Checklist:

### DIFF
--- a/.github/workflows/ci_mlir.yml
+++ b/.github/workflows/ci_mlir.yml
@@ -189,10 +189,10 @@ jobs:
           "{0}" %*' -f $targetPath
 
           # Write to a directory that's definitely in PATH
-          $batchContent | Out-File -FilePath "C:\Windows\quantum-opt.cmd" -Encoding ascii
+          $batchContent | Out-File -FilePath "C:\Windows\quantum-opt" -Encoding ascii
 
           # Verify the batch file
-          Get-Content "C:\Windows\quantum-opt.cmd"
-          Write-Output "Batch file created at C:\Windows\quantum-opt.cmd pointing to $targetPath"
+          Get-Content "C:\Windows\quantum-opt"
+          Write-Output "Batch file created at C:\Windows\quantum-opt pointing to $targetPath"
 
           cmake --build build --config Release --target check-quantum-opt
--- a/.github/workflows/ci_mlir.yml
+++ b/.github/workflows/ci_mlir.yml
@@ -184,7 +184,8 @@ jobs:
         run: |
           cmake --build build --config Release --target quantum-opt
           tree build /F /A
-          echo "${{ github.workspace }}\\build\\mlir\\tools\\quantum-opt\\Release" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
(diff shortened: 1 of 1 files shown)


DRovara:
> Thanks @DRovara for pushing this 🙏🏼 This was already looking pretty great. I just applied some small finishing touches here and there.
> 
> Let's see if CI passes and hope that caching works once this is on the `main` branch. Waiting two hours for CI certainly is not fun 🙃
> 
> I guess my only real question would be if we really want to check out the `X.1.0` version of MLIR or whether we would actually like to check out the latest `X.Y.Z` version, i.e., the latest minor and patch version, which would mirror what the Ubuntu runner is currently doing.
> 
> Potentially, we might even opt for doing both in the future, testing the lowest supported LLVM version and the latest one. Similar to how we do testing for the Python package at the moment. But I guess this is something to keep in mind for the future.

Sure, looks good to me!
I guess you're right about getting the latest minor patch version.
Is there any easy way to do that though or do we manually have to filter for it?

burgholzer:
> Sure, looks good to me! I guess you're right about getting the latest minor patch version. Is there any easy way to do that though or do we manually have to filter for it?

Good question. I do not really know to be honest.
We could simply specify the version for LLVM explicitly down to the patch version for the Windows CI runs. That way, we still need to manually update that once a new version comes out, but we don't need to do any complicated GitHub API parsing and filtering. What do you say?

burgholzer:
> I'm not so sure about that. First of all, that could lead to a lot of manual maintenance work, and also that could cause confusion because suddenly after an update, the Linux CI may fail while the one for Windows still passes because we didn't manually update it yet. Before we do that, I'd rather try to invest half an hour more to see if there's an automated approach for that.

Fine with me. And I generally agree with your statement. I just became a bit cautious over the years to demand such extra miles ;o)

DRovara:
Alright, I'm not sure if this is the most idiomatic/cleanest way to do it (would be cool if you could have another look at it @burgholzer), but this solution seems to work.

I use `git ls-remote --tags https://github.com/llvm/llvm-project.git "llvmorg-${{ matrix.llvm-version }}.*"`, that allows us to check the tags without needing to clone the repo. Then we extract the newest revision for the given version using string manipulation, and use that as key for the cache.

This seems to work (let's see if the CI fully passes)

burgholzer (approved):
Thanks for your work on this @DRovara 🙏🏼 
The test execution command looks slightly cryptic to me, but I currently do not have the time myself to search for better ways of runnning the tests and this seems to work.

I think some of these commands would potentially benefit from 
```
shell: bash
```
because then the syntax is much more uniform between the systems.

Let's tackle this in a potential follow-up PR though and get this in 😌

PR #975, opened 2025-05-30
✨ Random State Generation

MatthiasReumann:
## Description

This function adds a bunch of utility functions to generate random vector decision diagrams and accompanying unit tests.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - 📝 Add documentation page for MLIR ([#931]) ([**@ystade**])
 - ✨ Initial implementation of the mqtdyn Dialect ([#900]) ([**@DRovara**], [**@ystade**])
 - ✨ Add function to approximate decision diagrams ([#908]) ([**@MatthiasReumann**])
+- ✨ Add functions to generate random vector DDs ([#975]) ([**@MatthiasReumann**])
 
 ### Fixed
 
@@ -110,6 +111,7 @@ _📚 Refer to the [GitHub Release Notes](https://github.com/munich-quantum-tool
 
 <!-- PR links -->
 
+[#975]: https://github.com/munich-quantum-toolkit/core/pull/975
 [#973]: https://github.com/munich-quantum-toolkit/core/pull/973
 [#964]: https://github.com/munich-quantum-toolkit/core/pull/964
 [#934]: https://github.com/munich-quantum-toolkit/core/pull/934


MatthiasReumann:
Any idea why this fails, @burgholzer? 

---
🙋🏼‍♂️ (Besides the new functionality) What's your take on moving the `dd.make<name>State(...)` methods from the `dd::Package` into `StateGeneration.cpp` and refactoring the methods to functions?

burgholzer:
> Any idea why this fails, @burgholzer? 
> 

Never seen that failure before. 
If I had a guess, it is some bad interplay between macOS 15 and the gcc downloaded from homebrew.
Although I have no idea why that would be showing up here and now.
Quick googling also did not really reveal much.
I'd say ignore these two for now. They are in the extended test suite anyways.
If you happen to have a macOS computer around, it would be nice if you could tr to reproduce this locally.

Maybe @ystade can help here?

> 
> ---
> 
> 🙋🏼‍♂️ (Besides the new functionality) What's your take on moving the `dd.make<name>State(...)` methods from the `dd::Package` into `StateGeneration.cpp` and refactoring the methods to functions? 

Very much like the idea. I might, at some point, even have started such a refactoring, but never followed through.
I'd probably do this in a separate PR though to keep the PRs as isolated as they can be.

burgholzer:
> Indeed, I also had troubles with gcc-14 on Mac even though I could not reproduce the exact same error. For me it was already complaining at the end of the CMake build and giving me a warning that an option is not supported. THis led to the fix I just pushed. With that fix and after updating gcc-14 because apparently there was a bug recently, see [here](https://github.com/Homebrew/homebrew-core/issues/219174), it compiled. Let's see whether this also changes something in the CI.
> 
> 
> 
> @burgholzer This was just a quick fix in the cmake configuration, I am quite sure the proper solution looks different.

I faced that problem once before. That's so strange to me. The CMake function we call before adding the flag is an explicit feature check, which passes...

The underlying error here almost must be due to a recent update. I remember running the extended CI not too long ago for some PR of mine and everything worked.

Truly strange..

burgholzer:
> Note: The `windows-2022` and `windows-2025` versions fail due to ambiguity in the approximation algorithm, or the problem itself. Essentially, `1/sqrt(2)` is added twice to the `std::priority_queue` and (as it seems) some times the order of the _to-be-deleted_ terminal edge changes. I will look into fixing this.
> 
> 

Fairly sure that this is due to some numerical inaccuracy, where there are two slightly different values of 1/sqrt(2). That could lead to another node being picked for approximation.

MatthiasReumann:
I just could reproduce the exact issue both with `gcc@14` and `gcc@15` (via homebrew) using the `MacOSX15.2` and `MacOSX14.5` (can be specified via `-DCMAKE_OSX_SYSROOT`) SDKs on my M1 Air. 

[This](https://www.reddit.com/r/cpp_questions/comments/1kxmlsn/c_boundsh_not_found_after_upgrading_to_gcc15/) reddit post describes the same issue, I think. No idea how one should fix this 🤔

burgholzer:
> I just could reproduce the exact issue both with `gcc@14` and `gcc@15` (via homebrew) using the `MacOSX15.2` and `MacOSX14.5` (can be specified via `-DCMAKE_OSX_SYSROOT`) SDKs on my M1 Air. 
> 
> 
> 
> [This](https://www.reddit.com/r/cpp_questions/comments/1kxmlsn/c_boundsh_not_found_after_upgrading_to_gcc15/) reddit post describes the same issue, I think. No idea how one should fix this 🤔

Hm.. that's a tough one..
But given how macOS-15 is not in our default test matrix, gcc is not the main compiler under macOS, and this seems to be unrelated to our own code, I would be willing to just ignore this for now, disable the extensive cpp tests for this PR and hope that this resolves itself over time.

burgholzer:
> > > I just could reproduce the exact issue both with `gcc@14` and `gcc@15` (via homebrew) using the `MacOSX15.2` and `MacOSX14.5` (can be specified via `-DCMAKE_OSX_SYSROOT`) SDKs on my M1 Air.
> 
> > > [This](https://www.reddit.com/r/cpp_questions/comments/1kxmlsn/c_boundsh_not_found_after_upgrading_to_gcc15/) reddit post describes the same issue, I think. No idea how one should fix this 🤔
> 
> > 
> 
> > Hm.. that's a tough one.. But given how macOS-15 is not in our default test matrix, gcc is not the main compiler under macOS, and this seems to be unrelated to our own code, I would be willing to just ignore this for now, disable the extensive cpp tests for this PR and hope that this resolves itself over time.
> 
> 
> 
> Out of curiosity I tested the `MacOSX15.2` sdk with `gcc@13` - and it works. The workflow uses `gcc@14`.

Good to know. Maybe we should (temporarily) switch the workflows to that gcc version. But I suppose as long as it is not affecting our default CI platforms, we are still good either way.
However, it could be worth creating an issue for this, so that people running into the issue know that we are aware of it and that there are alternatives.
@MatthiasReumann would you mind creating such an issue?

include/mqt-core/dd/StateGeneration.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @brief Generate exponentially large vector DD. Use @p seed for randomization.
+ * @param levels The number of levels in the vector DD.
+ * @param seed The seed used for randomization.
+ * @param dd The DD package to use for generating the vector DD.
+ * @return The exponentially large vector DD.
+ */
+VectorDD generateExponentialState(std::size_t levels, std::size_t seed,
+                                  Package& dd);
```
burgholzer: Minor nitpick and probably style question, but I would have probably put the `seed` parameter last so that the seeded and the non-seeded version can be called the same way, with one just having an additional parameter.

Same argument holds for the other set of public functions here.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+                 Package& dd) {
+  constexpr double eps = std::numeric_limits<double>::epsilon();
+
+  const auto alpha = randomComplexOnUnitCircle(gen, dist);
+  const auto beta = randomComplexOnUnitCircle(gen, dist);
+  const auto norm = std::sqrt(2);
+
+  const std::array<vEdge, RADIX> edges{
+      vEdge::terminal(dd.cn.lookup(alpha / norm)),
+      vEdge::terminal(dd.cn.lookup(beta / norm)),
```
burgholzer: Minor: I think it could be slightly preferential to use our dd::SQRT2_2 constant and multiply with it instead of dividing by `sqrt(2)`.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+  };
+
+  // Check: Properly normalized.
+  assert(std::sqrt(ComplexNumbers::mag2(edges[0].w) +
+                   ComplexNumbers::mag2(edges[1].w)) < 1 + eps);
```
burgholzer: This is true by design, right?
Could just remove this entirely.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+VectorDD generateExponentialState(const std::size_t levels, Package& dd) {
+  std::random_device rd;
+  return generateExponentialState(levels, rd(), dd);
+}
+
+VectorDD generateExponentialState(const std::size_t levels,
+                                  const std::size_t seed, Package& dd) {
+  std::vector<std::size_t> nodesPerLevel(levels - 1); // [2, 4, 8, ...]
+  std::generate(nodesPerLevel.begin(), nodesPerLevel.end(),
+                [exp = 1]() mutable { return std::pow(2, exp++); });
```
burgholzer: Truly minor, but given how generating exponential DDs of >64 qubits is completely unrealistic anyway, the logic for the powers of two can be realized using bitshifts alone, I.e., `1ULL << exp++`

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+      e.p->e[0].p = curr[indices[2 * i]].p;
+      e.p->e[1].p = curr[indices[(2 * i) + 1]].p;
+    }
+
+    prev = std::move(curr);
+  }
+
+  vEdge ret{root.p, Complex::one()};
+  dd.incRef(ret);
+  return ret;
```
burgholzer: Hm, I think there is one fundamental problem with the (top-down) approach here: none of these nodes will actually be placed in the unique table of the DD package.
This would require calls to `makeDDNode`, which performs the respective logic (normalization + lookup). For this to work properly though, the construction has to be bottom-up from terminal to root.
That's essentially how all DD construction methods work because the normalization of a node theoretically affects all nodes above it.

That is something I definitely missed in my original branch on this.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+  return generateRandomState(levels, nodesPerLevel, strategy, rd(), dd);
+}
+
+VectorDD generateRandomState(const std::size_t levels,
+                             const std::vector<std::size_t>& nodesPerLevel,
+                             const GenerationWireStrategy strategy,
+                             const std::size_t seed, Package& dd) {
+  assert(levels > 0U && "Number of levels must be greater than zero");
+  assert(nodesPerLevel.size() == levels - 1 &&
+         "Number of levels - 1 must match nodesPerLevel size");
```
burgholzer: I believe this method is also missing the safety checks for the node configuration from the original branch. Specifically that there can never be more than twice as many nodes as in the previous level. The condition each level can have at most a power of two in nodes is implied by that given how there is only one root node.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+                    [&n, r = std::size_t{0}]() mutable { return (r++) % n; });
+      break;
+    }
+    case RANDOM: {
+      IndexDistribution indexDist{0, curr.size() - 1};
+      std::generate(indices.begin(), indices.end(),
+                    [&indexDist, &gen]() { return indexDist(gen); });
+      break;
+    }
+    }
```
burgholzer: I think this is missing one important part of the original branch, but I might be wrong about that: it needs to be ensured that each node is actually somewhat connected to the root, meaning that there are no hanging nodes which are entirely disconnected.

The round-robin one ensures that (I think). The random one does not. It could very well end up not selecting a particular node at all.

test/dd/test_state_generation.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <memory>
+#include <vector>
+
+using namespace dd;
+
+namespace {
+double norm(const std::vector<std::complex<double>>& v) {
+  double sum{};
+  for (const auto& entry : v) {
+    sum += std::pow(std::abs(entry), 2);
```
burgholzer:
```suggestion
    sum += std::norm(entry);
```
There's an stdlib function for that 😉

test/dd/test_state_generation.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto dd = std::make_unique<Package>(nq);
+  auto state = generateRandomState(nq, nodesPerLevel, ROUNDROBIN, *dd);
+
+  EXPECT_NEAR(norm(state.getVector()), 1., 1e-6);
+  EXPECT_EQ(state.size(), 2); // Node plus Terminal.
+
+  dd->decRef(state);
+  dd->garbageCollect(true);
+
+  EXPECT_EQ(dd->vUniqueTable.getNumEntries(), 0);
```
burgholzer, line 59: I am fairly sure this is where one would notice that the nodes are not added to the DD Package at all.
You could place an assertion after the creation of the state that the unique table contains exactly as many nodes as configured.
That would probably be helpful everywhere in the tests.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,145 @@
...
+      const vEdge& e = prev[i];
+      e.p->e[0].p = curr[indices[2 * i]].p;
+      e.p->e[1].p = curr[indices[(2 * i) + 1]].p;
+    }
+
+    prev = std::move(curr);
+  }
+
+  vEdge ret{root.p, Complex::one()};
+  dd.incRef(ret);
```
burgholzer, line 159: Truly a micro-optimization, but given how you know exactly how many nodes will lead to one particular other node, the reference counting here could be done by-construction.
Specifically, every time you add a path to a node, you can directly increase its reference count by one.
Then the whole DD does not have to be traversed again. 
(You can keep this in mind when moving over the other state generation routines. They all still have the post-construction reference counting.

test/dd/test_state_generation.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // Expect: The size of the resulting vector DD must be exponentially large.
+  // Expect: Properly increase and decrease the ref counts.
+
+  constexpr std::size_t nq = 3;
+
+  auto dd = std::make_unique<Package>(nq);
+  auto state = generateExponentialState(nq, *dd);
+
+  EXPECT_NEAR(norm(state.getVector()), 1., 1e-6);
+  EXPECT_EQ(state.size(), std::pow(2, nq));
```
burgholzer:
```suggestion
  EXPECT_EQ(state.size(), 1ULL << nq);
```
Always more efficient than floating point arithmetic. Although most compilers will actually perform
The above transformation automatically.

test/dd/test_state_generation.cpp
burgholzer, line 1: These tests are great already. It would be nice to also test the "failure" cases, where the function call is misconfigured.

burgholzer, line 1: Also, could it be worth verifying that the Round Robin strategy is really doing its thing? Or would that test just be tracing the implementation of the feature almost one to one?

burgholzer (changes requested):
Many thanks @MatthiasReumann for pushing this forward. Looking forward to seeing this being available in the main branch. 🙌

As you probably expected by now, I got a couple comments on the implementation as well as the tests.
Turns out I overlooked one kind of critical aspect in my original implementation: that the nodes should also be in the unique table for the resulting DDs to actually be useful.
But I believe this should not be too hard to add to the existing implementation.

src/dd/StateGeneration.cpp
MatthiasReumann: I was wondering about this. Shouldn't a call to `dd.garbageCollect()` remove these hanging nodes (since they have a ref count of 0)?

MatthiasReumann: Nevertheless, I just figured, if one specifies a particular level to have 3 nodes and the resulting DD has only 2 - that would be quite surprising and unintuitive. I will look into this.

burgholzer: If they were correctly added to the unique table, a forced garbage collection would remove them.

However, as you also noticed yourself, it is kind of unintuitive for a function that claims to produce a DD with a certain structure to, well, not produce a DD with said structure. 🙂

src/dd/StateGeneration.cpp
MatthiasReumann: I'm actually using the `makeDDNode` function instead of the `memoryManager<vNode>`, here. The unit tests agree (I've added a check as recommended by you somewhere below). 

The only thing I'm really confused about is the normalization of the resulting edge. I would've assumed that the ingoing edge weight is 1, but it is a `commonFactor` (?). 

Currently, I just ignore this factor. The sum of squared magnitudes of the edge weights - the norm - is still 1 for the node, if I do so.

burgholzer: Yeah. I expected something like that because of the way our normalization works.
One would have to check the exact implementation of the normalization scheme to see how the common factor is derived. Just out of memory I think it is something like:
Given weight `a` and `b` the normalization first computes the norm
```
norm = sqrt(|a|^2 + |b|^2)
```
And a common factor
```
factor = norm / |a|
```
Assuming that `a` has a larger magnitude than `b`.
And then does something like 
```
top = a * factor
a' = |a| / norm
b' = b / top
```
One could think about adapting the weight generation scheme so that it follows the normalization. However, I believe that this would actually not add too much value given how the numbers themselves are randomized.
The only drawback of using `makeDDNode` and discarding the top edge weight is that its values will have been added to the complex value unique table, while not actually being used anywhere, which is a waste.

Edit: you could work around this by working with CachedEdges here. They are similar to regular edges with the main difference being that their top edge weight is not yet looked up in the unique table.

MatthiasReumann: Do I understand the CachedEdges correctly? 

A DD built with CachedEdges is a temporary DD - no ref counts have been increased. Only when one calls `dd.incRef(e)`, where `e` is a `vEdge`, at the end (keeping the discussion below out of mind) the DD becomes "alive"?

---

Edit: I've pushed a working solution (as far as unit tests go). Feedback would be great.

test/dd/test_state_generation.cpp
MatthiasReumann: Done. 

I've replaced the assertions with cpp exceptions - yes / no? I feel like if there is a place for exceptions it's use cases such as this one. (Moreover: Exceptions also fix the problem that the `assert` is removed in release builds).

MatthiasReumann: Mhm, good question. I've added such a check. You be the judge - I'm not entirely sure if these checks make a lot of sense.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,183 @@
...
+                             const std::size_t seed) {
+  if (levels <= 0U) {
+    throw std::invalid_argument("Number of levels must be greater than zero");
+  }
+  if (nodesPerLevel.size() != levels - 1) {
+    throw std::invalid_argument(
+        "Number of levels - 1 must match nodesPerLevel size");
+  }
+
+  [[maybe_unused]] bool inc{};
```
burgholzer: Is there any value in this variable?
If you don't want to use the return value of some nodiscard function, you can simply use `std::ignore = ...`

test/dd/test_state_generation.cpp
burgholzer: Exceptions are certainly the way to go here. I think they fit quite well in the way they are placed here at the moment.

burgholzer: Yeah, to be honest, I guess it does not add too much value 🙃 might as well remove it to keep the PR a little smaller. Sorry for that.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,183 @@
...
+  const std::array<vCachedEdge, RADIX> edges{
+      vCachedEdge::terminal(alpha / SQRT2_2),
+      vCachedEdge::terminal(beta / SQRT2_2),
+  };
+  const vCachedEdge ret = dd.makeDDNode(v, edges);
+
+  dd.cn.incRef(ret.p->e[0].w);
+  dd.cn.incRef(ret.p->e[1].w);
+
+  return ret;
```
burgholzer: Related to the main comment below.
This method should also take two `vNode*` for which the function generates two random edge weights.
These nodes are supposed to be used instead of the terminals here.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,183 @@
...
+ *        Due to the use of cached edges, the weight of the returned edge is not
+ *        stored in the lookup table.
+ */
+vCachedEdge randomNode(Qubit v, Generator& gen, AngleDistribution& dist,
+                       Package& dd) {
+  const auto alpha = randomComplexOnUnitCircle(gen, dist);
+  const auto beta = randomComplexOnUnitCircle(gen, dist);
+  const std::array<vCachedEdge, RADIX> edges{
+      vCachedEdge::terminal(alpha / SQRT2_2),
+      vCachedEdge::terminal(beta / SQRT2_2),
```
burgholzer:
```suggestion
      vCachedEdge::terminal(alpha * SQRT2_2),
      vCachedEdge::terminal(beta * SQRT2_2),
```
The constant `SQRT2_2` is `sqrt(2)/2=1/sqrt(2)`.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+      left.p = curr[indices[2 * i]].p;
+      right.p = curr[indices[(2 * i) + 1]].p;
+
+      inc = dd.getUniqueTable<vNode>().incRef(left.p) &&
+            dd.getUniqueTable<vNode>().incRef(right.p);
+    }
+
+    prev = std::move(curr);
+  }
```
burgholzer: At present, this will not quite work just right yet as none of the nodes will be placed in the right unique table bucket. That's because you are altering the successors _after_ calling `makeDDNode`.
You would probably notice this if you would turn the resulting DDs into vectors and then into DDs again from the vector. Ideally (if round off errors don't interfere), you would expect both DDs to be equivalent, that is their root node pointing to the same node and the top edge weight being exactly one.
I am very very certain that the current implementation would yield a second DD that differs to the first one in all but the lowest level nodes.

The proper solution here starts to construct the DDs bottom up, starting from terminals and constructing as many unique `vCachedEdge` as requested for the last level. Then, it iterates upward until the root in a level-by-level fashion. Similarly to the top-down approach here. The next layer is always constructed by creating the respective number of new nodes with the `randomNode` method, where each call is fed with two node pointers according to the indexing strategy.
At the end, the top edge weight is replaced by One, as in the current implementation.

The reference count for the complex numbers is updated in the randomNode function as it is currently after the makeDDNode call. The reference count for the nodes themselves should also be updated there. That is, after the makeDDNode call, the reference count of the two successor nodes should be increased by one in the randomNode function.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+  vEdge ret{root.p, Complex::one()};
+
+  dd.cn.incRef(ret.w);
```
burgholzer: This is not necessary, as the constant one is ignored during reference counting and will never be collected anyway.

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+      // Note that the following assignments ignore the edge weights of the
+      // current layer. Since these aren't in the lookup table, we can
+      // safely do so.
+
+      left.p = curr[indices[2 * i]].p;
+      right.p = curr[indices[(2 * i) + 1]].p;
+
+      inc = dd.getUniqueTable<vNode>().incRef(left.p) &&
+            dd.getUniqueTable<vNode>().incRef(right.p);
```
burgholzer: I am not 100% sure that this form of using the existing reference counting is the most efficient way to do things/saves that much.
What I had in mind was to actually bump the reference count of the respective nodes manually.

However, we should not waste too much time on this, as there is an ongoing PR that is part of unitaryHACK, which removes reference tracking altogether.
So this might not even be as important in the middle term.

src/dd/StateGeneration.cpp
burgholzer: Caching and alive/dd are two separate concepts for the most part.

Alive and dead just refer to a node having non-zero or zero reference count.
A cached edge is simply one where the respective top weight has not been looked up in the unique table for the numbers yet.
This is helpful for many bottom up constructions, the cached weights go into the next higher level, where they will be subject to change due to normalization. So many of the cached numbers are truly temporary. Adding them to the unique table would be a waste of unique table space.

Hope that makes sense.

src/dd/StateGeneration.cpp
MatthiasReumann: this is great, thanks! didn't know something like this existed🙇🏻

src/dd/StateGeneration.cpp
MatthiasReumann: Yes that makes a lot of sense. Thanks!

src/dd/StateGeneration.cpp
MatthiasReumann: Many thanks 🙏 

At some point, I might need to dive deeper into how the hashing exactly works. I feel like this issue, my misunderstanding of things, keeps popping up. 

---

> You would probably notice this if you would turn the resulting DDs into vectors and then into DDs again from the vector.

I've added the respective check to the unit tests.

src/dd/StateGeneration.cpp
MatthiasReumann: As of now, I've returned to a single `dd.incRef(...)` at the end of the function. Should we keep it that way for now, or do you think it's worth the effort to in-place increase the ref counts (given that ref counts for nodes might be history soon)?

burgholzer: Yeah. That's totally fine and adequate!

src/dd/StateGeneration.cpp
```diff
@@ -0,0 +1,164 @@
...
+ * stored in the lookup table.
+ */
+vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
+                       AngleDistribution& dist, Package& dd) {
+  const auto alpha = randomComplexOnUnitCircle(gen, dist) * SQRT2_2;
+  const auto beta = randomComplexOnUnitCircle(gen, dist) * SQRT2_2;
+
+  const std::array<vCachedEdge, RADIX> edges{
+      vCachedEdge(left, ComplexValue(alpha)),
+      vCachedEdge(right, ComplexValue(beta))};
```
burgholzer:
```suggestion
ComplexValue randomComplexOnUnitCircle(Generator& gen,
                                               AngleDistribution& dist) {
  const double angle = dist(gen);
  return {std::cos(angle), std::sin(angle)};
}

/**
 * @brief Generate node with edges pointing at @p left and @p right.
 * Initialize edge weights randomly with the constraint that their norm is one.
 * @note The CachedEdge ensures that the weight of the resulting edge is not
 * stored in the lookup table.
 */
vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
                       AngleDistribution& dist, Package& dd) {
  const auto alpha = randomComplexOnUnitCircle(gen, dist) * SQRT2_2;
  const auto beta = randomComplexOnUnitCircle(gen, dist) * SQRT2_2;

  const std::array<vCachedEdge, RADIX> edges{
      vCachedEdge(left, alpha),
      vCachedEdge(right, beta)};
```
Just slightly more ergonomic.

burgholzer (approved):
This is looking phenomenal now! Many thanks for your work on this! 🙏

I'll just apply the last suggestion here directly.

Could you please also add a Changelog entry for this new feature?

(at some point we probably also want to expose this functionality to Python. But that can wait for a separate PR)

burgholzer (approved):
Thanks for cleaning up behind me 😉🧹

burgholzer (approved):

PR #980, opened 2025-06-02
✨ Implement mark-and-sweep garbage collection

### LINKED ISSUE #644: ✨ Mark and Sweep Garbage Collection for DD Package

q-inho:
## Description

The DD package now uses a mark-and-sweep garbage collector instead of reference counting. 
Nodes no longer store reference counts. 
The `incRef` and `decRef` functions now manage the root set for the new garbage collector.

Resolves #644


include/mqt-core/dd/Node.hpp
```diff
@@ -40,13 +38,23 @@ struct NodeBase : LLBase {
...
+   * 0b10000 = mark flag used for mark-and-sweep garbage collection,
+   *  0b1000 = marks a reduced dm node,
+   *   0b100 = marks a dm (tmp flag),
+   *    0b10 = mark first path edge (tmp flag),
+   *     0b1 = mark path is conjugated (tmp flag)
    */
   std::uint16_t flags = 0;
 
+  /// Mark flag used for mark-and-sweep garbage collection
+  static constexpr std::uint16_t MARK_FLAG = 0x10U;
```
burgholzer:
```suggestion
  static constexpr std::uint16_t MARK_FLAG = 0b10000U;
```
To match the docstring from above.

include/mqt-core/dd/Node.hpp
```diff
@@ -30,8 +30,6 @@ namespace dd {
  * Data Layout (8)|(4|2|2) = 16B.
```
burgholzer: These Data layout comments throughout the file will need to be updated.

Just out of interest: could you check what the byte alignment of the resulting structs is?
That would be interesting to know because it influences if this change actually saves memory.

include/mqt-core/dd/Package.hpp
```diff
@@ -169,6 +169,11 @@ class Package {
   RealNumberUniqueTable cUniqueTable{cMemoryManager};
   ComplexNumbers cn{cUniqueTable};
 
+  // Root sets for mark-and-sweep garbage collection
+  std::unordered_set<vNode*> vectorRoots{};
+  std::unordered_set<mNode*> matrixRoots{};
+  std::unordered_set<dNode*> densityRoots{};
```
burgholzer: I am pretty sure that you want to be tracking `vEdge` etc here because the edge weights are also important and need to be kept track of.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +199,100 @@ class Package {
...
+   * @details The complex edge weight and all complex numbers reachable from the
+   * node are marked. The edge itself is added to the appropriate root set so it
+   * is considered live for the mark phase of garbage collection.
+   *
    * @tparam Node The node type of the edge.
-   * @param e The edge to increase the reference count of
+   * @param e The edge to mark and add to the root set
    */
   template <class Node> void incRef(const Edge<Node>& e) noexcept {
     cn.incRef(e.w);
```
burgholzer: This PR will also have to refactor the handling for complex numbers. The edge weights need to be similarly marked/unmarked.
There is no need to create an unordered_set for the numbers though, as the edge weights are always tied to specific edges, which are tracked anyway.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +199,100 @@ class Package {
...
+    }
+    for (const auto& child : p->e) {
+      if (inc) {
+        cn.incRef(child.w);
+      } else {
+        cn.decRef(child.w);
       }
+      markComplexRec(child.p, inc);
     }
   }
```
burgholzer: As per the other comment, I believe to would be best to handle this differently. Specifically, to handle numbers similarly to the nodes here.
Note that the `RealNumber` class does not have any space for flags.
So you probably have to rely on a similar trick to the one we use for marking negative numbers, manipulating the lower bits of the pointer to the numbers.
A 1 in the LSB of the pointer address indicates the number is negative.
One could use the second bit to mark entries for example.

include/mqt-core/dd/statistics/UniqueTableStatistics.hpp
```diff
@@ -19,8 +19,7 @@ namespace dd {
 /// \brief A class for storing statistics of a unique table
 struct UniqueTableStatistics : public TableStatistics {
   /**
-   * @brief The total number of active entries
-   * @details An entry is considered active if it has a non-zero reference count
+   * @brief The total number of active entries.
    */
   std::size_t numActiveEntries = 0U;
```
burgholzer: Active entries do not make any sense any more now that they are not being tracked through continuous reference counting anymore.
So any mentions of it can essentially be removed.

It could be helpful to provide a method in the DD package that still allows to compute this by marking all active entries (using the new mark method) and then counting in the unique table.

src/dd/Package.cpp
```diff
@@ -101,15 +104,23 @@ bool Package::garbageCollect(bool force) {
...
+  for (auto* p : densityRoots) {
+    markNodes(p);
+  }
+
+  const auto cCollect = cUniqueTable.garbageCollect(force);
+  // Collecting complex numbers forces collection of the node tables. The
+  // following calls explicitly use `true` to enforce this.
+  const auto vCollect = vUniqueTable.garbageCollect(true);
+  const auto mCollect = mUniqueTable.garbageCollect(true);
+  const auto dCollect = dUniqueTable.garbageCollect(true);
```
burgholzer: `force` should only be set to true here if complex numbers have been collected. Similar to the original code.

src/dd/Simulation.cpp
```diff
@@ -120,7 +120,7 @@ std::map<std::string, std::size_t> sample(const qc::QuantumComputation& qc,
       auto measurement = dd.measureAll(e, false, mt);
       counts.operator[](measurement) += 1U;
     }
-    // reduce reference count of measured state
+    // remove the measured state edge from the root set
     dd.decRef(e);
```
burgholzer, line 126: Just based on the changes to the docstrings throughout the files I believe these methods deserve a more appropriate name now. Maybe something like `track` and `untrack`?

The rename should also help to revisit all the places where these functions are called, to see whether these calls are still necessary. One of the goals here would actually be to require as little as possible manual intervention.
The garbage collection would ideally check in periodic intervals how full the unique tables are. If one of them could be collected, the marking and collection would start.

src/mqt/core/dd.pyi
```diff
@@ -139,15 +139,15 @@ def simulate(qc: QuantumComputation, initial_state: VectorDD, dd_package: DDPack
         qc: The quantum computation.
             Must only contain unitary operations.
         initial_state: The initial state as a DD. Must have the same number of qubits
-                       as the quantum computation. The reference count of the initial
-                       state is decremented during the simulation, so the caller must
-                       ensure that the initial state has a non-zero reference count.
+                       as the quantum computation. The state is removed from the
+                       root set during the simulation, so the caller must ensure that
+                       it is rooted beforehand.
```
burgholzer: This is actually something I would have hoped to simplify via this PR. Let's see whether we can achieve that.

test/dd/test_package.cpp
```diff
@@ -959,13 +951,55 @@ TEST(DDPackageTest, PackageReset) {
...
+  dd->decRef(vec);
+
+  // matrix root
+  auto mat = getDD(qc::StandardOperation(0, qc::X), *dd);
+  dd->incRef(mat);
+  dd->incRef(mat);
+  EXPECT_EQ(dd->matrixRoots.size(), 1U);
+  dd->decRef(mat);
+  EXPECT_TRUE(dd->matrixRoots.empty());
+  dd->decRef(mat);
```
burgholzer: I think this highlights one of the problems with the current approach: assume that two separate computations currently refer to the same DD and add it to the root set. Then one of them calls the decRef method and runs GC. Suddenly the instance from the other computation will be gone as well.
I believe the solution here is to use an unordered_map instead of the unordered_set, where the values are counts of how many times a particular DD is referred to from the outside.
Increasing the ref count adds an entry to the map with a count of one or increases the count by one if it is already in the map. Decreasing works the opposite way around. Decreases by one if it's already in there. Removes if the count hits zero.

q-inho:
Thank you for the detailed review. I began with a minimal core implementation, but realized later many other codes are interacted, resulting existing code paths to be escaped from my purpose. I really appreciate for your feedback.

include/mqt-core/dd/Node.hpp
q-inho: To verify the memory footprint, I compiled a small test program and checked the results:
```cpp
sizeof(NodeBase)=16 align=8
sizeof(vNode)=64 align=8
sizeof(mNode)=112 align=8
sizeof(dNode)=112 align=8
```
Each structure is aligned to 8 bytes. Because `NodeBase` still contains a pointer, it remains 16 bytes after padding. Vector nodes are 64 bytes, and both matrix and density nodes are 112 bytes.

burgholzer: Thanks for checking. I almost expected this to be the case. This means that the removal of the `ref` member does not effectively reduce the memory footprint in the nodes.
So the main benefit of this PR is essentially the delayed reference counting that only kicks in once the unique tables are "full enough".
That still seems like a reasonable advantage of this mark-and-sweep approach. 🙂

include/mqt-core/dd/Package.hpp
q-inho: Yes. I intentionally store `vNode*`, `mNode*`, and `dNode*` in the root sets. The weight of each edge is already accounted for in `incRef` and `decRef`, which call `cn.incRef(e.w)`/`cn.decRef(e.w)` before adding or removing the node from the root set. Thus, the edge’s complex weight remains tracked, while the root sets only serve to keep the nodes alive for garbage collection.

burgholzer: As per one of the other comments, these functions for handling complex number references counts are to be refactored/removed as well as part of this PR.
The current approach would still work for all but the top edge weights.
I see two options:
- switching back to tracking Edges in these sets (preferred)
- adding a separate set for RealNumber* that tracks the real and imaginary parts of the top-level edge weights.

include/mqt-core/dd/Edge.hpp
```diff
@@ -395,6 +395,23 @@ template <class Node> struct Edge {
...
+  }
+};
+
+template <class Node> struct EdgePtrEqual {
+  bool operator()(const Edge<Node>& lhs, const Edge<Node>& rhs) const noexcept {
+    return lhs.p == rhs.p && lhs.w.r == rhs.w.r && lhs.w.i == rhs.w.i;
+  }
+};
+} // namespace dd
+
```
burgholzer, line 420: This change should not be necessary if I am not overlooking things.
The Edge class defines a hash function at the very end of this file.

And the equality operator defined for the Edge class should also be sufficient here.

include/mqt-core/dd/Package.hpp
```diff
@@ -169,6 +169,18 @@ class Package {
...
+  // reference counting for externally tracked DDs
+  std::unordered_map<vEdge, std::size_t, EdgePtrHash<vNode>,
+                     EdgePtrEqual<vNode>>
+      vectorRoots{};
+  std::unordered_map<mEdge, std::size_t, EdgePtrHash<mNode>,
+                     EdgePtrEqual<mNode>>
+      matrixRoots{};
+  std::unordered_map<dEdge, std::size_t, EdgePtrHash<dNode>,
+                     EdgePtrEqual<dNode>>
+      densityRoots{};
```
```suggestion
  std::unordered_map<vEdge, std::size_t> vectorRoots{};
  std::unordered_map<mEdge, std::size_t> matrixRoots{};
  std::unordered_map<dEdge, std::size_t> densityRoots{};
```
I would have assumed this to "just work". Any particular reason it would not?

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
+      if (it != densityRoots.end()) {
+        if (--it->second == 0U) {
+          densityRoots.erase(it);
+          return true;
+        }
+        return false;
+      }
+    }
+    return false;
+  }
```
burgholzer: I am fairly sure both of these functions can simply be `void` functions.
Their return type is not used in the main library and this would also remove any need for casts to void. In the existing incRef and decRef functions.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
   }
 
+  template <class Node> bool addRoot(const Edge<Node>& e) {
+    if constexpr (std::is_same_v<Node, vNode>) {
+      auto it = vectorRoots.find(e);
+      if (it == vectorRoots.end()) {
+        vectorRoots.emplace(e, 1U);
+        return true;
+      }
+      ++it->second;
```
burgholzer: I am not 100% sure, but if I remember correctly, this can be simplified to
```
vectorRoots[e]++;
```
I might be wrong though.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
+      auto it = densityRoots.find(e);
+      if (it != densityRoots.end()) {
+        if (--it->second == 0U) {
+          densityRoots.erase(it);
+          return true;
+        }
+        return false;
+      }
+    }
+    return false;
```
burgholzer: I believe this function should handle the case where the Edge is not in the root set and error out in that case.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
+  }
+
```
burgholzer: The three branches share quite some code between them.
This could be more compact as something like
```
std::unordered_map<Edge<Node>, size_t> roots;
if constexpr (...) {
  roots = ...
} else if constexpr(...) {
...
} else {
...
}
roots[e]++;
```
Or similar 😌

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
+   * triggered if the unique tables report that a collection might be
+   * necessary.
    *
-   * @param force
-   * @return
+   * @param force If `true`, perform a collection regardless of whether any
+   *              table reports that it may need collecting.
+   * @returns `true` if at least one vector or matrix node, or a complex number,
+   *          was reclaimed. Collections of density-matrix nodes do not affect
+   *          the return value.
```
burgholzer: Actually, density matrix nodes should also influence the return value here.
I'd consider that an oversight in the current implementation.

That should also allow to simplify this sentence a bit.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
   template <class Node> void incRef(const Edge<Node>& e) noexcept {
-    cn.incRef(e.w);
-    const auto& p = e.p;
-    const auto inc = getUniqueTable<Node>().incRef(p);
-    if (inc && p->ref == 1U) {
-      for (const auto& child : p->e) {
-        incRef(child);
-      }
-    }
+    static_cast<void>(addRoot(e));
```
burgholzer: Given how these methods now only consist of single function calls, you might as well inline the definition of the addRoot and removeRoot methods respectively.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
+    }
+    for (const auto& child : p->e) {
+      if (inc) {
+        cn.incRef(child.w);
+      } else {
+        cn.decRef(child.w);
+      }
+      markComplexNumbersRec(child.p, inc, visited);
+    }
+  }
```
burgholzer: You did not yet touch/refactor the reference counting for the `RealNumber` class used in the edge weights as brought up in the previous review.

That essentially boils down to removing 

https://github.com/munich-quantum-toolkit/core/blob/main/include/mqt-core/dd/RealNumber.hpp#L227-L233

And dealing with the fallout from that. For example, removing the respective incRef and decRef functions.

The marking of complex numbers for garbage collection in the respective table can probably be handled similarly to https://github.com/munich-quantum-toolkit/core/blob/main/src/dd/RealNumber.cpp#L32-L35, but with the second-least-significant bit.

Then these helper functions here can be removed.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,60 +206,147 @@ class Package {
...
+  template <class Node> void markNodes(Node* p) noexcept {
+    if (Node::isTerminal(p) || p->marked()) {
+      return;
+    }
+    p->mark();
+    for (const auto& child : p->e) {
+      cn.incRef(child.w);
+      markNodes(child.p);
     }
   }
```
burgholzer: I believe this function should be marking Edges now. Because otherwise, you will never be marking the top edge weight.

src/dd/Package.cpp
```diff
@@ -101,15 +104,53 @@ bool Package::garbageCollect(bool force) {
...
+  }
+  for (const auto& [edge, _] : densityRoots) {
+    cn.incRef(edge.w);
+    markNodes(edge.p);
+  }
+
+  // first sweep all node tables
+  const auto vCollect = vUniqueTable.garbageCollect(true);
+  const auto mCollect = mUniqueTable.garbageCollect(true);
+  const auto dCollect = dUniqueTable.garbageCollect(true);
```
burgholzer: As pointed out in the previous review, this is not yet the same logic as before this PR with regard to the `force` flag.

src/dd/Package.cpp
```diff
@@ -101,15 +104,53 @@ bool Package::garbageCollect(bool force) {
...
+      cn.decRef(edge.w);
+      markComplexNumbersRec(edge.p, false, &visited);
+    }
+  }
+  {
+    std::unordered_set<dNode*> visited{};
+    for (const auto& [edge, _] : densityRoots) {
+      cn.decRef(edge.w);
+      markComplexNumbersRec(edge.p, false, &visited);
+    }
```
burgholzer: I would assume this is no longer necessary once the reference counting for the numbers is adjusted.
Maybe the top level edge-weights need to be unmarked, but no recursive traversal should be necessary.

src/dd/Package.cpp
```diff
@@ -163,7 +204,6 @@ dEdge Package::makeZeroDensityOperator(const std::size_t n) {
     f = makeDDNode(static_cast<Qubit>(p),
                    std::array{f, dEdge::zero(), dEdge::zero(), dEdge::zero()});
   }
-  incRef(f);
```
burgholzer: Based on where this PR is heading, this `incRef` needs to stay.

src/dd/Package.cpp
```diff
@@ -179,7 +219,6 @@ vEdge Package::makeZeroState(const std::size_t n, const std::size_t start) {
   for (std::size_t p = start; p < n + start; p++) {
     f = makeDDNode(static_cast<Qubit>(p), std::array{f, vEdge::zero()});
   }
-  incRef(f);
```
burgholzer: Based on where this PR is heading, this `incRef` needs to stay.

q-inho:
Thanks a lot for detailed comments. I am really learning a lot from this PR about cpp projects. I will reach out for further questions if I am not clear.

include/mqt-core/dd/Edge.hpp
q-inho: You are right. `Edge` already provides them.

burgholzer:
@q-inho may I ask for an update on this? Is there anything blocking your progress?

burgholzer:
> @burgholzer Thank you for asking. I am currently not passing the test for Grover, so I am reviewing the changes I made, where it is going wrong. I am reviewing the codes myself. 😅

Have you already implemented the changes to the handling of the complex number ref counts? If so, it might be worth pushing the changes even if tests fail. Maybe I can quickly point you towards the underlying error.

q-inho:
> > @burgholzer Thank you for asking. I am currently not passing the test for Grover, so I am reviewing the changes I made, where it is going wrong. I am reviewing the codes myself. 😅
> 
> Have you already implemented the changes to the handling of the complex number ref counts? If so, it might be worth pushing the changes even if tests fail. Maybe I can quickly point you towards the underlying error.

I believed I made correct implementation on complex number ref counts. I will check little bit more and will push soon. Thank you for your help

q-inho:
@burgholzer Apologize for delay due to some works. I just updated current progress. I think this now implement mark & sweep and replaces node reference counts with a mark flag and manages roots through `Package::incRef/decRef`. Complex numbers are kept in RealNumberUniqueTable and marked via pointer bits.

q-inho:
The Windows build failed because it requires `clearMark` to be declared as `constexpr`, while passing on Mac and Linux. However, defining `clearMark` as `constexpr` require the use of `reinterpret_cast`, which is not permitted in constant expressions. As a temporary workaround, I have replaced `constexpr` with `inline`.

q-inho:
@burgholzer Thanks for your help! I've already enabled `Allow edits by maintainers` for this PR. Did you mean I should grant you direct access by adding you as a collaborator on my repository branch instead?

burgholzer:
Yeah, thanks!

I just pushed a work-in-progress commit with some fixups and corrections.
This is not yet functional unfortunately, but I need to get some sleep now.
I'll investigate more tomorrow. If you want to look at in the meantime, feel free to go ahead.
Something about the complex number tracking is still off. The mark and unmark does not seem to work fully.

burgholzer:
The respective feature has been implemented in #1020. Closing here.

PR #982, opened 2025-06-02
Draft: Streamline Python bindings

denialhaag:
## Description

Please include a summary of the change and, if applicable, which issue is fixed.
Please also include relevant motivation and context.
List any dependencies that are required for this change.

## Checklist:


ystade:
@denialhaag I just pushed a commit that may solve the linter issue, let's see what the CI says.

.readthedocs.yaml
```diff
@@ -17,7 +17,7 @@ build:
       - (git --no-pager log --pretty="tformat:%s -- %b" -1 | grep -viq "skip ci") || exit 183
       # Skip docs build if there are no changes related to docs
       - |
-        if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/main -- docs/ .readthedocs.yaml src/mqt/ src/python include/*/python .github/contributing* .github/support*;
+        if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/main -- docs/ .readthedocs.yaml bindings python include/*/python .github/contributing* .github/support*;
```
ystade: The change you made is correct in this context. Apart from this specific PR, I am wondering, whether this was/is actually correct. Since we also generate Docs via doxygen from C++ files, all C++ files or at least the header files in `include/` should also be included here, right, @burgholzer ?

burgholzer:
> Besides the comments that will be included in the review, I have one comment that I could not put in the review.
> 
> 
> 
> @burgholzer added the following file https://github.com/munich-quantum-toolkit/core/blob/643a18f8ee033b85b6388070e7990cb938e24368/include/mqt-core/python/pybind11.hpp#L1-L22
> 
> 
> 
> This did not exist in MQT QMAP when I restructured the bindings. The state right now feels not quite satisfactory as the source files are now clearly separated into `src/` and `bindings/` but the header files are still mixed in the `include/` directory. I would argue that the only Python header is superfluous and the few lines can be added to every C++ source file defining a Python module. However, @burgholzer had surely some good motivation to create this header, so maybe he can give his thoughts on that issue.

The main motivation was that these includes always need to come first in order for the STL bindings to work as expected. The common header makes that fairly easy to enforce.
Given that there is a limited amount of bindings files in this project, I would be fine with eliminating the header and adding the appropriate includes (including a comment stating that these need to come first) to every file that needs them.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Initial implementation of the mqtdyn Dialect ([#900]) ([**@DRovara**], [**@ystade**])
 - ✨ Add function to approximate decision diagrams ([#908]) ([**@MatthiasReumann**])
 - ✨ Add functions to generate random vector DDs ([#975]) ([**@MatthiasReumann**])
+- ✨ Add a new CMake function `add_mqt_core_binding` to add a Python binding library ([#982]) ([**@denialhaag**])
```
ystade: Rather a nitpick and I am not certain whether the other PRs that added entries here did that correctly but the entry at the top should correspond to the PR that was merged last. So this bullet point should go to the top of the list.

CHANGELOG.md
```diff
@@ -26,8 +27,10 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 - ♻️ Outsource definition of standard operations from MLIR dialects to reduce redundancy ([#933]) ([**@ystade**])
 - ♻️ Unify operands and results in MLIR dialects ([#931]) ([**@ystade**])
-- ⏪️ Restore support for (MLIR and) LLVM v19 ([#934]) ([**@flowerthrower**]), [**@ystade**])
+- ⏪️ Restore support for (MLIR and) LLVM v19 ([#934]) ([**@flowerthrower**], [**@ystade**])
 - ⬆️ Update nlohmann_json to `v3.12.0` ([#921]) ([**@burgholzer**])
+- ♻️ Move the C++ code for the Python bindings to the top-level `bindings` directory ([#982]) ([**@denialhaag**])
+- ♻️ Move all Python code (no tests) to the top-level `python` directory ([#982]) ([**@denialhaag**])
```
ystade: Same here, those should be added to the top of the list. To prevent confusion in the future, perhaps it makes sense to add a comment somewhere at the top of the file along those lines:
```html
<!-- The entries in each category are sorted by the time they are merged, i.e., the PR merged last should be at the top of the list -->
```
Maybe needs some rephrasing (by GPT) or so.

CMakeLists.txt
```diff
@@ -72,14 +72,34 @@ project(
...
   DESCRIPTION "MQT Core - The Backbone of the Munich Quantum Toolkit")
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(ExternalDependencies)
+include(AddMQTCoreBinding)
 
 # set the include directory for the build tree
 set(MQT_CORE_INCLUDE_BUILD_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include/mqt-core")
 
+set(MQT_CORE_TARGET_NAME "mqt-core")
```
ystade, line 81: I'll just add that here as a comment on what we discussed offline so that we do not forget about it.

We were not sure, whether this setup of the `RPATH` is only needed for the bindings when they are built as shared libs or whether it is needed for all libs when they are built as shared libs. We moved it here to have this setup available globally. Perhaps, @burgholzer can comment whether it should better be moved to `bindings/CMakeLists.txt`.

ystade (commented):
Thanks @denialhaag for your work on this. This looks already very good and clean. I added very few comments and one reminder within the code. After those comments are resolved, I think this PR is ready to be reviewed again by @burgholzer .

.readthedocs.yaml
burgholzer: Indeed. C++ headers should be included here as well so that RtD builds are not skipped if only C++ headers change.
Source files do not need to be included here in general as, I believe, we do not document them via doxygen, and it only really makes sense to document public interfaces. So headers only it should be 🙂

CHANGELOG.md
burgholzer: Yeah. The last couple of PRs definitively introduced some inconsistencies here.
Since no other PR is about to go in just now, we should fix that here and add a note to the top of the Changelog, as Yannick proposed.

CMakeLists.txt
burgholzer: The RPATH setup is needed for every binding library so that it can find the main shared libraries, which get installed to different directories within the wheel.
The other/main libraries do not need such handling as they all get placed in the same directory when the project is being installed.

This kind of handling is really finicky and fragile. It took quite a while to get the settings to the working state in which they currently are. Whether that is the optimal configuration is very much up for debate.

.github/workflows/ci.yml
```diff
@@ -134,9 +134,11 @@ jobs:
     if: fromJSON(needs.change-detection.outputs.run-cpp-linter)
     uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.9
```
ystade:
```suggestion
    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.10
```
@denialhaag Just saw in the logs that the CI did not run because the workflow could not be parsed. This makes the dummy change unnecessary that I demnaded in another comment. The version update here should fix the issue.

denialhaag: Ahh, I see. Done. 👍

CMakeLists.txt
denialhaag: I moved it to `bindings/CMakeLists.txt`. 👍

CHANGELOG.md
denialhaag: Done, it became very concise. 🤖

CHANGELOG.md
denialhaag: Done.

.readthedocs.yaml
denialhaag: Done.

denialhaag:
> > Besides the comments that will be included in the review, I have one comment that I could not put in the review.
> > @burgholzer added the following file https://github.com/munich-quantum-toolkit/core/blob/643a18f8ee033b85b6388070e7990cb938e24368/include/mqt-core/python/pybind11.hpp#L1-L22
> > 
> > This did not exist in MQT QMAP when I restructured the bindings. The state right now feels not quite satisfactory as the source files are now clearly separated into `src/` and `bindings/` but the header files are still mixed in the `include/` directory. I would argue that the only Python header is superfluous and the few lines can be added to every C++ source file defining a Python module. However, @burgholzer had surely some good motivation to create this header, so maybe he can give his thoughts on that issue.
> 
> The main motivation was that these includes always need to come first in order for the STL bindings to work as expected. The common header makes that fairly easy to enforce. Given that there is a limited amount of bindings files in this project, I would be fine with eliminating the header and adding the appropriate includes (including a comment stating that these need to come first) to every file that needs them.

I removed `pybind11.hpp`. Let me know if this is what you had in mind.

.github/workflows/ci.yml
```diff
@@ -132,11 +132,14 @@ jobs:
     name: 🇨‌ Lint
     needs: change-detection
     if: fromJSON(needs.change-detection.outputs.run-cpp-linter)
-    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.9
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@fix-linter
```
denialhaag:
```suggestion
    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.10
```

.github/workflows/ci.yml
```diff
@@ -132,11 +132,14 @@ jobs:
...
     with:
-      cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON
+      cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON -DBUILD_MQT_QMAP_BINDINGS=ON
       clang-version: 20
       build-project: true
+      files-changed-only: false
```
denialhaag:
```suggestion
```

denialhaag (commented):
Undo these changes before merging.

.github/workflows/ci.yml
```diff
@@ -132,11 +132,14 @@ jobs:
...
     if: fromJSON(needs.change-detection.outputs.run-cpp-linter)
-    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.9
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.10
     with:
-      cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON
-      clang-version: 20
       build-project: true
+      clang-version: 20
+      cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON -DBUILD_MQT_CORE_BINDINGS=ON
+      files-changed-only: false
```
denialhaag:
```suggestion
      files-changed-only: true
```

.github/workflows/ci.yml
```diff
@@ -134,12 +134,12 @@ jobs:
...
     uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.10
     with:
-      cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON -DBUILD_MQT_CORE_BINDINGS=ON
-      clang-version: 20
       build-project: true
-      files-changed-only: true
-      setup-python: true
+      clang-version: 20
+      cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON -DBUILD_MQT_CORE_BINDINGS=ON
+      files-changed-only: false
```
denialhaag:
```suggestion
      files-changed-only: true
```

.github/workflows/ci.yml
burgholzer, line 1: The changes here are totally fine, but eventually, they are just reordering the existing entries. We might as well leave this as it is.

CHANGELOG.md
```diff
@@ -9,11 +11,12 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add a new CMake function `add_mqt_core_binding` to add a Python binding library ([#982]) ([**@denialhaag**])
```
```suggestion
```
I'd personally not add this to the changelog here as it is not really something that users of mqt-core need to care about and it is more of an internal change to the library's structure.

CHANGELOG.md
```diff
@@ -24,12 +27,14 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Changed
 
+- ♻️ Move the C++ code for the Python bindings to the top-level `bindings` directory ([#982]) ([**@denialhaag**])
+- ♻️ Move all Python code (no tests) to the top-level `python` directory ([#982]) ([**@denialhaag**])
 - ♻️ Improve performance of getNqubits for StandardOperations ([#959]) ([**@ystade**])
```
```suggestion
- ⚡ Improve performance of getNqubits for StandardOperations ([#959]) ([**@ystade**])
```

CMakeLists.txt
```diff
@@ -72,14 +72,23 @@ project(
   VERSION ${MQT_CORE_VERSION}
   DESCRIPTION "MQT Core - The Backbone of the Munich Quantum Toolkit")
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
burgholzer: This is redundant and actually already included at the top of this file.
```suggestion
```

bindings/dd/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+  add_mqt_core_binding(dd ${DD_SOURCES} LINK_LIBS MQT::CoreDD)
+
+  # install the Python stub file in editable mode for better IDE support
+  if(SKBUILD_STATE STREQUAL "editable")
+    install(
+      FILES ${PROJECT_SOURCE_DIR}/python/mqt/core/dd.pyi
+      DESTINATION .
+      COMPONENT ${MQT_CORE_TARGET_NAME}_Python)
+  endif()
+endif()
```
burgholzer: @ystade should we do the nice trick you started in QMAP with "properly" naming the targets for the bindings so that the target name includes "mqt-core"? Same holds for the IR module.

bindings/ir/register_symbolic.cpp
```diff
@@ -30,4 +30,5 @@ void registerSymbolic(pybind11::module& m) {
   registerTerm(m);
   registerExpression(m);
 }
+
```
burgholzer: Seems like a stray change here.
```suggestion
```

src/CMakeLists.txt
```diff
@@ -19,15 +18,6 @@ if(MQT_CORE_INSTALL)
...
   set(MQT_CORE_CMAKE_PROJECT_TARGETS_FILE "${MQT_CORE_CMAKE_CONFIG_DIR}/mqt-core-targets.cmake")
-
-  if(APPLE)
-    set(BASEPOINT @loader_path)
-  else()
-    set(BASEPOINT $ORIGIN)
-  endif()
-  set(CMAKE_INSTALL_RPATH ${BASEPOINT} ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
-  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
```
burgholzer, line 30: I am not 100% sure if moving this code away from here actually works. This file contains a large part of the installation instructions. If the RPATH handling is only defined in a subdirectory, the respective values will not be available in this directory because they are not exported to the parent scope.
This is a bit hard to test, but presumably the best way would be to build the Python wheel from this PR (`uv build --wheel`) and installing that version for a no-build-isolation build of QMAP and then running the tests. If that works properly, I am convinced this works properly.

burgholzer (changes requested):
Hey @denialhaag 👋🏼 
Many thanks for your work on this! I just did a fairly quick review of the whole PR to keep moving this forward. Nothing really critical, so this should hopefully be easy to address. Otherwise, this is looking really good already and should be done "soon".

.github/workflows/ci.yml
denialhaag: The linter passed. 👍

.github/workflows/ci.yml
denialhaag: I undid the changes. 👍

bindings/dd/CMakeLists.txt
denialhaag: Done. We can revert the commit if the change is not desired in the end.

src/CMakeLists.txt
denialhaag: I ran `uv build --wheel` and then created a new Python environment in which I installed the generated wheel and the dependencies for running the tests (`pandas`, `pytest`, `pytest-console-scripts`, and `qiskit`). The tests passed. Is that what you meant? 🤔

burgholzer: I assume you just ran the _mqt-core_ tests here, right?
I was actually talking about running the tests of one of the libraries that depends on mqt-core (like QMAP) because this is where error's typically manifest.

For that you would need to run something along the lines of what the nox session for testing is running in QMAP with the only difference that you install mqt-core from the wheel that has been created.

denialhaag: That is indeed not working. However, I also cannot get this to work when I check out `main` in all three repositories (`qcec` is a dependency of the `qmap` tests). I always get the following error code originating from `qcec`: 

```
Symbol not found: [...]makeZeroState[...]
```

denialhaag: I'm assuming this is related to the fact that `qcec` has not been updated after #984. 🤔

burgholzer: That is definitely related. That PR introduced a breaking change and moved some methods around that will require changes in QCEC and DDSIM at least.

I would expect QMAP to actually be fine. But the problem likely comes up when building the QCEC dependency there.

However, I you could try to not build QCEC from source but use the PyPI version instead and see if that works.

QCEC and DDSIM will definitely need updates to work with mqt-core 3.1.

denialhaag: Still the same problem. :/

I can only get the `qmap` tests to work if I build the `core` wheel after checking out `v3.0.0`.

I think the problem is that `main` of `core` is incompatible with the `qcec` version distributed via PyPi.

denialhaag: But you are right; the problem is not `qmap` itself. The problem comes up when`mqt.qcec` is imported in the `qmap` tests.

denialhaag: Maybe related, in the `qcec` repository, for example, the RPATH is also not set in the root directory, see [here](https://github.com/munich-quantum-toolkit/qcec/blob/main/src/python/CMakeLists.txt#L9). On that topic, is there a reason so many paths are being appended? Should this be the same for all repositories? 🤔

denialhaag: That said, we could move it to the root directory in all repositories. It would be consistent, and we can be sure.

burgholzer: I just created https://github.com/munich-quantum-toolkit/qcec/pull/622, which adapts QCEC to be compatible with the latest mqt-core version. That should go through smoothly (I hope).

Once that is merged, testing this PR here should be as easy as adjusting the mqt-core ref in QCEC to the reference here and doing a similar thing in QMAP. That should ensure that this is working as expected.

> Maybe related, in the qcec repository, for example, the RPATH is also not set in the root directory, see [here](https://github.com/munich-quantum-toolkit/qcec/blob/main/src/python/CMakeLists.txt#L9). On that topic, is there a reason so many paths are being appended? Should this be the same for all repositories? 🤔

This should be the same for all repositories depending on mqt-core on the C++ and Python side. And I believe it actually is the same for all. At least for the main ones (ddsim, qmap, qcec).
The main reason for so many paths being added is to ensure that the mqt-core libraries can be found under most/all circumstances. There are subtle differences between the different platforms and where they are placing libraries. Essentially, this is a best-effort to make sure everything works. Not all of that might be needed all the time, but it certainly does not hurt.

Side comment (maybe safe that somewhere): At the moment, the wheels we are producing for the top-level repositories depending on mqt-core do not actually vendor the mqt-core libraries. Instead, they depend on the mqt-core package being installed and the RPATH to properly resolve and allow finding the shared libraries. While this leads to very lean wheels for these top-level libraries, it is not particularly considered a best practice and wheels should generally be self contained. The way to do this is via wheel repair steps applied as part of running cibuildwheel (auditwheel for linux, delocate for macOS, delvewheel for Windows). These would copy ideally vendor the shared libraries of mqt-core as part of the top-level wheels. For that to work, these tools need to be able to find the shared libraries to be vendored in. And that's actually what was fairly painful to set up when I last tried. The current RPATH management is a consequence of exactly that. Furthermore, all the exclusions [here](https://github.com/munich-quantum-toolkit/qcec/blob/0f62bcee3e150332364b1e6159b2a6034d0ba23b/pyproject.toml#L304-L325) are necessary for the wheel repair steps to pass in the top-level repositories.
I believe it would be nice to investigate whether this setup could be improved. I just remember that I was fairly happy once I eventually got this setup working and needed to move on to different problems at the time.

burgholzer: The QCEC PR has been merged. So you should be able to test this properly now.
It should be sufficient to simply test this PR with QCEC instead of QMAP. That should also simplify things.

denialhaag: It worked now; the `qcec` tests passed with a `core` wheel built from this branch! 🥳 

Thanks for updating `qcec`!

denialhaag: I'll have a better look at this RPATH stuff tomorrow! Thanks for the explanation!

burgholzer: > It worked now; the `qcec` tests passed with a `core` wheel built from this branch! 🥳 
> 
> 
> 
> Thanks for updating `qcec`! 

That is great to hear 😎
But I guess this also means that we can actually move forward with this for now, right?

We can tackle the RPATH stuff in a follow-up. The current setup is at least working, so there is no immediate need to change this here and now.

denialhaag: Yes, I agree. From my side, it should be good to go! 👍

burgholzer (approved):
LGTM 👍🏼 Let's get this in now, to not create too many conflicts with other ongoing PRs.
Thanks for your work on this!

PR #983, opened 2025-06-04
🚨 Resolve Qiskit 2.1 Deprecation Warnings

### LINKED ISSUE #974: ❌ Qiskit Upstream Tests Failure
## Qiskit Upstream Tests Failed on 2025-05-28

The weekly Qiskit upstream test has failed.

### Workflow Details

- **Workflow Run**: [View Logs and Details](https://github.com/munich-quantum-toolkit/core/actions/runs/15303151793)
- **Result**: `failure`
- **Triggered by**: pull_request
- **Qiskit Commit Tested**: https://github.com/Qiskit/qiskit/commit/278b46e2da95210e8184ba1e33b4e1612a7bd542 (2025-05-28T13:19:30Z)

Please investigate and fix this issue to ensure compatibility with the latest version of Qiskit.

> This issue was automatically generated by a GitHub Action.

ystade:
## Description

Fixes #974 

## Checklist:


burgholzer:
Are we sure that these are Qiskit internal and not from our use of Qiskit?
If they are from our usage, then I'd much rather fix the reason behind the deprecation instead of ignoring it until it hits us.

ystade:
> Are we sure that these are Qiskit internal and not from our use of Qiskit? If they are from our usage, then I'd much rather fix the reason behind the deprecation instead of ignoring it until it hits us.

Is it now in the way you imagined it?

src/mqt/core/plugins/qiskit/qiskit_to_mqt.py
```diff
@@ -207,7 +204,7 @@ def _emplace_operation(
     if name in {"i", "id", "iden"}:
         return _add_operation(qc, OpType.i, qargs, params, qubit_map)
 
-    if name in {"x", "cx", "ccx", "mcx", "mcx_gray"}:
+    if name in {"x", "cx", "ccx", "mcx"}:
```
ystade:
```suggestion
    # we keep "mcx_gray" here for compatibility reasons with older Qiskit versions, e.g., 1.0.0 where "noancilla" 
    # is the default for the argument "mode" which leads to the gate name "gray_code" 
    if name in {"x", "cx", "ccx", "mcx", "mcx_gray"}:
```

src/mqt/core/plugins/qiskit/qiskit_to_mqt.py
```diff
@@ -150,9 +150,6 @@ def qiskit_to_mqt(circ: QuantumCircuit) -> QuantumComputation:
     "sxdg",
     "csx",
     "mcx",
-    "mcx_gray",
-    "mcx_recursive",
-    "mcx_vchain",
     "mcphase",
```
ystade:
```suggestion
    "mcx_gray",  # we keep "mcx_gray" here for compatibility reasons with older Qiskit versions, e.g., 1.0.0 
                 # where "noancilla" is the default for the argument "mode" which leads to the gate name "gray_code"
    "mcphase",
```

ystade (commented):
@burgholzer After removing the code handling the deprecated decompositions of the mcx gates, the tests with the minimal supported Qiskit version throws an error. In this version, "noancilla" is default for the mode parameter (which we omit now in the tests) leading to a gate name "mcx_gray" which is not correctly translated anymore. What do you think about the following fix?

burgholzer:
> @burgholzer After removing the code handling the deprecated decompositions of the mcx gates, the tests with the minimal supported Qiskit version throws an error. In this version, "noancilla" is default for the mode parameter (which we omit now in the tests) leading to a gate name "mcx_gray" which is not correctly translated anymore. What do you think about the following fix?

I would have actually kept all of that logic until Qiskit removes the logic and our minimum version is higher than that.
Otherwise we are giving up compatibility. Given the deprecation is being raised in 2.1, that's 2.2 at the earliest, which is another 3 months away.

My above comment was mainly about checking whether there is a Qiskit 2.1 way (without deprecations) to create the special versions of the MCX gate so that we can still test the functionality.
And I am pretty sure that's possible.
Sorry for the poor communication here. I could have articulated that better..

burgholzer:
> > And I am pretty sure that's possible.
> 
> 
> 
> Honestly, I do not see how this is possible without raising deprecation warnings. 
> 
> 
> 
> E.g., these synthesis functions here return the full decomposed circuit
> 
> https://docs.quantum.ibm.com/api/qiskit/synthesis#synth_mcx_n_dirty_i15

Hm.. damn. Thought that would work.

I'd favor longer compatibility here. So how about we leave the code in, explicitly silence the deprecation warning for the test module (pythons warning filters allow you to specify this. I believe it's the last argument).
And place a dedicated commented above it stating that this is kept for compatibility and might be removed once Qiskit removes the respective functionality.

burgholzer (approved):
Alright. I am back at my laptop again.
And took the liberty to push this across the finish line.
In the meantime, one more deprecation warning came up (about Python 3.9) and something about the order of terms in expressions changed. I also addressed these.
After this, the Qiskit upstream workflow should run smoothly again and the v1.10 workflow update PR should be ready as well 😌

PR #984, opened 2025-06-05
♻️ Move Make-State Functionality To StateGeneration.cpp

MatthiasReumann:
## Description

This pull request moves the make-state functionality to the `StateGeneration.[hpp, cpp]` files and adds unit tests for the respective functions. 

## Checklist:


src/dd/StateGeneration.cpp
```diff
@@ -62,8 +64,194 @@ vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
...
+
+    curr = std::move(next);
+  }
+
+  const vCachedEdge state = curr.at(0);
+  const vEdge e{state.p, dd.cn.lookup(state.w)};
+  dd.incRef(e);
+
+  return e;
+}
```
burgholzer: Just a quick side note from browsing through the PR:
I believe this solution is inferior to the recursive solution as it will always create exponentially-sized vectors of nodes, which the recursive version does not. The runtime is exponential in any case because each amplitude needs to be looked at once.

src/dd/StateGeneration.cpp
```diff
@@ -62,8 +64,194 @@ vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
...
+  auto leftSubtree = vEdge::zero();
+  if ((1. / sqrt(static_cast<double>(n))) < RealNumber::eps) {
+    throw std::runtime_error(
+        "Requested qubit size for generating W-state would lead to an "
+        "underflow due to 1 / sqrt(n) being smaller than the currently set "
+        "tolerance " +
+        std::to_string(RealNumber::eps) +
+        ". If you still wanna run the computation, please lower "
+        "the tolerance accordingly.");
+  }
```
MatthiasReumann: @burgholzer How would you force this to throw in a unit test? 🤔 Many thanks!

burgholzer: That's not too easy.
Although it should be doable. And I think we actually force that in an existing unit test.
You simply have to raise the default tolerance to a rather high level so that `1/sqrt(n)` becomes smaller than that tolerance.

I'd anyway recommend to simply remove the existing functions to see where the functionality breaks. That should also point you towards existing tests (which can be removed/refactored/freshly designed as you see fit).

src/dd/StateGeneration.cpp
MatthiasReumann: You are absolutely right. Sorry for that. I reverted back to the recursive algorithm!

include/mqt-core/dd/StateGeneration.hpp
burgholzer, line 1: Is there any particular reason for the `noexcept(false)` throughout the file?
That's the default for any function not explicitly marked noexcept.

src/dd/StateGeneration.cpp
```diff
@@ -62,8 +65,208 @@ vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
 
   return dd.makeDDNode(v, edges);
 }
+
+/**
+ * @brief Validate that the package is suitable for the use with up to @p n
+ * qubits.
+ * @throw `std::invalid_argument`, if not suitable.
+ */
+void suitablePackage(const std::size_t n, Package& dd) {
```
burgholzer:
```suggestion
void suitablePackage(const std::size_t n, const Package& dd) {
```
I'd suppose this can be const.

src/dd/StateGeneration.cpp
```diff
@@ -62,8 +65,208 @@ vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
...
+  const auto oneSuccessor = makeStateFromVector(pivot, end, v - 1, dd);
+  return dd.makeDDNode<vNode, CachedEdge>(v, {zeroSuccessor, oneSuccessor});
+}
+
 } // namespace
 
+VectorDD makeZeroState(const std::size_t n, Package& dd,
+                       const std::size_t start) {
+  const std::vector<BasisStates> state(n - start, BasisStates::zero);
+  return makeBasisState(n, state, dd, start);
```
burgholzer: This is not quite the same logic as the original method.
```suggestion
VectorDD makeZeroState(const std::size_t n, Package& dd,
                       const std::size_t start) {
  const std::vector<BasisStates> state(n, BasisStates::zero);
  return makeBasisState(n, state, dd, start);
```

src/dd/StateGeneration.cpp
```diff
@@ -62,8 +65,208 @@ vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
...
+
+  if (sz == 0) {
+    return vEdge::one();
+  }
+
+  if (sz == 1) {
+    return vEdge::terminal(dd.cn.lookup(vec[0]));
+  }
+
+  const auto v = static_cast<Qubit>(std::log2(sz) - 1);
```
burgholzer, line 256: This could also use a suitablePackage check.

include/mqt-core/dd/StateGeneration.hpp
MatthiasReumann: I kind of wanted to make it explicit that these functions throw. 
But now if I think about it: If the assumption is that any function not explicitly marked with `noexcept` can throw... 🤔

What's your opinion on trading these `noexcept(false)` statements against a sentence (or so) in the function descriptions?

src/dd/StateGeneration.cpp
MatthiasReumann: Many thanks! There should have been a unit-test noticing this error.

src/dd/StateGeneration.cpp
```diff
@@ -62,8 +65,208 @@ vCachedEdge randomNode(Qubit v, vNode* left, vNode* right, Generator& gen,
...
+
+  if (state.size() < n) {
+    throw std::invalid_argument(
+        "Insufficient qubit states provided. Requested " + std::to_string(n) +
+        ", but received " + std::to_string(state.size()));
+  }
+
+  vCachedEdge f = vCachedEdge::one();
+  for (std::size_t p = start; p < n + start; ++p) {
+    switch (state[p]) {
```
MatthiasReumann: @burgholzer Am I stupid right now - or shouldn't the loop iterate over `[0, n]` and only the qubit value for the new nodes should be `p + start`?

Given the safety check above (`size() < n`) this should be an out-of-bounds for any `start > 0`, right?

include/mqt-core/dd/StateGeneration.hpp
burgholzer: Doxygen comments have an explicit special command for that in the docs `@throws`.

That would be the appropriate way of explicitly documenting that these functions may throw, and which exceptions they throw for whatever reason.
I'd be in favor of adding such documentation over the kind of redundant noexcept statements here.

src/dd/StateGeneration.cpp
burgholzer: Yeah. That's undefined behavior 😅 (if start != 0)
Probably not noticeable in the unit tests because they all use the default value for start.

(I'd actually like to get rid of that `start` parameter at some point.. it is only used in one particular method in our simulator DDSIM and I am pretty sure there would be other ways to realize that functionality)

include/mqt-core/dd/StateGeneration.hpp
```diff
@@ -16,6 +16,72 @@
 #include <vector>
 
 namespace dd {
+/**
+ * @brief Construct the all-zero state \f$|0...0\rangle\f$
+ * @param n The number of qubits.
+ * @param dd The DD package to use for making the vector DD.
+ * @param start The starting qubit index. Default is 0.
+ * @throws std::invalid_argument
```
burgholzer: Typically these would also contain some small explanation why/when the exception might be thrown. 🙂 otherwise the information isn't worth too much in the docstring.

burgholzer (approved):
Awesome. Many thanks 🙏

PR #1000, opened 2025-06-16
🚸 Make CMake function for adding Python bindings reusable

denialhaag:
## Description

This PR makes the function for adding Python bindings reusable across MQT repositories. The function has been added in 

## Checklist:

- [x] ~~I have added entries to the changelog for any noteworthy additions, changes, fixes or removals~~.

### DIFF
--- a/cmake/AddMQTPythonBinding.cmake
+++ b/cmake/AddMQTPythonBinding.cmake
@@ -30,6 +30,12 @@ function(add_mqt_python_binding package_name target_name)
   if(ARG_MODULE_NAME)
     # the library name must be the same as the module name
     set_target_properties(${target_name} PROPERTIES OUTPUT_NAME ${ARG_MODULE_NAME})
+    target_compile_definitions(${target_name}
+                               PRIVATE MQT_${package_name}_MODULE_NAME=${ARG_MODULE_NAME})
+  else()
+    # use the target name as the module name
+    target_compile_definitions(${target_name}
+                               PRIVATE MQT_${package_name}_MODULE_NAME=${target_name})
   endif()
 
   # add project libraries to the link libraries
@@ -42,5 +48,5 @@ function(add_mqt_python_binding package_name target_name)
   install(
     TARGETS ${target_name}
     DESTINATION ${ARG_INSTALL_DIR}
-    COMPONENT MQT_${package_name}_MODULE_NAME_Python)
+    COMPONENT ${MQT_${package_name}_TARGET_NAME}_Python)
 endfunction()
--- a/cmake/AddMQTPythonBinding.cmake
+++ b/cmake/AddMQTPythonBinding.cmake
@@ -6,15 +6,13 @@
 #
 # Licensed under the MIT License
 
-function(add_mqt_core_binding target_name)
+function(add_mqt_python_binding package_name target_name)
   # parse the arguments
   cmake_parse_arguments(ARG "" "MODULE_NAME;INSTALL_DIR" "LINK_LIBS" ${ARGN})
   set(SOURCES ${ARG_UNPARSED_ARGUMENTS})
 
-  # set default "." for INSTALL_DIR
-  if(NOT ARG_INSTALL_DIR)
-    set(ARG_INSTALL_DIR ".")
-  endif()
+  set(MQT_PACKAGE_MODULE_NAME MQT_${package_name}_MODULE_NAME)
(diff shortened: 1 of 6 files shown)


cmake/AddMQTPythonBinding.cmake
```diff
@@ -6,15 +6,13 @@
...
   # parse the arguments
   cmake_parse_arguments(ARG "" "MODULE_NAME;INSTALL_DIR" "LINK_LIBS" ${ARGN})
   set(SOURCES ${ARG_UNPARSED_ARGUMENTS})
 
-  # set default "." for INSTALL_DIR
-  if(NOT ARG_INSTALL_DIR)
-    set(ARG_INSTALL_DIR ".")
-  endif()
+  set(MQT_PACKAGE_MODULE_NAME MQT_${package_name}_MODULE_NAME)
+  set(MQT_PACKAGE_TARGET_NAME MQT_${package_name}_TARGET_NAME)
```
burgholzer: Are these variables truly helpful?
I don't see `MQT_PACKAGE_MODULE_NAME` being used somewhere. It is just set as a compiler definition down below.

The `MQT_PACKAGE_MODULE_NAME` could be inlined as well I suppose at it is only used in a single place.

denialhaag: Good point, I removed them. 👍

cmake/AddMQTPythonBinding.cmake
```diff
@@ -27,13 +22,14 @@ function(add_mqt_core_binding target_name)
...
 
+  # set default "." for INSTALL_DIR
+  if(NOT ARG_INSTALL_DIR)
+    set(ARG_INSTALL_DIR ".")
+  endif()
+
   if(ARG_MODULE_NAME)
     # the library name must be the same as the module name
     set_target_properties(${target_name} PROPERTIES OUTPUT_NAME ${ARG_MODULE_NAME})
-    target_compile_definitions(${target_name} PRIVATE MQT_CORE_MODULE_NAME=${ARG_MODULE_NAME})
```
burgholzer, line 33: This removal here now made me wonder, why this was here in the first place. Which led me to https://github.com/munich-quantum-toolkit/qmap/blob/355c8459c4f321b1d0c8d8cdb3316c55ba59b2ef/bindings/na/nasp/nasp.cpp#L31
That is something we have not fully implemented in MQT Core yet.

I am ok with leaving this here as it is right now. I just wanted to bring it up. Maybe @ystade also has some quick comments here as he was the one implementing this in QMAP.

cmake/AddMQTPythonBinding.cmake
```diff
@@ -46,5 +42,5 @@ function(add_mqt_core_binding target_name)
   install(
     TARGETS ${target_name}
     DESTINATION ${ARG_INSTALL_DIR}
-    COMPONENT ${MQT_CORE_TARGET_NAME}_Python)
+    COMPONENT MQT_${package_name}_MODULE_NAME_Python)
```
burgholzer: This doesn't really work, does it?
Previously (before this PR), this expanded to `mqt-core_Python`.
which, in QMAP, is also configured in the `pyproject.toml` file:
https://github.com/munich-quantum-toolkit/qmap/blob/355c8459c4f321b1d0c8d8cdb3316c55ba59b2ef/pyproject.toml#L96-L97

@ystade maybe also here some quick input from your side could help to truly generalise this.

denialhaag: You have a point; the current implementation doesn't make a whole lot of sense. I guess we could make `mqt-core` (or `mqt-qmap`, and so on) an input of the function, especially if we don't need the "core" for the compile definitions. 🤔

cmake/AddMQTPythonBinding.cmake
denialhaag: I changed it so that we are also using `MQT_CORE_MODULE_NAME` now, and it seems to be working with the proposed solution. 👍

cmake/AddMQTPythonBinding.cmake
denialhaag: Ignore my previous comment.

I double checked; `${MQT_${package_name}_TARGET_NAME}_Python}` resolves to what we want. 👍

burgholzer: Yeah. With the extra `${...}` evaluation now, this makes sense and seems to be working fine 🙌🏼

burgholzer:
Just noticed that this was PR #1000 to this repository 🎉

Issue #960, opened 2025-05-21
🧪 MLIR - Explicitly add Tests for MQTDyn Dialect Language Features


DRovara:
### Problem Statement

Since #948, MQT Core now explicitly tests all main features of the MQTOpt dialect.
For MQTDyn, on the other hand, we only have a single big test file at the moment.

To better test MQTDyn, we should improve its test suite in a similar fashion.

### Proposed Solution

A single test file `dialect_features.mlir` for the MQTDyn dialect. It should contain (more or less) the same test cases as the MQTOpt `dialect_features.mlir` introduced in #948, just reformulated for MQTDyn.

MatthiasReumann:
@DRovara (cc: @burgholzer) Do you mind if I work on this? I've just started to learn about MLIR and I think this would be a great starting point for some real world application.

DRovara:
We don't mind at all! I think this is an excellent issue to get started with the topic, and #948 provides a really good template for it. Thanks a lot!

PR #1004, opened 2025-06-18
⚗️ Test MQTDyn Dialect

MatthiasReumann:
## Description

Resolves Issue #960. Follows in the footsteps of #948 and adds tests for the MQTDyn MLIR dialect. 

## Checklist:


mlir/test/Dialect/MQTDyn/IR/dialect_features.mlir
```diff
@@ -0,0 +1,633 @@
...
+        //===------------------------------------------------------------------===//
+        //       ┌──────┐
+        // q0:  ┤      ├ q0
+        //       │ SWAP │
+        // q1:  ┤      ├ q1
+        //       └───┬──┘
+        // q2:  ────■─── q2
+        //           │
+        // q3:  ────■─── q3
+        //===----------------------------------------------------------------===//
```
DRovara: This diagram looks a bit messed up

```suggestion
        //===------------------------------------------------------------------===//
        //      ┌──────┐
        // q0:  ┤      ├ q0
        //      │ SWAP │
        // q1:  ┤      ├ q1
        //      └───┬──┘
        // q2:  ────■─── q2
        //          │
        // q3:  ────■─── q3
        //===----------------------------------------------------------------===//
```

DRovara (changes requested):
Thanks a lot for the effort! Looks pretty good to me. I didn't really find any problems other than what I mentioned below.

Although, honestly, I have to say, the more I look at those `CHECK-NOT` lines in all the test cases, the more I wonder if they really are necessary or just unneeded boilerplate. Why would we ever expect e.g. two `x` Operations to suddenly appear in the output when originally there was only one? I know this is not your fault here, as we already had them in the last tests as well - just wanted to mention the topic, since we are already at the tests. @burgholzer, do you have an opinion on that?

MatthiasReumann:
> Just a fairly superficial opinion: seems like boilerplate to me. The more concise we can get the tests, the better. To be fair, the MLIR tests will probably only get interesting, when we actually transform code. 🙂

Should I remove them in both test files?

burgholzer:
> > Just a fairly superficial opinion: seems like boilerplate to me. The more concise we can get the tests, the better. To be fair, the MLIR tests will probably only get interesting, when we actually transform code. 🙂
> 
> Should I remove them in both test files?

Yeah. If you don't mind that would be appreciated. Also fine as part of this PR 😌

DRovara (approved):
I didn't go through everything line by line again, but I don't think there's too much more that even can be wrong, so I would say this looks good now.
Thanks again for the effort!

burgholzer (approved):
This is looking great. Many thanks! 🙏🏼

PR #1005, opened 2025-06-18
👷 Enable templating

denialhaag:
## Description

This PR adds a workflow that uses an action introduced in https://github.com/munich-quantum-toolkit/templates/pull/1 to synchronize files across repositories. 

## Checklist:

- [x] ~~I have added entries to the changelog for any noteworthy additions, changes, fixes or removals.~~

### DIFF
--- a/.github/workflows/templating.yml
+++ b/.github/workflows/templating.yml
@@ -1,8 +1,5 @@
 name: Templating
 on:
-  schedule:
-    # Run every Monday at 00:00 UTC
-    - cron: "0 0 * * 1"
   pull_request:
     paths:
       - ".github/workflows/templating.yml"
@@ -24,5 +21,5 @@ jobs:
       - uses: munich-quantum-toolkit/templates@main
         with:
           token: ${{ steps.create-token.outputs.token }}
-          organization: munich-quantum-toolkit
-          repository: core
+          organization: ${{ github.repository_owner }}
+          repository: ${{ github.event.repository.name }}


denialhaag:
@burgholzer, this is only a first proposal, but it's already working kind of nicely in my opinion. I would be very interested in your opinions and ideas already. 🙂 

For now, the templating is only implemented for the pull-request template and the security policy, but it is easy to extend this to other files. The templates are added in https://github.com/munich-quantum-toolkit/.github/pull/44. At this point, I believe it would make sense to create a separate repository. 🤔 

To test the templating, the job is triggered by pushing into this branch, but it should be easy to make it a scheduled job. Furthermore, the job doesn't create a new PR but pushes the changes into this branch as well, but this would also be an easy change. I'm annotating the file to leave some comments.

If you want to play around with it a bit, you can change the templates in https://github.com/munich-quantum-toolkit/.github/pull/44 and then just push into this branch.

.github/workflows/ci.yml
```diff
@@ -3,9 +3,9 @@ on:
   push:
     branches:
       - main
-  pull_request:
-  merge_group:
-  workflow_dispatch:
+  # pull_request:
+  # merge_group:
+  # workflow_dispatch:
```
denialhaag, line 8: This has to be undone. I did this because I didn't want to block the CI.

```suggestion
  pull_request:
  merge_group:
  workflow_dispatch:
```

.github/workflows/templating.yml
```diff
@@ -0,0 +1,32 @@
+name: Templating
```
denialhaag, line 1: I think this could be moved to the `workflows` repository as a reusable workflow once we like it. 🤔

.github/workflows/templating.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+  workflow_dispatch:
```
denialhaag: This has to change to make the job scheduled.

.github/workflows/templating.yml
```diff
@@ -0,0 +1,32 @@
...
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+      - name: Commit and push changes
+        run: |
+          git add .
+          if ! git diff --cached --quiet; then
+            git commit -m "Update template"
+            git push
+          else
+            echo "No changes to commit."
+          fi
```
denialhaag: This has to be changed to make the job create a PR.

.github/workflows/ci.yml
denialhaag: This has to be undone. I did this because I didn't want to block the CI.

```suggestion
  pull_request:
  merge_group:
  workflow_dispatch:
```

.github/workflows/templating.yml
burgholzer:
```suggestion
on:
  schedule:
    # Run every Monday at 00:00 UTC
    - cron: "0 0 * * 1"
  pull_request:
    paths:
      - ".github/workflows/templating.yml"
  workflow_dispatch: # Allow manual triggering
```
I suppose this can be fairly similar to how we handle upstream Qiskit testing.

.github/workflows/templating.yml
burgholzer: Yeah. That would be similar to how we are doing that for the MQT Core update workflow
In QCEC: https://github.com/munich-quantum-toolkit/qcec/blob/main/.github/workflows/update-mqt-core.yml
In the workflow repo: https://github.com/munich-quantum-toolkit/workflows/blob/main/.github/workflows/reusable-mqt-core-update.yml
Although I suppose that this will actually be an action instead of a workflow. So it might as well be that we do not add anything to the workflows repository.

.github/workflows/templating.yml
burgholzer: This can be heavily inspired by https://github.com/munich-quantum-toolkit/workflows/blob/main/.github/workflows/reusable-mqt-core-update.yml
Especially when it comes to creating the PR.

That also means that this code snippet can be entirely moved to the reusable workflow/action.

burgholzer (changes requested):
Thanks @denialhaag for churning this out so quickly. This is already a great start. I'll simultaneously write a review in the `.github` repo PR to summarize my thoughts.
Some minor inline thoughts and inspiration already here.

.github/workflows/templating.yml
denialhaag: Yes, I agree; now that the action does everything, we can just copy-paste this file to all repositories where we want to use the action. 👍

.github/workflows/templating.yml
denialhaag: Thanks! I'll commit the suggestion once I'm no longer using this PR as the trigger for testing. 😄

burgholzer:
This currently fails silently with
```
fatal: invalid object name 'main'.
fatal: invalid object name 'main'.
Warning: Failed to retrieve old template for pull_request_template.md. Error: Command '['git', 'show', 'main:.templates/pull_request_template.md']' returned non-zero exit status 128.
Warning: Failed to merge files for pull_request_template.md. Error: Command '['git', 'merge-file', '.github/pull_request_template.md', '/tmp/tmp940_eb3w/pull_request_template.md', '.templates/pull_request_template.md']' returned non-zero exit status 1.
Warning: Failed to retrieve old template for SECURITY.md. Error: Command '['git', 'show', 'main:.templates/SECURITY.md']' returned non-zero exit status 128.
```

So it seems like some more work will be necessary here.
Also check https://github.com/munich-quantum-toolkit/templates/pull/4#issuecomment-3044938428 which is related to this.

denialhaag:
> This currently fails silently with
> 
> ```
> fatal: invalid object name 'main'.
> fatal: invalid object name 'main'.
> Warning: Failed to retrieve old template for pull_request_template.md. Error: Command '['git', 'show', 'main:.templates/pull_request_template.md']' returned non-zero exit status 128.
> Warning: Failed to merge files for pull_request_template.md. Error: Command '['git', 'merge-file', '.github/pull_request_template.md', '/tmp/tmp940_eb3w/pull_request_template.md', '.templates/pull_request_template.md']' returned non-zero exit status 1.
> Warning: Failed to retrieve old template for SECURITY.md. Error: Command '['git', 'show', 'main:.templates/SECURITY.md']' returned non-zero exit status 128.
> ```
> 
> So it seems like some more work will be necessary here. Also check [munich-quantum-toolkit/templates#4 (comment)](https://github.com/munich-quantum-toolkit/templates/pull/4#issuecomment-3044938428) which is related to this.

Yeah, I think I should have been more careful with Copilot's suggestions. Those should really only be considered as warnings. The action didn't fail; it has its intended output: #1040

`git show` fails if a file doesn't exist and creates an empty file: exactly what we want to happen.
 `git merge-file` fails if there are merge conflicts and leaves merge markers in the target file: exactly what we want to happen.

burgholzer:
Yeah. We should probably tweak the action in that regard.

Looking at #1040, I noticed another thing: It is kind of inconvenient that the merge markers are part of the PR. This makes it quite cumbersome to resolve the merge conflicts. It would be way more convenient if the merge conflicts would simply show up in GitHub as proper merge conflicts to be resolved.
I am not yet 100% sure how to achieve that, but from a developer experience point of view, I would want to be able to resolve the conflicts directly on GitHub instead of having to find out how to tell my local IDE that it should support me in the conflict resolution.

Does that make sense?

.github/workflows/templating.yml
```diff
@@ -0,0 +1,28 @@
...
+      - id: create-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+      - uses: munich-quantum-toolkit/templates@main
+        with:
+          token: ${{ steps.create-token.outputs.token }}
+          organization: munich-quantum-toolkit
+          repository: core
```
burgholzer: I believe we can even make this workflow a step more generic so that it can also be more easily templated.
```suggestion
          organization: ${{ github.repository_owner }}
          repository: ${{ github.event.repository.name }}
```

.github/workflows/templating.yml
```diff
@@ -0,0 +1,28 @@
+name: Templating
+on:
+  schedule:
+    # Run every Monday at 00:00 UTC
+    - cron: "0 0 * * 1"
+  pull_request:
+    paths:
+      - ".github/workflows/templating.yml"
+  workflow_dispatch: # Allow manual triggering
```
burgholzer: Let's briefly revisit the triggering conditions here.
The underlying templates are only supposed to change when a new version of the action is released.
The respective update PRs will be generated by Renovate automatically.
So I believe the `schedule` component can be removed here.
The `workflows_dispatch` can remain here for manual triggering.
The `pull_request` one is probably the right one to use here as it will trigger then run on any updates to this file.
The only thing we should try to avoid is creating endless loops of updates, if we should ever decide to template this workflow file (which seems likely in my opinion). But maybe the below is fine already.

```suggestion
  pull_request:
    paths:
      - ".github/workflows/templating.yml"
  workflow_dispatch: # Allow manual triggering
```

burgholzer (changes requested):
Just two small comments on fine details of this workflow.

.github/workflows/templating.yml
denialhaag: Yes, sounds good!

PR #996, opened 2025-06-13
✨ Neutral Atom QDMI Device

ystade:
## Description

This PR adds a universal QDMI device implementation for neutral atom-based quantum computers. The device itself is specified in a JSON file whose structure is given by a protobuf definition. During compilation time, the JSON file is read and translated to C++ code similar to how TableGen works in the LLVM project.

Subsequent PRs will add a QDMI driver utilizing the new device and a FoMaC tailored to those neutral atom devices.

## Checklist:

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.
- [x] As mentioned in one thread, we want to open an issue for 3D support on NA devices, i.e., including the z-variable.

PR #1010, opened 2025-06-20
✨ MQT QDMI Driver

ystade:
## Description

This is the second PR in a series of PRs adding QDMI implementations to MQT. #996 has already implemented a QDMI device. Still missing are QDMI clients and FoMaC making use of the device and the driver that is contained in this PR.

Additionally, this PR fixes some issues in the setup of the QDMI Device itself, such that it is contained in the Python distribution.

Furthermore, it adds a copy of the NA Device that is always built as a Dynamic Library (only on non-Windows Systems).

## Checklist:


ystade:
@burgholzer We might want to introduce a new label `QDMI`.

burgholzer:
> @burgholzer We might want to introduce a new label `QDMI`.

Yes, definitely. Doing that real quick. That also warrants a release drafter section, I suppose. Would you mind adding that in one of the PRs?

ystade:
The current "correct" Files Changed view is the following:

https://github.com/munich-quantum-toolkit/core/pull/1010/files/c9bb86d4e5299ec230779da3882edc044ed6e82a..4dbaa0d0e9b1a094e5a7b9a770fa72f0723aa438

include/mqt-core/qdmi/Macros.hpp
```diff
@@ -0,0 +1,334 @@
...
+        memcpy(static_cast<void*>(value),                                      \
+               static_cast<const void*>((prop_values).data()),                 \
+               (prop_values).size() * sizeof(prop_type));                      \
+      }                                                                        \
+      if ((size_ret) != nullptr) {                                             \
+        *size_ret = (prop_values).size() * sizeof(prop_type);                  \
+      }                                                                        \
+      return QDMI_SUCCESS;                                                     \
+    }                                                                          \
+  }
```
burgholzer: Should we be documenting the expected inputs for these macros?
In this case, for example, that the container needs to provide a `size()` and a `data()` method?
Not strictly necessary of course, but could be helpful in the long run.

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace qc {
+struct Library {
+  std::string prefix;
+  std::string path;
+};
```
burgholzer: Long term thinking:
Should this include the authentication options? (`baseURL`, `token`, etc.) as optionals?
If, at some point, we are connecting to actual QDMI devices with a real backend behind them, I am pretty sure that this will be necessary.
Internally, the driver would just use the provided credentials for opening the device session.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ * Copyright (c) 2025 Munich Quantum Software Company GmbH
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+/** @file
+ * An example driver implementation in C++.
```
burgholzer: This is no longer just an example driver. It is the MQT QDMI driver.
Maybe this even deserves a couple more lines of documentation here.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ */
+
+// todo: Figure out how to set those from CMake.
+#define DEVICE_LIST_UPPERCASE (MQT_NA)
+#define DEVICE_LIST_LOWERCASE (mqt_na)
```
burgholzer: Pretty sure that is just a `target_compile_definition`

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+
+#include "qdmi/Driver.hpp"
+
+#include "qdmi/Macros.hpp"
+#include "qdmi/client.h"
+#include "qdmi/device.h"
```
burgholzer:
```suggestion
#include <qdmi/client.h>
#include <qdmi/device.h>
```
Technically, these are not internal includes.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+#include INCLUDE(NTH_MAX(1, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(2, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(3, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(4, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(5, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(6, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(7, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(8, DEVICE_LIST_LOWERCASE))
+#include INCLUDE(NTH_MAX(9, DEVICE_LIST_LOWERCASE))
+// NOLINTEND(readability-duplicate-include)
```
burgholzer: Hm. I see where this is coming from, but it's not really too pretty to be honest.

Given that we know how many devices we are statically linking here, we could just place that many includes here and avoid the clang-tidy warning.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+
+namespace qc {
```
burgholzer: Should this be `namespace qdmi`?

include/mqt-core/qdmi/Driver.hpp
burgholzer, line 1: In some previous efforts on devolving a QDMI driver, I took this one step further and defined the driver in an object oriented fashion as a singleton class.
At least to me, that still feels like a natural fit and makes working with the driver slightly more elegant.
I'll send you some pointers in a direct message. 🙂

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+  }
+};
+
+/**
+ * Definition of the QDMI Session.
+ */
+struct QDMI_Session_impl_d {
+  /// The status of the session.
+  qc::SessionStatus status = qc::SessionStatus::ALLOCATED;
+};
```
burgholzer: Maybe I am overlooking things. But something seems to be missing here.
A QDMI session should maintain potentially multiple Device Sessions that it has access to, shouldn't it? At least I would think that there needs to be some kind of connection.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ * Definition of the QDMI Job.
+ */
+struct QDMI_Job_impl_d {
+  QDMI_Device_Job deviceJob = nullptr; ///< The device job handle.
+  QDMI_Device device = nullptr; ///< The device handle associated with the job.
+  void* program = nullptr;      ///< Pointer to the program data
+  std::string id;               ///< Unique identifier of the job
+  size_t shotsNum = 1;          ///< The number of shots
+  /// The format of the program to be executed.
+  QDMI_Program_Format format = QDMI_PROGRAM_FORMAT_MAX;
```
burgholzer: I am not 100% sure whether these are necessary as the respective information should already be present and/or forwarded to the Device Job, shouldn't it?

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+  library.device_session_query_operation_property =                            \
+      prefix##_##device_session_query_operation_property;                      \
+  /* initialize the device */                                                  \
+  library.device_initialize();
+
+/**
+ * Returns a reference to the map of library handles returned by `dlopen`
+ * to the dynamic device libraries.
+ * @returns a reference to a map library handle to unique pointer of
+ * DeviceLibrary objects.
```
burgholzer: This docstring is kind of repeating itself. Typically, docstirmgs like these are fine as one-liners starting with "Returns"

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+                             const std::string& prefix) {
+  auto* libHandle = dlopen(libName.c_str(), RTLD_NOW | RTLD_LOCAL);
+  if (libHandle == nullptr) {
+    throw std::runtime_error("Couldn't open the device library: " + libName);
+  }
+  if (const auto it = dynamicDeviceLibraries().find(libHandle);
+      it != dynamicDeviceLibraries().end()) {
+    // dlopen uses reference counting, so we need to decrement the reference
+    // count that was increased by dlopen.
+    dlclose(libHandle);
```
burgholzer: Shouldn't this throw an error or at least a warning? Adding the same library twice can hardly ever be desirable.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ * @param library is the DeviceLibrary object that provides the device interface
+ * functions.
+ */
+auto addDevice(const DeviceLibrary& library) -> void {
+  auto& device =
+      *devices().emplace_back(std::make_unique<QDMI_Device_impl_d>());
+  device.library = &library;
+  device.library->device_session_alloc(&device.session);
+  device.library->device_session_init(device.session);
+}
```
burgholzer: I am not quite if this handling here is correct yet.
Devices are always supposed to be associated with a QDMI Session.
In my understanding, they shouldn't be stored globally, but rather on a per-session basis.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ * given library.
+ * @param library is the DeviceLibrary object that provides the device interface
+ * functions.
+ */
+auto addDevice(const DeviceLibrary& library) -> void {
+  auto& device =
+      *devices().emplace_back(std::make_unique<QDMI_Device_impl_d>());
+  device.library = &library;
+  device.library->device_session_alloc(&device.session);
+  device.library->device_session_init(device.session);
```
burgholzer: Wouldn't we have to check the return code of these methods for success?

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ * Returns a reference to the map of jobs. The map is used to store the
+ * jobs created by the QDMI library.
+ * @returns a reference to a map of QDMI_Job to their corresponding unique
+ * pointers of QDMI_Job_impl_d objects.
+ */
+[[nodiscard]] auto jobs()
+    -> std::unordered_map<QDMI_Job, std::unique_ptr<QDMI_Job_impl_d>>& {
+  static std::unordered_map<QDMI_Job, std::unique_ptr<QDMI_Job_impl_d>> jobs;
+  return jobs;
+}
```
burgholzer: Depending on the outcome of the above comment, I also believe that jobs should be on a per-session basis instead of global.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+    addDynamicDeviceLibrary(path, prefix);
+  }
+  // Add all static and dynamic device libraries to the device list
+  for (const auto& lib : staticDeviceLibraries()) {
+    addDevice(*lib);
+  }
+  for (const auto& [handle, lib] : dynamicDeviceLibraries()) {
+    addDevice(*lib);
+  }
+}
```
burgholzer: I believe the functionality for adding another dynamic library should be exposed as part of the public interface of the driver.
Mainly because this is something that might not only be relevant during initialization but that could be done dynamically at runtime.

We haven't standardized this part yet as part of the Client interface, but I think this could be worth to play around with.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+
+auto finalize() -> void {
+  // Free all existing sessions
+  while (!sessions().empty()) {
+    QDMI_session_free(sessions().begin()->first);
+  }
+  // By clearing the devices, we ensure that all device sessions are freed and
+  // the dynamic libraries are closed, see the destructor of `DeviceLibrary`
+  // and `QDMI_Device_impl_d` for details.
+  devices().clear();
```
burgholzer: The same technique could be applied to sessions as well.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+  dynamicDeviceLibraries().clear();
+  // Clear the static device libraries in the same way just that there is no
+  // `clear` method for the array.
+  for (auto& lib : staticDeviceLibraries()) {
+    lib.reset();
+  }
```
burgholzer: Couldn't one just do
```
staticDeviceLibraries() = {};
```
?

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+ * Returns a reference to the map of sessions. The map is used to store the
+ * sessions allocated by the QDMI library.
+ * @returns a reference to a map of QDMI_Session to their corresponding unique
+ * pointers of QDMI_Session_impl_d objects.
+ */
+[[nodiscard]] auto sessions()
+    -> std::unordered_map<QDMI_Session, std::unique_ptr<QDMI_Session_impl_d>>& {
+  static std::unordered_map<QDMI_Session, std::unique_ptr<QDMI_Session_impl_d>>
+      sessions;
+  return sessions;
```
burgholzer: I am just putting this comment here to ensure that this is a decision made on purpose:
Does this additional level of indirection (using a hashmap over a plain vector of unique pointers) bring that much of an advantage? If yes, what is it?
This does not only hold here, but much more generally.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+  }
+  // Query the job ID from the device job
+  size_t size = 0;
+  result = dev->library->device_job_query_property(
+      uniqueJob->deviceJob, QDMI_DEVICE_JOB_PROPERTY_ID, 0, nullptr, &size);
+  if (result != QDMI_SUCCESS || size == 0) {
+    dev->library->device_job_free(uniqueJob->deviceJob);
+    return QDMI_ERROR_FATAL;
+  }
+  uniqueJob->id.resize(size);
```
burgholzer: Not 100% sure, but like 99% sure that this needs to be `size-1` because of qdmi also reporting the `\0` terminator.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+  if (job == nullptr) {
+    return QDMI_ERROR_INVALIDARGUMENT;
+  }
+  ADD_POINTER_PARAMETER(QDMI_JOB_PARAMETER_PROGRAM, void, job->program, param,
+                        size, value)
+  ADD_SINGLE_VALUE_PARAMETER(QDMI_JOB_PARAMETER_PROGRAMFORMAT,
+                             QDMI_Program_Format, job->format, param, size,
+                             value)
+  ADD_SINGLE_VALUE_PARAMETER(QDMI_JOB_PARAMETER_SHOTSNUM, size_t, job->shotsNum,
+                             param, size, value)
```
burgholzer: Shouldn't this just forward to the device job parameters setter? Same for the getter down below? Otherwise these properties will never be set on the device job.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,688 @@
...
+  return QDMI_ERROR_INVALIDARGUMENT;
```
burgholzer: Fairly sure this should return not supported.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+ * Copyright (c) 2025 Munich Quantum Software Company GmbH
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#include "qdmi/Driver.hpp"
+#include "qdmi/client.h"
```
burgholzer:
```suggestion
#include <qdmi/client.h>
```

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+TEST_P(DriverTest, SessionSetParameterImplemented) {
+  EXPECT_EQ(QDMI_session_set_parameter(session, QDMI_SESSION_PARAMETER_MAX, 0,
+                                       nullptr),
+            QDMI_ERROR_INVALIDARGUMENT)
+      << "Devices must implement `QDMI_session_set_parameter`.";
+}
+
+TEST_P(DriverTest, JobCreateImplemented) {
+  QDMI_Job job = nullptr;
+  EXPECT_EQ(QDMI_device_create_job(device, &job), QDMI_ERROR_NOTSUPPORTED)
```
burgholzer: Not quite sure I follow this assertion.
Job creation should be successful on any QDMI device, right?

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+      QDMI_job_query_property(job, QDMI_JOB_PROPERTY_MAX, 0, nullptr, nullptr),
+      QDMI_ERROR_INVALIDARGUMENT)
+      << "Devices must implement `QDMI_job_set_parameter`.";
+}
+
+TEST_P(DriverTest, JobSubmitImplemented) {
+  QDMI_Job job = nullptr;
+  EXPECT_EQ(QDMI_job_submit(job), QDMI_ERROR_INVALIDARGUMENT)
+      << "Devices must implement `QDMI_job_submit`.";
+}
```
burgholzer: This test does not particularly test whether the function is implemented.
For that, a non-null job, would need to be passed.
This can still be a separate test that ensures the right failure conditions.
Same holds for the other job tests down below.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+      << "Devices must implement `QDMI_job_set_parameter`.";
+}
+
+TEST_P(DriverTest, JobQueryPropertyImplemented) {
+  QDMI_Job job = nullptr;
+  EXPECT_EQ(
+      QDMI_job_query_property(job, QDMI_JOB_PROPERTY_MAX, 0, nullptr, nullptr),
+      QDMI_ERROR_INVALIDARGUMENT)
+      << "Devices must implement `QDMI_job_set_parameter`.";
+}
```
burgholzer: I think I would slightly rephrase all of these test names that use MAX values and assert that the return value is Invalid Parameter.
That is a very specific test that does not particularly indicate that the full function has been implemented.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+      << "Devices must provide the number of qubits.";
+  EXPECT_GT(numQubits, 0) << "Number of qubits must be greater than 0.";
+}
+
+TEST_P(DriverTest, QueryDeviceProperties) {
+  EXPECT_EQ(QDMI_device_query_device_property(device, QDMI_DEVICE_PROPERTY_MAX,
+                                              0, nullptr, nullptr),
+            QDMI_ERROR_INVALIDARGUMENT)
+      << "The MAX property is not a valid value for any device.";
+}
```
burgholzer: As it stands, this is a redundant test, as the respective call is already tested a few tests above.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+            QDMI_SUCCESS)
+      << "Failed to get sites.";
+  std::unordered_set<size_t> ids;
+  for (auto* site : sites) {
+    uint64_t id = 0;
+    EXPECT_EQ(QDMI_device_query_site_property(device, site,
+                                              QDMI_SITE_PROPERTY_INDEX,
+                                              sizeof(uint64_t), &id, nullptr),
+              QDMI_SUCCESS)
+        << "Devices must provide a site id";
```
burgholzer: IDs as a term should be replaced by "index" here as this was renamed in QDMI as well.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+    EXPECT_EQ(QDMI_device_query_site_property(device, site,
+                                              QDMI_SITE_PROPERTY_T2,
+                                              sizeof(double), &t2, nullptr),
+              QDMI_SUCCESS)
+        << "Devices must provide a site T2 time.";
+    EXPECT_GT(t2, 0) << "Devices must provide a site T2 time larger than 0.";
+    EXPECT_EQ(QDMI_device_query_site_property(
+                  device, site, QDMI_SITE_PROPERTY_MAX, 0, nullptr, nullptr),
+              QDMI_ERROR_INVALIDARGUMENT)
+        << "The MAX property is not a valid value for any device.";
```
burgholzer, line 378: This seems to be a much better place to put these assertions than the "_Implememted" tests.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+                                              sizeof(double), &t1, nullptr),
+              QDMI_SUCCESS)
+        << "Devices must provide a site T1 time.";
+    EXPECT_GT(t1, 0) << "Devices must provide a site T1 time larger than 0.";
+    EXPECT_EQ(QDMI_device_query_site_property(device, site,
+                                              QDMI_SITE_PROPERTY_T2,
+                                              sizeof(double), &t2, nullptr),
+              QDMI_SUCCESS)
+        << "Devices must provide a site T2 time.";
+    EXPECT_GT(t2, 0) << "Devices must provide a site T2 time larger than 0.";
```
burgholzer: I would not force these to be defined and rather rewrite these tests so that they check whether the properties are defined. And if they are, assert that they are larger than 0.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+                  QDMI_OPERATION_PROPERTY_DURATION, sizeof(double), &duration,
+                  nullptr),
+              QDMI_SUCCESS)
+        << "Failed to query duration for operation " << name << ".";
+    EXPECT_EQ(QDMI_device_query_operation_property(
+                  device, op, 0, nullptr, numParams, params.data(),
+                  QDMI_OPERATION_PROPERTY_FIDELITY, sizeof(double), &fidelity,
+                  nullptr),
+              QDMI_SUCCESS)
+        << "Failed to query fidelity for operation " << name << ".";
```
burgholzer: Also here, I think we cannot enforce these to be defined (IQM devices for example won't define a duration for the foreseeable future). But if they are defined, they should have sensible values. Durations should be greater equal zero and fidelities should be between 0. And 1.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,379 @@
...
+      << "Device must return `INVALIDARGUMENT` if the buffer is too small.";
+}
+
+TEST_P(DriverTest, QueryNeedsCalibration) {
+  size_t needsCalibration = 0;
+  const auto ret = QDMI_device_query_device_property(
+      device, QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION, sizeof(size_t),
+      &needsCalibration, nullptr);
+  EXPECT_EQ(ret, QDMI_SUCCESS);
+  EXPECT_EQ(needsCalibration, 0);
```
burgholzer: This is not an assertion we can make in general.
The only thing we can assert is that this will return 0 or 1.

test/qdmi/test_driver.cpp
burgholzer, line 1: In order to get good coverage of the code we are adding here, I believe it would be important to also attest a dynamically loaded device library here.

I was thinking whether we could add something that rather directly matches with Qiskit's Target object or the `GenericBackendV2` interface (rather the latter), where one can maybe even build a device at runtime and populate it with values as desired. Definitely something for a separate PR in this series.
Maybe there is also some step in between, where we can test a simpler, dynamically-loaded device library. Potentially even the NA one, but simply compiled as a shared library.

burgholzer, line 1: Another high-level comment mainly intended for brainstorming: are there any more reasonable device-agnostic tests that we could be including here?
The tests here aren't only testing the driver. They are actually mostly sanity checking the devices connected to the driver.

src/qdmi/Driver.cpp
burgholzer, line 1: Fairly general observation here:
There are quite a couple of places throughout the driver where all the QDMI functions are exhaustively listed. How do we best make sure that we aren't missing any of those places when QDMI changes?
Where do we put the respective documentation?

burgholzer (changes requested):
Many thanks for your work on this! This looks great already. I completed a first round of review comments and collected quite some thoughts. I still feel like this isn't too far off where we want it to be 🙂 great work on that!

include/mqt-core/qdmi/Driver.hpp
ystade: Yes, this is probably about right. However, I would wait until we actually have the use case.

src/qdmi/Driver.cpp
ystade: Guess what, I tried that already. The parentheses cause a bit of an issue here with the escaping of CMake. However, I need the parentheses otherwise the nice macros for iterating over the list (so the above macros could also be like `#define DEVICE_LIST_UPPERCASE (MQT_NA)(MQT_SC)(MQT_IT)`) would not work anymore.

src/qdmi/Driver.cpp
ystade: Hm, but then I propose to replace a couple of other code snippets as well through static code that are currently generated with the help of the `ITERATE` macro. Is that your intention?

src/qdmi/Driver.cpp
ystade: I was using the _default_ namespace of MQT here. But I think `qdmi` is better fitting. Just acknowledge and I'll change it.

src/qdmi/Driver.cpp
ystade: I wanted to make the MQT Driver intentionally simpler than that for now. This driver that you see here manages its device connections globally and every (client) session has access to all devices. In particular, there is only one device session between each device and the driver. There are not separate device sessions for every (client) session.

src/qdmi/Driver.cpp
ystade: I knew that you will comment on that. This originated from the fact that there is a `Device_Job` and a `Job`. I was asking myself how the `query_job_property` and `set_job_paramter` function should look like. If one forwards the querys and setters to the device job, then one must translated the enum values or one stores all required information in the driver's job. I have actually no preference at all, that was just one solution. What do you prefer?

src/qdmi/Driver.cpp
ystade: I'll resolve this thread here, because the underlying issue is covered by the thread above.

src/qdmi/Driver.cpp
ystade: It's the same reason as for the analog handling of device sessions in the NA device. This way one does not need to use `new` and `delete` and can fully rely on owning memory via `unique_ptr`s. I know that it is an extra level of indirection and would suggest to change that as soon as we discover performance issues.

src/qdmi/Driver.cpp
ystade: As argued recently orally, I do not think that this should be part of the client interface as, e.g., compiler passes should not be able to add new device libraries. I rather think, this needs an additional interface that controls the driver, e.g., `control` or `management` interface.

src/qdmi/Driver.cpp
ystade: I intentionally did not do that because the deidcated function freeing sessions could do something more then just freeing the sessions (which it currently does not). On the other side, there is no dedicated function to free devices.

src/qdmi/Driver.cpp
ystade: I'd prefer `  staticDeviceLibraries().fill(nullptr);` Any objections?

src/qdmi/Driver.cpp
ystade: As mentioned above, forwarding includes translating the enum values. However, I agree that the parameters should nevertheless be forwarded to the device job. Let's discuss the rest in the thread above and I'll close this thread.

test/qdmi/test_driver.cpp
ystade: But it is not on our NA device, this indeed reports NOTSUPPORTED

test/qdmi/test_driver.cpp
ystade: This is from the tests in the QDMI repo. We might want to change then there as well what is required and what not.

src/qdmi/Driver.cpp
burgholzer: Hm, I am conflicted on this. I get the urge for simplicity here. At the same time, this sets a precedent for an open-source QDMI driver implementation that does not faithfully implement the intended concepts of QDMI.
At the same time, one could argue that QDMI does not dictate how to manage sessions and devices, so the current solution is just fine.
In the spirit of getting this merged, I'd be fine with keeping the simpler driver version for now and adding the complexity when it turns out to be necessary.

src/qdmi/Driver.cpp
burgholzer: Thinking about this now, I believe that the added redundancy could actually be helpful. In case a device job is "lost" for whatever reason, the debugging information is already available in the `Job` and need not be retrieved somehow from the `Device_Job`.
Given how it does not add much complexity, let's keep this for now.

src/qdmi/Driver.cpp
burgholzer: I agree that this is not formally part of the client interface.
However, it can already be part of the public interface of the driver. That interface is not standardized yet in the sense that QDMI offers a header that people can implement and use.
Our implementation here could lay the groundwork for such an interface though. Let's expose something that works for us here and afterwards investigate how we can channel this back to the QDMI specification itself.

src/qdmi/Driver.cpp
burgholzer: By now, I am convinced that this is a valid solution for now. Some of the code patterns around this still look slightly strange to me, but I suppose I will get used to that.

test/qdmi/test_driver.cpp
burgholzer: I suppose this is outdated by now. Even the NA device reports this as supported.

test/qdmi/test_driver.cpp
burgholzer: I agree that it would make a lot of sense to change these items in the QDMI repo as well.

src/qdmi/Driver.cpp
ystade: I do not see any long-term problem when the driver has exactly one session with each device. In my opinion, this would be a valid implementation of a driver.

burgholzer: Yeah. Let's stick with this for now. Resolving this here.

test/qdmi/test_driver.cpp
ystade: I added now the capability that the NA device can additionally be loaded as a dynamic library. For that I had to implement a thin wrapper with another prefix for the QDMI function since otherwise the `dlsym` binds to the global functions that come from the statically linked NA device. Anything further is definitely something for another PR.

src/qdmi/Driver.cpp
ystade: I would argue that this is hopefully covered by appropriate tests, especially the test that checks whether all function definitions exist. In the end, I would say, it is up to us, to check in new PRs that all functions are covered. No matter whether there is documentation on that or not, people may ignore it. If you want, you can already now derive a list of all function from QDMI's documentation. For us, I would say, the best place to look for is the `cmake/prefix_Defs.txt` file in the QDMI repo. I mean, in the end this is a consequence of the decision that the QDMI is not generated by a script rather than hardcoded files, which, I think, is currently still the better option.

burgholzer: Agreed 👍🏼

test/qdmi/test_driver.cpp
ystade: I agree that we *also*  test the device's functionality. However, these tests ensure that all queries are handled correctly by the driver and that the driver also return the correct values from the device. In this regard, I would argue, that we indeed also test the driver with those tests. In the end, we just have the public client interface to test the driver against the QDMI spec. I cannot imagine any better tests.

burgholzer: Fair enough. I see the argument 👍🏼

ystade:
I know, it is not good practice, I also included minor changes in the device to reduce/remove some compiler warnings about shadowed names. I hope that is fine.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,476 @@
...
+                      "MQT NA Dynamic QDMI Device"),
+    [](const testing::TestParamInfo<std::string>& info) {
+      std::string name = info.param;
+      // Replace spaces with underscores for valid test names
+      std::ranges::replace(name, ' ', '_');
+      // Remove parentheses for valid test names
+      std::erase(name, '(');
+      std::erase(name, ')');
+      return name;
+    });
```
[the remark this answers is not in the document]
ystade: The tests in that test suite _do_ execute.

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,423 @@
...
+  virtual ~DeviceLibrary() = default;
+};
+
+#ifndef _WIN32
+/**
+ * @brief Definition of the dynamic device library.
+ * @details This class is used to load the QDMI device interface functions
+ * from a dynamic library at runtime. It inherits from DeviceLibrary and
+ * overrides the constructor and destructor to open and close the library.
+ */
```
burgholzer: I suppose this could benefit from a comment that this is not available on windows systems.

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,423 @@
...
+  ALLOCATED,  ///< The session has been allocated but not initialized
+  INITIALIZED ///< The session has been initialized and is ready for use
+};
+} // namespace qdmi
+
+/**
+ * @brief Definition of the QDMI Device.
+ */
+// Since we treat this struct as a class, we apply also the naming scheme for
+// classes, i.e., an underscore at the end of member names.
```
burgholzer: This mixes a doxygen docstring with a regular code comment. I am fairly sure this will not render as expected.
You could include the code comment as a note or move the comment.
Also applies to further places down below.

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+public:
+  /**
+   * @brief Constructor for the QDMI device.
+   * @details This constructor initializes the device session and allocates
+   * the device session handle.
+   * @param lib is a unique pointer to the device library that provides the
+   * device interface functions.
+   */
+  explicit QDMI_Device_impl_d(std::unique_ptr<qdmi::DeviceLibrary>&& lib);
```
burgholzer, line 184: Is this an rvalue reference on purpose?

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+/**
+ * @brief Definition of the QDMI Job.
+ */
+// Since we treat this struct as a class, we apply also the naming scheme for
+// classes, i.e., an underscore at the end of member names.
+struct QDMI_Job_impl_d {
+private:
+  /// @brief The device library that provides the device interface functions.
+  const qdmi::DeviceLibrary* library_ = nullptr;
```
burgholzer: Why does this need to have the device library when it already has the pointer to the device, which contains the library?

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,423 @@
...
+                  size_t* sizeRet) const -> int;
+
+  /**
+   * @brief Frees the job.
+   * @note This function just forwards to the device's @ref Device::freeJob
+   * function. This function is needed because the interface only provides the
+   * job handle to the @ref QDMI_job_free function and the job's device handle
+   * is private.
+   */
+  auto free() -> void;
```
burgholzer, line 323: I am probably opening Pandora's box here, but I still wanted to bring this up for discussion:
The definitions of these classes are very object-oriented already. We already discussed that we want to eventually expose this to Python as well. For that, we will be building an object-oriented interface around the QDMI client interface _again_, so that we can conveniently bind it to Python and comfortably use it.

I can't it help but feel like we are doing things twice here.

So, just to entertain the thought:
What would be the drawback if we would simply build the C++ interface on top of the device interface and ignore the client interface entirely.
Assuming that, in the future, the whole MQSS will be exposed as a QDMI device itself, we won't be interacting with its driver anyway.
Could it be, that the device interface is entirely sufficient as a standard and that we might not even need the C interface for the driver?
I don't expect the top-levels above QDMI to be implemented in C. Various interfaces along the software stack, yes. Things such as compiler pass interfaces, but the general implementation of the stack, no.
If we wanted to expose our MQT compiler infrastructure and the devices to the world, we would do something very similar to the envisioned MQSS situation: we would wrap our C++ code in a QDMI device implementation.

To follow that line of thought even further: at that point, we could entirely drop the term "device interface" as there is only a single interface then. We could simply refer to this as QDMI. This seems to eliminate complexity, redundancy (no more duplicate enums, no more duplicate structs), and like a fairly nice solution.

However, this almost sounds too good to be true. So I must be overlooking things here. I would appreciate a discussion on this. I fully understand that this is out of scope for this PR here, so what I'd propose is the following:
Let's briefly discuss this in a private meeting before moving it to a separate issue (probably in the QDMI repository) and continuing the discussion there.

The PR here can stay like it is and be merged once all the other comments have been addressed. It is an entirely valid driver implementation as it stands.

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -0,0 +1,423 @@
...
+  /**
+   * @brief Vector of unique pointers to QDMI_Device_impl_d objects.
+   */
+  std::vector<std::unique_ptr<QDMI_Device_impl_d>> devices_;
+
+public:
+  // Delete copy constructors and assignment operators to prevent copying the
+  // singleton instance.
+  Driver(const Driver&) = delete;
+  Driver& operator=(const Driver&) = delete;
```
burgholzer, line 394: I am fairly sure this should define (and default) the move and move assignment operators as per the rule of 5.

src/na/device/CMakeLists.txt
```diff
@@ -143,9 +147,55 @@ if(NOT TARGET ${TARGET_NAME})
       PARENT_SCOPE)
 
   # Make QDMI and MQT Core Version available
-  target_compile_definitions(${TARGET_NAME} PRIVATE QDMI_VERSION="${QDMI_VERSION}")
-  target_compile_definitions(${TARGET_NAME} PRIVATE MQT_CORE_VERSION="${MQT_CORE_VERSION}")
+  target_compile_definitions(${TARGET_NAME} PRIVATE QDMI_VERSION="${QDMI_VERSION}"
+                                                    MQT_CORE_VERSION="${MQT_CORE_VERSION}")
+  if(NOT BUILD_MQT_CORE_SHARED_LIBS)
+    target_compile_definitions(${TARGET_NAME} PRIVATE MQT_CORE_QDMI_NA_STATIC_DEFINE)
+  endif()
```
burgholzer: I don't think this really matters here as we do not define an export header for the qdmi device. So I believe this has no effect.

src/na/device/CMakeLists.txt
```diff
@@ -143,9 +147,55 @@ if(NOT TARGET ${TARGET_NAME})
...
+        PUBLIC qdmi::qdmi
+        PRIVATE ${TARGET_NAME} MQT::ProjectOptions MQT::ProjectWarnings spdlog::spdlog)
+      # set c++ standard
+      target_compile_features(${DYN_TARGET_NAME} PRIVATE cxx_std_20)
+      # set versioning information
+      set_target_properties(
+        ${DYN_TARGET_NAME}
+        PROPERTIES VERSION ${PROJECT_VERSION}
+                   SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
+                   EXPORT_NAME CoreQDMI)
```
burgholzer: Shouldn't this have a different export name than the other QDMI device. Furthermore: should this name be more specific and at least include `na` in there? Most likely this should be `CoreDynNaQDMI` and the top one should be `CoreNaQDMI`

src/qdmi/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+  # Set C++ standard
+  target_compile_features(${TARGET_NAME} PRIVATE cxx_std_20)
+
+  # set versioning information
+  set_target_properties(
+    ${TARGET_NAME}
+    PROPERTIES VERSION ${PROJECT_VERSION}
+               SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
+               EXPORT_NAME CoreQDMI)
```
burgholzer: I suppose this should also be more specific. `CoreQDMIDriver` maybe?

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,479 @@
...
+ */
+
+#include "qdmi/Driver.hpp"
+
+#include "mqt_na_qdmi/device.h"
+
+#include <cassert>
+#include <cstddef>
+#include <cstring>
+#include <dlfcn.h>
```
burgholzer: The seeds to be guarded by `#ifndef _WIN32` because it is not available on Windows.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,479 @@
...
+  if (libHandle_ == nullptr) {
+    throw std::runtime_error("Couldn't open the device library: " + libName);
+  }
+  if (openLibHandles().contains(libHandle_)) {
+    // dlopen uses reference counting, so we need to decrement the reference
+    // count that was increased by dlopen.
+    dlclose(libHandle_);
+    throw std::runtime_error("Device library already loaded: " + libName);
+  }
+  openLibHandles().emplace(libHandle_);
```
burgholzer: I think this can be slightly more idiomatic by using `try_emplace`, which combines the check with the emplacement in case it is not present and gives you a success indicator as a return value.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,479 @@
...
+// @param symbol is the name of the symbol to load.
+#define LOAD_DYNAMIC_SYMBOL(symbol)                                            \
+  {                                                                            \
+    const std::string symbolName = std::string(prefix) + "_QDMI_" + #symbol;   \
+    (symbol) = reinterpret_cast<decltype(symbol)>(                             \
+        dlsym(libHandle_, symbolName.c_str()));                                \
+    if ((symbol) == nullptr) {                                                 \
+      throw std::runtime_error("Failed to load symbol: " + symbolName);        \
+    }                                                                          \
+  }
```
burgholzer, line 103: this could use some whitespace after the macro. 
I accidentally thought that the code after the macro actually belonged to the macro.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,479 @@
...
+  QDMI_Device_Job deviceJob = nullptr;
+  auto result =
+      library_->device_session_create_device_job(deviceSession_, &deviceJob);
+  if (result != QDMI_SUCCESS) {
+    return result;
+  }
+  auto uniqueJob =
+      std::make_unique<QDMI_Job_impl_d>(library_.get(), deviceJob, this);
+  *job = jobs_.emplace(uniqueJob.get(), std::move(uniqueJob)).first->first;
```
burgholzer: I will keep commenting on that pattern until we find a more readable solution. this `emplace(...).first->first` is hard to grasp every time I look at it in a GitHub diff.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,479 @@
...
+  switch (param) {
+  case QDMI_JOB_PARAMETER_PROGRAM:
+    return library_->device_job_set_parameter(
+        deviceJob_, QDMI_DEVICE_JOB_PARAMETER_PROGRAM, size, value);
+  case QDMI_JOB_PARAMETER_PROGRAMFORMAT:
+    return library_->device_job_set_parameter(
+        deviceJob_, QDMI_DEVICE_JOB_PARAMETER_PROGRAMFORMAT, size, value);
+  case QDMI_JOB_PARAMETER_SHOTSNUM:
+    return library_->device_job_set_parameter(
+        deviceJob_, QDMI_DEVICE_JOB_PARAMETER_SHOTSNUM, size, value);
```
burgholzer: Why exactly can't this simply forward to the device function as well as most of the functions above?
Same goes for the function below.

src/qdmi/Driver.cpp
```diff
@@ -0,0 +1,479 @@
...
+  *session = sessions_.emplace(uniqueSession.get(), std::move(uniqueSession))
+                 .first->first;
+  return QDMI_SUCCESS;
+}
+
+auto Driver::sessionFree(QDMI_Session session) -> void {
+  if (session != nullptr) {
+    if (const auto& it = sessions_.find(session); it != sessions_.end()) {
+      sessions_.erase(it);
+    }
```
burgholzer: I believe there is a C++ addition that does this in one go. might be worth checking.

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,548 @@
...
+  ss << a << b << c << d << e;
+  return ss.str();
+}
+// NOLINTBEGIN(readability-identifier-naming,cppcoreguidelines-avoid-const-or-ref-data-members)
+MATCHER_P2(IsBetween, a, b,
+           stringConcat5(negation ? "isn't" : "is", " between ",
+                         PrintToString(a), " and ", PrintToString(b))) {
+  return a <= arg && arg <= b;
+}
+// NOLINTEND(readability-identifier-naming,cppcoreguidelines-avoid-const-or-ref-data-members)
```
burgholzer, line 41: are these ignores still needed? especially the last one?

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,548 @@
...
+#ifndef _WIN32
+  static void SetUpTestSuite() {
+    // avoid loading the same dynamic library twice as it would cause an
+    // exception; this only happens, because the `DriveJobTest` inherits from
+    // `DriverTest` and both call `SetUpTestSuite()` when instantiated
+    static bool added = false;
+    if (!added) {
+      qdmi::Driver::get().addDynamicDeviceLibrary(DYN_DEV_LIB, "MQT_NA_DYN");
+      added = true;
+    }
```
burgholzer: Should loading the same library twice really cause an exception? Shouldn't this "just work"?

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,548 @@
...
+TEST_P(DriverTest, QueryNeedsCalibration) {
+  size_t needsCalibration = 0;
+  const auto ret = QDMI_device_query_device_property(
+      device, QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION, sizeof(size_t),
+      &needsCalibration, nullptr);
+  EXPECT_EQ(ret, QDMI_SUCCESS);
+  EXPECT_THAT(needsCalibration, testing::AnyOf(0, 1));
+}
+#ifdef _WIN32
+const std::array<std::string, 2> DEVICES{"MQT NA Default QDMI Device"};
```
burgholzer:
```suggestion
const std::array<std::string, 1> DEVICES{"MQT NA Default QDMI Device"};
```
?

test/qdmi/test_driver.cpp
```diff
@@ -0,0 +1,548 @@
...
+    testing::ValuesIn(DEVICES),
+    [](const testing::TestParamInfo<std::string>& paramInfo) {
+      std::string name = paramInfo.param;
+      // Replace spaces with underscores for valid test names
+      std::ranges::replace(name, ' ', '_');
+      // Remove parentheses for valid test names
+      std::erase(name, '(');
+      std::erase(name, ')');
+      return name;
+    });
```
burgholzer, line 541: This feels a bit redundant, couldn't this just be one test suite? feel free to just say no and resolve this. I was just curious.

burgholzer (changes requested):
Thanks for all the hard work on this @ystade really great to see the progress here.

I basically just have minor comments that should be easily addressed. One will probably fix at least the Windows build errors. Another one might fix the Python one where the installation fails.
Other than that, there is one pretty big one; but you'll see it when you get there that it is definitely not something to tackle as part of this PR.

include/mqt-core/qdmi/Driver.hpp
ystade: Because the library in the device is private and cannot be accessed here. Would you change that?

src/qdmi/Driver.cpp
ystade: Because the `JOB_PARAMETER` and the `DEVICE_JOB_PARAMETER` have a different type.

include/mqt-core/qdmi/Driver.hpp
burgholzer: I think it would make sense to either make it public or to provide a getter so that it can be used.

src/qdmi/Driver.cpp
ystade: In the next commit, I'll have a dedicated translation function. Let me know if you like that solution more. I think, it is slightly more readable.

burgholzer: Ah, yeah. It would be a stretch to assume that one can simply reinterpret one enum value as the other similar to how your small wrapper for the device does that, right?

src/qdmi/Driver.cpp
ystade: After seraching, this is apparently as good as it gets. :/

src/qdmi/Driver.cpp
ystade: In the device, it refers at least to the same enum. Only the prefixing generated two versions out of the same enum. Here it is a (completely) different enum that could potentially contain more or less values (even though unlikely).

test/qdmi/test_driver.cpp
ystade: No, the nice thing about the second test suite is that it already allocated a job on top of the session. However, I want to split the test suites because if it would be one test suite and the job creation fails the entire test suite would fail. This way I can test a lot without creating a job and more pinpoiting to the error cause if there is one.

src/qdmi/Driver.cpp
burgholzer: Technically, we have full control over both though and can certainly check ourselves that the assumption is true. But I don't actually want to argue too much here. The current solution is fine. I was just wondering.

test/qdmi/test_driver.cpp
burgholzer: I see that. Would it make sense to split this into two test files?

include/mqt-core/qdmi/Driver.hpp
ystade: Yes, the device is supposed to take over ownership of the DeviceLibraray.

src/qdmi/Driver.cpp
burgholzer: Just checked. You should be able to write all of this as just 
```
sessions_.erase(session);
```
Without the if and find and so on.

ystade: True, stackoverflow told me something differrent but according to the c++ reference, you are right.

ystade:
@burgholzer Everything is green now. I just updated the changelog one last time. Should we merge now?

PR #1019, opened 2025-06-23
✨ Add MLIR pass for merging rotation gates

### LINKED ISSUE #898: ✨ MLIR - Rotation Gate Merge Pass/Pattern

We recently introduced the first optimization passes for the MLIR ecosystem we are building #892.
Now this opens up a flood of options for implementing further optimization passes.
One that would be nice to have is an optimization pass that merges rotation gates wherever possible.

Example (qasm-style):

```
rx(pi/2) q[0];
rx(pi/2) q[0];
```

could be merged into

```
rx(pi) q[0];
```

Currently, we only have a pass that cancels consecutive inverse gates (see https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td#L31-L37, https://github.com/munich-quantum-toolkit/core/blob/main/mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInverses.cpp, and https://github.com/munich-quantum-toolkit/core/blob/main/mlir/lib/Dialect/MQTOpt/Transforms/CancelConsecutiveInversesPattern.cpp).
However, that pass does not handle parametrized gates such as rotation gates.

I see two possible ways to implement this:
- This new feature could be added as a new pattern to that pass as the features are related and might use similar helper methods (in that case, the pass would have to be renamed).
- This new feature could be added as an entirely new transformation pass.

Both strategies seem reasonable at first glance.

In any case, the pass should at least cover the following rotation gates:
- rx, ry, rz, phase
- rxx, ryy, rzz, rzx
- xx_minus_yy, xx_plus_yy
- U, U2

Most likely, an optional parameter for defining a tolerance value would be useful for controlling, when two parametrized gates cancel one another.

Cancellation rules could be derived from https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/dd/Operations.cpp#L74-L142

denialhaag:
## Description

This PR adds a pass for merging simple rotation gates.

Fixes #898 

## Checklist:

### DIFF
--- a/mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
@@ -93,6 +93,41 @@ struct MergeRotationGatesPattern final
   }
 
   /**
+   * @brief Creates a new rotation gate.
+   *
+   * The new rotation gate is created by adding the angles of two compatible
+   * rotation gates.
+   *
+   * @tparam OpType The type of the operation to create.
+   * @param op The first instance of the rotation gate.
+   * @param user The second instance of the rotation gate.
+   * @param rewriter The pattern rewriter.
+   * @return A new rotation gate.
+   */
+  template <typename OpType>
+  static UnitaryInterface
+  createOpAdditiveAngle(UnitaryInterface op, UnitaryInterface user,
+                        mlir::PatternRewriter& rewriter) {
+    auto loc = user->getLoc();
+
+    auto userInQubits = user.getInQubits();
+    auto userPosCtrlInQubits = user.getPosCtrlInQubits();
+    auto userNegCtrlInQubits = user.getNegCtrlInQubits();
+
+    auto opParam = op.getParams()[0];
+    auto userParam = user.getParams()[0];
+    auto add = rewriter.create<mlir::arith::AddFOp>(loc, opParam, userParam);
+    const llvm::SmallVector<mlir::Value, 1> newParamsVec{add.getResult()};
+    const mlir::ValueRange newParams(newParamsVec);
+
+    return rewriter.create<OpType>(
+        loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
+        userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
+        mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
+        userPosCtrlInQubits, userNegCtrlInQubits);
+  }
+
+  /**
    * @brief Merges two consecutive rotation gates into a single gate.
    *
    * The function supports gphase, p, rx, ry, rz, rxx, ryy, rzz, and rzx.
@@ -107,74 +142,26 @@ struct MergeRotationGatesPattern final
     auto const type = op->getName().stripDialect().str();
 
     auto user = mlir::dyn_cast<UnitaryInterface>(*op->getUsers().begin());
-    auto loc = user->getLoc();
-
-    auto opParam = op.getParams()[0];
-    auto userParam = user.getParams()[0];
-
-    auto add = rewriter.create<mlir::arith::AddFOp>(loc, opParam, userParam);
-    const llvm::SmallVector<mlir::Value, 1> newParamsVec{add.getResult()};
-    const mlir::ValueRange newParams(newParamsVec);
-
-    auto userInQubits = user.getInQubits();
-    auto userPosCtrlInQubits = user.getPosCtrlInQubits();
-    auto userNegCtrlInQubits = user.getNegCtrlInQubits();
 
     UnitaryInterface newUser;
     if (type == "gphase") {
-      newUser = rewriter.create<GPhaseOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<GPhaseOp>(op, user, rewriter);
     } else if (type == "p") {
-      newUser = rewriter.create<POp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<POp>(op, user, rewriter);
     } else if (type == "rx") {
-      newUser = rewriter.create<RXOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RXOp>(op, user, rewriter);
     } else if (type == "ry") {
-      newUser = rewriter.create<RYOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RYOp>(op, user, rewriter);
     } else if (type == "rz") {
-      newUser = rewriter.create<RZOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RZOp>(op, user, rewriter);
     } else if (type == "rxx") {
-      newUser = rewriter.create<RXXOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RXXOp>(op, user, rewriter);
     } else if (type == "ryy") {
-      newUser = rewriter.create<RYYOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RYYOp>(op, user, rewriter);
     } else if (type == "rzz") {
-      newUser = rewriter.create<RZZOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RZZOp>(op, user, rewriter);
     } else if (type == "rzx") {
-      newUser = rewriter.create<RZXOp>(
-          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
-          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
-          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
-          userPosCtrlInQubits, userNegCtrlInQubits);
+      newUser = createOpAdditiveAngle<RZXOp>(op, user, rewriter);
     } else {
       throw std::runtime_error("Unsupported operation type: " + type);
     }


denialhaag:
So far, this PR adds a pass that can merge the following gates:

- gphase
- rx, ry, rz
- rxx, ryy, rzz, rzx

I'm not sure about how to merge the remaining gates mentioned in #898. I have the feeling that they are fundamentally different and cannot be merged as easily because the respective gates cannot be decomposed into something that commutes.

I also still have to figure out why the tests are failing on Windows.

I already texted you privately, but if you have any input, I would be very interested, @burgholzer, @DRovara, and @ystade. :)

DRovara:
> I'm not sure about how to merge the remaining gates mentioned in #898. I have the feeling that they are fundamentally different and cannot be merged as easily because the respective gates cannot be decomposed into something that commutes.

I did some more thinking on this topic and I might have an idea: Rotations by Euler Angles are known to be difficult to compose from multiple rotations. That's why quaternions (https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation) are often used by game engines to represent rotations.

I think we might be able to do the same thing here: first translate the three Euler angles to quaternions, then we combine the multiple rotations into one single rotation and then we translate back into Euler angles. If you want we can also chat about this in person tomorrow, @denialhaag.

burgholzer:
> > I'm not sure about how to merge the remaining gates mentioned in #898. I have the feeling that they are fundamentally different and cannot be merged as easily because the respective gates cannot be decomposed into something that commutes.
> 
> 
> 
> I did some more thinking on this topic and I might have an idea: Rotations by Euler Angles are known to be difficult to compose from multiple rotations. That's why quaternions (https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation) are often used by game engines to represent rotations.
> 
> 
> 
> I think we might be able to do the same thing here: first translate the three Euler angles to quaternions, then we combine the multiple rotations into one single rotation and then we translate back into Euler angles. If you want we can also chat about this in person tomorrow, @denialhaag.
> 
> 
> 
> 

Have a look at the current Qiskit code in that regard. I remember that they are doing something very similar in the code since recently. Might serve as some good inspiration.

The rules for `u` gates should be fairly straight forward, I believe. `u2` is only a special case of `u` so that should work similarly.

The two qubit gates are a bit more special.
I think that two of these with the same beta value just add up the angles of the first parameter. I am not aware of any more general rules for that at the moment.

In general, merging might not be particularly easy here. But cancellation is fairly easy to decide as also indicated by the code linked in the issue description. Maybe those gates should be included as special rules there.

Just my two cents here though.

denialhaag:
> > I'm not sure about how to merge the remaining gates mentioned in #898. I have the feeling that they are fundamentally different and cannot be merged as easily because the respective gates cannot be decomposed into something that commutes.
> 
> I did some more thinking on this topic and I might have an idea: Rotations by Euler Angles are known to be difficult to compose from multiple rotations. That's why quaternions (https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation) are often used by game engines to represent rotations.
> 
> I think we might be able to do the same thing here: first translate the three Euler angles to quaternions, then we combine the multiple rotations into one single rotation and then we translate back into Euler angles. If you want we can also chat about this in person tomorrow, @denialhaag.



> > I'm not sure about how to merge the remaining gates mentioned in #898. I have the feeling that they are fundamentally different and cannot be merged as easily because the respective gates cannot be decomposed into something that commutes.
> 
> I did some more thinking on this topic and I might have an idea: Rotations by Euler Angles are known to be difficult to compose from multiple rotations. That's why quaternions (https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation) are often used by game engines to represent rotations.
> 
> I think we might be able to do the same thing here: first translate the three Euler angles to quaternions, then we combine the multiple rotations into one single rotation and then we translate back into Euler angles. If you want we can also chat about this in person tomorrow, @denialhaag.

Yes, an in-person chat would be nice! I'll be in the office in a bit! :)

denialhaag:
> Have a look at the current Qiskit code in that regard. I remember that they are doing something very similar in the code since recently. Might serve as some good inspiration.
> 
> The rules for `u` gates should be fairly straight forward, I believe. `u2` is only a special case of `u` so that should work similarly.
> 
> The two qubit gates are a bit more special. I think that two of these with the same beta value just add up the angles of the first parameter. I am not aware of any more general rules for that at the moment.
> 
> In general, merging might not be particularly easy here. But cancellation is fairly easy to decide as also indicated by the code linked in the issue description. Maybe those gates should be included as special rules there.
> 
> Just my two cents here though.

Thanks for the ideas! I'll have a look at the Qiskit implementation!

denialhaag:
For reference, the Qiskit pass is implemented [here](https://github.com/Qiskit/qiskit/blob/stable/2.1/qiskit/transpiler/passes/optimization/optimize_1q_gates.py#L34).

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -36,6 +36,13 @@ def CancelConsecutiveInverses : Pass<"cancel-consecutive-inverses", "mlir::Modul
   }];
 }
 
+def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
+  let summary = "This pass searches for consecutive applications of rotation gates that can be merged.";
+  let description = [{
+    This pass searches for consecutive applications of rotation gates that can be merged.
```
DRovara: I know the existing code base is not a good standard for that, but I think we should use good summaries and descriptions if we can. Maybe you can check if there is some best practice on what the summary should be?

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include <stdexcept>
+#include <string>
+#include <unordered_set>
+namespace mqt::ir::opt {
+
+/**
+ * @brief This pattern attempts to merge consecutive rotation gates.
+ */
+struct MergeRotationGatesPattern final
+    : mlir::OpInterfaceRewritePattern<UnitaryInterface> {
```
DRovara, line 39: I'm not sure if this is possible, but couldn't we instead make this pattern targets the parameter-traits rather than the full unitary interface? That way, we would not have to to implement `getParams()` for the `UnitaryInterface` which doesn't always make sense.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
```diff
@@ -0,0 +1,236 @@
...
+      auto add =
+          rewriter.create<mlir::arith::AddFOp>(loc, opParams[0], userParams[0]);
+      const llvm::SmallVector<mlir::Value, 1> newParamsVec{add.getResult()};
+      const mlir::ValueRange newParams(newParamsVec);
+      return rewriter.create<RYOp>(
+          loc, inQubits.getType(), controlQubitsPositive.getType(),
+          controlQubitsNegative.getType(), mlir::DenseF64ArrayAttr{},
+          mlir::DenseBoolArrayAttr{}, newParams, inQubits,
+          controlQubitsPositive, controlQubitsNegative);
+    }
```
DRovara: There's quite a lot of repetitive code here. Maybe there's some way to merge parts of these branches?

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+// Copyright (c) 2025 Munich Quantum Software Company GmbH
+// All rights reserved.
+//
+// SPDX-License-Identifier: MIT
+//
+// Licensed under the MIT License
+
+// RUN: quantum-opt %s -split-input-file --merge-rotation-gates | FileCheck %s
```
DRovara, line 9: This will not quite work as intended. Note how in e.g. https://github.com/munich-quantum-toolkit/core/blob/5605dde5b288ec66ba5752cb0f59aa5299599d72/mlir/test/Dialect/MQTOpt/Transforms/consecutive-inverses.mlir#L11 individual test cases are separated by `//-----`. This might look unnecessary, but it is actually required to make `-split-input-file`work.

Also, I'd like to keep up a standard of writing short test descriptions below the `//-----` separators.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
```diff
@@ -0,0 +1,88 @@
...
+
+    // CHECK: %[[Reg_4:.*]] = "mqtopt.insertQubit"(%[[Reg_3]], %[[Q01_3]]#0)  <{index_attr = 0 : i64}>
+    %reg_4 = "mqtopt.insertQubit"(%reg_3, %q01_3#0) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
+    // CHECK: %[[Reg_5:.*]] = "mqtopt.insertQubit"(%[[Reg_4]], %[[Q01_3]]#1)  <{index_attr = 1 : i64}>
+    %reg_5 = "mqtopt.insertQubit"(%reg_4, %q01_3#1) <{index_attr = 1 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
+    // CHECK: %[[Reg_6:.*]] = "mqtopt.insertQubit"(%[[Reg_5]], %[[Q12_2]]#1)  <{index_attr = 2 : i64}>
+    %reg_6 = "mqtopt.insertQubit"(%reg_5, %q12_2#1) <{index_attr = 2 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
+    // CHECK: "mqtopt.deallocQubitRegister"(%[[Reg_6]])
+    "mqtopt.deallocQubitRegister"(%reg_6) : (!mqtopt.QubitRegister) -> ()
+    return
```
DRovara, line 360: Maybe we could also add some negative test cases here that contain situations where no merging is supposed to happen.

DRovara (changes requested):
Thanks a lot for the effort! I made some comments with suggestions, feel free to look into them.

One more thing though: As far as I see, this will only work if the parameters are passed as operands and won't work if the parameters are static attributes. This is okay for now, but we should probably keep this in mind and at the very least open a follow-up issue for that.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: I was able to reduce it a bit now by computing the sum in a single place, but I cannot find a way to make the creation nicer.

My Python brain wants to do something like this (very pseudo code):

```
ops = {"rx": RXOp, "ry": RYOp, "rz": RZOp, ...}
rewriter.create<ops[type]>(loc, inQubits.getType(), ...)
```

It looks like `RXOp` and friends cannot be put in a C++ map, though. 

Do you have a nice idea?

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
denialhaag: Ahh, maybe that fixes the Windows tests. 🤞 

I put them back now and added brief comments.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
denialhaag: Each test now has a gate that is left as is because it has no consecutive partner. Do you have something else in mind? I would be happy to add it!

denialhaag:
> Thanks a lot for the effort! I made some comments with suggestions, feel free to look into them.
> 
> One more thing though: As far as I see, this will only work if the parameters are passed as operands and won't work if the parameters are static attributes. This is okay for now, but we should probably keep this in mind and at the very least open a follow-up issue for that.

Thank you for the review!

Regarding your second point, I think I am still missing some context (and knowledge) to understand. 😬

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: I will have a look at this next week!

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
denialhaag: In the [docs](https://mlir.llvm.org/docs/PassManagement/#tablegen-specification), it says that the `summary` should be a one-liner, whereas the `description` should be more detailed (somewhere else on the linked page, it says that one can describe the constraints there).

I'll improve the description next week!

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
DRovara: Yeah, I'm pretty sure something like that cannot be done without some complex macros. I guess in that situation, it is okay to keep the repeating structure where it really cannot be reduced, unless @burgholzer maybe has an idea?

burgholzer: No particular idea at the top of my head. I think it should be fine to keep that. We can always reiterate/refactor if we find a better solution at some point.

DRovara:
> `xxminusyy`, `xxminusyy`, `u`, and `u2` still cannot be merged, though. I don't know if it maybe makes sense to move this to a different issue. 🤔

I'm fine with making that a separate step. In that case, let's not forget to create the corresponding issue if we don't plan to tackle that situation immediately.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
DRovara: Generally I prefer many small test cases over few big ones.

E.g. the situation where a test case tests for merging two gates, and then as a side effect also tests that some other gate is not also merged with them is not as nice as having two test cases, one which only tests merging and one which only tests that merging does not happen in some situation.

With the state of the test system right now, it barely makes a difference. But if we ever do end up having a more rigorous test system that actually manages to differentiate between test cases in a nice way, that would be a useful distinction.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
denialhaag: Done. Let me know what you think. :)

DRovara: Looks a lot better already! Just to point out the typo: I believe it should be "angles", not "angels".

Also, no strong opinion, but is it a good idea to have the description start with the same sentence as the summary? I know this situation for python docstrings, where some tooling just displays the detailed description directly below the general summary. In that case, it looks weird when the same sentence appears twice.

burgholzer: The general trend I see across the ecosystem is to _not_ repeat the brief in the detailed summary.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
denialhaag: Done. Let me know what you think!

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
denialhaag: Fixed. 👍

denialhaag:
After an offline discussion with @DRovara, I removed the feature of erasing gates whose parameters meet the necessary conditions. This should only be supported for gates with static parameters. To this end, I created #1030.

DRovara:
> After an offline discussion with @DRovara, I removed the feature of erasing gates whose parameters meet the necessary conditions. This should only be supported for gates with static parameters. To this end, I created #1030.

Just for context, the idea was

`rx(a), rx(b) ==> rx(a + b)`, then if `a + b === 0`, the gate can be removed.

However, since this PR works with dynamic operands, we can't really know if `a + b === 0`. 
The previous solution was to check if `a` and `b` were `arith.constant` expressions and getting their numeric values from there, but in my opinion, it makes sense to have this idea covered by a different constant folding path that converts the arguments from dynamic operands to static attributes.

Once we have that, `a + b === 0` can be checked in the merge process for rotation gates that use static attributes (see #1030).

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: So I found [`OpTraitRewritePattern`](https://mlir.llvm.org/doxygen/classmlir_1_1OpTraitRewritePattern.html#details), but I'm struggling to use it. Is it possible that `OneParameter` is currently not being exported but only `ParameterArityTrait`? 🤔

DRovara: Now that you say that, I remember encountering a similar issue once.
I believe MLIR might not generate classes for each of these traits specifically.

In that case, for the sake of not spending too much time with this, I think we can leave it as is right now. It's not necessarily a bad solution, even if the trait approach might have been a little bit cleaner.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
```diff
@@ -0,0 +1,221 @@
...
+    // Erase op
+    rewriter.eraseOp(op);
+  }
+
+  void rewrite(UnitaryInterface op,
+               mlir::PatternRewriter& rewriter) const override {
+    auto const type = op->getName().stripDialect().str();
+
+    if (type == "gphase" || type == "rx" || type == "ry" || type == "rz" ||
+        type == "rxx" || type == "ryy" || type == "rzz" || type == "rzx") {
```
DRovara: Can we use the `MERGEABLE_GATES` constant for this check instead?

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
```diff
@@ -0,0 +1,323 @@
...
+// SPDX-License-Identifier: MIT
+//
+// Licensed under the MIT License
+
+// RUN: quantum-opt %s -split-input-file --merge-rotation-gates | FileCheck %s
+
+// -----
+// This test checks that consecutive rx gates are merged correctly.
+
+module {
```
DRovara, line 14: I think it would be best if we still did some smaller clean-up with the CHECK strings here.

While writing [the MQTOpt `dialect features` tests](https://github.com/munich-quantum-toolkit/core/blob/main/mlir/test/Dialect/MQTOpt/IR/dialect_features.mlir), we decided for the standard of collecting all `CHECK` strings at the start of the test function body. Maybe we should do that here, too, rather than interleaving the checks.

Also, we had some discussions on what parts of the code actually warrant `CHECK` strings and decided that only those functions relevant to the tested feature do. That means that something like the register allocations/deallocations or qubit extractions/insertions don't really require CHECK strings.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
```diff
@@ -0,0 +1,323 @@
...
+
+    // CHECK: "mqtopt.deallocQubitRegister"(%[[Reg_4]])
+    "mqtopt.deallocQubitRegister"(%reg_4) : (!mqtopt.QubitRegister) -> ()
+
+    return
+  }
+}
+
+// -----
+// This test checks that incompatible multi-qubit gates are not merged.
```
DRovara, line 273: I believe there might still be some scenarios that are worthy of test cases. Here are some ideas:
- `rx(a) q1; rx(b) q2` should not be merged as they are different qubits
- similarly, two `rxx` gates that only share one qubit should not be merged
- maybe we can even test something like `rzz q1, q2; rzz q2, q1` which should not be merged based on the code (but could in theory also be merged, just so that we have a documentation of that behaviour in the tests)
- it would also be interesting to test how merging works if the angle parameters are not constants, but e.g. arguments to the test function. In that situation, there would probably be an actual "add" operation to add up the angles rather than just the single, already added up constant that is generated by MLIR's internal optimizations.

DRovara (changes requested):
I gave the PR another look and I just have some small comments left. I hope they should be addressable without too much effort.

Especially regarding the test cases, I think it is good to set up a good standard early.

Thanks a lot!

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
denialhaag: Done. 👍

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
denialhaag: Done. Let me know what you think! :)

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: Good point. Done. 👍

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <mlir/Support/LLVM.h>
+#include <mlir/Support/LogicalResult.h>
+#include <stdexcept>
+#include <string>
+#include <unordered_set>
+
+namespace mqt::ir::opt {
+
+static const std::unordered_set<std::string> MERGEABLE_GATES = {
+    "gphase", "rx", "ry", "rz", "rxx", "ryy", "rzz", "rzx"};
```
burgholzer: this is missing the `p` (phase) gate, which is the same as the `rz` gate up to global phase. Adding that might require a couple of docs and test updates throughout this PR.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
```diff
@@ -0,0 +1,219 @@
...
+    auto unitaryUser = mlir::dyn_cast<UnitaryInterface>(user);
+    if (op.getAllOutQubits() != unitaryUser.getAllInQubits()) {
+      return mlir::failure();
+    }
+    if (op.getPosCtrlInQubits().size() !=
+            unitaryUser.getPosCtrlInQubits().size() ||
+        op.getNegCtrlInQubits().size() !=
+            unitaryUser.getNegCtrlInQubits().size()) {
+      // We only need to check the sizes, because the order of the controls was
+      // already checked by the previous condition.
```
burgholzer, line 89: This is really more of a meta-question; probably for @DRovara:
Do we actually sort the controls of operations somehow canonically?
Because `ctrl(0) @ ctrl(1) @ x(2)` is totally the same as `ctrl(1) @ ctrl(0) @ x(2)`, that is, controls commute with one another.
The comment here in the code triggered me slightly. 
If these operations are not yet considered the same, I would argue we should do something about that in a separate PR.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
```diff
@@ -0,0 +1,219 @@
...
+          userPosCtrlInQubits, userNegCtrlInQubits);
+    } else if (type == "rzx") {
+      newUser = rewriter.create<RZXOp>(
+          loc, userInQubits.getType(), userPosCtrlInQubits.getType(),
+          userNegCtrlInQubits.getType(), mlir::DenseF64ArrayAttr{},
+          mlir::DenseBoolArrayAttr{}, newParams, userInQubits,
+          userPosCtrlInQubits, userNegCtrlInQubits);
+    } else {
+      throw std::runtime_error("Unsupported operation type: " + type);
+    }
```
burgholzer: Technically, the redundancy here could be removed a bit by wrapping this in a template function that takes the OpType as a template parameter and the `user` as well as the `add` value as a parameter and returns the new user.
That should reduce this by quite a bit.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
```diff
@@ -0,0 +1,337 @@
...
+    // CHECK: %[[ANY:.*]]:2 = mqtopt.rxx(%[[Res_2]]) %[[Q0_1]]#1, %[[Q0_1]]#0 : !mqtopt.Qubit, !mqtopt.Qubit
+
+    %reg_0 = "mqtopt.allocQubitRegister"() <{size_attr = 3 : i64}> : () -> !mqtopt.QubitRegister
+    %reg_1, %q0_0 = "mqtopt.extractQubit"(%reg_0) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+    %reg_2, %q1_0 = "mqtopt.extractQubit"(%reg_1) <{index_attr = 1 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
+
+    %c_0 = arith.constant 1.000000e+00 : f64
+    %c_1 = arith.constant 2.000000e+00 : f64
+    %q01_1:2 = mqtopt.rxx(%c_0) %q0_0, %q1_0 : !mqtopt.Qubit, !mqtopt.Qubit
+    %q01_2:2 = mqtopt.rxx(%c_1) %q01_1#1, %q01_1#0 : !mqtopt.Qubit, !mqtopt.Qubit
```
burgholzer, line 354: I really like the tests throughout this PR 👍🏼 
This one brought up something though that is related to the other comment about the order of controls and the respective effect.
A similar question goes here, but the answer is clearer: `rxx`, `ryy`, `rzz`, (not `rzx`), `swap`, `iswap` are all symmetrical. The order of qubits does not make a difference for these gates. As such, they should probably be canonicalized.

And just to have it said out loud, this technically also holds for gates like a (multi-)controlled-Z and other diagonal gates. It does not matter which qubit is the control and which qubit is the target in that regard.

@ystade @DRovara any comments on that?

burgholzer (changes requested):
Many thanks for all the work on this 🙏🏼  great to see more and more MLIR passes making it into the repository.

I have a couple of comments, which hopefully are fairly easy to address/comment on. One potential simplification, a missing gate, and two more general comments/questions on potential canonicalizations that we are missing.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
DRovara: That's a good point I didn't really think about.
For clarification, what I meant with "the order of the controls" in the other pass (where this check and comment was taken from) was not actually the order inside the groups of control qubits, but order as in "the first x controls are positive controls, the next y controls are negative".

But inside the control groups, the order really shouldn't matter. The easy solution would be a canonicalisation pass that keeps the order consistent. We could also just change the lines highlighted by this comment, but then we would also need to change the inverse cancel pass. Probably best to keep this for a canonicalisation pass.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
DRovara: Fair point, I don't see any reason against that. However, I just found something interesting in [MLIR's guide on Canonicalisations](https://mlir.llvm.org/docs/Canonicalization/):

>Pass pipelines should not rely on the canonicalizer pass for correctness. They should work correctly with all instances of the canonicalization pass removed.

That's a bit confusing because I thought the point of canonicalisations was to make writing optimisations easier. Dan Gohman's article on Canonicalisations (which is suggested reading by the MLIR guide) even says explicitly:

>It [using canonicalizations] means that subsequent optimizations that look for specific patterns of code only have to look for the canonical forms, rather than all forms

My intuition says that "correct" here means that it should never produce "wrong" code, not that it should always work, but that's just my best guess. In any case, it should be reasonable to ignore that first quote from MLIR (unless I misunderstand something here).

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
burgholzer: Ah, thanks for the clarification. I missed that part.

I think a canonicalization pass for the controls still makes sense, though.
This should adequately address the case mentioned above.

I am not sure if something like `ctrl(2) @ negctrl(1) @ ctrl(0) @ x(3)` could occur in the code. That could be canonicalized to either
- `ctrl(0) @ negctrl(1) @ ctrl(2) @ x(3)` (ascending qubits)
- `ctrl(0) @ ctrl(2) @ negctrl(1) @ x(3)` (grouping positive and negative controls)
- ...

although the second option probably does not make too much sense.

In any case, can we move this discussion to a separate/new issue and continue over there? This is definitely not something to be tackled as part of this PR.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
burgholzer: My intuition also says that the whole point of canonicalization is that it ensures you can rely on certain pre-conditions in optimizations.
Maybe also something to add to the new issue, to collect some further information in a central place.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
DRovara: I can set up the issue and add my two cents there. But just for quick information: Our MLIR format does not allow something like `ctrl - negctrl - ctrl`. It always requires positive controls first, then negative controls - so there really isn't much room for design decisions anyways, should be quite straightforward.

Edit: #1031 now describes this situation.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: Ahh, that's a nice idea. Done.

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: Thanks for creating the issue! :)

mlir/lib/Dialect/MQTOpt/Transforms/MergeRotationGatesPattern.cpp
denialhaag: Done.

mlir/test/Dialect/MQTOpt/Transforms/merge-rotation-gates.mlir
denialhaag: I added a comment that this test should fail when #1031 is implemented. 👍

PR #1020, opened 2025-06-24
♻️ Switch from reference counting to mark-and-sweep garbage collection in decision diagram package

MatthiasReumann:
## Description

Continues the work from #980 and resolves issue #644. 

## Checklist:


include/mqt-core/dd/Node.hpp
```diff
@@ -40,12 +37,23 @@ struct NodeBase : LLBase {
...
-   *    0b1 = mark path is conjugated (tmp flag)
+   * 0b10000 = mark flag used for mark-and-sweep garbage collection,
+   *  0b1000 = marks a reduced dm node,
+   *   0b100 = marks a dm (tmp flag),
+   *    0b10 = mark first path edge (tmp flag),
+   *     0b1 = mark path is conjugated (tmp flag)
    */
-  std::uint16_t flags = 0;
+  std::uint16_t flags = 0; // TODO: Would it make sense to use a larger datatype
+                           // here since Ref is gone?
```
burgholzer: Yeah. I was thinking about that already.
Initially, I would have hoped that we get some space savings for the nodes here, but that turned out to be wishful thinking. So we might as well make the padding explicit.
However that's also kind of awkward as there is no 48bit type.

include/mqt-core/dd/Operations.hpp
```diff
@@ -101,8 +101,8 @@ MatrixDD getInverseDD(const qc::Operation& op, Package& dd,
  * @brief Apply a unitary operation to a given vector DD.
  *
  * @details This is a convenience function that realizes @p op times @p in and
- * correctly accounts for the permutation of the operation's qubits as well as
- * automatically handles reference counting.
```
burgholzer: I think I would personally prefer to keep the old wording in most of the places where this was changed. Essentially, we are still doing some kind of reference counting, but only at the top level.

include/mqt-core/dd/Package.hpp
```diff
@@ -194,63 +194,142 @@ class Package {
...
+      }
+    }
+
+    /// @brief Unmark edges contained in @p roots.
+    template <class Edge> static void unmark(const RootSet<Edge>& roots) {
+      for (auto& [edge, _] : roots) {
+        auto e = edge;
+        e.unmark();
+      }
+    }
```
burgholzer: Just so that I noted it down: I am still not quite sure if this trick of copying the hashmap key is actually valid.

include/mqt-core/dd/UniqueTable.hpp
```diff
@@ -195,7 +170,7 @@ class UniqueTable {
 
         while (p != nullptr) {
           std::cout << "\t\t" << std::hex << reinterpret_cast<std::uintptr_t>(p)
-                    << std::dec << " " << p->ref << std::hex;
+                    << std::dec;
```
burgholzer: This can be dropped entirely, otherwise the subsequent pointers would not be printed in hex.

src/dd/Complex.cpp
```diff
@@ -29,6 +29,16 @@ bool Complex::approximatelyZero() const noexcept {
...
 
+void Complex::mark() const noexcept {
+  r->mark();
+  i->unmark();
+}
+
+void Complex::unmark() const noexcept {
+  r->mark();
+  i->unmark();
+}
```
burgholzer:
```suggestion
void Complex::mark() const noexcept {
  r->mark();
  i->mark();
}

void Complex::unmark() const noexcept {
  r->unmark();
  i->unmark();
}
```
I suppose this was merely an oversight

src/dd/RealNumber.cpp
```diff
@@ -13,16 +13,20 @@
...
 
 namespace dd {
 
-static constexpr std::size_t LSB = static_cast<std::uintptr_t>(1U);
+static constexpr std::uintptr_t LSB = 1U;
+static constexpr std::uintptr_t LSB2 = 2U;
+
+RealNumber* RealNumber::next() const noexcept {
+  const std::uintptr_t ptr = reinterpret_cast<std::uintptr_t>(next_) & ~LSB2;
+  return ptr == 0U ? nullptr : reinterpret_cast<RealNumber*>(ptr);
```
burgholzer: Is that if really necessary? I would have assumed that casting 0 to any pointer type would yield nullptr 🧐

burgholzer (commented):
Just briefly had some time to look over this and wanted to leave some comments just in case. Many thanks for working on this 🙏

include/mqt-core/dd/Node.hpp
MatthiasReumann: What's your take on using a [bit field](https://en.cppreference.com/w/cpp/language/bit_field.html)? All I know is that they exist.

burgholzer: Hm. Just read through the linked page.
Sounds reasonable in principle. I am a bit worried that the standard text mentions the implementation defined nature of the allocation details quite often.
We should, at least, make sure that on the platforms that we commonly test with, the packing and alignment of the resulting struct is as we would expect it.

MatthiasReumann: Yes. Quick googling tells us that this might not be the most portable and efficient solution. 

So I guess we stick to 32 bit for now? 🤔

src/dd/Complex.cpp
MatthiasReumann: Yes. Thanks 🤦🏻

include/mqt-core/dd/Package.hpp
MatthiasReumann: I was wondering that too. Seems to work - but looks awkward. I think something like a `const_cast` could be useful here.

include/mqt-core/dd/Node.hpp
burgholzer: Yeah. Let's stick with that for now. We don't use most of the flag bitfield anyway.
So this is highly likely to be the most portable solution for now.

include/mqt-core/dd/Package.hpp
burgholzer: Maybe one could also just separately mark `edge.p` and `edge.w`. Maybe that would work without the copy because only the pointers are const, but not the data they point to 🤷🏼

src/dd/RealNumber.cpp
MatthiasReumann: Nope, definitely not.

MatthiasReumann:
👋🏻 @burgholzer 

I think this is a good point to ask you for some input. The `TODO:`s in the code highlight the discussion points.

1. The current implementation holds a static `.5` in the real unique table. Not used, this number will certainly be deleted - some tests have failed as a consequence of that. But: What's more interesting is that other tests - Approximation - fail due to numeric issues when removing this static value. For example, the `TwoQubitCorrectlyRebuilt` doesn't approximate anything with the exact budget of `.25` but works for `.26`. Similarly, `ThreeQubitRemoveNodeWithChildren`.
2. For some reason the `Grover/Grover.Functionality/15_qubits_2` is the only test that fails inconsistently on my system and those of the CI. I want to believe this is also due to numerical issues - but I'm not entirely sure either. In fact, this also happens for the recursive version of the test.
3. Using the `track` and `untrack` semantic requires a different philosophy for referencing counting in the project, I think. As of now it feels kind of awkward to use `track` and `untrack` throughout. As of now, it also seems a bit inconsistent in the way it is used. 
    
    Here's a suggestion: It is the callees job to `track` and `untrack` states. When returning states, the state should always be untracked. The function itself can change the tracked states but must always "clean up their own garbage". Essentially, this bowls then down to the end-user calling `track` and `untrack` for DD's they want to - well - track. I could even imagine that is possible to refactor `VectorDD` to a struct with constructor (`track`) and destructor (`untrack`) for something similar to the RAII idiom. 

    This is a major refactoring - for sure. 

Thanks 🙇🏻

include/mqt-core/dd/Node.hpp
MatthiasReumann: Thinking about BitFields made me wondering: Wouldn't it be nice to have something like:

```cpp
struct Flags { // Size: 4 bytes, alignment 4 bytes
  uint32_t mark : 1;
  uint32_t reduced : 1;
  uint32_t dm : 1;
  uint32_t firstPath : 1;
  uint32_t conjugated : 1;
};

Flags f;
```

The compiler would give us the respective masks for free. LLVM uses it in a similar fashion too, see [here](https://llvm.org/doxygen/IR_2Value_8h_source.html).

Anyhow, probably something for a follow-up PR.

test/algorithms/test_grover.cpp
```diff
@@ -89,96 +84,80 @@ TEST_P(Grover, Functionality) {
...
   for (std::size_t i = 0U; i < iterations - 1U; ++i) {
-    auto f = dd->multiply(iteration, e);
-    dd->incRef(f);
-    dd->decRef(e);
-    e = f;
-    dd->garbageCollect();
+    const auto next = dd->multiply(iterationOp, iteration);
+    dd->track(next);
+    dd->untrack(iteration); // This will automatically untrack the iterationOp.
+    iteration = next;
```
MatthiasReumann: Interestingly, adding `garbageCollect()` here causes the numerical issues.

include/mqt-core/dd/Node.hpp
burgholzer: That looks really reasonable. and convenient.
Seems to be something for a separate PR, but I definitely like the idea! 👍🏼

test/algorithms/test_grover.cpp
burgholzer: I suppose that might be because the tables are actually full enough so that collection happens. Without garbage collection, "dead" entries might become alive again in subsequent computations. With garbage collection, these entries might be gone and the computation might result in slightly different results due to tolerances and such.
The interesting thing is that I would not expect things to actually change based on the changes in this PR. We are still using the same criterion for when to collect garbage and we should be tracking the same DDs as previously with the more fine-grained reference counting.

include/mqt-core/dd/Package.hpp
MatthiasReumann: Much easier solution: Make `Edge::mark` and `Edge::unmark` const functions 🫠

test/algorithms/test_grover.cpp
MatthiasReumann: Okay got it. Mistake on my side. The problem was that `iterationOp` was garbage collected after the first loop iteration. Hence, following iterations had (probably) to work with slightly different numbers. 

Does this sound reasonable to you? 🤔

bindings/dd/register_dd_package.cpp
```diff
@@ -173,20 +173,24 @@ void registerDDPackage(const py::module& mod) {
         const auto data = v.unchecked<1>();
         const auto length = static_cast<size_t>(data.shape(0));
         if (length == 0) {
-          return dd::vEdge::one();
+          const auto state = dd::vEdge::one();
+          p.track(state);
+          return state;
```
burgholzer: At first glance, this seems kind of wasteful.
Neither the terminal node, nor the static numbers will ever be collected.
Would it make sense to leave this here as
```
return dd::vEdge::one();
```
? 
I am a bit split on this. Maybe it also makes sense to simply track edges consistently; terminal or not.

eval/eval_dd_package.cpp
```diff
@@ -161,22 +161,23 @@ benchmarkFunctionalityConstructionGrover(
...
     if (iterBits[j]) {
       if (zero) {
-        dd.incRef(f);
-        dd.decRef(e);
+        dd.track(f);
+        dd.untrack(e);
         e = f;
         zero = false;
       } else {
+        dd.track(f);
```
burgholzer: Just looking for differences here. This is a `track` call that was not here before.
And I do not think it should be necessary.
```suggestion
```

include/mqt-core/dd/Package.hpp
```diff
@@ -194,63 +194,156 @@ class Package {
...
+
+    /// @brief Remove from respective root set.
+    template <class Node> void removeFromRoots(const Edge<Node>& e) {
+      auto& set = getRoots<Node>();
+      auto it = set.find(e);
+      if (it == set.end()) {
+        throw std::invalid_argument("Edge is not part of the root set.");
+      }
+      if (--it->second == 0U) {
+        set.erase(it);
```
burgholzer, line 244: Just to have it said out loud: I suspect, that this might actually be another potential difference between the previous approach and the new one. The dictionaries here are using the hash functions of the `Edge` class and their equality operator. However, the Edge class implements a fuzzy equality operator instead of an explicit one. Edges are considered equal if their Pointers are the same and the edge weight is approximately equal up to the defined tolerance.
While I believe it should not happen, a potential scenario could be that two edges are considered equal, while they are actually not.
Due to the way the tolerance works, this should actually never happen. The second edge would never be created because the lookup of its edge weight would result in the weight of the first edge.

src/dd/FunctionalityConstruction.cpp
```diff
@@ -23,12 +23,14 @@
...
-    return MatrixDD::one();
+    auto e = MatrixDD::one();
+    dd.track(e);
+    return e;
   }
 
   auto permutation = qc.initialLayout;
   auto e = dd.createInitialMatrix(qc.getAncillary());
-
+  dd.track(e);
```
burgholzer: This one also seems like an additional track that was not there before.
```suggestion
```

src/dd/RealNumber.cpp
```diff
@@ -13,37 +13,71 @@
...
   if (exactlyZero(e)) {
     return &constants::zero;
   }
   return reinterpret_cast<RealNumber*>(reinterpret_cast<std::uintptr_t>(e) ^
-                                       LSB);
+                                       NEG_FLAG);
+}
+
+void RealNumber::mark(RealNumber* e) noexcept {
+  RealNumber* p = isNegativePointer(e) ? getAlignedPointer(e) : e;
```
burgholzer: This can be simplified. The `isNegativePointer` check already performs one check, so that's the bare minimum. Might as well spend that one operation in the `getAlignedPointer` function that directly yields the result.
```suggestion
  RealNumber* p = getAlignedPointer(e);
```
Also holds for the functions below.

src/dd/Simulation.cpp
```diff
@@ -95,6 +95,7 @@ std::map<std::string, std::size_t> sample(const qc::QuantumComputation& qc,
     // simulate once and measure all qubits repeatedly
     auto permutation = qc.initialLayout;
     auto e = in;
+    dd.track(e);
```
burgholzer: Another additional `track` here. In order to preserve the behaviour from before this PR, this should not be necessary.
```suggestion
```

test/algorithms/test_grover.cpp
```diff
@@ -89,96 +84,77 @@ TEST_P(Grover, Functionality) {
...
-        dd->decRef(e);
-        e = g;
-        dd->garbageCollect();
-      }
-    }
+
+  auto iteration = iterationOp;
+  dd->track(iteration);
+  for (std::size_t i = 0U; i < iterations - 1U; ++i) {
+    iteration = dd->applyOperation(iterationOp, iteration);
```
burgholzer: Maybe I am stupid, but this completely changed the original code?
This is no longer computing the functionality recursively.

burgholzer (changes requested):
Hey @MatthiasReumann 👋🏼 
Thanks for your continued work on this! This should be fairly close to being done I believe.
I very carefully reviewed the code diff again and identified a couple of places that seem fishy. 
In the end, I would expect this PR to reduce the memory footprint of the `RealNumber` class and, thus, the total memory allocated.
In addition, I would expect benchmarks (especially those that do not need garbage collection) to improve in runtime because of the savings in only tracking the root edges and completely skipping the continuous tracking of the edge weights.
The rest of the system should behave exactly the same (modulo the typical fluctuations due to memory addresses being used in the hashes).

@MatthiasReumann the numbers you provided me with roughly confirm this. Maybe one of the suggestions in the review helps to level out the last negative impacts. Would be nice to re-run the benchmarks after the changes. You can also directly upload them here as part of the PR.

Eventually, a changelog entry as well as an upgrade guide notice would be great. In the end, this contains breaking changes for the DD package due to the renaming of the reference counting functions. We could think about reverting the renaming; then there would be no real need for an upgrade note and for changes in consuming libraries.

include/mqt-core/dd/Package.hpp
MatthiasReumann: Yes! I was wondering about this. 

For example, in the `Package::measureOneCollapsing` function one has to set the weight of the state before tracking it - awkward, right?  Previously it didn't matter.

test/algorithms/test_grover.cpp
MatthiasReumann: You are absolutely right. Sorry.

include/mqt-core/dd/Package.hpp
burgholzer: Technically, it also mattered before, I suppose. And I would not rule out that the tracking in some of the functions (especially those expliciltly messing with the weights) is not perfectly implemented. Although the unit tests provide a reasonbly high confidence that everything is roughly working as expected.

eval/eval_dd_package.cpp
MatthiasReumann: Actually, this one is necessary. If you run the `eval` script in debug (which no one probably did previously - for good reason) this will fail with _"Reference count of Node must not be zero before decrement"_. In the release build the preprocessor simply removes the assert.

burgholzer: Can we cross match that with the `test_grover.cpp` code? this should, more or less, be a copy of that.

bindings/dd/register_dd_package.cpp
MatthiasReumann: Yes, I was having similar thoughts on this. 

As of now, the problem with removing this `track` statement is that `untrack` will throw an exception if the edge is found in the root set - which will be the case for these states. (This is also part of the same reason why there are additional `track` statements throughout the pull request).

The only way I think this could be circumvented is by adding a bit of logic in `track` and `untrack` so that these states won't be tracked generally. 

The idea of tracking states consistently seems reasonable. However, this would probably require a bunch of additional `track` statements in helper functions. Then, most of the additional `track`s of this PR could be removed. 

---

For now, I've added such an `if` statement to `track` and `untrack`. Not sure if I like it, honestly. Any opinions?

eval/eval_dd_package.cpp
MatthiasReumann: Yes, it's the same code with one subtle difference:

```cpp
// test_grover
if (iterBits[j]) {
    if (zero) {
      [same]
    } else {
      auto tmp = dd->multiply(e, f);  // operation := e
      dd->incRef(tmp);
      dd->decRef(e); // !difference: untrack(e := operation) | track (tmp)
      e = tmp;
      dd->garbageCollect();
    }
}

// eval
if (iterBits[j]) {
    if (zero) {
      [same]
    } else {
      e = dd.applyOperation(e, f); // untrack(f) | track(tmp)
    }
}
```

bindings/dd/register_dd_package.cpp
burgholzer: Hm. Also not too sure, if I really like it. But it gets the job done. And it's just three additional checks per add/removal.
Maybe adding an additional (static) member function to the `Edge` class that encapsulates this check is a sweet solution. We already had a `ReferenceCountingNeeded` function or similar.

eval/eval_dd_package.cpp
burgholzer: Ah. That's indeed a tiny, but important difference.
the `eval` code should likely be
```
e = dd.applyOperation(f, e);
```
in order for the tracking to be correct.
In a similar fashion, the `test_grover.cpp` code can be adjusted to using `applyOperation`

MatthiasReumann:
@burgholzer I've adjusted the code accordingly. Thanks for the feedback 🙌🏼

<hr>

The results of the benchmarks didn't change. However, there's probably still some bug in the code. Take a look at the following excerpt from the `mqt-core-dd-compare` utility:
```
|           before |            after |      ratio |           algo |          task |    n |                         component |                       metric |
|-----------------:|-----------------:|-----------:|---------------:|--------------:|-----:|----------------------------------:|-----------------------------:|
|      1           | 113988           | 113988     |         Grover | Functionality |   36 |             matrix_memory_manager | num_available_for_reuse_peak |
|   2015           | 239605           |    118.911 |         Grover | Functionality |   36 |             matrix_memory_manager |          num_available_total |
|   2015           | 125617           |     62.341 |         Grover | Functionality |   36 |             matrix_memory_manager |    num_available_from_chunks |
|    825           |  30798           |     37.331 |         Grover | Functionality |   32 |       real_numbers_memory_manager |    num_available_from_chunks |
|    825           |  30798           |     37.331 |         Grover | Functionality |   32 |       real_numbers_memory_manager |          num_available_total |
|   1133           |  31324           |     27.647 |         Grover |    Simulation |   36 |       real_numbers_memory_manager |          num_available_total |
|   1133           |  31324           |     27.647 |         Grover |    Simulation |   36 |       real_numbers_memory_manager |    num_available_from_chunks |
```

First of all, note that the algorithm is Grover's algorithm - the one that got significantly slower. 

Then, the ratios are all over the place. I can't figure out what, but something is definitely wrong here. Profiling didn't reveal anything out of the ordinary. Do you have any idea want could've caused this behavior? 

I've attached the evaluation results. 

<hr>

Lastly, should we revert back to `incRef` and `decRef` or keep `track` and `untrack`? 

<hr>

[Results JSON of Mark&Sweep](https://github.com/user-attachments/files/20993503/mark_and_sweep.json)
[Results JSON of Reference Counting](https://github.com/user-attachments/files/20993504/refcount.json)


Thanks 🙇🏻

include/mqt-core/dd/Edge.hpp
```diff
@@ -88,6 +88,16 @@ template <class Node> struct Edge {
...
   }
 
+  /**
+   * @brief Check whether an edge requires tracking.
+   * @param e The edge to check.
+   * @return Whether the edge requires tracking.
+   */
+  [[nodiscard]] static constexpr bool trackingRequired(const Edge& e) {
+    return !(e.isTerminal() && constants::isStaticNumber(e.w.r) &&
+             constants::isStaticNumber(e.w.i));
```
burgholzer:
```suggestion
    return !e.isTerminal() || !constants::isStaticNumber(e.w.r) ||
             !constants::isStaticNumber(e.w.i);
```
Might as well apply DeMorgan here directly.

src/dd/RealNumber.cpp
```diff
@@ -79,32 +105,6 @@ bool RealNumber::approximatelyZero(const RealNumber* e) noexcept {
   return e == &constants::zero || approximatelyZero(val(e));
 }
 
-bool RealNumber::noRefCountingNeeded(const RealNumber* const num) noexcept {
-  assert(!isNegativePointer(num));
-  return num == nullptr || constants::isStaticNumber(num) ||
-         num->ref == std::numeric_limits<RefCount>::max();
-}
```
burgholzer, line 86: Would it make sense to even preserve this function? and use it in the corresponding `mark()` function? Just thinking about making the diff of this PR as small as possible.

test/algorithms/test_grover.cpp
```diff
@@ -140,44 +134,38 @@ TEST_P(Grover, FunctionalityRecursive) {
...
+        dd->untrack(e);
         e = f;
         zero = false;
       } else {
-        auto g = dd->multiply(e, f);
-        dd->incRef(g);
-        dd->decRef(e);
-        e = g;
+        e = dd->applyOperation(f, e);
         dd->garbageCollect();
```
burgholzer: this call is superfluous now that we use `applyOperation` here.
```suggestion
```

test/algorithms/test_grover.cpp
burgholzer, line 1: In the spirit of keeping the diff as small as possible and understanding the differences more clearly, would it make sense to reduce this diff as far as possible/feasible?

eval/eval_dd_package.cpp
```diff
@@ -161,22 +161,22 @@ benchmarkFunctionalityConstructionGrover(
...
       if (zero) {
-        dd.incRef(f);
-        dd.decRef(e);
+        dd.track(f);
+        dd.untrack(e);
         e = f;
         zero = false;
       } else {
-        e = dd.applyOperation(e, f);
+        e = dd.applyOperation(f, e);
```
burgholzer, line 175: Just to point out the obvious (or maybe not): this line here is responsible for messing with the statistics for the Grover functionality construction.
Given how it's actually a bugfix though, these can be safely ignored.
It should be more correct now.
In the previous case, it could happen that numbers with a zero reference count got decremented once more, which effectively immortalised them through an underflow of the ref count.

burgholzer (changes requested):
Many thanks for your continued work on this 🙏🏼 
We are really getting close to the finish line.
I have a couple of comments left, which mostly go towards the goal of reducing the code diff in this PR.
Part of that is to keep using `incRef` and `decRef` over `mark` and `unmark`. That should also very clearly show the necessary changes for this feature.

Then, we can run the comparison once again; with the main focus being on the runtime differences. Secondary focus on the total memory used/allocated.

src/dd/RealNumber.cpp
MatthiasReumann: Essentially this check is done in the newly added `trackingRequired` function. Theoretically, we could keep it. But I'm not sure if it brings any benefit.

eval/eval_dd_package.cpp
MatthiasReumann: I've rerun both benchmarks with both fixed. The results are now more reasonable. Interestingly, the above behavior for the `matrix_memory_manager` swapped. Probably, as we discussed yesterday, due to being close to the garbage collection threshold but right under. Otherwise, most benchmarks have improved in runtime. Grover being the exception. 

[results_mark_and_sweep.json](https://github.com/user-attachments/files/21014031/results_mark_and_sweep.json)
[results_ref_counting.json](https://github.com/user-attachments/files/21014033/results_ref_counting.json)

burgholzer: Thanks for running these again! 👍🏼 
I think it really boils down to the difference in garbage collection runs. If an additional run triggers, this will considerably increase the runtime due to the complexity of the collection. This phenomenon is independent of the PR changes here though and it might as well happen if you run the benchmarks multiple times on the main branch.

Given how it seems to be a 5-15% runtime improvement for most of the remaining benchmarks, I would consider this a success overall 💪🏼

burgholzer (approved):
Nice job 👏 thanks for putting in the work!

I just pushed some finishing touches and cleanups.
Let's get this in 🚀 

Thanks also to @q-inho for the initial work on this! 🙏🏼

PR #1021, opened 2025-06-25
♻️ Specify Install Components for Wheel

ystade:
## Description

This PR does not immediately solve an issue but contributes to avoiding an issue that arose in #996. In particular, it allows to select the install components explicitly and, hence, avoid installing components that were added by dependencies where install instructions cannot be deactivated, e.g., RE2.

Also in general, I think, it is a good idea to set the install components for the wheel explicitly.

## Checklist:


pyproject.toml
```diff
@@ -92,6 +92,9 @@ build.targets = [
   "mqt-core-dd-bindings",
 ]
 
+# Only install the Python package component
```
burgholzer: This comment here is slightly incorrect by now. Maybe 
```suggestion
# Only install the mqt-core library and Python package components
```

burgholzer (changes requested):
This is looking reasonable. From my past experience working on the installation instructions, this really deserves some proper testing. I am especially worried since this will probably no longer install `nlohmann_json` now.
Can we make sure this works in other repositories that use mqt-core? Most importantly `mqt-qmap` and especially on Linux.
Running `cibuildwheel` locally, installing that wheel locally, building mqt-qmap from source without build-isolation (to ensure the local wheel is used), and running the tests.
If that works, I am fairly convinced that this can be merged.
Maybe this also helps with the MLIR Plugin and the conflict that we saw between Catalyst and mqt-core on `nlohmann_json`.

burgholzer (approved):
Let's go with this for now 🚀 We can always fix stuff, if this turns out to be breaking something.

PR #1042, opened 2025-07-08
🐛 Fix parsing of `GPhase`

ystade:
## Description

This PR adds a—currently failing—test case that tests the parsing of the global phase operation in the `MQTOpt` dialect.

## Checklist:


ystade:
I want to use this PR to consolidate on a solution for how to represent global phases in MLIR, i.e., the newly added test case should succeed. The issue is that a global phose has no target qubits, however, MLIR cannot deal with the situation when the place for the target qubits remain empty.

ystade:
The most promising idea right now is to define the custom assembly format separately for the global phase operation. This includes defining the global phase operation within each dialect and not in the shared table gen file that is used by both dialects.

ystade:
@burgholzer @DRovara The latest commit contains my proposal how I would imagine a separate definition of a global phase. However, this does not work and requires more changes because a gphase cannot be handled as a unitary operation anymore, see the compilation failure in the merge rotation pass. I think this is not desirable, hence, I am happy to take input on how we solve this issue.

burgholzer:
Hm. It seems to fail because the `getInQubits` method of the unitary interface is not defined.
Wasn't that exactly what the `0-arity` trait was for initially? Basically indicating that all the input and output methods would return an empty vector?

ystade:
I totally agree with you both: (1) The global phase should still implement the unitary interface and (2) the in- and out-qubits should be an empty vector. However, this causes an issue, I do not know how to solve yet: For the implementation of the unitary interface, the operation is required to have an `in_qubits` and only for the opt dialect `out_qubits` attribute (which will not be used in the end. Since it will not be used, it should not appear in the custom assembly format, however this causes an error in the compilation as you can see in the build logs of the last CI jobs. Any ideas welcome. :)

mlir/lib/Dialect/MQTOpt/IR/MQTOptOps.cpp
```diff
@@ -9,6 +9,14 @@
...
 
 #include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h" // IWYU pragma: associated
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/OpImplementation.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/Types.h"
+#include "mlir/Support/LogicalResult.h"
+
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
```
burgholzer:
```suggestion
#include <mlir/IR/BuiltinTypes.h>
#include <mlir/IR/OpImplementation.h>
#include <mlir/IR/Operation.h>
#include <mlir/IR/Types.h>
#include <mlir/Support/LogicalResult.h>

#include <llvm/ADT/STLExtras.h>
#include <llvm/ADT/SmallVector.h>
```
To my understanding, these are all system includes.

mlir/lib/Dialect/MQTOpt/IR/MQTOptOps.cpp
```diff
@@ -55,3 +63,115 @@ void mqt::ir::opt::MQTOptDialect::initialize() {
...
+      parser.emitError(parser.getCurrentLocation(),
+                       "expected at least one type after `ctrl` keyword");
+      return mlir::failure();
+    }
+  }
+
+  if (parser.parseOptionalKeyword("nctrl").succeeded()) {
+    if (parser.parseTypeList(negCtrlOutQubits).failed()) {
+      parser.emitError(parser.getCurrentLocation(),
+                       "expected at least one type after `ctrl` keyword");
```
burgholzer:
```suggestion
                       "expected at least one type after `nctrl` keyword");
```

burgholzer (approved):
Despite two small corrections, this looks great to me.
@DRovara if you are happy with it, feel free to apply the suggestions and merge.

This might be worth a changelog entry (in the "Fixed") category.

PR #1059, opened 2025-07-17
🧪 Add additional tests to improve coverage

DRovara:
## Description

This PR adds additional test cases. In particular, it mainly focuses on trait and assembly verification, which has not been thoroughly tested before. Several tests were added that intentionally violate traits or assembly formats. Furthermore, it adds some tests for transformation passes.

The following types of uncovered lines were ignored:
- The general structure:
```
if (mlir::failed(APPLY_PATTERNS_GREEDILY(op, std::move(patterns)))) {
      signalPassFailure();
}
```
  This is a repeating pattern that comes from MLIR and that is not really easy to test, so I kept it uncovered. Mainly, the `then` block of this `if` is just a safety check to make sure future changes don't break anything, currently, none of these `then` blocks should ever be reached.
- Most code in the `MQT Core Roundtrip`. This pass will be updated by the Catalyst Conversion PR anyways. I just fixed one bug that I clearly saw while looking through the uncovered lines and added a related fix.
- All code in the `Quantum Sink` pass. This pass will be updated in the future.
- The `unsupported operation type` check in `MergeRotationGatesPattern.cpp`. This check is not strictly necessary, but it's still good for future-proofing.

## Checklist:


denialhaag:
@DRovara, thank you for finding this! I pushed a fix that fixes the reason for the crash without fixing the underlying problem: as implemented, the pass can only merge gates that have users. I would propose that I look into this in a separate PR. I rewrote your test to check that user-less `gphase` gates are **not** merged. What do you think? 🤔

DRovara:
> @DRovara, thank you for finding this! I pushed a fix that fixes the reason for the crash without fixing the underlying problem: as implemented, the pass can only merge gates that have users. I would propose that I look into this in a separate PR. I rewrote your test to check that user-less `gphase` gates are **not** merged. What do you think? 🤔

That's a good point. I just realised that without involved qubits, there is no def-use chain that can be accessed from within this pass, so it's not even compatible with the general logic. Therefore, it absolutely makes sense to have a separate PR for this issue. Maybe this might even be its own transformation pattern that could then also be added to the merge rotations pass.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
```diff
@@ -192,15 +202,30 @@ struct ToQuantumComputationPattern final : mlir::OpRewritePattern<AllocOp> {
...
+                   [](const size_t index) {
+                     return qc::Control{static_cast<qc::Qubit>(index),
+                                        qc::Control::Type::Pos};
+                   });
+    std::transform(negCtrlInsIndices.cbegin(), negCtrlInsIndices.cend(),
+                   std::inserter(controls, controls.end()),
+                   [](const size_t index) {
+                     return qc::Control{static_cast<qc::Qubit>(index),
+                                        qc::Control::Type::Neg};
+                   });
```
burgholzer: I can't really help myself, but I personally think this is expressed much clearer as a simple range-based for loop, that emplaces the respective controls.
Something like
```c++
for (const auto& index: posCtrlInsIndices) {
  controls.emplace(static_cast<qc::Qubit>(index), qc::Control::Type::Pos);
}
for (const auto& index: negCtrlInsIndices) {
  controls.emplace(static_cast<qc::Qubit>(index), qc::Control::Type::Neg);
}
```

mlir/test/Dialect/MQTDyn/IR/dialect_features.mlir
```diff
@@ -571,6 +571,8 @@ module {
     func.func @testAllocMissingSize() {
         // expected-error-re@+1 {{'mqtdyn.allocQubitRegister' op exactly one attribute ({{.*}}) or operand ({{.*}}) must be provided for 'size'}}
         %qreg = "mqtdyn.allocQubitRegister"() : () -> !mqtdyn.QubitRegister
+
+        return
```
burgholzer, line 575: Is that return not implicit? Or, asked differently, does it really have to be explicit?
I suspect it is good practice to keep it; if only for consistency.

burgholzer (changes requested):
This looks reasonable! Thanks for your work on this.
I just have two small comments. Aside from that, the conflicts with main need to be resolved, and I would suspect that this might also trigger a clang-tidy warning due to the shift to C++20.

mlir/test/Dialect/MQTDyn/IR/dialect_features.mlir
DRovara: In MLIR, every block needs a "terminator" instruction, function blocks do not explicitly return to the caller and might instead also end with a "jump"-like instruction to continue to a different block (although I can't really think of a use case for that).

Therefore, this is necessary. The only reason we didn't notice this before is that in this test case, execution already fails in the line before (because the test case is supposed to check for an error in `mqtdyn.allocQubitRegister` so it never reached the end of the block where the "missing terminator" error would have been thrown.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
DRovara: I have no strong preference for either approach. The for loop does look cleaner, so if you say that's still idiomatic for C++ I'm happy to change it.

mlir/test/Dialect/MQTDyn/IR/dialect_features.mlir
burgholzer: Ah. That makes sense! Thanks for the explanation.

mlir/lib/Dialect/MQTOpt/Transforms/ToQuantumComputationPattern.cpp
burgholzer: I'd very much argue that range-based for loops are idiomatic C++ 😌👍🏼 thanks for the changes.

burgholzer (approved):
LGTM 👍🏼 I'll tag this for auto-merge.

PR #1060, opened 2025-07-17
🐛👷 Fix condition for Python coverage

denialhaag:
## Description

This PR fixes the condition for triggering the Python coverage job. 

## Checklist:

- [x] ~~I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,7 +265,7 @@ jobs:
               && '' || 'cpp-linter,'
             }}
             ${{
-              fromJSON(needs.change-detection.outputs.run-python-tests) && !contains(github.event.pull_request.labels.*.name, 'extensive-python-ci')
+              fromJSON(needs.change-detection.outputs.run-python-tests)
               && '' || 'python-tests,'
             }}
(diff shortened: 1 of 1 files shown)


burgholzer (approved):
The previous fix would not have worked because you cannot depend on a job that is skipped. Thus, the coverage upload would have never been triggered.

I now refactored the logic slightly so that the Python tests and the extensive Python tests are now complementary again instead of exclusive.
Additionally, coverage uploads only depend on the regular test job, which now always runs (or is skipped for the same reason).

Should all be good now. But let's see what CI says.

denialhaag:
> The previous fix would not have worked because you cannot depend on a job that is skipped. Thus, the coverage upload would have never been triggered.
> 
> I now refactored the logic slightly so that the Python tests and the extensive Python tests are now complementary again instead of exclusive. Additionally, coverage uploads only depend on the regular test job, which now always runs (or is skipped for the same reason).
> 
> Should all be good now. But let's see what CI says.

Ahh, okay, I didn't know that. Thanks for the help!

PR #1098, opened 2025-08-07
✨ Transform `MQTDyn` to `MQTRef` Dialect

### LINKED ISSUE #1080: ✨ Definition of `MQTMem`
### Problem Statement

Compiling a quantum computation to a specific hardware requires that the specific hardware qubits can be addressed in the intermediate representation. Currently, there is no dialect in MLIR that allows for static addressing of qubits, as both allocate qubits dynamically.

### Proposed Solution

A new dialect `MQTMem` should be defined following the structure of `MQTDyn` and implementing all the quantum operations. The main difference is that there will not be a dynamic allocation of qubits; rather, the qubits are addressed statically.

MatthiasReumann:
## Description

This pull request adds static qubit addressing operations to the `MQTDyn` dialect and renames it to `MQTRef`. Effectively resolves #1080.

#### Additions and Changes
- Rename `mqtdyn` to `mqtref`
- Add `QubitOp` to `mqtref` ("Get static `QubitType`")
- Update tests and add new test cases accordingly

## Checklist:


MatthiasReumann:
## Rationale / Discussion

I've experimented a lot the last couple of days and I just want to put this somewhere for discussion. 

---

> Why two qubit types, i.e., why add `DynamicQubitType` and `StaticQubitType`? 

In theory, the `extract` as well as the `qubit` operation could return the already existing `QubitType`. In `mqtref` dynamic and static qubits are references to qubits. Static qubits are, simply put, "pinned" qubits to hardware.  However, not distinguishing between those two allows the mixed use in unitary and measurement operations. This should be avoided in my opinion. Hence, two types. 

**Consequence 1** 
Consequently, one has to specify _input_ qubit types for unitary ops also in the `mqtref` dialect. For consistency, this is achieved via the same custom directive (hence same custom assembly format) used in the `mqtopt` dialect for _outputs_. Moreover, I think it should be possible to refactor the unitary op such that this directive is applied to inputs rather than outputs in the `mqtopt` dialect as well. Implementing the [`InferTypeOpInterface`](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/InferTypeOpInterface.td) interface should resolve any typing issues. 

> Can't we do something like custom parsing to infer the input types automatically and keep the format the same?

It is possible to add custom `C++` parsing and printing functions via `hasCustomAssemblyFormat = 1`. Unfortunately, we require this logic for each instantiation of the unitary op  (`XOp`, `YOp`, ...). So either we place hundred of lines of code in the TableGen definition or use preprocessor directives and apply it for each instantiation. Both options are not optimal. 

Moreover, the only difference between the custom and auto-generated code will be the automatic type resolution for inputs. This is a lot of code duplication. 

> Aren't custom directives meant for exactly that purpose? 

Yes, I thought so too. 

Unfortunately (as far as I know) it isn't possible to resolve operand types using custom directives (can't access the `results` vector). The only way this could work is if the custom directive parses
```
$in_qubits ( `ctrl` $pos_ctrl_in_qubits^ )? ( `nctrl` $neg_ctrl_in_qubits^ )?
```

also. Yet again, this directive would implement the same code as the auto-generator. 

**Consequence 2** 
Furthermore, a custom verifier has to be added to ensure that all qubit types are the same. 

> What about the [`AllTypesMatch`](https://mlir.llvm.org/docs/DefiningDialects/Operations/#type-inference) trait? 

The use of variadic for the input qubits unfortunately doesn't allow the use of the `AllTypesMatch` trait. The problem is basically the following: 
```mlir 
mqtref.x () %q0 : !mqtref.DynamicQubit
```

will fail since the types are
```
in_qubits: {mqtref.DynamicQubit}, pos_ctrl: {/*None*/}, neg_ctrl: {/*None*/} 
```

MatthiasReumann:
> I could very well imagine a pass that maps parts of the circuit to static qubits while others are still dynamically allocated.

This is a valid point. I was more thinking about `mqtopt` and optimizations:

```mir
// clear that these two can be cancelled
... = mqtopt.x %static0 ctrl %static1
... = mqtopt.x %static0 ctrl %static1

// what now? maybe cancel after static assignments? 
// probably required anyways 🤔
... = mqtopt.x %static0 ctrl %static1
... = mqtopt.x %static0 ctrl %dyn1
```

> Having two types (without having looked into your code so far), I am wondering whether we then also have every operation twice?

Nope. That's exactly what I wanted to avoid. It is solved via the `AnyTypeOf` trait. That's also part of the reason why we need verification and explicit typing.  

> You mean the custom assembly format, right? 

Nope. [Custom directives](https://mlir.llvm.org/docs/DefiningDialects/Operations/#directives). Essentially a function which parses (and prints) part of the assembly. Nevertheless, your arguments still hold. I also think that one should keep the amount of custom code as minimal as possible, so I think the current solution is fine. 

> Having only one qubit type, this would not be necessary, right?

Yes.

burgholzer:
> > However, I am not sure if this is enough evidence to support going forward with two separate qubit types.
> 
> Agreed. I think one advantage of having separate types is that it is inherently known if the qubit is dynamically or statically addressed. A single `QubitType` would require additional logic to figure out if it came from a `QubitOp (static)` or `ExtractOp (dynamic)`.

Yeah. Just because it came to my mind now: Could it make sense (or is it even possible) to derive an abstract `QubitType` and then simply define two derived types for the `StaticQubitType` and `DynamicQubitType`? In essence: can one make the connection between both types a little more obvious/direct? In the end, both are qubits and operations can be applied to them.

> > I agree with Yannick that programs with mixed qubit types (static and dynamic) are entirely valid during their compilation. They are just invalid as input or (final) output of the compiler.
> 
> So, if we decide to keep the two types, remove the constraint that they have to be of the same type (static or dynamic)?

Yeah, I think I would remove that constraint. Technically, such a constraint is not even present in OpenQASM 3 (or QIR, iirc). At least, OpenQASM 3 actively describes a situation where some qubits are already assigned to hardware qubits, but some are not. I just can't find that at the moment.

MatthiasReumann:
> Honestly, assuming that we have just one qubit type, I think checking whether a qubit is statically or dynamically allocated is nothing more than getting its definition and checking which operation it is. It's constant time and requires just one more call compared to having two qubit types.
> 
> In the end, I do not see the advantage two qubit types bring us and as we can see already here in the PR it makes things more complicated (which we try to avoid if possible). In the end, both are qubit (references/addresses) no matter how they were allocated.

You are probably right. 

Moreover, QIR also has just one (opaque) qubit type. It's really simply a pointer to a qubit regardless of whether it's static or dynamic. Worst case it requires an (rather trivial) analysis pass to collect static and dynamic qubits.

So, back to a single `QubitType` - do we agree on that?

burgholzer:
I suppose simplicity is key here. So let's go with the single type solution for now.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
burgholzer, line 1: Most likely, I am simply overlooking things, but isn't this conversion missing the conversion for the new static qubit addressing? Same holds for the other direction (most likely).

burgholzer (changes requested):
Thanks for the work on this @MatthiasReumann 🙏🏼 this is looking great!
I tried to resolve the merge conflicts after #1106 and really hope that I did not mess anything up. Might be worth checking.
On the code itself, I only have one inline comment.

Other than that, it would be great to add corresponding changelog entries for the renaming of the dialects and the addition of the static qubit addressing.
Then this should be ready to merge

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
MatthiasReumann: Yes, exactly. My plan was to add the `QubitOp` for the `mqtopt` dialect and the associated conversion logic (`ref <-> opt`) in a follow-up pull request.

MatthiasReumann:
> Might be worth checking. 

Besides a left-over comment and a few variables having still the `dyn` naming, looks fine. 

> Other than that, it would be great to add corresponding changelog entries for the renaming of the dialects and the addition of the static qubit addressing. 

Done.

mlir/include/mlir/Dialect/MQTRef/IR/MQTRefOps.td
```diff
@@ -218,4 +221,24 @@ def ExtractOp : ResourceOp<"extractQubit", [UniqueIndexDefinition]> {
...
+        Example:
+        ```mlir
+        // Static (hardware) qubit with address '0'.
+        %q = "mqtref.qubit"() <{index = 0 : i64}> : () -> !mqtref.Qubit
+
+        // Using custom assembly.
+        %q = mqtref.qubit 0
+        ```
+    }];
+    let arguments = (ins OptionalAttr<ConfinedAttr<I64Attr, [IntNonNegative]>>:$index);
```
ystade: Quick question: Why do we use `OptionalAttr` here? Do we allow not to specify the index and then just increment the qubit index?

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
ystade: Yes, this is a very good idea. You may want to open a new issue for that if not already happened.

mlir/test/Dialect/MQTRef/IR/dialect_features.mlir
```diff
@@ -0,0 +1,1237 @@
...
+}
+
+// -----
+// This test checks if the MeasureOp applied to a single static qubit is parsed and handled correctly.
+module {
+    // CHECK-LABEL: func.func @testMeasureOpStatic
+    func.func @testMeasureOpStatic() {
+        // CHECK: [[M0:.*]] = "mqtref.measure"(%[[ANY:.*]]) : (!mqtref.Qubit) -> i1
+
+        %q0 = mqtref.qubit 0
```
ystade, line 122: Just pointing out here that we never test to allocate a static qubit without an index. I think, in the end, we also should make the index mandatory.

ystade (approved):
@MatthiasReumann nice work here. I also looked through the files and everything looks pretty good to me. I put one question inline.

burgholzer (approved):
I agree with Yannick's comments.
Once the new issue is created and the last comment here is resolved, we can merge this 🎉

mlir/include/mlir/Dialect/MQTRef/IR/MQTRefOps.td
MatthiasReumann: First of all, yes, the `Optional` part here, makes absolutely no sense. That's a very good catch. 

Would incrementing qubit indices bring any benefit? 

Essentially we require static qubits for routing. Inspired by [Qiskit](https://quantum.cloud.ibm.com/docs/en/api/qiskit/transpiler#layout-stage), the first step will be to _"expand the circuit to have as many qubits as the target architecture"_. In other words, somewhere at the beginning of the quantum routine we will have something a long the lines:

```mlir 
// Assume 5-Qubit Architecture
%q0 = mqtref.qubit 0
%q1 = mqtref.qubit 1 
%q2 = mqtref.qubit 2 
%q3 = mqtref.qubit 3
%q4 = mqtref.qubit 4  
```

Any previous quantum register and respective `extract` ops will be replaced with the static qubits (probably a separate pass?). Given that the number of qubits of an architecture is a "compile-time constant", I'm not sure if using operands here makes sense. I might be wrong though.

ystade: Then let's remove the `optional`.

I had the same thought, and I agree with you, operands do not make sense here in my opinion.

ystade: Feel free to resolve the conversation when you removed the `optional` and then this can go in 🚀

burgholzer: I'd rather make the index explicit and just get rid of the optional.

As for the layout: layout or placement should definitely be a dedicated pass (similar to Qiskit). This is were the circuit qubits are replaced by hardware qubits after all. And there are many variations for doing that. QMAPs own strategies (which are fairly outdated) mostly treat this separately from the actual routing afterwards. Some of Qiskit's layout passes do that too.
Qiskit's default pass combines layout and routing by doing an iterative refinement of the layout through routing the circuit back and forth multiple times.

Just something to note here in addition to the above: it would be very beneficial to somewhat keep track of the chosen layout (similar to Qiskit's layout class) as metadata somewhere. This is essential for performing verification on the circuits.

However, I think none of that is immediately import here. Just good to keep in mind I suppose.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
MatthiasReumann: Created #1114

MatthiasReumann:
> This is still happening at compile time though, but not necessarily constant.

Appendix: The same arguments hold if we specify architectures with `.json` files.

burgholzer:
Merging this now. Everything else can be tackled in a follow-up PR ☺️

PR #1091, opened 2025-08-01
✨ Add MQTDyn-QIR conversions

li-mingbao:
## Description

This PR introduces conversion passes between the `mqtdyn` dialect and the `LLVM IR` that adheres to the QIR specifications for dynamic qubit allocations. Currently it mainly focuses on the conversion within MLIR between the `mqtdyn` dialect and the `llvmlir` dialect and vice versa. Once the IR consists only of llvm dialect operations it can be convert to LLVM IR with the built-in passes from the `mlir-translate` module and back.

## Checklist:

- [x] The pull request only contains commits that are focused and relevant to this change.
- [ ] I have added appropriate tests that cover the new/changed functionality.
- [ ] I have updated the documentation to reflect these changes.
- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

PR #1099, opened 2025-08-07
✨ Add translation from `QuantumComputation` to `MQTRef`

### LINKED ISSUE #1097: ✨ MLIR - Add translation from `QuantumComputation`
### Problem Statement

Currently, we only have an MLIR round trip between the `QuantumComputation` IR and our `Opt` dialect. This is not enough to generate valid MLIR output code.

### Proposed Solution

To have valid MLIR code, we should add a [translation](https://mlir.llvm.org/getting_started/Glossary/#translation) from the `QuantumComputation` IR to one of our dialects.

denialhaag:
## Description

This PR adds a translation from `QuantumComputation` to `MQTDyn`.

Fixes #1097 

## Checklist:

### DIFF
--- a/mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
+++ b/mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
@@ -52,17 +52,27 @@ mlir::Value allocateQreg(mlir::OpBuilder& builder, mlir::MLIRContext* context,
 }
 
 /**
+ * @brief Deallocates a quantum register in the MLIR module.
+ *
+ * @param builder The MLIR OpBuilder used to create operations
+ * @param quantumRegister The quantum register to deallocate
+ */
+void deallocateQreg(mlir::OpBuilder& builder, mlir::Value qreg) {
+  builder.create<mqt::ir::ref::DeallocOp>(builder.getUnknownLoc(), qreg);
+}
+
+/**
  * @brief Extracts individual qubits from a quantum register.
  *
  * @param builder The MLIR OpBuilder used to create operations
  * @param context The MLIR context in which types are created
- * @param quantumRegister The quantum register from which to extract qubits
+ * @param qreg The quantum register from which to extract qubits
  * @param numQubits The number of qubits to extract
  * @return llvm::SmallVector<mlir::Value> Vector of extracted qubit values
  */
 llvm::SmallVector<mlir::Value> extractQubits(mlir::OpBuilder& builder,
                                              mlir::MLIRContext* context,
-                                             mlir::Value quantumRegister,
+                                             mlir::Value qreg,
                                              const std::size_t numQubits) {
   const auto& qubitType = mqt::ir::ref::QubitType::get(context);
   llvm::SmallVector<mlir::Value> qubits;
@@ -72,7 +82,7 @@ llvm::SmallVector<mlir::Value> extractQubits(mlir::OpBuilder& builder,
   for (std::size_t qubit = 0; qubit < numQubits; ++qubit) {
     auto indexAttr = builder.getI64IntegerAttr(static_cast<int64_t>(qubit));
     auto extractOp = builder.create<mqt::ir::ref::ExtractOp>(
-        loc, qubitType, quantumRegister, nullptr, indexAttr);
+        loc, qubitType, qreg, nullptr, indexAttr);
     qubits.emplace_back(extractOp.getResult());
   }
 
@@ -82,10 +92,10 @@ llvm::SmallVector<mlir::Value> extractQubits(mlir::OpBuilder& builder,
 /**
  * @brief Adds a single quantum operation to the MLIR module.
  *
- * @tparam OpType The type of the operation to create.
- * @param builder The MLIR OpBuilder.
- * @param operation The quantum operation to add.
- * @param qubits The qubits of the quantum register.
+ * @tparam OpType The type of the operation to create
+ * @param builder The MLIR OpBuilder
+ * @param operation The quantum operation to add
+ * @param qubits The qubits of the quantum register
  */
 template <typename OpType>
 void addOperation(mlir::OpBuilder& builder, const qc::Operation& operation,
@@ -128,9 +138,9 @@ void addOperation(mlir::OpBuilder& builder, const qc::Operation& operation,
 /**
  * @brief Adds a measure operation to the MLIR module.
  *
- * @param builder The MLIR OpBuilder.
- * @param operation The measure operation to add.
- * @param qubits The qubits of the quantum register.
+ * @param builder The MLIR OpBuilder
+ * @param operation The measure operation to add
+ * @param qubits The qubits of the quantum register
  */
 void addMeasureOp(mlir::OpBuilder& builder, const qc::Operation& operation,
                   const llvm::SmallVector<mlir::Value>& qubits) {
@@ -146,9 +156,9 @@ void addMeasureOp(mlir::OpBuilder& builder, const qc::Operation& operation,
 /**
  * @brief Adds a reset operation to the MLIR module.
  *
- * @param builder The MLIR OpBuilder.
- * @param operation The reset operation to add.
- * @param qubits The qubits of the quantum register.
+ * @param builder The MLIR OpBuilder
+ * @param operation The reset operation to add
+ * @param qubits The qubits of the quantum register
  */
 void addResetOp(mlir::OpBuilder& builder, const qc::Operation& operation,
                 const llvm::SmallVector<mlir::Value>& qubits) {
@@ -264,6 +274,8 @@ mlir::OwningOpRef<mlir::ModuleOp> translateQuantumComputationToMLIR(
     // Even if operations fail, return the module with what we could translate
     emitError(loc) << "Failed to translate some quantum operations";
   }
+
+  deallocateQreg(builder, qreg);
 
   // Create terminator
   builder.create<mlir::func::ReturnOp>(loc);
--- a/mlir/unittests/translation/test_translation.cpp
+++ b/mlir/unittests/translation/test_translation.cpp
@@ -99,7 +99,7 @@ TEST_F(ImportTest, EntryPoint) {
   ASSERT_TRUE(checkOutput(checkString, outputString));
 }
 
-TEST_F(ImportTest, Allocation) {
+TEST_F(ImportTest, AllocationAndDeallocation) {
   const qc::QuantumComputation qc(3);
 
   auto module = translateQuantumComputationToMLIR(context.get(), qc);
@@ -110,6 +110,7 @@ TEST_F(ImportTest, Allocation) {
     CHECK: "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 0 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
     CHECK: "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 1 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
     CHECK: "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 2 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: "mqtref.deallocQubitRegister"(%[[Reg]]) : (!mqtref.QubitRegister) -> ()
   )";
 
   ASSERT_TRUE(checkOutput(checkString, outputString));
@@ -741,6 +742,7 @@ TEST_F(ImportTest, GHZ) {
     CHECK: %[[M0:.*]] = mqtref.measure %[[Q0]]
     CHECK: %[[M1:.*]] = mqtref.measure %[[Q1]]
     CHECK: %[[M2:.*]] = mqtref.measure %[[Q2]]
+    CHECK: "mqtref.deallocQubitRegister"(%[[Reg]]) : (!mqtref.QubitRegister) -> ()
   )";
 
   ASSERT_TRUE(checkOutput(checkString, outputString));


denialhaag:
I am locally testing the translation with the blob below. The goal is that a valid MLIR representation of the defined circuit is printed. 

```
#include "ir/QuantumComputation.hpp"
#include "mlir/Dialect/MQTDyn/IR/MQTDynDialect.h"
#include "mlir/Dialect/MQTDyn/Translation/ImportQuantumComputation.h"

#include <mlir/IR/BuiltinOps.h>
#include <mlir/IR/MLIRContext.h>

int main() {
  qc::QuantumComputation qc(2);
  qc.h(0);
  qc.cx(0, 1);
  qc.print(std::cout);

  mlir::MLIRContext context;
  mlir::DialectRegistry registry;

  registry.insert<mqt::ir::dyn::MQTDynDialect>();
  context.appendDialectRegistry(registry);
  context.loadAllAvailableDialects();

  auto module = translateQuantumComputationToMLIR(context, qc);

  module->print(llvm::outs());

  return 0;
}
```

ystade:
For the wrapping of the quantum computation with a `main` function required for a valid QIR program, the code here https://github.com/li-mingbao/core/blob/545b66116222e1e665beec4cabb8728804eb3ccc/mlir/lib/Conversion/MQTDynToQIR/MQTDynToQIR.cpp#L322-L337
could also be interesting. In general, the outcome of #1091 is highly related to this task as well.

CMakeLists.txt
```diff
@@ -111,6 +111,9 @@ if(BUILD_MQT_CORE_BENCHMARKS)
   add_subdirectory(eval)
 endif()
 
+# FIXME: Remove this option before merging
+add_compile_options(-Wno-dangling-assignment-gsl)
+
```
denialhaag: Remove this again before merging.

```suggestion
```

mlir/lib/Dialect/MQTDyn/Translation/ImportQuantumComputation.cpp
```diff
@@ -0,0 +1,200 @@
...
+  for (const auto& operation : quantumComputation) {
+    if (operation->getType() == qc::OpType::H) {
+      addOperation<mqt::ir::dyn::HOp>(builder, *operation, qubits);
+    } else if (operation->getType() == qc::OpType::RX) {
+      addOperation<mqt::ir::dyn::RXOp>(builder, *operation, qubits);
+    } else if (operation->getType() == qc::OpType::SWAP) {
+      addOperation<mqt::ir::dyn::SWAPOp>(builder, *operation, qubits);
+    } else if (operation->getType() == qc::OpType::X) {
+      addOperation<mqt::ir::dyn::XOp>(builder, *operation, qubits);
+    }
```
denialhaag: Currently, only these four operations are supported, but it will be easy to extend this. I will do this when adding unit tests.

burgholzer:
> If I understand the [MLIR docs](https://mlir.llvm.org/getting_started/TestingGuide/#c-unit-tests) correctly, the tests for this feature should go in `mlir/unittests`.

That seems about right: https://github.com/llvm/llvm-project/tree/main/mlir/unittests
There don't seem to be that many tests (compared to the FileCheck ones) but i suppose it makes sense to have a look into how they are setting this up.
For example: https://github.com/llvm/llvm-project/blob/main/mlir/unittests/Transforms/DialectConversion.cpp or https://github.com/llvm/llvm-project/blob/main/mlir/unittests/Target/LLVM/SerializeToLLVMBitcode.cpp

mlir/lib/Dialect/MQTDyn/Translation/ImportQuantumComputation.cpp
denialhaag: This comment is outdated. Resolving the conversation.

ystade:
> If I understand the [MLIR docs](https://mlir.llvm.org/getting_started/TestingGuide/#c-unit-tests) correctly, the tests for this feature should go in `mlir/unittests`.

Probably already too late and you figured that out already but just for the full picture: Xanadu has set up also some unit tests in Catalyst, see for an example here: https://github.com/PennyLaneAI/catalyst/blob/main/mlir/unittests/Example/Example.cpp

burgholzer:
@denialhaag okay, I think I have experimented enough with this and would now hand this back to you.
I iterated all of the code and the CMake setup around it.
I think I like the FileCheck+Googletest setup. At least it felt way closer to home than the FileCheck+lit setup.
Please look through all the changes and see if they make sense. If so, then adding all the missing operations would be a natural next step. 
- For the quantum operations that should be moderately easy.
- Then, we are still missing the `measure` operation as well as the `reset` operation (which will be introduced in #1106).
- Additionally, the `ClassicControlledOperation` from MQT Core could also be supported directly I suppose. It basically translates to an `IfElseOp`. we can also tackle this in a follow-up PR and create an issue for that in the meantime.
- MQT Core IR also has the `CompoundOperation`, which is simply a collection of other operations. I am not yet 100% sure what the right concept for that is in MLIR, but we can also tackle this in a follow-up PR and create an issue for that in the meantime.
- Last, but not least, MQT Core IR also has symbolic operations, where the rotation angles are not known at compile time. This would require us to write translations for the Expression class to classical MLIR code. I'd propose to also open an issue for this and tackle this in a follow-up

Hope that makes sense.

mlir/lib/Dialect/MQTDyn/Translation/CMakeLists.txt
```diff
@@ -6,6 +6,9 @@
 #
 # Licensed under the MIT License
 
+# FIXME: This seems to be needed for the build to succeed
+add_compile_options(-fexceptions)
+
```
denialhaag: @burgholzer, I needed to add this again for the build to succeed (see also [the CI](https://github.com/munich-quantum-toolkit/core/actions/runs/16921464000/job/47948367235)). I understand that you already added the compile option below using `target_compile_options()`, but it looks like something is wrong. Do you know what the issue could be? 🤔

burgholzer: Classic.. I obviously tested with one of the combinations that turns out to work 🙄 
In case this causes real problems, just replace the `target_compile_options()` with the general `add_compile_options`. This was just another try to reduce the use of global CMake constructs.

denialhaag: Sounds good! Done.

denialhaag:
> @denialhaag okay, I think I have experimented enough with this and would now hand this back to you. I iterated all of the code and the CMake setup around it. I think I like the FileCheck+Googletest setup. At least it felt way closer to home than the FileCheck+lit setup. Please look through all the changes and see if they make sense. If so, then adding all the missing operations would be a natural next step.

Thanks a lot for the improvements! I agree that the `FileCheck` setup is quite nice and also doesn't require a lot of overhead.

> * For the quantum operations that should be moderately easy.

It was! Done.

> * Then, we are still missing the `measure` operation as well as the `reset` operation (which will be introduced in [♻️ MLIR - Update Measure Functionalities  #1106](https://github.com/munich-quantum-toolkit/core/pull/1106)).

Done.

> * Additionally, the `ClassicControlledOperation` from MQT Core could also be supported directly I suppose. It basically translates to an `IfElseOp`. we can also tackle this in a follow-up PR and create an issue for that in the meantime.

I created #1109. I'm happy to work on this next if it makes sense!

> * MQT Core IR also has the `CompoundOperation`, which is simply a collection of other operations. I am not yet 100% sure what the right concept for that is in MLIR, but we can also tackle this in a follow-up PR and create an issue for that in the meantime.

I created #1110.

> * Last, but not least, MQT Core IR also has symbolic operations, where the rotation angles are not known at compile time. This would require us to write translations for the Expression class to classical MLIR code. I'd propose to also open an issue for this and tackle this in a follow-up

I created #1111.

denialhaag:
> 0. If I have time, I have to have a closer look on how you use `FileCheck` from unit tests. We tried that in the beginning because unit tests integrated much better into our framework than the lit tests that we are using currently for the MLIR part. So, I find that very interesting. Maybe you can already give your opinion on whether all present lit tests could be rewritten in a sensible way to unit tests. The ideal solution would be to still keep the `.mlir` files and build unit test out of them. Maybe you have some insights to share here.

I can have a look at this tomorrow. I think it should be possible in principle. As I also just said in the other comment, the configuration is rather simple and doesn't require a lot of overhead.

> 1. We had that in other places already, where we had to copy & paste code for the several gates. Often, this could be slightly more concise by using macros, like here. Perhaps that would be something for [here](https://github.com/munich-quantum-toolkit/core/blob/17a39782eb2499dacd1c00ec1b2473e965f6dbdd/mlir/lib/Conversion/MQTDynToMQTOpt/MQTDynToMQTOpt.cpp#L13) as well.

Thanks! I'll have a look at the macro tomorrow!

mlir/lib/Dialect/MQTDyn/Translation/ImportQuantumComputation.cpp
```diff
@@ -0,0 +1,331 @@
...
+void addMeasureOp(mlir::OpBuilder& builder, const qc::Operation& operation,
+                  const llvm::SmallVector<mlir::Value>& qubits) {
+  const auto& bitType = mlir::IntegerType::get(builder.getContext(), 1);
+  const auto& targets = operation.getTargets();
+  for (const auto& target : targets) {
+    const mlir::Value inQubit = qubits[target];
+    builder.create<mqt::ir::dyn::MeasureOp>(builder.getUnknownLoc(), bitType,
+                                            inQubit);
+  }
+}
```
burgholzer: I am not quite sure that this is correct yet as it is.
In the context of the `QuantumComputation`, there exist classical registers into which the measurements put their classical results.
these are addressed very similarly to quantum registers.
Right now, the code would generate a new classical value for each measurement.
This would lead to problems if a program measures into the same bit of a classical register multiple times.
Simply imagine the (non-sense) computation
```
qc.addQubitRegister(2);
qc.addClassicalRegister(1);
qc.measure(0, 0);
qc.measure(1, 0);
```

Right now, this will be realised as two classical results.

The second we talk about ClassicalControlledOperations, it would have also become obvious that we currently do not have a way to reference a classical result from one of the registers in the QuantumComputation when defining the condition for the `if` instruction.
The classic-controlled operations have essentially two modes:
- bit == true|false
- register == integer value

Overall, we have to find a way to handle classical registers defined as part of the QuantumComputation. I am fairly sure that there are already some best practices on how to allocate these classical entities.

mlir/unittests/translation/test_translation.cpp
burgholzer, line 1: I think it could make sense to add at least one test that uses more than two gates and ensures that the sequence of operations is correct (who know's 🤷🏼).

burgholzer (approved):
@denialhaag thanks for the additions here. This looks pretty good already!

This will have to be slightly updated after #1098 (basically replacing `Dyn` with `Ref`.

There is one small comment in the tests where I see a small opening for adding a couple more tests (or at least one).

There is one bigger open issue regarding how to deal with measurements and classical outputs of quantum computations.
We might even want to postpone that for a follow-up PR in order to get this merged now.

mlir/lib/Dialect/MQTDyn/Translation/ImportQuantumComputation.cpp
ystade: I think, we actually have to refactor our measure operation. In the `opt` dialect, the current behaviour is fine, because we can `insert` the returned classical value into a register, i.e., if we want that, we can also treat the measured classical results in value semantics. However, in the `MQTRef` dialect here, there is nothing like the `insert` operation. Probably, the measure operation must also take a reference to a classical bit besides the qubit reference and not return anything. This is similar to the measure operation in QIR (in some code examples at least):
```llvm
call void @__quantum__qis__mz__body(ptr %9, ptr %11)
```

burgholzer: Having the measure operation take a reference to a bit makes a lot of sense to me. That way, one can allocate the bits at the beginning of the program and then reference them in the measurement operations.
What's the appropriate type for the register? Array of i1? It's definitely dense and should be indexable. Is there any precedent for that in either QIR or regular MLIR or LLVM?

burgholzer:
#1098 has just been merged and renamed `MQTDyn` to `MQTRef`. This also needs to be reflected here, but should hopefully not cause too much disruption.

mlir/unittests/translation/test_translation.cpp
denialhaag: Makes sense! Let me know what you think.

I also added a test that the entry point exists. Hope that makes sense.

mlir/lib/Dialect/MQTDyn/Translation/ImportQuantumComputation.cpp
denialhaag: You are right; I implemented the `MeasureOp` without caring about the result for now.

I also like the idea of a classical register.

I checked the QIR specification a bit, but they don't seem to make strong statements about [measurement results](https://github.com/qir-alliance/qir-spec/blob/main/specification/v0.1/1_Data_Types.md#measurement-results).

In the MLIR docs, I only saw that `DenseI1ArrayAttr` apparently [doesn't exist](https://mlir.llvm.org/docs/Dialects/Builtin/#densearrayattr). They even say that "[b]ool elements are stored as bytes."

burgholzer: I think the key takeaway from the QIR documentation is that they define a separate opaque data type for the result of a measurement.
Catalyst also seems to do this in their Quantum dialect: https://github.com/PennyLaneAI/catalyst/blob/90e5d0bc3bcff98e51c55ca217fb3519d7b8be8e/mlir/include/Quantum/IR/QuantumDialect.td#L72-L74
They convert this to QIR here: https://github.com/PennyLaneAI/catalyst/blob/90e5d0bc3bcff98e51c55ca217fb3519d7b8be8e/mlir/lib/Quantum/Transforms/quantum_to_llvm.cpp#L65-L69

Their `MeasurementOp` does return an `i1` though: https://github.com/PennyLaneAI/catalyst/blob/90e5d0bc3bcff98e51c55ca217fb3519d7b8be8e/mlir/include/Quantum/IR/QuantumOps.td#L900-L918

I don't quite see yet, how this matches up.

denialhaag: Not 100 percent sure, but this pattern also seems relevant: https://github.com/PennyLaneAI/catalyst/blob/90e5d0bc3bcff98e51c55ca217fb3519d7b8be8e/mlir/lib/Quantum/Transforms/ConversionPatterns.cpp#L780

denialhaag: This discussion has been moved to #1115.

I'm resolving this thread.

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
```diff
@@ -0,0 +1,276 @@
...
+      ADD_OP_CASE(RXX)
+      ADD_OP_CASE(RYY)
+      ADD_OP_CASE(RZZ)
+      ADD_OP_CASE(RZX)
+    case qc::OpType::XXminusYY:
+      addOperation<mqt::ir::ref::XXminusYY>(builder, *operation, qubits);
+      break;
+    case qc::OpType::XXplusYY:
+      addOperation<mqt::ir::ref::XXplusYY>(builder, *operation, qubits);
+      break;
```
burgholzer: I am very sure I am overlooking things, but why are these two operations not covered by a similar macro?

mlir/unittests/translation/test_translation.cpp
```diff
@@ -0,0 +1,747 @@
...
+  sm.AddNewSourceBuffer(std::move(outputBuffer), llvm::SMLoc());
+  return fc.checkInput(sm, outputBufferBuffer);
+}
+
+} // namespace
+
+using namespace qc;
+
+TEST_F(ImportTest, EntryPoint) {
+  const qc::QuantumComputation qc(1);
```
burgholzer: couldn't this test even use `0` qubits? so essentially
```suggestion
  const qc::QuantumComputation qc{};
```

mlir/unittests/translation/test_translation.cpp
```diff
@@ -0,0 +1,747 @@
...
+  const qc::QuantumComputation qc(3);
+
+  auto module = translateQuantumComputationToMLIR(context.get(), qc);
+
+  const auto outputString = getOutputString(&module);
+  const auto* checkString = R"(
+    CHECK: %[[Reg:.*]] = "mqtref.allocQubitRegister"() <{size_attr = 3 : i64}> : () -> !mqtref.QubitRegister
+    CHECK: "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 0 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 1 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 2 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
```
burgholzer, line 112: I only now noticed this while carefully reading through the tests. The current setup is missing one crucial thing and that is the deallocation of the qubit register just before the return statement.
This could be considered boilerplate code, but it is essential in my opinion to form a valid and clean program.

mlir/unittests/translation/test_translation.cpp
```diff
@@ -0,0 +1,747 @@
...
+
+  auto module = translateQuantumComputationToMLIR(context.get(), qc);
+
+  const auto outputString = getOutputString(&module);
+  const auto* checkString = R"(
+    CHECK: %[[Reg:.*]] = "mqtref.allocQubitRegister"() <{size_attr = 2 : i64}> : () -> !mqtref.QubitRegister
+    CHECK: %[[Q_0:.*]] = "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 0 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: %[[Q_1:.*]] = "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 1 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: "mqtref.measure"(%[[Q_0]]) : (!mqtref.Qubit) -> i1
+    CHECK: "mqtref.measure"(%[[Q_1]]) : (!mqtref.Qubit) -> i1
```
burgholzer: This does not quite look like the right check to me yet. Shouldn't the measurement return an `%i1`? can we also include that return value here in the check? holds for all measurement tests.

mlir/unittests/translation/test_translation.cpp
```diff
@@ -0,0 +1,747 @@
...
+    CHECK: %[[Reg:.*]] = "mqtref.allocQubitRegister"() <{size_attr = 3 : i64}> : () -> !mqtref.QubitRegister
+    CHECK: %[[Q_0:.*]] = "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 0 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: %[[Q_1:.*]] = "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 1 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: %[[Q_2:.*]] = "mqtref.extractQubit"(%[[Reg]]) <{index_attr = 2 : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit
+    CHECK: mqtref.h() %[[Q_0]]
+    CHECK: mqtref.x() %[[Q_1]] ctrl %[[Q_0]]
+    CHECK: mqtref.x() %[[Q_2]] ctrl %[[Q_0]]
+    CHECK: "mqtref.measure"(%[[Q_0]]) : (!mqtref.Qubit) -> i1
+    CHECK: "mqtref.measure"(%[[Q_1]]) : (!mqtref.Qubit) -> i1
+    CHECK: "mqtref.measure"(%[[Q_2]]) : (!mqtref.Qubit) -> i1
```
burgholzer: Just looking at this triggered the thought: do we already define a custom assembly format for the measurement and reset operations so that they can be written more compactly?
I would have expected the check string to look something like this 
```
%[[C_0]] = mqtref.measure %[[Q_0]]
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add translation from `QuantumComputation` to the `MQTRef` MLIR dialect ([#1099]) ([**@denialhaag**])
```
```suggestion
- ✨ Add translation from `QuantumComputation` to the `MQTRef` MLIR dialect ([#1099]) ([**@denialhaag**], [**@burgholzer**])
```
Hope you do not mind that. I think it helps to keep track of the code owners.

burgholzer (changes requested):
Thanks for the additional changes here.
This is extremely close to being mergeable. However, my devil's advocate checks revealed some small further points to address.
After that, we should be done here.

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
denialhaag: `XXminusYY` and `XXplusYY` are the only operations in our dialects not ending in `Op`, which the macro is appending. I already thought this might be a mistake. Shall I fix that here or in a separate PR?

CHANGELOG.md
denialhaag: Not at all, I should have done that myself!

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
burgholzer: Certainly an oversight.
Might as well fix this here and add a separate Changelog entry.

mlir/unittests/translation/test_translation.cpp
denialhaag: Good point. Done.

mlir/unittests/translation/test_translation.cpp
denialhaag: I defined an assembly format for `mqtref.measure` now, and the check string looks like you say.

That said, I realized that all operations besides the `UnitaryOp`s are missing assembly formats. We should probably define those rather soon, before updating all the tests becomes too big of a headache. 🤔

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
denialhaag: Done.

mlir/unittests/translation/test_translation.cpp
denialhaag: Ahh, I had forgotten about that. Done.

mlir/include/mlir/Dialect/MQTRef/IR/MQTRefOps.td
```diff
@@ -134,7 +138,7 @@ def ResetOp : GateOp<"reset", []> {
 
         Example:
         ```mlir
-        "mqtdyn.reset"(%0) : (!mqtopt.Qubit) -> ()
+        "mqtref.reset"(%0) : (!mqtopt.Qubit) -> ()
```
burgholzer, line 141: Shouldn't the reset op also get a custom assembly format?

PR #1100, opened 2025-08-08
✨ Support more Properties in NA QDMI Device

ystade:
## Description

#996 implemented the first version of MQT's NA QDMI Device. This PR adds support for more properties to the device. In particular, it enables querying operations for the supported sites. Meanwhile, it uses the newly introduced zone sites of QDMI.

## Checklist:


include/mqt-core/na/device/Generator.hpp
```diff
@@ -262,16 +251,83 @@ struct Device {
   };
 
   /// @brief The unit of measurement for lengths in the device.
-  Unit lengthUnit;
+  Unit lengthUnit = {1.0, "um"}; ///< Default is micrometers (um).
   /// @brief The unit of measurement for time in the device.
-  Unit timeUnit;
+  Unit durationUnit = {1.0, "us"}; ///< Default is microseconds (us).
 
+  // Before we used the macro NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT here,
```
burgholzer, line 262: Given how much boilerplate code this has generated, would there by any downside to simply having the ID be part of the input JSON?

src/na/device/Device.cpp
```diff
@@ -279,29 +321,166 @@ auto MQT_NA_QDMI_Site_impl_d::queryProperty(const QDMI_Site_Property prop,
...
-auto MQT_NA_QDMI_Operation_impl_d::isShuttling(const Type type) -> bool {
-  switch (type) {
-  case Type::ShuttlingLoad:
-  case Type::ShuttlingMove:
-  case Type::ShuttlingStore:
-    return true;
-  default:
-    return false;
-  }
+MQT_NA_QDMI_Operation_impl_d::MQT_NA_QDMI_Operation_impl_d(
```
burgholzer, line 347: CLion is pointing out here that the `numQubits` argument here will always be `1`, which makes the argument redundant. However, removing it creates a clash with the constructor intended for the `load` instruction. The only difference between both is that `load` does not set the number of qubits. However, isn't the number of qubits for a load instruction also `1`? If so, then one could eliminate one of the constructors here.

src/na/device/Device.cpp
```diff
@@ -310,23 +489,73 @@ auto MQT_NA_QDMI_Operation_impl_d::queryProperty(
...
+    } // this device does not support operations with more than two qubits
+  }
   ADD_STRING_PROPERTY(QDMI_OPERATION_PROPERTY_NAME, name_.c_str(), prop, size,
                       value, sizeRet)
   ADD_SINGLE_VALUE_PROPERTY(QDMI_OPERATION_PROPERTY_PARAMETERSNUM, size_t,
                             numParameters_, prop, size, value, sizeRet)
-  if (type_ != Type::ShuttlingMove) {
-    ADD_SINGLE_VALUE_PROPERTY(QDMI_OPERATION_PROPERTY_DURATION, double,
-                              duration_, prop, size, value, sizeRet)
+  const std::vector sitesVec(supportedSites_.cbegin(), supportedSites_.cend());
```
burgholzer: Why the copy here and not simply pass `supportedSites_` to the macro?

src/na/device/Generator.cpp
```diff
@@ -118,24 +82,109 @@ auto writeName(const Device& device, std::ostream& os) -> void {
...
      << "UL\n";
 }
 
+/**
+ * @brief Writes the length unit from the device object.
+ * @param device is the device object containing the length unit.
+ * @param os is the output stream to write the length unit to.
+ */
+auto writeLengthUnit(const Device& device, std::ostream& os) -> void {
+  os << "#define INITIALIZE_LENGTHUNIT(var) var = {\"" << device.lengthUnit.unit
```
burgholzer, line 98: Truly a nitpick, but the ordering of the writes seems inconsistent to me. Here, the unit is written as "unit, scalefactor", while in the respective class of the device the order is "scalefactor, unit". Not the end of the world, but I am a big fan of consistency where easily possible.

burgholzer (approved):
Looked through everything here and this really looks good.
I added a couple of minor things directly. That leaves just four small comments to be addressed (or ignored 😉).
Then this can go in.
Since the device has not been released yet, when preparing the next release, we should consolidate the changelog entries a little bit and merge them where appropriate.

include/mqt-core/na/device/Generator.hpp
ystade: I agree, the id could go into the JSON file. Then the validation code for the units remain. I thought I can implement the validation diffrerently but I couldn't. I propose to still keep the current solution even if it adds boilerplate code that is copied from the expanded macro.

src/na/device/Device.cpp
ystade: A load operation can load arbitrary many qubits at the same time. Or from a different perspective, numQubits denotes the number of operands, since a load does not perform a gate on a qubit, the notion numQubits does not really make sense and should not be supported in my opinion.

src/na/device/Device.cpp
ystade: That's a leftover from times where the supported sites where a different data-structure.

src/na/device/Device.cpp
burgholzer: Yeah. Guess that makes sense. Resolving for now.

include/mqt-core/na/device/Generator.hpp
burgholzer: Let's keep it then. 👍🏼

burgholzer (approved):
Let's get this in for now 🚀 

(I don't particularly think the order of the changelog is that essential. We will probably shift some things around before the next release anyway)

PR #1106, opened 2025-08-11
:recycle: MLIR - Update Measure Functionalities 

### LINKED ISSUE #1092: ✨ MLIR - Support for Reset operation
### Problem Statement

Both of our MLIR dialects currently do not support a `Reset` operation.
We do support such an operation for our `QuantumComputation` IR.

### Proposed Solution

This should be fairly straightforward to add. It probably deserves its own Operation, because, as opposed to other single-target operations, it is non-unitary and may not be controlled or inverted. In that sense, it is more similar to the `Measure` operation.
It might make sense to introduce a tag for unitary operations to distinguish between unitary and non-unitary operations.

Besides the operation itself, conversions between the dialects as well as conversions from/to `QuantumComputation` should be added.

DRovara:
## Description

This PR updates the existing measure functionalities in `MQTDyn`/`MQTOpt` by:
- Changing the `measure` signatures so that a single measurement can only ever measure a single qubit.
- Adding a new `reset` operation to reset qubit values and adding the corresponding conversion patterns

Resets are executed as shown by this example:

```mlir
%r0 = "mqtopt.allocQubitRegister"() <{size_attr = 1 : i64}> : () -> !mqtopt.QubitRegister
%r1, %q0 = "mqtopt.extractQubit"(%r0) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister) -> (!mqtopt.QubitRegister, !mqtopt.Qubit)
%q1 = "mqtopt.reset"(%q0) : (!mqtopt.Qubit) -> (!mqtopt.Qubit)
%r2 = "mqtopt.insertQubit"(%r1, %q1) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
"mqtopt.deallocQubitRegister"(%r2) : (!mqtopt.QubitRegister) -> ()
```

Fixes #1092

## Checklist:


mlir/lib/Conversion/MQTDynToMQTOpt/MQTDynToMQTOpt.cpp
```diff
@@ -238,32 +238,56 @@ struct ConvertMQTDynMeasure final
...
-    const auto& dynQubits = op.getInQubits();
+    const auto& dynQubit = op.getInQubit();
 
     // get the latest opt qubit from the map and add them to the vector
-    std::vector<Value> optQubits;
-    for (const auto& dynQubit : dynQubits) {
-      optQubits.emplace_back(getState().qubitMap[dynQubit]);
-    }
+    const Value optQubit = getState().qubitMap[dynQubit];
     // create the result types
```
burgholzer, line 245: this comment seems superfluous now.

mlir/lib/Conversion/MQTDynToMQTOpt/MQTDynToMQTOpt.cpp
```diff
@@ -238,32 +238,56 @@ struct ConvertMQTDynMeasure final
...
+
+    // prepare result type
+    const auto& qubitType = opt::QubitType::get(rewriter.getContext());
+
+    const auto& dynQubit = op.getInQubit();
+
+    // get the latest opt qubit from the map and add them to the vector
+    const Value optQubit = getState().qubitMap[dynQubit];
+    // create the result types
+
```
burgholzer: superfluous comment
```suggestion
```

mlir/lib/Conversion/MQTOptToMQTDyn/MQTOptToMQTDyn.cpp
```diff
@@ -136,20 +136,20 @@ struct ConvertMQTOptMeasure final : OpConversionPattern<opt::MeasureOp> {
...
+    const auto& newBit = mqtdynOp.getOutBit();
 
     // concatenate the dyn qubits and the bits
     SmallVector<Value> newValues;
-    newValues.reserve(dynQubits.size() + newBits.size());
-    newValues.append(dynQubits.begin(), dynQubits.end());
-    newValues.append(newBits.begin(), newBits.end());
+    newValues.reserve(2);
+    newValues.emplace_back(dynQubit);
+    newValues.emplace_back(newBit);
```
burgholzer: Couldn't these lines be combined to one?
Something like 
```c++
SmallVector<Value> newValues{dynQubit, newBit};

File: mlir/lib/Conversion/MQTOptToMQTDyn/MQTOptToMQTDyn.cpp [hunk 4]
```diff
@@ -158,6 +158,25 @@ struct ConvertMQTOptMeasure final : OpConversionPattern<opt::MeasureOp> {
...
+                  ConversionPatternRewriter& rewriter) const override {
+
+    const auto& dynQubit = adaptor.getInQubit();
+
+    // create new operation
+    auto mqtdynOp = rewriter.create<dyn::ResetOp>(op.getLoc(), dynQubit);
+
+    // replace the results of the old operation with the new results and delete
+    // old operation
+    rewriter.replaceOp(op, mqtdynOp);
```
burgholzer: IIRC, this can be simplified to a one liner using `rewriter.replaceOpWithNewOp` similar to how this is handled for the Extract operation above. I might be mistaken though.

burgholzer (changes requested):
Thanks for the changes here @DRovara 🙏🏼 
This is looking great already. I left some inline comments on potential smaller code refactors. Should be pretty straight forward hopefully. The tests still seem to be failing here, so I suspect this will still need some kind of fix. Afterwards, this should be good to go.

A changelog entry could be useful here.

mlir/lib/Conversion/MQTOptToMQTDyn/MQTOptToMQTDyn.cpp
DRovara: As far as I understand, `replaceOpWithNewOp` can only be used if the return types are the same for the original and new operation. This is not the case when converting resets, so I don't think it works.

That being said, `replaceOp` was the reason why the test cases were failing anyways and as it turns out, just calling `eraseOp` on the original operation works too, so I changed the `replaceOp` calls to that.

DRovara:
Thanks for the feedback! Sorry about the stray comments, I rushed through updating the conversion patterns without thinking too much about how the changes would influence the validity of the comments.
I checked them again and they should be fine now. 

As the issue that caused the checks to fail seems to be resolved now, I guess we can set this PR to automerge now.

PR #990, opened 2025-06-10
WIP: Support for `IfElseOp` operation

lavanya-m-k:
## Description

- Introduced a new IfElseOperation class representing conditional branches with optional register or bit comparisons
- Modified the QASM3 importer to generate a single IfElseOperation when an if statement has both then and else blocks
- Extended operation types with an IfElse entry and added string conversions for it
- Updated parser tests to match the new single-if-else output format

Fixes #923  
## Checklist:

- [x] The pull request only contains commits that are focused and relevant to this change.
- [ ] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

PR #1117, opened 2025-08-13
✨ Support Qiskit's `IfElseOp`

### LINKED ISSUE #923: ✨ Support for Qiskit's `IfElseOp`

It's been a while since Qiskit has introduced ["Classical feedforward and control flow"](https://qiskit.qotlabs.org/guides/classical-feedforward-and-control-flow), which matches OpenQASM 3's ["Looping and branching"](https://openqasm.com/language/classical.html#looping-and-branching).

Within the MQT, support for these dynamic circuit primitives is still limited.
We support mid-circuit measurements, reset operations, as well as a limited set of _classically-controlled_ operations, which roughly match the capabilities of OpenQASM 2's `if` statement (see https://arxiv.org/abs/1707.03429v2).

The goal of this issue is to add proper support for Qiskit's `IfElseOp` to the MQT Core IR so that Qiskit circuits using such instructions can be imported into the MQT, used within it, as well as exported back to Qiskit.
This requires 
- adapting the MQT Core IR to be capable of properly representing an `if (...) { ... } else { ... }` construct
- adapting the Qiskit import and export routines to handle the newly-supported operation
- adapting the OpenQASM3 parser and exporter to handle the newly supported operation.
- refactoring any dependant code to use the new functionality.

Any solution to this issue may take inspiration from the way Qiskit is handling these operations. Specifically, the [`ControlFlowOp` class](https://github.com/Qiskit/qiskit/blob/511ecbf74e5f251e5d1e0607e646e132c9d87d5b/qiskit/circuit/controlflow/control_flow.py#L28) as well as the [`IfElseOp` class](https://github.com/Qiskit/qiskit/blob/511ecbf74e5f251e5d1e0607e646e132c9d87d5b/qiskit/circuit/controlflow/if_else.py#L42).
It probably makes sense to replicate a similar operation hierarchy in MQT Core.
So instead of
> Operation --> ClassicControlledOperation

it is probably reasonable to refactor this to
> Operation --> ControlFlowOperation --> IfElseOperation

For the purpose of this PR, it is fine if the `IfElseOperation` only supports conditions in the form they are currently supported throughout MQT Core, that is
- either a full classical register is compared to an integer, or
- a single bit is compared against a Boolean value.

### Where to find the respective code?

The classically-controlled operations are currently defined here: https://github.com/munich-quantum-toolkit/core/blob/main/include/mqt-core/ir/operations/ClassicControlledOperation.hpp and contain the following members
https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/include/mqt-core/ir/operations/ClassicControlledOperation.hpp#L132-L136

They are exposed as utility functions in the `QuantumComputation` class
https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/include/mqt-core/ir/QuantumComputation.hpp#L336-L362

In a similar fashion, they are exposed to Python 
- https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/python/ir/operations/register_classic_controlled_operation.cpp#L22-L81
- https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/python/ir/register_quantum_computation.cpp#L421-L473

Neither the Qiskit import nor the export handles that type of gate yet:
- https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/mqt/core/plugins/qiskit/qiskit_to_mqt.py#L113-L171
- https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/mqt/core/plugins/qiskit/mqt_to_qiskit.py#L267-L269

The OpenQASM 3 importer breaks down if-else statements into two separate `ClassicControlledOperation`s: https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/qasm3/Importer.cpp#L892-L929
The corresponding exporter works for the limited scope of the `ClassicControlledOperation`: https://github.com/munich-quantum-toolkit/core/blob/d88d1f091b1257d8b7b4ae48521f19ed60602b76/src/ir/operations/ClassicControlledOperation.cpp#L130-L153

Further places where the ClassicControlledOperation class is used can easily be found via code search.

denialhaag:
## Description

This PR adds an `IfElseOperation` to the C++ library and the Python package to support Qiskit's `IfElseOp`. The new operation replaces the `ClassicControlledOperation`.

This PR is a continuation of #990.

Fixes #923

## Checklist:

### DIFF
--- a/test/python/ir/test_if_else_operation.py
+++ b/test/python/ir/test_if_else_operation.py
@@ -10,11 +10,24 @@
 
 from __future__ import annotations
 
+import pytest
+
 from mqt.core.ir import QuantumComputation
 from mqt.core.ir.operations import ComparisonKind, OpType, StandardOperation
 
 
-def test_if_else_operation_register_eq() -> None:
+@pytest.mark.parametrize(
+    ("comparison_kind", "comparison_kind_string"),
+    [
+        (ComparisonKind.eq, "=="),
+        (ComparisonKind.neq, "!="),
+        (ComparisonKind.lt, "<"),
+        (ComparisonKind.leq, "<="),
+        (ComparisonKind.gt, ">"),
+        (ComparisonKind.geq, ">="),
+    ],
+)
+def test_if_else_operation_register(comparison_kind: ComparisonKind, comparison_kind_string: str) -> None:
     """Test the creation of an if-else operation."""
     qc = QuantumComputation()
     qc.add_qubit_register(1)
@@ -24,176 +37,22 @@ def test_if_else_operation_register_eq() -> None:
         then_operation=StandardOperation(0, OpType.x),
         else_operation=StandardOperation(0, OpType.y),
         control_register=c,
+        comparison_kind=comparison_kind,
     )
 
     qasm = qc.qasm3_str()
-    expected = """
+    expected = f"""
         // i 0
         // o 0
         OPENQASM 3.0;
         include "stdgates.inc";
         qubit[1] q;
         bit[1] c;
-        if (c == 1) {
+        if (c {comparison_kind_string} 1) {{
             x q[0];
-        } else {
+        }} else {{
             y q[0];
-        }
-    """
-    # Remove all whitespace from both strings before comparison
-    assert "".join(qasm.split()) == "".join(expected.split())
-
-
-def test_if_else_operation_register_neq() -> None:
-    """Test the creation of an if-else operation."""
-    qc = QuantumComputation()
-    qc.add_qubit_register(1)
-    c = qc.add_classical_register(1)
-
-    qc.if_else(
-        then_operation=StandardOperation(0, OpType.x),
-        else_operation=StandardOperation(0, OpType.y),
-        control_register=c,
-        comparison_kind=ComparisonKind.neq,
-    )
-
-    qasm = qc.qasm3_str()
-    expected = """
-        // i 0
-        // o 0
-        OPENQASM 3.0;
-        include "stdgates.inc";
-        qubit[1] q;
-        bit[1] c;
-        if (c != 1) {
-            x q[0];
-        } else {
-            y q[0];
-        }
-    """
-    # Remove all whitespace from both strings before comparison
-    assert "".join(qasm.split()) == "".join(expected.split())
-
-
-def test_if_else_operation_register_lt() -> None:
-    """Test the creation of an if-else operation."""
-    qc = QuantumComputation()
-    qc.add_qubit_register(1)
-    c = qc.add_classical_register(1)
-
-    qc.if_else(
-        then_operation=StandardOperation(0, OpType.x),
-        else_operation=StandardOperation(0, OpType.y),
-        control_register=c,
-        comparison_kind=ComparisonKind.lt,
-    )
-
-    qasm = qc.qasm3_str()
-    expected = """
-        // i 0
-        // o 0
-        OPENQASM 3.0;
-        include "stdgates.inc";
-        qubit[1] q;
-        bit[1] c;
-        if (c < 1) {
-            x q[0];
-        } else {
-            y q[0];
-        }
-    """
-    # Remove all whitespace from both strings before comparison
-    assert "".join(qasm.split()) == "".join(expected.split())
-
-
-def test_if_else_operation_register_leq() -> None:
-    """Test the creation of an if-else operation."""
-    qc = QuantumComputation()
-    qc.add_qubit_register(1)
-    c = qc.add_classical_register(1)
-
-    qc.if_else(
-        then_operation=StandardOperation(0, OpType.x),
-        else_operation=StandardOperation(0, OpType.y),
-        control_register=c,
-        comparison_kind=ComparisonKind.leq,
-    )
-
-    qasm = qc.qasm3_str()
-    expected = """
-        // i 0
-        // o 0
-        OPENQASM 3.0;
-        include "stdgates.inc";
-        qubit[1] q;
-        bit[1] c;
-        if (c <= 1) {
-            x q[0];
-        } else {
-            y q[0];
-        }
-    """
-    # Remove all whitespace from both strings before comparison
-    assert "".join(qasm.split()) == "".join(expected.split())
-
-
-def test_if_else_operation_register_gt() -> None:
-    """Test the creation of an if-else operation."""
-    qc = QuantumComputation()
-    qc.add_qubit_register(1)
-    c = qc.add_classical_register(1)
-
-    qc.if_else(
-        then_operation=StandardOperation(0, OpType.x),
-        else_operation=StandardOperation(0, OpType.y),
-        control_register=c,
-        comparison_kind=ComparisonKind.gt,
-    )
-
-    qasm = qc.qasm3_str()
-    expected = """
-        // i 0
-        // o 0
-        OPENQASM 3.0;
-        include "stdgates.inc";
-        qubit[1] q;
-        bit[1] c;
-        if (c > 1) {
-            x q[0];
-        } else {
-            y q[0];
-        }
-    """
-    # Remove all whitespace from both strings before comparison
-    assert "".join(qasm.split()) == "".join(expected.split())
-
-
-def test_if_else_operation_register_geq() -> None:
-    """Test the creation of an if-else operation."""
-    qc = QuantumComputation()
-    qc.add_qubit_register(1)
-    c = qc.add_classical_register(1)
-
-    qc.if_else(
-        then_operation=StandardOperation(0, OpType.x),
-        else_operation=StandardOperation(0, OpType.y),
-        control_register=c,
-        comparison_kind=ComparisonKind.geq,
-    )
-
-    qasm = qc.qasm3_str()
-    expected = """
-        // i 0
-        // o 0
-        OPENQASM 3.0;
-        include "stdgates.inc";
-        qubit[1] q;
-        bit[1] c;
-        if (c >= 1) {
-            x q[0];
-        } else {
-            y q[0];
-        }
+        }}
     """
     # Remove all whitespace from both strings before comparison
     assert "".join(qasm.split()) == "".join(expected.split())
@@ -229,7 +88,7 @@ def test_if_else_operation_bit() -> None:
     assert "".join(qasm.split()) == "".join(expected.split())
 
 
-def test_if_operation_register_eq() -> None:
+def test_if_operation_register() -> None:
     """Test the creation of an if-else operation."""
     qc = QuantumComputation()
     qc.add_qubit_register(1)
--- a/test/python/plugins/test_qiskit.py
+++ b/test/python/plugins/test_qiskit.py
@@ -10,7 +10,7 @@
 
 from __future__ import annotations
 
-from typing import cast
+from typing import TYPE_CHECKING, Any, cast
 
 import pytest
 from packaging import version
@@ -38,6 +38,9 @@ from mqt.core.ir.operations import (
 )
 from mqt.core.ir.symbolic import Expression
 from mqt.core.plugins.qiskit import mqt_to_qiskit, qiskit_to_mqt
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
 
 
 def test_empty_circuit() -> None:
@@ -476,6 +479,62 @@ def test_symbolic_two_qubit() -> None:
         mqt_to_qiskit(mqt_qc)
 
 
+@pytest.mark.parametrize(
+    ("condition_callable", "comparison_kind"),
+    [
+        (expr.equal, ComparisonKind.eq),
+        (expr.not_equal, ComparisonKind.neq),
+        (expr.less, ComparisonKind.lt),
+        (expr.less_equal, ComparisonKind.leq),
+        (expr.greater, ComparisonKind.gt),
+        (expr.greater_equal, ComparisonKind.geq),
+    ],
+)
+def test_if_else_operation_register(
+    condition_callable: Callable[[Any, Any], Any],
+    comparison_kind: ComparisonKind,
+) -> None:
+    """Test import of if-else operation on register."""
+    from qiskit.qasm3 import dumps
+
+    then_qc = QuantumCircuit(1)
+    then_qc.x(0)
+
+    else_qc = QuantumCircuit(1)
+    else_qc.y(0)
+
+    qc = QuantumCircuit(1, 1)
+    qc.append(
+        instruction=IfElseOp(condition_callable(qc.cregs[0], 0), then_qc, else_qc),
+        qargs=qc.qubits,
+    )
+    print(qc)
+
+    mqt_qc = qiskit_to_mqt(qc)
+    print(mqt_qc)
+    assert mqt_qc.num_qubits == 1
+    assert mqt_qc.num_ops == 1
+    if_else_operation = mqt_qc[0]
+    assert isinstance(if_else_operation, IfElseOperation)
+    assert if_else_operation.control_register is not None
+    assert if_else_operation.control_bit is None
+    assert if_else_operation.expected_value_register == 0
+    assert if_else_operation.comparison_kind == comparison_kind
+    then_operation = if_else_operation.then_operation
+    assert isinstance(then_operation, StandardOperation)
+    then_operation.type_ = OpType.x
+    else_operation = if_else_operation.else_operation
+    assert isinstance(else_operation, StandardOperation)
+    else_operation.type_ = OpType.y
+
+    qiskit_qc = mqt_to_qiskit(mqt_qc)
+    print(qiskit_qc)
+
+    qasm_input = dumps(qc)
+    qasm_output = dumps(qiskit_qc)
+    assert qasm_input == qasm_output
+
+
 def test_if_else_operation_bit() -> None:
     """Test import of if-else operation."""
     from qiskit.qasm3 import dumps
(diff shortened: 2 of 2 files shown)


include/mqt-core/ir/operations/IfElseOperation.hpp
```diff
@@ -0,0 +1,148 @@
...
+class IfElseOperation final : public Operation {
+public:
+  IfElseOperation(std::unique_ptr<Operation>&& thenOp,
+                  std::unique_ptr<Operation>&& elseOp,
+                  ClassicalRegister controlReg, std::uint64_t expectedVal = 1U,
+                  ComparisonKind kind = Eq);
+
+  IfElseOperation(std::unique_ptr<Operation>&& thenOp,
+                  std::unique_ptr<Operation>&& elseOp, Bit cBit,
+                  std::uint64_t expectedVal = 1U, ComparisonKind kind = Eq);
```
burgholzer: We could think about changing the expected Val to `bool` here, as it is the only valid value range.

include/mqt-core/ir/operations/IfElseOperation.hpp
```diff
@@ -0,0 +1,148 @@
...
+  void dumpOpenQASM(std::ostream& of, const QubitIndexToRegisterMap& qubitMap,
+                    const BitIndexToRegisterMap& bitMap, std::size_t indent,
+                    bool openQASM3) const override;
+
+  void invert() override {
+    thenOp->invert();
+    if (elseOp) {
+      elseOp->invert();
+    }
+  }
```
burgholzer: I think this might not be entirely valid. The IfElse operation is most definitely a non-unitary operation. In my opinion it makes more sense to leave this undefined and have consumers of the class decide whether they want to invert either of the conditions.

I also believe "invert" is ambiguous here. It could also refer to inverting the condition of the if.

include/mqt-core/ir/operations/IfElseOperation.hpp
```diff
@@ -0,0 +1,148 @@
...
+        "Cannot set operation type of an IfElseOperation.");
+  }
+
+  void setParameter(const std::vector<fp>&) override {
+    throw std::runtime_error("An IfElseOperation cannot be parameterized.");
+  }
+
+  void apply(const Permutation&) override {
+    throw std::runtime_error("Cannot apply permutation to an IfElseOperation.");
+  }
```
burgholzer: Apply could technically work by applying the permutation to both operation blocks.

include/mqt-core/ir/QuantumComputation.hpp
```diff
@@ -333,33 +334,13 @@ class QuantumComputation {
...
-                         ComparisonKind cmp = Eq,
-                         const std::vector<fp>& params = {});
-  void classicControlled(OpType op, Qubit target, Control control, Bit cBit,
-                         std::uint64_t expectedValue = 1U,
-                         ComparisonKind cmp = Eq,
-                         const std::vector<fp>& params = {});
-  void classicControlled(OpType op, Qubit target, const Controls& controls,
-                         Bit cBit, std::uint64_t expectedValue = 1U,
-                         ComparisonKind cmp = Eq,
-                         const std::vector<fp>& params = {});
```
burgholzer, line 362: I think it would be nice to still provide these shortcuts for an IfElse with only an `if` and no else, given how this is the most common way to use the operation.
Those convenience shortcuts could be called `if` (instead of the previous classicControlled)

src/algorithms/QFT.cpp
```diff
@@ -79,13 +81,15 @@ auto createIterativeQFT(const Qubit nq) -> QuantumComputation {
     for (Qubit j = 1; j <= i; ++j) {
       const auto d = nq - j;
       if (j == i) {
-        qc.classicControlled(S, 0, d, 1U);
+        qc.ifElse(std::make_unique<StandardOperation>(0, S), nullptr, d, 1U);
```
burgholzer: If the suggestion in the QuantumComputation class is adopted, this should be simpler to write again.
Also applies to other places where the syntax was changed.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -339,17 +339,22 @@ void removeDiagonalGatesBeforeMeasureRecursive(
...
-      auto* cop = dynamic_cast<ClassicControlledOperation*>(op)->getOperation();
+    } else if (op->isIfElseOperation()) {
+      auto* thenOp = dynamic_cast<IfElseOperation*>(op)->getThenOp();
+      auto* elseOp = dynamic_cast<IfElseOperation*>(op)->getElseOp();
+      if (thenOp == nullptr || elseOp != nullptr) {
+        throw std::runtime_error("If-else operations with non-trivial else "
+                                 "branches are currently not supported.");
+      }
+
+      // treat then branch as above
```
burgholzer: This is fine. Handling this properly would require quite a bit of work that I do not want to invest in the circuit optimizer if it is bound to be replaced by the MLIR infrastructure anyway.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -722,12 +727,25 @@ void CircuitOptimizer::eliminateResets(QuantumComputation& qc) {
...
+            } else if (auto* ifElse =
+                           dynamic_cast<IfElseOperation*>(compOpIt->get());
+                       ifElse != nullptr) {
+              auto* thenOp = ifElse->getThenOp();
+              auto* elseOp = ifElse->getElseOp();
+              if (thenOp == nullptr || elseOp != nullptr) {
+                throw std::runtime_error(
+                    "If-else operations with non-trivial else branches are "
+                    "currently not supported.");
+              }
```
burgholzer: This should not be necessary. The else branch can be handled just like the then branch here.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -736,12 +754,23 @@ void CircuitOptimizer::eliminateResets(QuantumComputation& qc) {
...
         changeTargets(targets, replacementMap);
         changeControls(controls, replacementMap);
+      } else if (auto* ifElse = dynamic_cast<IfElseOperation*>(it->get());
+                 ifElse != nullptr) {
+        auto* thenOp = ifElse->getThenOp();
+        auto* elseOp = ifElse->getElseOp();
+        if (thenOp == nullptr || elseOp != nullptr) {
+          throw std::runtime_error("If-else operations with non-trivial else "
+                                   "branches are currently not supported.");
+        }
```
burgholzer: This should not be necessary. The else branch can be handled just like the then branch here.

src/circuit_optimizer/CircuitOptimizer.cpp
```diff
@@ -830,26 +859,32 @@ void CircuitOptimizer::deferMeasurements(QuantumComputation& qc) {
...
-            classicOp != nullptr) {
-          const auto& expectedValue = classicOp->getExpectedValue();
+        if (auto* ifElse = dynamic_cast<IfElseOperation*>(opIt->get());
+            ifElse != nullptr) {
+          auto* thenOp = ifElse->getThenOp();
+          auto* elseOp = ifElse->getElseOp();
+          if (thenOp == nullptr || elseOp != nullptr) {
+            throw std::runtime_error("If-else operations with non-trivial else "
+                                     "branches are currently not supported.");
+          }
```
burgholzer: This should also be fairly straightforward to handle (I hope).
The then branch is inserted as before. The else branch is inserted with X gates being added on the new control before and after adding all operations of the else branch.

src/dd/Operations.cpp
```diff
@@ -184,10 +184,15 @@ MatrixDD getDD(const qc::Operation& op, Package& dd,
...
-    return getDD(*classicOp.getOperation(), dd, permutation, inverse);
+  if (op.isIfElseOperation()) {
+    const auto& ifElse = dynamic_cast<const qc::IfElseOperation&>(op);
+    auto* thenOp = ifElse.getThenOp();
+    auto* elseOp = ifElse.getElseOp();
+    if (thenOp == nullptr || elseOp != nullptr) {
+      throw std::runtime_error("If-else operations with non-trivial else "
+                               "branches are currently not supported.");
+    }
+    return getDD(*thenOp, dd, permutation, inverse);
```
burgholzer: I think this has to be handled differently.
IfElse is a proper non-unitary operation now and I'd rather have the respective code for getting the DD of the then or else branch where the value of the classical bit (register) is known, i.e., in the simulator or in the applyIfElse function

src/ir/operations/IfElseOperation.cpp
```diff
@@ -0,0 +1,192 @@
...
+  }
+  if (openQASM3) {
+    of << "}";
+    if (elseOp) {
+      of << " else {\n";
+      elseOp->dumpOpenQASM(of, qubitMap, bitMap, indent + 1, openQASM3);
+      of << "}";
+    }
+    of << "\n";
+  }
```
burgholzer: This is a nitpick, but the code here does not handle OpenQASM exports correctly as it silently drops the else branch.
I think this should be fairly easy to correct though. For OpenQASM 2, one would simply emit two `if` statements, one with a negated condition.

src/ir/operations/Operation.cpp
```diff
@@ -76,10 +65,7 @@ std::ostream& Operation::print(std::ostream& os, const Permutation& permutation,
   for (std::size_t i = 0; i < nqubits; ++i) {
     const auto q = static_cast<Qubit>(i);
     if (std::ranges::find(actualTargets, q) != actualTargets.cend()) {
-      if (type == ClassicControlled) {
-        const auto reducedName = name.substr(2);
-        os << "\033[1m\033[35m" << std::setw(4) << reducedName;
-      } else if (type == Barrier) {
```
burgholzer, line 82: This lost the special color printing for if instructions. Not the end of the world though.

src/ir/operations/Operation.cpp
```diff
@@ -27,25 +27,14 @@
...
-
-    os << "  c[" << parameter[0];
-    if (parameter.size() == 2) {
-      os << "] == " << parameter[1];
-    } else {
-      os << "..." << (parameter[0] + parameter[1] - 1)
-         << "] == " << parameter[2];
-    }
-    return os;
-  }
```
burgholzer, line 40: It's slightly unfortunate that we no longer print the condition in our textual output. We should be able to restore that (I think).

src/qasm3/Importer.cpp
```diff
@@ -894,36 +894,33 @@ void Importer::visitIfStatement(
...
-      const auto& [creg, comparisonKind, rhs] = std::get<
-          std::tuple<qc::ClassicalRegister, qc::ComparisonKind, uint64_t>>(
-          condition);
-      qc->emplace_back<qc::ClassicControlledOperation>(std::move(thenOps), creg,
-                                                       rhs, comparisonKind);
-    }
+    thenOps = translateBlockOperations(ifStatement->thenStatements);
+  } else {
+    thenOps = std::make_unique<qc::CompoundOperation>();
   }
```
burgholzer: I think you can simply drop this else.
An empty block should be represented perfectly fine as `nullptr`.
Similarly below.

include/mqt-core/ir/operations/IfElseOperation.hpp
denialhaag: Good point! Done. As a result, there are now `getExpectedValueRegister()` and `getExpectedValueBit()` methods. Let me know what you think!

include/mqt-core/ir/QuantumComputation.hpp
denialhaag: Also good point! I added them back again. The Python bindings for the respective methods are still missing, but I'll do that tomorrow.

src/algorithms/QFT.cpp
denialhaag: Done! The tests are a lot cleaner again now. 😌

src/ir/operations/IfElseOperation.cpp
denialhaag: I fixed it now and added unit tests for all eight cases. 🤠

src/qasm3/Importer.cpp
denialhaag: Yes, that just worked!

> Similarly below.

Regarding this, is there another instance? I couldn't find one. 🤔

include/mqt-core/ir/QuantumComputation.hpp
denialhaag: I also added the `if_` method to the Python package now.

src/ir/operations/Operation.cpp
denialhaag: I put it back by overriding the `print()` method for the `IfElseOperation`. 👍

src/ir/operations/Operation.cpp
denialhaag: I decided to override the `print()` method for the `IfElseOperation`. I'm not super sure how to do this best. For now, it prints the then-operation in blue and the condition in pink. I didn't find a way to change the color of the then-operation without adding the color as a method parameter. I also didn't find a nice way to add the else-operation. Let me know if you have a nice idea!

src/circuit_optimizer/CircuitOptimizer.cpp
denialhaag: Perfect, I left it as is. 👍

src/circuit_optimizer/CircuitOptimizer.cpp
denialhaag: You are right, it was easy. I also added an appropriate unit test.

src/circuit_optimizer/CircuitOptimizer.cpp
denialhaag: See comment above.

src/circuit_optimizer/CircuitOptimizer.cpp
denialhaag: Yes, it worked as you said. I opted for using a negative control instead of two X gates.

While doing that, I ran into a bug with printing negative controls: because we were using a `Qubit` to check for inclusion, the printing only worked for positive qubits, because a `Qubit` becomes a positive control. I fixed the bug in https://github.com/munich-quantum-toolkit/core/pull/1117/commits/0c32fef0c14ea3bdd5bd08ac254cbda942d61a98.

include/mqt-core/ir/operations/IfElseOperation.hpp
denialhaag: Makes sense! I removed it now! 👍

include/mqt-core/ir/operations/IfElseOperation.hpp
denialhaag: Also makes sense! Done.

src/dd/Operations.cpp
denialhaag: That makes a lot of sense! If I understand correctly, it's fine to simply remove this then. 🤔  Done.

burgholzer (approved):
I believe this is looking great now. From my side, there is not that much more to do. It would be great to get the test coverage up a bit so that CI is all green ✅

denialhaag:
> I believe this is looking great now. From my side, there is not that much more to do. It would be great to get the test coverage up a bit so that CI is all green ✅  (An upgrade guide paragraph would probably also be adequate)

Perfect, I will do that later and then also update the upgrade guide!

denialhaag:
@burgholzer, thanks a lot for your improvements! I went through them, and everything looked good to me.

 The coverage should be good now. When improving the Python coverage, I ended up finding that we didn't allow for all comparison kinds when translating from Qiskit to MQT. To get the C++ coverage across the required value, I added a test that checks that the runtime errors are thrown. I don't love it, but I hope that that is fine.

I'll request another review because I ended up making more changes than anticipated.

test/python/ir/test_if_else_operation.py
```diff
@@ -0,0 +1,277 @@
...
+        qubit[1] q;
+        bit[1] c;
+        if (c >= 1) {
+            x q[0];
+        } else {
+            y q[0];
+        }
+    """
+    # Remove all whitespace from both strings before comparison
+    assert "".join(qasm.split()) == "".join(expected.split())
```
burgholzer: This is truly a nitpick, but I think we could be saving some code duplication here by combining these tests (either into a parametrized test or in a test that just loops over the different comparison kinds). The tests essentially differ in 2 localized places that should be easy to parametrize.

test/python/plugins/test_qiskit.py
```diff
@@ -562,9 +563,167 @@ def test_if_else_operation_register() -> None:
...
+    if_else_operation = mqt_qc[0]
+    assert isinstance(if_else_operation, IfElseOperation)
+    assert if_else_operation.comparison_kind == ComparisonKind.neq
+
+    qiskit_qc = mqt_to_qiskit(mqt_qc)
+    print(qiskit_qc)
+
+    qasm_input = dumps(qc)
+    qasm_output = dumps(qiskit_qc)
+    assert qasm_input == qasm_output
```
burgholzer: The same comment as in the other file applies here. These tests really don't differ by much. It would be great to reduce the duplicate code.
I am pretty sure that is something LLMs can already do for you if prompted appropriately.

test/python/ir/test_if_else_operation.py
denialhaag: No, I agree. The parameterization worked perfectly! 👍

test/python/plugins/test_qiskit.py
denialhaag: Done!

Issue #1122, opened 2025-08-15
✨ MLIR - Add more passes


flowerthrower:
### Problem Statement

In order to have a compiler of value, we need to have at least a handful of optimization passes available.

### Proposed Solution

These passes should ideally be atomar and independent of other passes. @taminob already found a few options
https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/passes/optimization
https://github.com/CQCL/tket/tree/main/tket/src/Transformations

The ones selected for implementation are listed below:
- [ ] placeholderPass

flowerthrower:
Thank you very much @burgholzer. This will give @taminob plenty of passes to choose from. I would suggest starting with sth. simple (pure quantum / no classical control flow) and increasing complexity from there. Basically, feel free to pick whatever pass seems most easy to implement for you, @taminob, to get a feeling for the code and test setup.

burgholzer:
Just from past experience, the easiest passes to accommodate are the gate cancellation and gate merging passes.
Maybe a good first start would be to look into the MLIR passes we already implemented and check whether there are easy ways to add functionality to them or to adapt them for further use cases.

Just for awareness: #916 already contains some open discussion on how particular optimizations might be realized.
If I recall correctly, @DRovara has a student who is planning to work on canonicalization in general. So it might be worth coordinating.

burgholzer:
Based on previous discussions, I am adding some pointers to relevant passes here.
In general, I would expect the Rust code in Qiskit to be the best reference for the lessons the IBM team has learned over the years that are now manifested in the Rust rewrite.

## general pointers

- unitary synthesis: https://github.com/Qiskit/qiskit/blob/stable/2.1/crates/transpiler/src/passes/unitary_synthesis.rs
- basis translation: https://github.com/Qiskit/qiskit/tree/stable/2.1/crates/transpiler/src/passes/basis_translator

## single-qubit gate decomposition

- https://github.com/Qiskit/qiskit/blob/stable/2.1/crates/transpiler/src/passes/optimize_1q_gates_decomposition.rs

- https://github.com/Qiskit/qiskit/blob/stable/2.1/crates/synthesis/src/euler_one_qubit_decomposer.rs

- https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.synthesis.OneQubitEulerDecomposer

##  two-qubit gate decomposition

[...]

PR #1140, opened 2025-08-20
:sparkles: Implement simple identity removal pass in MLIR

taminob:
## Description

This PR adds a simple MLIR pass which removes all identity gates from a circuit.
The main purpose is to get to know the code base regarding MLIR passes.

Related to #1122 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 ### Added
 
 - ✨ Add new pattern to MLIR pass `GateElimination` for identity gate removal ([#1140]) ([**@taminob**])
-- ✨ Rename MLIR pass `CancelConsecutiveInverses` to `GateElimination`([#1140]) ([**@taminob**])
 - ✨ Add Clifford block collection pass to `CircuitOptimizer` module ([#885]) ([**jannikpflieger**], [**@burgholzer**])
 - ✨ Add `isControlled()` method to the `UnitaryInterface` MLIR class ([#1157]) ([**@taminob**], [**@burgholzer**])
 - 📝 Integrate generated MLIR documentation ([#1147]) ([**@denialhaag**], [**@burgholzer**])
@@ -26,6 +25,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Changed
 
+- 🚚 Rename MLIR pass `CancelConsecutiveInverses` to `GateElimination`([#1140]) ([**@taminob**])
 - 🚚 Rename `xxminusyy` to `xx_minus_yy` and `xxplusyy` to `xx_plus_yy` in MLIR dialects ([#1071]) ([**@BertiFlorea**], [**@denialhaag**])
 - 🚸 Add custom assembly format for operations in the MLIR dialects ([#1139]) ([**@burgholzer**])
 - 🚸 Enable `InferTypeOpInterface` in the MLIR dialects to reduce explicit type information ([#1139]) ([**@burgholzer**])


mlir/test/Dialect/MQTOpt/Transforms/remove-identities.mlir
```diff
@@ -0,0 +1,48 @@
...
+    // ========================== Check for operations that should not be canceled ==========================
+    // CHECK: %[[ANY:.*]] = mqtopt.x() %[[ANY:.*]] : !mqtopt.Qubit
+    // CHECK: %[[Q0_1:.*]] = mqtopt.z() %[[Q0_0]] : !mqtopt.Qubit
+    // CHECK: %[[Q1_5:.*]] = mqtopt.h() %[[ANY:.*]] : !mqtopt.Qubit
+
+    // ========================== Check for operations that should be canceled ==============================
+    // CHECK-NOT: %[[ANY:.*]] = mqtopt.i() %[[ANY:.*]] : !mqtopt.Qubit
+
+    %q1_1 = mqtopt.x() %q1_0 : !mqtopt.Qubit
+    %q1_2 = mqtopt.x() %q1_1 : !mqtopt.Qubit
```
flowerthrower: Any particular reason for double X gate here? The check only looks for one appearance.

taminob: No real reason, I just created any dummy circuit. I can remove it when I incorporate Lukas' feedback - thanks :+1:

mlir/lib/Dialect/MQTOpt/Transforms/RemoveIdentitiesPattern.cpp
```diff
@@ -0,0 +1,96 @@
...
+  matchAndRewrite(UnitaryInterface op,
+                  mlir::PatternRewriter& rewriter) const override {
+    if (!isIdentity(*op)) {
+      return mlir::failure();
+    }
+
+    // Attach users to previous operation in order to be able to remove the
+    // operation. Note: There might be multiple users of an operation. The
+    // qubits itself can only be used once (linear typing), but may be used
+    // as control line in CX gates multiple times (TODO: correct?).
```
taminob: I just remembered that I wasn't entirely sure about this comment - did I understand the note in `CancelConsecutiveInverses` correctly? I reworded it to be clearer (in my opinion).

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -28,6 +28,14 @@ def MQTCoreRoundTrip : Pass<"mqt-core-round-trip", "mlir::ModuleOp"> {
   }];
 }
 
+def RemoveIdentities : Pass<"remove-identities", "mlir::ModuleOp"> {
+  let summary = "This pass removes all explicit identity gates.";
+  let description = [{
+    This pass searches for all explicit identity gates in the circuit and removes them since they have no influence
+    on the result.
+  }];
+}
```
burgholzer: I think the logic here is so simple that it might not warrant its own pass.
However, I believe we could add this as a dedicated pattern to the "CancelConsecutiveInverses" pass.

That way, the pattern executor of MLIR will include this simplification as part of the pass.

We might want to rename the pass at that point to something a little more suitable. "GateCancellation" for example.

Separate note: most of what is done here and in the existing cancellation pass would probably be considered a canonicalization instead of a transformation pass. But that is a topic for another day 😌

mlir/lib/Dialect/MQTOpt/Transforms/RemoveIdentitiesPattern.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <mlir/Support/LogicalResult.h>
+#include <string_view>
+
+namespace mqt::ir::opt {
+
+/**
+ * @brief This pattern removes all identity (noop) operations.
+ */
+struct RemoveIdentitiesPattern final
+    : mlir::OpInterfaceRewritePattern<UnitaryInterface> {
```
burgholzer: I think the implementation of this pattern can be improved and simplified.
Instead of manually doing a string based match, you should be able to specify the Identity operation instead of the `UnitaryInterface` here, which should allow you to only match identity operations without writing any more code.

mlir/lib/Dialect/MQTOpt/Transforms/RemoveIdentitiesPattern.cpp
```diff
@@ -0,0 +1,96 @@
...
+          continue;
+        }
+        const auto idx = std::distance(outQubits.begin(), found);
+        rewriter.modifyOpInPlace(
+            childUser, [&] { childUser->setOperand(i, inQubits[idx]); });
+      }
+    }
+
+    rewriter.eraseOp(op);
+    return mlir::success();
```
burgholzer: I am fairly sure that there is a more compact way to write this based on existing MLIR concepts. This might be a good opportunity to start to dive a little deeper into what MLIR has to offer.

Qubits in the MQT Opt dialect are linear types. As such, they may only have one user and may only be used once.
The replacement pattern here should probably simply remove the operation and, in the process, replace all input qubits of its users with the input qu its of the operation itself.

I personally wouldn't rely too much on the existing passes in the MLIR part of the code base as we are still learning on how to best approach this.
Reimplementing or refactoring existing passes is totally on the table here (in a separate PR).

mlir/test/Dialect/MQTOpt/Transforms/remove-identities.mlir
```diff
@@ -0,0 +1,48 @@
...
+    %q1_4 = mqtopt.i() %q1_3 : !mqtopt.Qubit
+    %q1_5 = mqtopt.h() %q1_4 : !mqtopt.Qubit
+
+    // CHECK: %[[Reg_3:.*]] = "mqtopt.insertQubit"(%[[Reg_2]], %[[Q0_1]])  <{index_attr = 0 : i64}>
+    %reg_3 = "mqtopt.insertQubit"(%reg_2, %q0_2) <{index_attr = 0 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
+    // CHECK: %[[Reg_4:.*]] = "mqtopt.insertQubit"(%[[Reg_3]], %[[Q1_5]])  <{index_attr = 1 : i64}>
+    %reg_4 = "mqtopt.insertQubit"(%reg_3, %q1_5) <{index_attr = 1 : i64}> : (!mqtopt.QubitRegister, !mqtopt.Qubit) -> !mqtopt.QubitRegister
+    // CHECK: "mqtopt.deallocQubitRegister"(%[[Reg_4]])
+    "mqtopt.deallocQubitRegister"(%reg_4) : (!mqtopt.QubitRegister) -> ()
+    return
```
burgholzer: Couple of comments here:
- you should be able to avoid pretty much all of the boilerplate code (register allocation, extraction, etc) by using the newly added "mqtopt.qubit" static qubits or the "mqtopt.allocQubit" single-qubit allocation functions; the former being even less code to write.
- we typically aggregate the check strings at the top of the test. See some of the dialect feature test files for proper inspiration.
- it would be great to break this down into real unit tests that each test a particular aspect of the pass/pattern Instead of throwing everything into one bigger test.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: Yes, I already thought so and didn't expect this PR to get merged - I just wanted to get going with something really easy that I can test.

I will probably close this PR and re-open a new one once I've adapted the `CancelConsecutiveInverses` (GateCancellation sounds like a good name to me).

mlir/lib/Dialect/MQTOpt/Transforms/RemoveIdentitiesPattern.cpp
taminob: Thank you for the hint - I'm still new to MLIR and need to get used to the different types and what is supposed to do what.
However, this comment might not make it into the new PR since I need all gates for the more generic "GateCancellation" pattern. I'll look into it.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
burgholzer: Feel free to keep this PR open and add the respective changes here. No actual need to open a new PR in my opinion.

mlir/lib/Dialect/MQTOpt/Transforms/RemoveIdentitiesPattern.cpp
taminob: Great, thank you - looks like some of the existing passes really need some rework. :)

mlir/test/Dialect/MQTOpt/Transforms/remove-identities.mlir
taminob: I'll incorporate your bullet points in the update :+1:

mlir/lib/Dialect/MQTOpt/Transforms/RemoveIdentitiesPattern.cpp
burgholzer: Let's see what you come up with. I think it would be totally fair to have the identity removal as a separate _pattern_ that is part of the gate cancellation _pass_ and only matches on the `Id` operation.

taminob:
@flowerthrower @burgholzer I adapted the pattern to be a part of the `GateCancellation` pass now together with the `CancelConsecutiveInversesPattern` - thanks for the tip. Is this how you imagined it?

If you have any other hints on how to solve this better, feel free to give me another pointer - I did some digging in the LLVM source code and I didn't find a better fit than `OpRewritePattern`. Only using the `rewrite()` method did not work since it is required to have a `match()` method which is final, unfortunately. Luckily, I came across `getOpOperands()` which is used internally by `getOperands()` anyway and saves us the logic with the indices.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.h
```diff
@@ -25,7 +25,8 @@ namespace mqt::ir::opt {
 #define GEN_PASS_DECL
 #include "mlir/Dialect/MQTOpt/Transforms/Passes.h.inc" // IWYU pragma: export
 
-void populateCancelInversesPatterns(mlir::RewritePatternSet& patterns);
+void populateRemoveIdentitiesPatterns(mlir::RewritePatternSet& patterns);
```
burgholzer: I am failry sure that this function does not exist anymore.
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
```diff
@@ -128,13 +128,52 @@ struct CancelConsecutiveInversesPattern final
...
+  explicit RemoveIdentitiesPattern(mlir::MLIRContext* context)
+      : OpRewritePattern(context) {}
+
+  mlir::LogicalResult
+  matchAndRewrite(IOp op, mlir::PatternRewriter& rewriter) const override {
+    auto inQubits = op.getAllInQubits();
+    auto outQubits = op.getAllOutQubits();
+
+    // an identity operation only has exactly one in- and output
+    assert(inQubits.size() == 1);
```
burgholzer: Would be nice to fix the warning(s) reported by clang-tidy. This one is particularly easy, as it is simply a missing `#include <cassert>` at the top of the file.

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
```diff
@@ -128,13 +128,52 @@ struct CancelConsecutiveInversesPattern final
 };
 
 /**
- * @brief Populates the given pattern set with the
- * `CancelConsecutiveInversePattern`.
+ *
+ */
```
burgholzer: This is missing an appropriate doxygen comment for documentation purposes.

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
```diff
@@ -128,13 +128,52 @@ struct CancelConsecutiveInversesPattern final
...
+  mlir::LogicalResult
+  matchAndRewrite(IOp op, mlir::PatternRewriter& rewriter) const override {
+    auto inQubits = op.getAllInQubits();
+    auto outQubits = op.getAllOutQubits();
+
+    // an identity operation only has exactly one in- and output
+    assert(inQubits.size() == 1);
+    assert(outQubits.size() == 1);
+    auto inQubit = inQubits.front();
+    auto outQubit = outQubits.front();
```
burgholzer: I do not think that this assertion is valid.
While the Identity gate is a single-target gate, it may still have arbitrarily many controls. That might not be the most useful gate, but still. It might occur at some point due to some optimization.

You should be able to see this if you create a test that adds a controlled identity gate.

And in that case, it can, of course happen that there is more than one successor operation that needs to be modified. Assume a single-controlled identity gate that is followed by an X gate on the first qubit and another X gate on the second qubit. Then each of the qubits involved in the controlled identity gate is used only once (linear type), but both qubits are used by separate gates/users.
I am fairly sure the code below does not handle that properly yet.

Also, are you sure that you cannot simply use [`replaceAllOpUsesWith()`](https://mlir.llvm.org/doxygen/classmlir_1_1RewriterBase.html#af11a67366e4e1531f3bb0fbe6a67ec03) or any of its variants here? This seems like such a common pattern to me, that I would be highly surprised if there were no shorter, builtin solution to this.

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
```diff
@@ -211,3 +211,99 @@ module {
     return
   }
 }
+
+
+// -----
+// This test checks if a single identity operations is removed correctly.
```
burgholzer:
```suggestion
// This test checks if a single identity operation is removed correctly.
```

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
```diff
@@ -211,3 +211,99 @@ module {
...
+    // ========================== Check for operations that should not be canceled ==========================
+    // CHECK: %[[Q0_1:.*]] = mqtopt.x() %[[Q0_0]] : !mqtopt.Qubit
+    // CHECK: %[[Q0_3:.*]] = mqtopt.z() %[[Q0_1]] : !mqtopt.Qubit
+
+    // ========================== Check for operations that should be canceled ==============================
+    // CHECK-NOT: %[[ANY:.*]] = mqtopt.i() %[[ANY:.*]] : !mqtopt.Qubit
+
+    // CHECK: mqtopt.deallocQubit %[[Q0_3]]
+
+    %q0_0 = "mqtopt.allocQubit"() : () -> !mqtopt.Qubit
```
burgholzer: Since one of the latest PRs you can also use the shorter notation for these tests
```suggestion
    %q0_0 = mqtopt.allocQubit
```
safes some typing. (same for the dealloc operation).

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
```diff
@@ -211,3 +211,99 @@ module {
...
+
+    %q0_1 = mqtopt.x() %q0_0 : !mqtopt.Qubit
```
burgholzer: I am also reasonably sure (altough not 100%) that the `: !mqtopt.Qubit` at the end can be dropped throughout the tests.
```suggestion
    %q0_1 = mqtopt.x() %q0_0
```

burgholzer (changes requested):
@taminob thanks for another iteration.
Nice to see how quickly this is coming together.
I left some further comments inline that should help to further drive this forward.

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
taminob: I planned to do so after your feedback to avoid unnecessary CI time since this is such an easy fix :)

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
taminob: Thanks for the explanation - I didn't know that all gates can be controlled. I'll adjust the logic.

Also not sure how I missed `replaceAllOpUsesWith`, maybe my language server didn't list it for some reason.

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
taminob: That is useful, thanks - I saw this before, but was unsure and thus decided to keep it consistent instead :)

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
taminob: I thought that it's maybe considered good style (like explicit typing in Python), but probably not necessary in tests

burgholzer: Fair point, but I think in cases like these, I'd rather prefer less to type; plus the type is pretty obvious from the operation itself. 😌

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
burgholzer: Consistency is good; in principle. When it comes to MLIR, we are still learning and I think we can be a bit more lenient with adopting other styles. If there is anything that you find during your work on all of this that you think might make sense, feel free to simply propose it 😌

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
burgholzer: 👍🏼 CI is free (well, almost. excessively using it still potentially blocks other PRs due to the concurrency limit). Generally, don't hesitate to make use of it 😌

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.h
burgholzer: Just if you are not aware of this: For suggestions like these, you can directly apply them on GitHub. You can go to the "Files changed" tab and even add them to a batch there instead of adding them as single commits. The convenient feature there is that this will also automatically resolve the respective comment and credit the author of the commit.
Not a must too use that, but we have found it beneficial in the past.

mlir/test/Dialect/MQTOpt/Transforms/gate-cancellation.mlir
taminob: I disagree style-wise, but I'll remove it where it's possible :)

While this works for the `allocQubit`/`deallocQubit` operations, it doesn't for the others (`x`, `i`, ...). Maybe this is related to this comment (https://github.com/munich-quantum-toolkit/core/commit/8bb58888948246ab3486c678a595ca670c2eddc7#diff-011b16028e44c910bf3049b6c63f68938faf1908e25a7116e645b545948bedcbR182-R192) where it's mentioned that the "assembly format" needs to be set accordingly?

burgholzer: In the end, I have no strong opinion here and am happy to be convinced of the explicit typing. As long as it is consistent.

Pretty sure that's due to the assembly format, yes. Although I was almost certain that we also define one for the unitary operations. Maybe we don't define it "well enough" yet.

flowerthrower: I also like having the types there; MLIR can get tricky to read anyway, for me, this simplified reading test code (alternatively, this could be done by adding comments here and there, but for such simple programs not necessary). I would also agree to do it consistently.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Add translation from `QuantumComputation` to the `MQTRef` MLIR dialect ([#1099]) ([**@denialhaag**], [**@burgholzer**])
 - ✨ Add `reset` operations to the MLIR dialects ([#1106]) ([**@DRovara**])
 - ✨ Add MQT's implementation of a QDMI Device for neutral atom-based quantum computing ([#996], [#1010]) ([**@ystade**])
+- ✨ Add new pattern to `GateCancellation` (previously named `CancelConsecutiveInverses`) MLIR pass for identity gate removal ([#1140]) ([**@taminob**])
```
burgholzer: Some small remarks here:
- as noted at the top of this file, newer entries go first in the list
- you will need to add your username to the bottom of the file so that the link works correctly
- you will need to add the issue link to the bottom of the document so that the link works correctly.
- I'd argue that this should be split into separate changelog entries; one for the renaming from `CancelConsecutiveInverses` to `GateCancellation`, which should go in the `Changed` section and one for the addition of the identity removal pattern to the pass

Random side comment: maybe `GateElimination` would be an even more suitable name for the overall pass since technically, the identity removal is not really cancelling with anything.

mlir/lib/Dialect/MQTOpt/Transforms/GateCancellationPattern.cpp
```diff
@@ -128,13 +128,33 @@ struct CancelConsecutiveInversesPattern final
...
+ */
+struct RemoveIdentitiesPattern final : mlir::OpRewritePattern<IOp> {
+
+  explicit RemoveIdentitiesPattern(mlir::MLIRContext* context)
+      : OpRewritePattern(context) {}
+
+  mlir::LogicalResult
+  matchAndRewrite(IOp op, mlir::PatternRewriter& rewriter) const override {
+    auto inQubits = op.getAllInQubits();
+    auto outQubits = op.getAllOutQubits();
```
burgholzer: I am pretty sure this is not necessary as it is not used.
```suggestion
```

CHANGELOG.md
taminob: Oops, missed that - thanks :+1: 

I also added the rename to `GateElimination` in the latest version.

PR #1147, opened 2025-08-23
📝 Integrate generated MLIR documentation

### LINKED ISSUE #1127: 🧾 Docs - Integrate generated MLIR dialect docs into our docs site
## Background
CMake uses `add_mlir_doc(...)` to generate dialect and interface docs, but they aren’t surfaced in our docs site yet.

## Problem
Users can’t easily find dialect/op/interface documentation.

## Proposal:
- Update CMake/docs to copy or build generated docs into `docs/_build` (or Sphinx sources), and link them from `docs/mlir.md`.

## Examples:
```cmake
add_mlir_doc(MQTRefOps MLIRMQTRefDialect Dialects/ -gen-dialect-doc)
add_mlir_doc(MQTRefInterfaces MLIRMQTRefInterfaces Dialects/ -gen-op-interface-docs)
```

## Acceptance criteria:
- Generated artifacts included in published docs; navigation links present.

denialhaag:
## Description

This PR adds the generated MLIR documentation to our documentation. To this end, a custom CMake command is added that automatically copies the generated documentation. To generate and copy the docs, run the following command:

```bash
cmake --build build --target mlir-doc  
```

Fixes #1127

## Checklist:


docs/mlir/Dialects/MLIRMQTOptDialect.md
denialhaag, line 1: Generally speaking, most of these files (that is, the "docstrings" in the respective TableGen files) would benefit from some improvements. I'd be happy to scan the them for typos and obvious improvements so that the documentation is presentable from the beginning.

docs/mlir/Dialects/MLIRMQTOptDialect.md
```diff
@@ -0,0 +1,1862 @@
...
+(https://doi.org/10.1145/3491247)
+
+The `mqtopt` dialect supports dynamic as well as static qubit
+addressing.
+
+[TOC]
+
+## Operations
+
+[source](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/MQTOpt/IR/MQTOptOps.td)
```
denialhaag: Interestingly, this confidently points to the LLVM repository. I guess we could fix these links in the custom CMake command. 🤔

docs/mlir/Dialects/MLIRMQTOptDialect.md
burgholzer: That would be nice. And from then on out, we try to continuously add and improve the docs.

docs/mlir/Dialects/MLIRMQTOptDialect.md
burgholzer: I think this is an indication that we are not setting something correctly at the moment. I would assume it points to the main llvm repo because that is the default when working directly with LLVM.
Could you try and see if there is an option to change that?

docs/mlir.md
burgholzer, line 1: I think we could be dropping the sole FAQ entry here. It does not add too much value at the moment

burgholzer, line 1: A table of content would be nice here (probably at the very bottom). So that the markdown documents are more directly integrated.

CMakeLists.txt
```diff
@@ -115,6 +115,14 @@ option(BUILD_MQT_CORE_MLIR "Build the MLIR submodule of the MQT Core project" OF
...
   include(SetupMLIR)
   add_subdirectory(mlir)
+
+  # copy generated MLIR documentation
+  add_custom_command(
+    TARGET mlir-doc
+    POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_BINARY_DIR}/docs/
+            ${CMAKE_CURRENT_SOURCE_DIR}/docs/mlir/
+    COMMENT "Copying generated MLIR documentation")
```
burgholzer, line 125: Just some thoughts:
Initially I was thinking that we should not be checking in the copied markdown files. In the end, they are generated automatically and one should not edit them manually. They also very much mirror the text that is already checked in as part of the source code.
The advantage of having them checked in is that one can directly inspect them and inspect changes to those files, which might help to guard against certain unwanted changes.
However, I am still leaning towards not checking them in and just having them generated on demand (or rather, copied over) as part of the docs build (similar to how we handle the python api auto generation, where we also exclude the respective directory from VCS and have it generated at the docs build time)
It would also be great if there were some kind of index page so that one could include the whole mlir subfolder in the sphinx build. Maybe that already exists and just needs to be used.

denialhaag: We are generating the documentation in the Read the Docs job now. :tada:

I will look into the last point tomorrow. I would be nce to have an index page.

docs/mlir.md
denialhaag: Makes sense! Done.

docs/mlir/Dialects/MLIRMQTOptDialect.md
denialhaag: It took me a bit of time to find it, but it seems to be hardcoded: https://github.com/llvm/llvm-project/blob/main/mlir/tools/mlir-tblgen/OpDocGen.cpp#L283

Should I then just use `sed` in the custom CMake command to replace the links? 🤔

burgholzer: I just tried to see whether there is anything set in https://github.com/jmgorius/mlir-standalone-template for out of tree projects, but that was not the case.
So I suppose manual replacement is the only real option here.

Catalyst seems to be doing something else here if I look at their documentation page: https://docs.pennylane.ai/projects/catalyst/en/latest/code/dialects/quantum.html

denialhaag: Judging by this commit, they manually edited the Markdown files after the fact: https://github.com/PennyLaneAI/catalyst/pull/1959/commits/4612a3dab036cd2c928a4f39be38440442f5eece

(We have the same problem that we need to remove instances of `[TOC]` in some files.)

docs/mlir.md
denialhaag: I now sort of followed the approach in the Catalyst documentation and built more customized pages out of the generated pages. I think that might even be the idea because some of the generated pages' most top-level heading is `###`. Let me know what you think!

CMakeLists.txt
denialhaag: Regarding your second point, see comment above.

docs/mlir/Dialects/MLIRMQTOptDialect.md
denialhaag: I added two `sed` commands but something is still wrong. I'll fix it tomorrow. 👍

denialhaag: The`sed` commands are finally working. 😌

docs/mlir/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+This part of the MQT explores the capabilities of the Multi-Level Intermediate Representation (MLIR) in the context of compilation for quantum computing.
+
+We define multiple dialects, each with its dedicated purpose:
+
+- The {doc}`MQTOpt dialect <MQTOpt>` is designed for optimization.
+  We also provide a set of passes for the dialect.
+
+- The {doc}`MQTRef dialect <MQTRef>` uses reference semantics and is designed as a compatibility dialect that simplifies translations from and to existing languages such as QASM or QIR.
+  Also the MQTRef dialect is accompanied by passes.
```
burgholzer:
```suggestion
- The {doc}`MQTRef dialect <MQTRef>` uses reference semantics and is designed as a compatibility dialect that simplifies translations from and to existing languages such as Qiskit, OpenQASM, or QIR.

- The {doc}`MQTOpt dialect <MQTOpt>` uses value semantics and is mainly designed for running optimizations.

Both dialects define various transformation passes.
```
Or something like that.

docs/mlir/index.md
```diff
@@ -0,0 +1,30 @@
...
+See the [contribution guidelines](contributing.md) for more information.
+:::
+
+```{toctree}
+:maxdepth: 1
+:caption: Table of Contents
+
+MQTOpt Dialect <MQTOpt>
+MQTRef Dialect <MQTRef>
+Conversions
```
burgholzer: Is this somewhat missing the `Passes`? At least they are excluded in the `.gitignore`. If so, then these could be added as a link to the suggestion I put above as well.

burgholzer (approved):
I think this looks good for starters 🚀 
I just left a couple of comments inline that should be fairly easy to address. I'd propose to handle further updates to the docs themselves in subsequent PRs.

denialhaag:
> Just because I also looked at the actual RtD documentation a little bit now, I think it would be nice to add table of contents to the respective subpages. Or, if that is creating problems, to somehow tune the righthand side TOC on RtD so that it is a bit easier to read and use (limiting the depth for example).

This was somehow more complicated than I expected...

1. Furo doesn't allow one to change the depth of the contents sidebar: https://github.com/pradyunsg/furo/discussions/557
2. It only allows one to hide the contents sidebar altogether: https://pradyunsg.me/furo/customisation/toc/
3. When one wants to add a table of contents to the page, it yells at one to not do that. To suppress the error, the definition of the `this-will-duplicate-information-and-it-is-still-useful-here` class is necessary. 

Let me know what you think of the new dialect pages!

docs/mlir/index.md
denialhaag: There's no dedicated page for the passes now. They are just a subsection on the page of the respective dialect: https://mqt--1147.org.readthedocs.build/projects/core/en/1147/mlir/MQTRef.html#passes

If you prefer, we can change the depth of the `toctree` to 2. Then, there would also be links to the operations, types, passes, and interfaces of each dialect on the index page. 🤔

docs/mlir/MQTOpt.md
```diff
@@ -0,0 +1,36 @@
...
+:local:
+:backlinks:
+:class: this-will-duplicate-information-and-it-is-still-useful-here
+```
+
+## Operations
+
+```{include} Dialects/MLIRMQTOptDialect.md
+:start-after: "## Operations"
+```
```
denialhaag, line 23: Technically, this should work, but it doesn't. I have tried for half an hour now. 🤠 

````suggestion
```{include} Dialects/MLIRMQTOptDialect.md
:start-at: "## Operations"
```
````

denialhaag:
> Yeah, FURO makes this notoriously hard.. I like the direction; I'd slightly tweak it to
> 
> * include one more level on the main page
> * include one less level on the detailed pages
> * put the table of content after the summary (but I suppose that part is a bit harder to achieve). you could be reading the same file twice with different start and end markers supposedly

Done to all three points!

docs/mlir/index.md
denialhaag: Changed the depth to 2. 👍

burgholzer (approved):
Yeah, this looks like a great start 🎉🙌
Let's get this in once ci is green (no idea why the windows build caching is not working properly.. very high up on my list to actually improve that situation)

Issue #1125, opened 2025-08-15
🧊 MLIR - Eliminate qubit registers in MQTOpt (value semantics only)


burgholzer:
## Background
Registers make liveness/hoisting/reuse analysis harder and introduce boilerplate (`alloc → extract → use → insert → dealloc`).

## Problem
Analyses must reason through register indirections and partial reuse, increasing complexity and risk of missed optimizations.

## Proposal:
- In MQTOpt, allocate individual qubits and deallocate when done; convert from MQTRef register form during lowering to Opt.
- Update passes to operate on individual qubit values.

## Acceptance criteria:
- Conversion implemented; passes adjusted; tests show simpler IR and correct behavior.
- Measurement hoisting and qubit reuse analyses become simpler/clearer (tests or before/after diffs).

burgholzer:
I don't quite see the notion of registers going away in MQTRef. Basically every single language that currently comes to mind as an option for the input language has the concept of registers. 
I would also argue that this is not just out of nothing. These notions seem generally useful in those languages.

DRovara:
After some more offline discussions, we think it would be best to get rid of the `memref` usage for qubit registers completely.
Especially for `MQTOpt`, it drastically makes many optimizations more difficult, and for `MQTRef` the only advantage it really gives is the translation from other languages.

In the upcoming revamp of our MLIR dialects, we will remove any usages of `memref` for qubit registers.

PR #1148, opened 2025-08-24
🚧 Naive Quantum Program Mapping

### LINKED ISSUE #1081: ✨ MLIR - Naive Mapping Conversion from `MQTOpt` to `MQTMem`
### Problem Statement

A program represented in `MQTMem` implicitly carries the information that this program can be executed on the hardware that was specified during compilation. In particular, the resulting program must obey the coupling constraints of the hardware when accessing the statically addressed qubits. Hence, when translating `MQTOpt` to `MQTMem` the quantum computation must be mapped to the hardware which complicates the respective conversion.

### Proposed Solution

Implement a first naive conversion to `MQTMem` that naively inserts `SWAP` gates to make the quantum computation executable on the hardware.

MatthiasReumann:
## Description

This is the first attempt to solve the circuit routing problem for the MQT MLIR project. This pull request implements a naive mapping algorithm following shortest paths. Effectively resolves #1081. A follow-up pull request will implement an adjusted QMAP algorithm for MLIR. 

## Design Goals 
- Keeps open MQT-MLIR issues and developments in mind.
- Computes `head` and `look-ahead` layers on demand.
- Supports multiple "circuits" in a single module.
- Forwards as well as backwards mapping, a.k.a. the LightSABRE approach.
- Supports classically controlled operations (`scf.if`) and gate repetitions (`scf.for`).
- Supports initial layouts.

## Checklist:


MatthiasReumann:
## Discussion 

For later reference (and discussion) I will try to summarize my thoughts and ideas regarding the current implementation and how each of the design goals is addressed.  

### Design Goals 
#### Keeps open MQT-MLIR issues and developments in mind.
Firstly, the implementation assumes LLVM@21 in alignment with #1180. 

Secondly, the current algorithm ignores any `allocRegister` operations and assumes only the use of `AllocQubitOp`s and `DeallocQubitOp`s (aligned with #1125). Consequently, this strengthens the argument that registers can safely be removed from the `opt` dialect.

Lastly, as of now, only functions with an `entry_point` attribute are routed. However, with the current implementation it should _not be that difficult_ to route all other functions, if we apply the proposed strategy in #1152.

#### Supports multiple "circuits" in a single module and supports initial layouts.

As long as the qubit limit of the targeted architecture is not reached, the current implementation should be able to cope with any sequence of `alloc`s and `dealloc`'s. We achieve this by maintaining a queue of "unused" static qubit indices.

Initially, the queue is filled with the permutated indices of static qubits as defined by the initial layout. We re-generate (think: random layout) and re-apply the initial layout each time the number of allocated qubits hits zero. Essentially, this defines the boundaries of a "circuit" - even though we don't explicitly define it that way. 

#### Supports classically controlled operations (`scf.if`) and gate repetitions (`scf.for`).

Currently, this is not implemented. However, by walking the IR pre-order (visit parent first) - which we currently do - this _shouldn't be that troublesome_. An idea that I like (taken from the LightSABRE Paper) is that the `if` statement (and consequently the `for` as well) is simply another "gate" that requires mapping.

#### Forwards as well as backwards mapping, a.k.a. the LightSABRE approach.

This pull request implements forwards mapping. Backwards mapping should also be possible by using the `ReverseIterator` struct when walking the IR, swapping in and out qubits as well as `alloc`s and `dealloc`s. 

However, one thing to consider here is that we'll likely not require to commit changes made by the backwards mapping. This can work purely on a "logical" level without any rewrites taking place. Moreover, when applying routing back-and-forth only the very last forward pass will rewrite the IR. 

#### Computes `head` and `look-ahead` layers on demand.

Thanks to the Def-Use chain this task is easy.

MatthiasReumann:
## Discussion (Follow-Up)

> We achieve this by maintaining a queue of "unused" static qubit indices.

During testing I found out that the deallocated qubits might not be the same static qubits than initially allocated due to the inserted SWAPs (`makeExecutable`). To cope with this I added a set which tracks free static indices. However, instead of maintaining a queue + a set (which is totally doable) I utilized LLVM's SetVector which does this "for free". 

Essentially: The free static indices are now a stack instead of a queue. 

> Currently, this is not implemented. However, by walking the IR pre-order (visit parent first) - which we currently do - this shouldn't be that troublesome.

I've implemented the respective logic to deal with the `for` and `if` loop. 

Side note: When linear typing is not ensured, it is necessary to always replace SSA values with an additional if-check if the SSA Value is in the correct region. I'm not sure how much of an overhead this is - nonetheless noteworthy.

> We re-generate (think: random layout) and re-apply the initial layout each time the number of allocated qubits hits zero.

For simplicity the current implementation generates the initial layout once and re-applies it in every function (which for now is only the `entry_point`).

#### Using Pattern Drivers

I really, _really_, liked the idea of using patterns here. However, in the end I decided to go back to a simple walk & rewrite strategy without patterns. I've added the rationale to the description of the `route` function. I think, unfortunately, that patterns and the pattern drivers are simply the wrong tool to use here.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -83,4 +83,18 @@ def QuantumSinkPass : Pass<"quantum-sink", "mlir::ModuleOp"> {
   }];
 }
 
+def RoutingPass : Pass<"routing", "mlir::ModuleOp"> {
+  let summary = "This pass ensures that the connectivity constraints of a target architecture are met.";
+  let description = [{
+    This pass inserts SWAP operations to ensure two-qubit gates are executable on a given target architecture.
+  }];
+}
```
burgholzer: Should we give this a more specific name? This might not be the only routing pass we develop over here.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Architecture.h
burgholzer, line 1: I am not going to review this in very much detail as I strongly believe that this class will only be temporary.
It serves as a great blueprint though for functions/functionality that should go into the FoMaC library that builds on top of QDMI in MQT Core (https://github.com/munich-quantum-toolkit/core/blob/main/include/mqt-core/qdmi/FoMaC.hpp).
The coupling map is already available there, as well as the name and the number of qubits.
however, queries such as `areAdjacent`, or `shortestPathBetween` are not yet part of the library.
This is more of a mental note and an FYI than a request for changes here. Maybe some of this information could be put into the respective issue for incorporating architecture information.

mlir/lib/Dialect/MQTOpt/Transforms/QuantumSinkPass.cpp
burgholzer, line 1: Very minor, but this could go into a separate PR to not have to wait on this until the PR is merged.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Architecture.cpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+        const std::size_t sum = dist_[i][k] + dist_[k][j];
+        if (dist_[i][j] > sum) {
+          dist_[i][j] = sum;
+          prev_[i][j] = prev_[k][j];
+        }
+      }
+    }
+  }
+}
```
burgholzer, line 61: Should it be noted down somewhere that this has a time complexity of `O(|nqubits|^3)` and a space complexity of `O(|nqubits|^2)`? probably somewhere in the respective function docstring where this is used.
Separate thought: when should this be called in the future; at construction time of the FoMaC library or on demand the first time the respective information is queried? I'd lean to the latter.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+  /// @brief Pop the top from the stack.
+  void pop() { stack_.pop_back(); }
+  /// @brief Returns the size of the stack.
+  [[nodiscard]] std::size_t size() const { return stack_.size(); }
+  /// @brief Reset the state.
+  void reset() { stack_.clear(); }
+
+private:
+  llvm::SmallVector<QubitIndexMap> stack_;
+};
```
burgholzer: I should have probably gotten that from somewhere else, but couldn't this be a "proper" stack? Is `SmallVector` really the appropriate datastructure for this given how every `push_back` could trigger a complete reallocation of the stack?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+class VerificationStack {
+public:
+  /// @brief Put empty state on stack.
+  void push() { stack_.emplace_back(); }
+  /// @brief Put @p map on stack.
+  void push(const QubitIndexMap& map) { stack_.push_back(map); }
+  /// @brief Return top of stack.
+  [[nodiscard]] QubitIndexMap& top() { return stack_.back(); }
+  /// @brief Return the element below the top of stack.
+  [[nodiscard]] QubitIndexMap& belowTop() { return stack_[stack_.size() - 2]; }
```
burgholzer: Does this need some kind of access validation? at least an assertion would be reasonable I assume.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * @brief Verifies if the unitary acts on either zero, one, or two qubits:
+ * - Advances for zero qubit unitaries (Nothing to do)
+ * - Forwards SSA values for one qubit.
+ * - Verifies executability for two-qubit gates for the given architecture and
+ *   forwards SSA values.
+ */
+WalkResult handleUnitary(UnitaryInterface op, VerificationStack& stack,
+                         const Architecture& arch) {
+  const std::size_t nacts = op.getAllInQubits().size();
```
burgholzer: Nothing to change right here, but this is another aspect where we need to improve the design of our quantum dialect. This will construct the entire vector of qubits only to determine the size, which is wasteful. There should be a quick access function for getting the number of qubits without computing the union of targets and controls.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+  const Value in0 = op.getAllInQubits()[0];
+  const Value out0 = op.getAllOutQubits()[0];
+
+  if (nacts == 1) {
+    forwardOne(stack.top(), in0, out0);
+    return WalkResult::advance();
+  }
+
+  const Value in1 = op.getAllInQubits()[1];
+  const Value out1 = op.getAllOutQubits()[1];
```
burgholzer: Similar to the comment above, but I am simply not sure how clever the compiler is here.
I'd rather only get the InQubits and the OutQubits once and extract the respective indices separately.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+  void reset() { stack_.clear(); }
+
+private:
+  llvm::SmallVector<QubitIndexMap> stack_;
+};
+
+/**
+ * @brief Forward SSA values in map.
+ */
+void forwardOne(QubitIndexMap& map, const Value in, const Value out) {
```
burgholzer: This might deserve a better, more explicit name. Took me roughly half the file to understand what this is supposed to do. But maybe my brain is just slow ;o)

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
```diff
@@ -0,0 +1,276 @@
...
+              .Case<MeasureOp>([&](MeasureOp measure) {
+                return handleMeasure(measure, stack);
+              })
+              .Case<QubitOp>(
+                  [&](QubitOp op) { return handleQubit(op, stack, *arch); })
+              .Case<UnitaryInterface>([&](UnitaryInterface unitary) {
+                return handleUnitary(unitary, stack, *arch);
+              })
+              /// Skip the rest.
+              .Default([](auto) { return WalkResult::skip(); });
```
burgholzer, line 279: Should this explicitly handle `if` operations from the `scf` dialect?

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing-errors.mlir
```diff
@@ -0,0 +1,34 @@
...
+        mqtopt.deallocQubit %q0_0
+        mqtopt.deallocQubit %q1_0
+        mqtopt.deallocQubit %q2_0
+        mqtopt.deallocQubit %q3_0
+        mqtopt.deallocQubit %q4_0
+        mqtopt.deallocQubit %q5_0
+        mqtopt.deallocQubit %q6_0
+
+        return
+    }
```
burgholzer: This is not in particular for this issue, but again a more general comment that might be better placed in the respective issue. It would naturally be great to not have the architecture hard-coded for tests like these, but to programmatically specify a device where this leads to a violation.

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
```diff
@@ -0,0 +1,347 @@
...
+//
+// SPDX-License-Identifier: MIT
+//
+// Licensed under the MIT License
+
+// RUN: quantum-opt %s -split-input-file --pass-pipeline="builtin.module(routing,verify-routing)" -verify-diagnostics | FileCheck %s
+
+module {
+
+    // CHECK-LABEL: func.func @entrySABRE
```
burgholzer, line 13: I suppose this is only provisional anyway, but wouldn't these tests benefit from more elaborate `CHECK` strings? or is this essentially covered by the `verify-routing` pass? (which would make this fairly elegant I have to say)

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
```diff
@@ -0,0 +1,347 @@
...
+            %q1_2, %q0_2 = mqtopt.x() %q1_1 ctrl %q0_1 : !mqtopt.Qubit ctrl !mqtopt.Qubit
+
+            scf.yield %q0_2, %q1_2 : !mqtopt.Qubit, !mqtopt.Qubit
+        }
+
+        mqtopt.deallocQubit %q0_2
+        mqtopt.deallocQubit %q1_2
+
+        return
+    }
```
burgholzer, line 223: This is kind of missing a description, isn't it?

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
```diff
@@ -0,0 +1,347 @@
...
+
+    func.func @entryBranching() attributes { entry_point } {
+        %true = arith.constant 1 : i1
+
+        %q0_0 = mqtopt.allocQubit
+        %q1_0 = mqtopt.allocQubit
+
+        %q0_1 = mqtopt.h() %q0_0 : !mqtopt.Qubit
+        %q0_2, %q1_2 = scf.if %true -> (!mqtopt.Qubit, !mqtopt.Qubit) {
```
burgholzer: Would it make sense to test this with the `if` being conditioned on a measurement result?

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
```diff
@@ -0,0 +1,347 @@
...
+        return
+    }
+
+    // CHECK-LABEL: func.func @entryAll
+    func.func @entryAll() attributes { entry_point } {
+
+        //
+        // All of the above quantum computations in a single entry point.
+        // This test shows that the algorithm can handle multiple computations
+        // in a single function.
```
burgholzer, line 231: I like this very much. This would essentially be a demonstration of hybrid quantum classical computation during the coherence time of the quantum computer.

Would it make sense to complement this with a single module that contains multiple functions that are each marked with entry point?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+/**
+ * @brief Return output qubit pair for a two-qubit unitary.
+ * @param u A two-qubit unitary.
+ * @return Pair of SSA values consisting of the first and second out-qubits.
+ */
+[[nodiscard]] std::pair<Value, Value> getOuts(UnitaryInterface u) {
+  assert(isTwoQubitGate(u));
+  return {u.getAllOutQubits()[0], u.getAllOutQubits()[1]};
+}
```
burgholzer, line 115: Nothing to be changed in this PR (I suppose), but functions like these should be made available from the dialects themselves as they will be needed across multiple places.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+      /* out_qubits = */ resultTypes,
+      /* pos_ctrl_out_qubits = */ TypeRange{},
+      /* neg_ctrl_out_qubits = */ TypeRange{},
+      /* static_params = */ nullptr,
+      /* params_mask = */ nullptr,
+      /* params = */ ValueRange{},
+      /* in_qubits = */ inQubits,
+      /* pos_ctrl_in_qubits = */ ValueRange{},
+      /* neg_ctrl_in_qubits = */ ValueRange{});
+}
```
burgholzer, line 144: I also hate that this is so complicated at the moment in our dialects.
This should basically be as simple as
```
return rewriter.create<SWAPOp>(location, in0, in1);
```

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+ * @param nqubits The number of qubits.
+ * @param seed The seed used for randomization.
+ */
+[[nodiscard, maybe_unused]] SmallVector<std::size_t>
+getRandomLayout(const std::size_t nqubits, const std::size_t seed) {
+  std::mt19937_64 rng(seed);
+  auto layout = getIdentityLayout(nqubits);
+  std::shuffle(layout.begin(), layout.end(), rng);
+  return layout;
+}
```
burgholzer: from an interface point of view it might make more sense to feed this with a generator instead of with the seed. that way, this function won't always construct a generator and the generator can be an object of the pass itself.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+//===----------------------------------------------------------------------===//
+// State (Permutation) Management
+//===----------------------------------------------------------------------===//
+
+/**
+ * @brief Manage mapping between SSA values and physical hardware indices.
+ * This class maintains the bi-directional mapping between software and hardware
```
burgholzer: "software qubits" is an interesting term. I'd still rather refer to these as "program qubits" probably.
"Hardware qubits" is totally fine and established.
I like the note down below. These are not only C++ keywords; logical and physical qubits also have separate meanings in other contexts such as error correction. Hence, the use of different naming makes a lot of sense.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+   * index.
+   */
+  explicit LayoutState(ArrayRef<Value> qubits,
+                       ArrayRef<std::size_t> initialLayout)
+      : qubits_(qubits.size()), softwareToHardware_(qubits.size()) {
+    valueToMapping_.reserve(qubits.size());
+
+    for (const std::size_t softwareIdx : initialLayout) {
+      const QubitInfo info{.hardwareIdx = initialLayout[softwareIdx],
+                           .softwareIdx = softwareIdx};
```
burgholzer: I am probably stupid right now, but does this really do what it is supposed to do?
Judging from the docstring, `initialLayout` is supposed to be a map from `software` (index) to `hardware` (value). However
```
for (const std::size_t softwareIdx : initialLayout) 
```
iterates over the values and not the indices (if I am not overlooking something).

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+private:
+  struct StackItem {
+    explicit StackItem(LayoutState state) : state(std::move(state)) {}
+
+    LayoutState state;
+    SmallVector<SoftwareIndexPair, 32> history;
+  };
+
+  SmallVector<StackItem, 2> stack_;
```
burgholzer: Similar question to the other file: should these be proper stacks instead of vectors?
As far as I can tell neither of them really require random access.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+ *
+ * @param old A range of values to check for included qubits.
+ * @param qubits The hardware qubits.
+ * @return Vector of missing hardware qubits with increasing index.
+ */
+[[nodiscard]] SmallVector<Value> getMissingQubits(ValueRange old,
+                                                  ArrayRef<Value> qubits) {
+  const SmallVector<Value> oldResults(old.begin(), old.end());
+  llvm::DenseSet<Value> included;
+  included.insert(oldResults.begin(), oldResults.end());
```
burgholzer: This seems overly complex. Does 
```suggestion
  llvm::DenseSet<Value> included(old.begin(), old.end());
```
not work?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+  WalkResult handleUnitary(UnitaryInterface op, RoutingStack& stack,
+                           const Architecture& arch, HardwareIndexPool& pool,
+                           PatternRewriter& rewriter) {
+    // Global-phase or zero-qubit unitary: Nothing to do.
+    if (op.getAllInQubits().empty()) {
+      return WalkResult::advance();
+    }
+
+    // Single-qubit: Forward mapping.
+    if (!isTwoQubitGate(op)) {
```
burgholzer: Shouldn't this more explicitly check that this is a single-qubit operation? at this point, it could also be a 3qb+ operation.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    }
+
+    if (!isExecutable(op, stack, arch)) {
+      makeExecutable(op, stack, arch, pool, rewriter);
+    }
+
+    const auto [execIn0, execIn1] = getIns(op);
+    const auto [execOut0, execOut1] = getOuts(op);
+
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-do-while)
```
burgholzer: Where is the `do-while` in this?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    }
+  }
+};
+
+/**
+ * @brief Inserts SWAPs along the shortest path between two hardware
+ * qubits.
+ */
+class NaiveRouter final : public Router {
+protected:
```
burgholzer: Does it make sense to split this into a general `RoutingInterface` (in a separate file) and a dedicated implementation of a router in this file? Also relates to one of the other comments on the structural setup of the passes.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+      });
+
+      auto swap = createSwap(op->getLoc(), qIn0, qIn1, rewriter);
+      const auto [qOut0, qOut1] = getOuts(swap);
+
+      rewriter.setInsertionPointAfter(swap);
+      replaceAllUsesInRegionAndChildrenExcept(
+          qIn0, qOut1, swap->getParentRegion(), swap, rewriter);
+      replaceAllUsesInRegionAndChildrenExcept(
+          qIn1, qOut0, swap->getParentRegion(), swap, rewriter);
```
burgholzer, line 648: I suppose this is one of the places where the code would become much simpler if we were to assume linear types, right?
In a straight line program without control flow we very much enforce this by design. The only place where I am not quite sure this would hold is when `scf.for` and `scf.if` constructs are involved. I am not yet familiar with how the def-use chain looks like there.

However, this feels like more complexity than necessary to me at the moment.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+      rewriter.setInsertionPointAfter(swap);
+      replaceAllUsesInRegionAndChildrenExcept(
+          qIn0, qOut1, swap->getParentRegion(), swap, rewriter);
+      replaceAllUsesInRegionAndChildrenExcept(
+          qIn1, qOut0, swap->getParentRegion(), swap, rewriter);
+
+      stack.recordSwap(softwareIdx0, softwareIdx1);
+      stack.getState().swapSoftwareIndices(qIn0, qIn1);
+      stack.getState().remapQubitValue(qIn0, qOut0);
+      stack.getState().remapQubitValue(qIn1, qOut1);
```
burgholzer: Just because I am not sure how clever compilers are these days, but I would still probably write statements like these as
```suggestion
      auto& state = stack.getState();
      state.swapSoftwareIndices(qIn0, qIn1);
      state.remapQubitValue(qIn0, qOut0);
      state.remapQubitValue(qIn1, qOut1);
```

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+      : arch(std::move(arch)), router(std::move(router)) {
+    initialLayout = getIdentityLayout(this->arch->nqubits());
+  }
+
+  explicit RoutingContext(std::unique_ptr<Architecture> arch,
+                          std::unique_ptr<Router> router,
+                          const std::size_t seed)
+      : arch(std::move(arch)), router(std::move(router)) {
+    initialLayout = getRandomLayout(this->arch->nqubits(), seed);
+  }
```
burgholzer: I think it could make sense to separate the concept of an initial layout a little bit more from the routing conceptionally.
I'd imagine an interface for defining a layout, that is implemented in the `Identity` and `Random` Layout variants. Then, similar to how this function here takes a `std::unique_ptr<Router> router`, it would take a `std::unique_ptr<Layouter> layouter` (better name needed).

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    return op.emitOpError(
+        "requires one too many qubits for the targeted architecture");
+  }
+
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-do-while)
+  LLVM_DEBUG({ llvm::dbgs() << "alloc index: " << *index << '\n'; });
+  const Value q = stack.getState().lookupHardware(*index);
+
+  auto reset = rewriter.create<ResetOp>(op.getLoc(), q);
+  rewriter.replaceOp(op, reset);
```
burgholzer, line 912: This will unconditionally insert `Reset` operations. Would it make sense to somehow track if this is the first allocation of a qubit (which does not require an active reset) and only add resets for subsequent allocations of the same qubit within the same entry point?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
```diff
@@ -0,0 +1,1058 @@
...
+ * @note
+ * We consciously avoid MLIR pattern drivers: Idiomatic MLIR transformation
+ * patterns are independent and order-agnostic. Since we require state-sharing
+ * between patterns for the transformation we violate this assumption.
+ * Essentially this is also the reason why we can't utilize MLIR's
+ * `applyPatternsGreedily` function. Moreover, we require pre-order traversal
+ * which current drivers of MLIR don't support. However, even if such a driver
+ * would exist, it would probably not return logical results which we require
+ * for error-handling (similarly to `walkAndApplyPatterns`). Consequently, a
+ * custom driver would be require in any case, which adds unnecessary code to
```
burgholzer:
```suggestion
 * custom driver would be required in any case, which adds unnecessary code to
```

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: I just stumbled upon the terms _circuit qubits_ and _device qubits_ when reading up on the MQT Core IR: https://mqt.readthedocs.io/projects/core/en/latest/mqt_core_ir.html#layout-information

Maybe use these to be consistent with the MQT IR?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Yes! That's a great idea. Will look into it.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
MatthiasReumann: Yes exactly. With linear typing (and for `scf.for` alone) a simple `replaceAllUsesExcept` would suffice. It's branching ops that cause trouble. (Even though the `replaceAllUsesInRegionAndChildrenExcept` could be simplified to just check the direct parent)

I'm also not 100% sure how much of a "problem" this really is since the majority of runtime will eventually come from the `Router`.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Inside the preprocessor directive `LLVM_DEBUG`, unfortunately.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Ah, yes. This is probably a left over from an older version of this rewrite function. Will update!

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
MatthiasReumann: Yes, definitely! That's a bug. Good catch 🕵🏻‍♂️

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
burgholzer: Yeah. That was my try in the past to avoid logical and physical, which can still be found quite a bit throughout the MQT Core IR in C++.
However, there is a general trend in the community to move "beyond circuits" or at least beyond the notion of a circuit and rather refer to "quantum programs". Straight line programs without control flow would be the equivalent of a quantum circuit. Any program with any kind of control flow would be a "structured program".
These terms seem to establish themselves more and more. So I generally think that "program qubits" would be a very suitable name here. But I am open to other suggestions in principle 😌

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
burgholzer: That is really kind of unfortunate.
To be honest, I'd be willing to drop this rule entirely in the clang-tidy config. I am not sure that I am personally a big fan of it.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
burgholzer: This is probably not really a problem complexity wise. Although one shouldn't underestimate the impact of a bunch of linear or quadratic overheads that might accumulate.
Maybe something for our discussion tomorrow. I'd like to understand a little more where the limitations here are when it comes to simplification.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Yes I agree with that. I was going to say that "circuit" might not be that fitting anymore today.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: The function isn't required anymore altogether: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: The `std::stack` class internally uses a `std::deque` which we now also do. It makes a lot of sense to use deque here because of the copying of elements on resizing - especially since our items are rather large (1x`DenseMap`, 2x`SmallVector`).

Moreover, I've added a new `RoutingStack.h` file which implements a templated `RoutingStack<Item>`. This stack is now used both in the routing and the verification pass. Hence, reduces code duplication. 

Ergo: Done.

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
MatthiasReumann: Done.

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
MatthiasReumann: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
MatthiasReumann: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
MatthiasReumann: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
burgholzer: Very nice! 👏🏼

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
burgholzer: Nice!

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
MatthiasReumann: Yes. These should definitely also be optimized and not construct temporary vectors on each call.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
MatthiasReumann: Done.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
MatthiasReumann: `--route-sc`?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingVerificationPass.cpp
MatthiasReumann: Resolved by adding the `RoutingStack.h` file. Done.

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing-errors.mlir
MatthiasReumann: I agree on that. I've only added such a test to pass the coverage tests 🙈

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
MatthiasReumann: As discussed yesterday, the `verify-routing` pass should ensure the correctness of the routing pass.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Updated `software` to `program`. Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Yeah. There was a bug in there. I mixed up hardware and program indices. Should be fixed now. 

(Even though I think this constructor will go away some time soon - thinking about the dynamic allocation strategy we discussed yesterday)

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: I've updated the `.clang-tidy` file accordingly: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: I've added the `InitialLayoutGeneratorBase` (short: `ilg`) class to deal with that.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
burgholzer: Quick follow-up comment here: do we plan to only have one (superconducting) routing pass that has multiple options for layout and routing or do we rather want to split these into separate passes? I'd be slightly leaning towards the first option. At that point `--route-sc` seems reasonable. maybe also "RoutingPassSC" for the class?

MatthiasReumann: Sounds good. I like the option with the options `arch=` and `method=` too. However, I would deal with this in a separate PR.

burgholzer: Yeah, totally! Let's strive to get this merged asap 😌 just ping me whenever I can do another review.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
MatthiasReumann: I've added a respective check to only reset qubits that have already been used: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: The new random generator class takes the generator as argument: Done.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/RoutingPass.cpp
MatthiasReumann: Good idea. Maybe let's do this also in a follow-up PR.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Architecture.cpp
MatthiasReumann: I've added the time and space complexity to the doxygen. I really like the idea of lazy loading. One caveat though: when benchmarking the routing algorithm one should carefully consider this since other algorithms (SABRE, for example) doesn't include this computation in their benchmarks.

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
MatthiasReumann: I am not quite sure if I understand. Isn't this not already the case with the functions above the `@entryAll` function?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Architecture.cpp
burgholzer: Good point. It might make sense to very carefully split the timing information for the routing so that it is clear what is pre-processing, actual routing, and post-processing. Feel free to resolve this comment here for now.

mlir/test/Dialect/MQTOpt/Transforms/Transpilation/naive-routing.mlir
burgholzer: Ah, indeed it is. I was assuming that the above tests are split up by file check but I now realised that they are not and that this is simply one huge test. Sorry for the confusion here.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Architecture.h
```diff
@@ -0,0 +1,91 @@
...
+  [[nodiscard]] llvm::SmallVector<std::size_t>
+  shortestPathBetween(std::size_t u, std::size_t v) const;
+
+private:
+  using Matrix = llvm::SmallVector<llvm::SmallVector<std::size_t>>;
+
+  /**
+   * @brief Find all shortest paths in the coupling map between two qubits.
+   * @details Vertices are the qubits. Edges connected two qubits. Has a time
+   * and complexity of O(nqubits^3) and O(nqubits^2), respectively.
```
burgholzer:
```suggestion
   * and memory complexity of O(nqubits^3) and O(nqubits^2), respectively.
```

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/RoutingStack.h
```diff
@@ -0,0 +1,85 @@
...
+   */
+  [[nodiscard]] Item& getItemAtDepth(std::size_t depth) {
+    assert(depth < stack_.size() && "getItemAtDepth: depth out of bounds");
+    return stack_[stack_.size() - 1 - depth];
+  }
+
+  /**
+   * @brief Pushes a new item on to the stack.
+   */
+  void push(Item item) { stack_.emplace_back(item); }
```
burgholzer:
```suggestion
  void push(const Item& item) { stack_.emplace_back(item); }
```
Otherwise this will most likely create an unnecessary copy of `item`. The alternative would be
```
  void push(Item item) { stack_.emplace_back(std::move(item)); }
```
I am not 100% sure if the compiler is clever enough to turn the current code into the latter solution automatically.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+    for (const auto& qubit : llvm::drop_end(layout_)) {
+      llvm::dbgs() << qubit << " ";
+    }
+    llvm::dbgs() << layout_.back();
+  }
+#endif
+
+protected:
+  SmallVector<QubitIndex> layout_;
+};
```
burgholzer, line 224: Let's keep this interface for now. However, I'd imagine this will have to drastically change in the near term future.
The layout stage will want to have information about the architecture (connectivity, fidelities. etc.) and it will want to have information about the circuit/program (interaction graph, gate frequency, etc.).
I would propose to make these part of the general interface, even though they might not be used by the fairly static passes.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+   * @param rng A random number generator.
+   */
+  explicit RandomLayoutGenerator(const std::size_t nqubits, std::mt19937_64 rng)
+      : InitialLayoutGeneratorBase(nqubits), rng_(rng) {
+    generate();
+  }
+
+  void generate() override {
+    std::iota(this->layout_.begin(), this->layout_.end(), 0);
+    std::shuffle(this->layout_.begin(), this->layout_.end(), rng_);
```
burgholzer:
```suggestion
  void generate() override {
    std::iota(layout_.begin(), layout_.end(), 0);
    std::shuffle(layout_.begin(), layout_.end(), rng_);
```
The `this` is superfluous here supposedly and I personally think it is rather verbose. I see this as a matter of taste though (provided that this actually compiles without the `this`)

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+  /**
+   * @brief Set of free hardware indices.
+   *
+   * The SetVector ensures a deterministic iteration order.
+   */
+  llvm::SetVector<QubitIndex> freeHardwareIndices_;
+};
+
+/**
+ * @brief Returns true iff @p u is executable on the targeted architecture.
```
burgholzer:
```suggestion
 * @brief Returns true iff @p op is executable on the targeted architecture.
```

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+  }
+};
+
+/**
+ * @brief The necessary datastructures to route a quantum-classical module.
+ */
+struct RoutingContext {
+  explicit RoutingContext(std::unique_ptr<Architecture> arch,
+                          std::unique_ptr<RouterBase> router,
+                          std::unique_ptr<InitialLayoutGeneratorBase> ilg)
```
burgholzer, line 669: This is fine for now. I just wanted to note that I am not particularly sure if these will remain as unique pointers here in the near future. Is ownership of these objects really managed by the Context?
For the ILG and the Router, I could imagine that to be true. But for the Device handle, I fail to see that. The resources for that will be managed more centrally (e.g., in the QDMI Driver) and all you'll get here is a handle to be used.
But, as said, nothing to be changed here. So feel free to resolve this after acknowledging the fact.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+  llvm::SmallVector<Value> qubits(ctx.arch().nqubits());
+  for (std::size_t i = 0; i < qubits.size(); ++i) {
+    auto qubitOp =
+        rewriter.create<QubitOp>(rewriter.getInsertionPoint()->getLoc(), i);
+    rewriter.setInsertionPointAfter(qubitOp);
+    qubits[i] = qubitOp.getQubit();
+  }
+
+  ctx.pool().fill(ctx.ilg().getLayout());
+  ctx.stack().emplace(LayoutState(qubits, ctx.ilg().getLayout()));
```
burgholzer:
```suggestion
  ctx.stack().emplace(qubits, ctx.ilg().getLayout());
```
if set up correctly, this should also work and avoid the construction of a temporary.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+}
+
+/**
+ * @brief Defines the end of a region (and hence routing state) defined by a
+ * function. Consequently, we pop the region's state from the stack. Since
+ * we currently only route entry_point functions we do not need to return
+ * all static qubits here.
+ */
+WalkResult handleReturn(StateStack& stack) {
+  assert(!stack.empty() && "handleReturn: stack must not be empty");
```
burgholzer:
```suggestion
```
While it doesn't hurt, I believe this assertion is redundant as it is almost exactly replicated in the `stack.pop` method.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+ *
+ * Restores layout by uncomputation and replaces (invalid) yield.
+ *
+ * The results of the yield op are extended by the missing hardware
+ * qubits, similarly to the 'for' and 'if' op. This is only possible
+ * because we restore the layout - the mapping from hardware to program
+ * qubits (and vice versa).
+ *
+ * Using uncompute has the advantages of (1) being intuitive and
+ * (2) preserving the optimally of the original SWAP sequence.
```
burgholzer:
```suggestion
 * (2) preserving the optimality of the original SWAP sequence.
```

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
```diff
@@ -0,0 +1,1071 @@
...
+                     stack.getStateAtDepth(1).getCurrentLayout()) &&
+         "layouts must match after restoration");
+
+  const SmallVector<Value> newResults(stack.topState().getHardwareQubits());
+  rewriter.replaceOpWithNewOp<scf::YieldOp>(op, newResults);
+
+  stack.pop();
+
+  return WalkResult::advance();
+}
```
burgholzer, line 887: Again, nothing for this PR, but maybe for a slightly sooner follow-up:
This is not necessarily optimal. In many cases, there may be better ways to revert from the final permutation back to the initial permutation than to simply play back the SWAPs in reverse.
At the top of my head, I can't quite remember how hard the classical problem of permutation synthesis out of locally restricted SWAPs is, but I would guess that a different heuristic would perform a bit better here.
Such a heuristic already exists in MQT Core as part of the DD package: https://github.com/munich-quantum-toolkit/core/blob/5058297f58fb7a2efdea9c2ca6ef6caf39edfed1/include/mqt-core/dd/Operations.hpp#L248-L304
This transforms from one permutation to another in a greedy fashion without connectivity constraints.

burgholzer (approved):
This looks great! I left a couple of comments throughout. Some are purely cosmetic and you should be able to just batch them up and apply them directly.
Another bunch is just random comments and observations that are not be tackled as part of this PR, but rather as part of a follow up.
Lastly, it would be great to add a CHANGELOG entry for this feature before we merge this.
Would be cool to get this done today 🔥 🚀

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
MatthiasReumann: Totally agree with you. I think the routing context (similarly to the initial layout handling) will change dramatically at some point in the future.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/RoutingPass.cpp
MatthiasReumann: Yup. I think this is actually a quite interesting problem to research and eventually solve efficiently.

MatthiasReumann: Thinking about it. This could actually be part of the router implementation.

burgholzer: Yeah, there could be different strategies for this that one can choose from. Could be generalised to "structured program handling" or similar.

MatthiasReumann: Absolutely! Either way it would also be pretty nice to get rid of the SWAP history for the naive approach.

Issue #1134, opened 2025-08-15
🧩 QDMI - FoMaC C++ abstraction and Python bindings (device API and Qiskit backend)


burgholzer:
## Background
#996 and #1010 landed a first QDMI device (neutral atoms) and a QDMI driver, laying the groundwork for broader QDMI support as envisioned in #772. What’s still missing is a higher-level, object-oriented façade over the C-based client interface (FoMaC = Figures of Merit and Constraints) and Python bindings.

- FoMaC goal: Provide an ergonomic C++ abstraction over the low-level C client interface while retaining full coverage. The same C++ abstraction should be part of our MLIR compiler to query device capabilities/constraints for devices that are exposed to the compiler through QDMI.
- Python goal: Expose FoMaC to Python so devices can be surfaced to frontends (e.g., as Qiskit backends with a corresponding `Target`).
- Prior art: A functional-style FoMaC example exists in the QDMI repo but it’s limited and not suited for direct, object-oriented Python exposure: https://github.com/Munich-Quantum-Software-Stack/QDMI/tree/develop/examples/fomac
- Binding constraint: Neither pybind11 nor nanobind expose raw opaque pointers nicely; the C opaque handles must be wrapped behind C++ value/RAII types first, and only those types should be exposed to Python.

## Problem
We need a consistent, OO C++ FoMaC library that safely wraps the QDMI client interface (opaque handles, lifetimes, errors), models device capabilities/constraints in compiler-friendly types, and is easily bindable to Python. We also need a first Qiskit BackendV2 adapter that builds a `Target` from FoMaC properties and routes job submission via QDMI.

## Proposal:
Design and implement FoMaC in C++, add Python bindings, and provide a minimal Qiskit adapter. This section is mostly LLM generated. So take it with a grain of salt. It still shows the general direction and some options.

1) C++ FoMaC library (headers + src)
- Core types (value or RAII wrappers):
  - Driver (enumeration of devices, sessions), Device (properties, capabilities, constraints), Session (auth), Job (submit, poll/callback), Program (payloads/text/binary), Results (shots, counts, metadata).
  - Capabilities/constraints: qubits, topology, gate set, timing/latency, calibration FOMs, result schemas; prefer typed structs with JSON serde as needed.
- Ownership/lifetimes: Wrap C opaque handles in RAII classes with custom deleters; prohibit exposing raw pointers in public headers; use spans/views for buffers.
[...]

PR #1150, opened 2025-08-25
✨ Add MQT's implementation of a generic FoMaC with Python bindings

ystade:
## Description

This PR comes in a series of PRs (#996, #1010, #1100). It implements a generic FoMaC on top of the QDMI Driver that facilitates an object-oriented use of QDMI. On top of the C++ implementation, it adds Python bindings such that the object-oriented interface can also be used from Python.

Fixes: #1134 

## Checklist:


burgholzer:
> I do not know whether throwing/raising an exception is the best way to go if a property is not supported. Other ideas on that are very welcome.

One idea would be to use `std::optional` for that and return `std::nullopt_t` in case a property is not supported. In cases where another error code is reported, throwing exceptions still seems the way to go for me.

ystade:
@burgholzer, I increased the test coverage and improved the documentation on QDMI aspects within MQT. I'll put the reconstruction of the device, including the zones defined via lattices, in a different PR to (1) not increase the size of this PR unnecessarily and (2) to get this PR merged sooner rather than later.

docs/qdmi/driver.md
```diff
@@ -0,0 +1,38 @@
...
+
+A QDMI Driver manages the communication between QDMI devices, such as [MQT Core's NA QDMI Device](na_device.md), and QDMI clients, see the [QDMI specification](https://munich-quantum-software-stack.github.io/QDMI/).
+It is responsible for loading the device, forwarding requests from the client to the device, and sending back the results.
+The MQT Core's QDMI Driver, {cpp:class}`qdmi::Driver`, comes with the [MQT Core's NA QDMI Device](na_device.md) that is already statically linked into the driver and can directly be used.
+Other devices can be loaded dynamically at runtime via {cpp:func}`qdmi::Driver::addDynamicDeviceLibrary`.
+
+## Python Bindings
+
+The QDMI Driver is implemented in C++ and exposed to Python via [pybind11](https://pybind11.readthedocs.io).
+Direct binding of QDMI driver functions is not feasible due to technical limitations.
```
burgholzer: > QDMI driver functions

Should this be "QDMI Client interface" or similar?

python/mqt/core/qdmi/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+# Copyright (c) 2025 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""Package for Quantum Device Management Interface (QDMI) functionalities."""
```
burgholzer:
```suggestion
"""Submodule for Quantum Device Management Interface (QDMI) functionality."""
```

pyproject.toml
```diff
@@ -42,6 +42,9 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dynamic = ["version"]
+dependencies = [
+    "deepdiff>=8.6.0",
+]
```
burgholzer: I don't quite see where this dependency is coming from or coming in. I'd very much like to remove it (and update the lock file).

include/mqt-core/qdmi/FoMaC.hpp
```diff
@@ -0,0 +1,636 @@
...
+  case QDMI_ERROR_PERMISSIONDENIED:
+    return "Permission denied.";
+  case QDMI_ERROR_NOTSUPPORTED:
+    return "Not supported.";
+  case QDMI_ERROR_BADSTATE:
+    return "Bad state.";
+  case QDMI_ERROR_TIMEOUT:
+    return "Timeout.";
+  default:
+    return "Unknown status code.";
```
burgholzer: Just a consistency consideration: the other `toString` methods do not add a `.` to the end of the strings. Might be worth removing this here.

include/mqt-core/qdmi/FoMaC.hpp
```diff
@@ -0,0 +1,636 @@
...
+ * `queryProperty` method.
+ * @tparam T The type to check.
+ * @see Operation::queryProperty
+ */
+template <typename T>
+concept maybe_optional_value_or_string_or_vector =
+    value_or_string_or_vector<remove_optional_t<T>>;
+
+/// @returns the string representation of the given QDMI_STATUS.
+constexpr auto toString(QDMI_STATUS result) -> std::string {
```
burgholzer: Can we move the definitions of these functions to the CPP file to keep the header slightly cleaner?

include/mqt-core/qdmi/FoMaC.hpp
```diff
@@ -0,0 +1,636 @@
...
+    break;
+  case QDMI_WARN_GENERAL:
+    std::cerr << "Warning: " << msg << "\n";
+    break;
+  default:
+    throwError(result, msg);
+  }
+}
+
+class Site {
```
burgholzer, line 242: These classes could use slightly elaborate docstrings in my opinion.

include/mqt-core/qdmi/FoMaC.hpp
```diff
@@ -0,0 +1,636 @@
...
+  friend class Device;
+  friend class Operation;
```
burgholzer, line 244: This might just be me, but why exactly is this necessary? (holds for the other classes as well)

python/mqt/core/qdmi/fomac.pyi
```diff
@@ -0,0 +1,125 @@
...
+    def blocking_radius(self, sites: list[Site] = ..., params: list[float] = ...) -> int | None:
+        """Returns the blocking radius of the operation."""
+    def idling_fidelity(self, sites: list[Site] = ..., params: list[float] = ...) -> float | None:
+        """Returns the idling fidelity of the operation."""
+    def is_zoned(self, sites: list[Site] = ..., params: list[float] = ...) -> bool | None:
+        """Returns whether the operation is zoned."""
+    def sites(self, sites: list[Site] = ..., params: list[float] = ...) -> list[Site] | None:
+        """Returns the list of sites the operation can be performed on."""
+    def mean_shuttling_speed(self, sites: list[Site] = ..., params: list[float] = ...) -> int | None:
+        """Returns the mean shuttling speed of the operation."""
```
burgholzer: I am just noting this here, even though this is nothing for this PR. since pybind11 v3, there is a nice way of making these bindings slightly more generic by using `collections.abc` classes for the annotations. I would guess that most of the `list` here could be `Iterable` or similar.

test/python/qdmi/test_fomac.py
```diff
@@ -0,0 +1,369 @@
...
+
+    Returns:
+        tuple[Device, Operation]: A tuple containing a quantum device instance and one of its operations.
+    """
+    device = request.param
+    operation = device.operations()[0]
+    return device, operation
+
+
+def testdevice_name(device: Device) -> None:
```
burgholzer: very minor nitpick: I would appreeciate a `_` between the word "test" and the actual test name here.

burgholzer (changes requested):
This is much more of an approval than a request for changes. I only have minor comments left that should be fairly straightforward to address.
Really really great to see this! And I am very glad how this turned out (quite interesting to see concepts being used 😌)

include/mqt-core/qdmi/FoMaC.hpp
ystade: Then these functions cannot be constexpr anymore, but I guess they are not called in a constexpr context anyways, so this is actually not a concern. WIll do.

include/mqt-core/qdmi/FoMaC.hpp
ystade: The objective is that only the device can create instances of Sites and Operations since they can only exist as part of a device. Hence, the constructor is private and they declare the classes that must be able to construct them as friends. Do you have a better idea?

python/mqt/core/qdmi/fomac.pyi
ystade: I changed it to Iterable. Is there anything else I have to adapt?

ystade:
@burgholzer I should have addressed all comments and added to replies above.

docs/qdmi/driver.md
```diff
@@ -18,7 +18,7 @@ Other devices can be loaded dynamically at runtime via {cpp:func}`qdmi::Driver::
 ## Python Bindings
 
 The QDMI Driver is implemented in C++ and exposed to Python via [pybind11](https://pybind11.readthedocs.io).
-Direct binding of QDMI driver functions is not feasible due to technical limitations.
+Direct binding of the QDMI Clietn interface functions is not feasible due to technical limitations.
```
burgholzer:
```suggestion
Direct binding of the QDMI Client interface functions is not feasible due to technical limitations.
```

burgholzer (approved):
Let's get this in 😌 Great work on this!

python/mqt/core/qdmi/fomac.pyi
burgholzer: Should be good for now. I believe one can override the signature in the bindings file itself to also have Python's help and so on report this correctly. But that's something for a follow up.

include/mqt-core/qdmi/FoMaC.hpp
burgholzer: One could make the Site and Operation class nested classes of the device class. However, I am fine with the current solution here. Let's keep it for now and iterate on it if we should need that.

PR #1151, opened 2025-08-25
:sparkles: Implement elide permutations pass in MLIR

taminob:
## Description

This PR adds a MLIR pass to eliminate all SWAP gates. This can be used on unmapped circuits to reduce the amount of needed swaps or in combination with another optimization called swap reconstruction to remove repeated CNOTs on two qubits.

Related to https://github.com/munich-quantum-toolkit/core/issues/1122

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add new MLIR pass `ElidePermutations` for SWAP gate elimination ([#1151]) ([**@taminob**])
 - ✨ Add new pattern to MLIR pass `GateElimination` for identity gate removal ([#1140]) ([**@taminob**])
 - ✨ Add Clifford block collection pass to `CircuitOptimizer` module ([#885]) ([**jannikpflieger**], [**@burgholzer**])
 - ✨ Add `isControlled()` method to the `UnitaryInterface` MLIR class ([#1157]) ([**@taminob**], [**@burgholzer**])
@@ -181,6 +182,7 @@ _📚 Refer to the [GitHub Release Notes](https://github.com/munich-quantum-tool
 <!-- PR links -->
 
 [#1157]: https://github.com/munich-quantum-toolkit/core/pull/1157
+[#1151]: https://github.com/munich-quantum-toolkit/core/pull/1151
 [#1147]: https://github.com/munich-quantum-toolkit/core/pull/1147
 [#1140]: https://github.com/munich-quantum-toolkit/core/pull/1140
 [#1139]: https://github.com/munich-quantum-toolkit/core/pull/1139
--- a/mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
+++ b/mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
@@ -45,6 +45,25 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
   }];
 }
 
+def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
+  let summary = "This pass removes any permutation (i.e., SWAP gates) it encounters and modifies the order of the qubits instead.";
+  let description = [{
+    Elides permutations (i.e., SWAP gates) by removing them from the circuit and, instead, permuting the respective qubit values.
+
+    This optimization will not preserve the current qubit mapping and has no awareness of potential connectivity limitations.
+
+    Example:
+    ```
+    // before
+    %q0_1, %q1_1 = mqtopt.swap() %q0_0, %q1_0 : !mqtopt.Qubit, !mqtopt.Qubit
+    %q1_2, %q0_2 = mqtopt.x() %q1_1 ctrl %q0_1 : !mqtopt.Qubit ctrl !mqtopt.Qubit
+
+    // after
+    %q0_1, %q1_1 = mqtopt.x() %q0_0 ctrl %q1_0 : !mqtopt.Qubit ctrl !mqtopt.Qubit
+    ```
+  }];
+}
+
 def QuantumSinkPass : Pass<"quantum-sink", "mlir::ModuleOp"> {
   let summary = "This pass attempts to push down operations into branches for possible optimizations.";
   let description = [{
--- a/mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutations.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutations.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+ * Copyright (c) 2025 Munich Quantum Software Company GmbH
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#include "mlir/Dialect/Common/Compat.h"
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h"
+
+#include <mlir/IR/PatternMatch.h>
+#include <mlir/Support/LLVM.h>
+#include <utility>
+
+namespace mqt::ir::opt {
+
+#define GEN_PASS_DEF_ELIDEPERMUTATIONS
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h.inc"
+
+/**
+ * @brief This pattern attempts to remove SWAP gates by re-ordering qubits.
+ */
+struct ElidePermutations final
+    : impl::ElidePermutationsBase<ElidePermutations> {
+
+  void runOnOperation() override {
+    // Get the current operation being operated on.
+    auto op = getOperation();
+    auto* ctx = &getContext();
+
+    // Define the set of patterns to use.
+    mlir::RewritePatternSet patterns(ctx);
+    populateElidePermutationsPatterns(patterns);
+
+    // Apply patterns in an iterative and greedy manner.
+    if (mlir::failed(APPLY_PATTERNS_GREEDILY(op, std::move(patterns)))) {
+      signalPassFailure();
+    }
+  }
+};
+
+} // namespace mqt::ir::opt
--- a/mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+ * Copyright (c) 2025 Munich Quantum Software Company GmbH
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h"
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h"
+
+#include <cassert>
+#include <llvm/ADT/STLExtras.h>
+#include <mlir/IR/MLIRContext.h>
+#include <mlir/IR/PatternMatch.h>
+#include <mlir/IR/ValueRange.h>
+#include <mlir/Support/LLVM.h>
+#include <mlir/Support/LogicalResult.h>
+
+namespace mqt::ir::opt {
+
+/**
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Add translation from `QuantumComputation` to the `MQTRef` MLIR dialect ([#1099]) ([**@denialhaag**], [**@burgholzer**])
 - ✨ Add `reset` operations to the MLIR dialects ([#1106]) ([**@DRovara**])
 - ✨ Add MQT's implementation of a QDMI Device for neutral atom-based quantum computing ([#996], [#1010]) ([**@ystade**])
+- ✨ Add new MLIR pass `ElidePermutations` for SWAP gate elimination ([#1140]) ([**@taminob**])
```
burgholzer: Should be moved to the top of the list. (although we might really have to rethink this at some point. Apparently it feels more natural for people to extend the list at the bottom)

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -46,6 +46,14 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
   }];
 }
 
+def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
+  let summary = "This pass searches for consecutive applications of rotation gates that can be merged.";
```
burgholzer: That seems to be a copy&paste error.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -46,6 +46,14 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
...
+  let description = [{
+    Remove SWAP gates and permutate order of Qubits instead.
+    This optimization does only work for unmapped circuits which do not have any connectivity restraints.
```
burgholzer: A little more detailed explanation would not hurt here. This eventually ends up in our documentation (which will soon also be published as part of #1147).
What is here in the description could be the brief at the moment.

I'd turn the brief into something like
> Elides permutations (i.e., SWAP gates) by removing them from the circuit and, instead, permuting the respective qubit values.

Then, in the description, it would be great to have a fairly brief example of the pass in action. (MWE-style)
I'd also slightly rephrase the note about connectivity to rather have it point out that this optimization is not connectivity-preserving.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutations.cpp
```diff
@@ -0,0 +1,45 @@
...
+    // Define the set of patterns to use.
+    mlir::RewritePatternSet patterns(ctx);
+    populateElidePermutationsPatterns(patterns);
+
+    // Apply patterns in an iterative and greedy manner.
+    if (mlir::failed(APPLY_PATTERNS_GREEDILY(op, std::move(patterns)))) {
+      signalPassFailure();
+    }
+  }
+};
```
burgholzer, line 43: Just out of curiosity: could you try and check whether this kind of boilerplate code is actually necessary? we keep writing that for all of our passes and it feels so repetitive that I can't help but think this could be automated by tablegen. Maybe tablegen has some convenient shortcuts for defining passes and/or patterns. 

Would be nice if you could read into that a little bit.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
```diff
@@ -0,0 +1,64 @@
...
+ * @brief This pattern attempts to remove SWAP gates by re-ordering qubits.
+ */
+struct ElidePermutationsPattern final : mlir::OpRewritePattern<SWAPOp> {
+
+  explicit ElidePermutationsPattern(mlir::MLIRContext* context)
+      : OpRewritePattern(context) {}
+
+  mlir::LogicalResult
+  matchAndRewrite(SWAPOp op, mlir::PatternRewriter& rewriter) const override {
+    const bool isControlledSwap = !op.getAllCtrlInQubits().empty();
```
burgholzer: We should probably make queries like these a little simpler and more direct. Would you mind adding an `isControlled()` function to the unitary interface in a separate PR?

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (isControlledSwap) {
```
burgholzer: Best case, this simply becomes 
```suggestion
    if (op.isControlled()) {
```
after addressing the comment above.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    auto inQubits = op.getInQubits();
+
+    std::vector<mlir::Value> swappedQubits;
+    for (auto&& reversed : llvm::reverse(inQubits)) {
+      swappedQubits.push_back(reversed);
+    }
+
+    rewriter.replaceAllOpUsesWith(op, swappedQubits);
```
burgholzer: Given how it is compile-time known that the SWAP will always have two targets, I believe this logic could be even simpler. Something like
```suggestion
    rewriter.replaceAllOpUsesWith(op, {inQubits[1], inQubits[0]});
```
should just work and achieve the same result

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
```diff
@@ -0,0 +1,64 @@
...
+};
+
+/**
+ * @brief Populates the given pattern set with the `ElidePermutationsPattern`.
+ *
+ * @param patterns The pattern set to populate.
+ */
+void populateElidePermutationsPatterns(mlir::RewritePatternSet& patterns) {
+  patterns.add<ElidePermutationsPattern>(patterns.getContext());
+}
```
burgholzer, line 56: I know that this is only three lines of code, but I am still wondering if that boilerplate is necessary or if one could just include this as part of the tablegen definition somehow. Also totally fine if the answer to all of that is simply: no, it's necessary.
I just have the feeling that some of these patterns and passes are so simple that they could probably be written in tablegen itself (which has support for defining passes in some form).

mlir/test/Dialect/MQTOpt/Transforms/elide-permutations.mlir
```diff
@@ -0,0 +1,110 @@
...
+  func.func @testMultiSwapElidePermutation() {
+    // CHECK: %[[Q0_0:.*]] = mqtopt.allocQubit
+    // CHECK: %[[Q1_0:.*]] = mqtopt.allocQubit
+    // CHECK: %[[Q2_0:.*]] = mqtopt.allocQubit
+
+    // ========================== Check for operations that should not be canceled ==============================
+    // CHECK-NOT: %[[ANY:.*]] = mqtopt.swap()
+
+    // ========================== Check for operations that should not be canceled ==============================
+    // CHECK: %[[ANY:.*]], %[[ANY:.*]] = mqtopt.swap() %[[ANY:.*]], %[[ANY:.*]] ctrl %[[ANY:.*]]
```
burgholzer: I think it would be worth to check that the arguments of the remaining SWAP are correct instead of just asserting `ANY`.

burgholzer (changes requested):
Thanks @taminob for this further pass! Great to see how simple the implementation of that pass is in MLIR. That was quite painful to implement in MQT Core as part of the circuit optimizer.

CHANGELOG.md
taminob: It was me twice, but yes, it feels more intuitive to extend a list at the end. But once you know that new entries should be added to the front, that is more convenient since you need to scroll less if the list is already pretty long.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: Does the ASCII art you usually use for these examples work in the documentation? What kind of markup language is used (if it is not already monospaced by default)?

burgholzer: This will be translated to markdown. If you merge the main branch in, you can actually see the output as part of the docs build in the PR (just click on the RtD build once it completed and navigate to the MLIR section).
I think MLIR Code blocks would be totally fine for examples.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutations.cpp
taminob: I can have a look if I find anything, but I don't think it should become a part of this PR.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
taminob: Since the `GateElimination` in #1140 has two patterns which need to be added, I don't think that it is possible (or worth it) to make this even shorter. But I'll have a look at it (together with the `ElidePermutations` class).

mlir/test/Dialect/MQTOpt/Transforms/elide-permutations.mlir
taminob: I'll add it - it's just a little bit cumbersome because somehow the two qubits get bundled into one array by the operation (see https://github.com/munich-quantum-toolkit/core/blob/1d8a42a09a1f6b2dbacee222a8ad798223171673/mlir/test/Dialect/MQTOpt/Transforms/elide-permutations.mlir#L54). :)

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutations.cpp
burgholzer: I agree. This is supposedly a follow-up. The right pointers would probably be
- https://mlir.llvm.org/docs/DeclarativeRewrites/
- https://mlir.llvm.org/docs/PDLL/

Both look very promissing for reducing boilerplate code.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
burgholzer: I did some reading on the declarative pattern rewrites yesterday and what I saw there looked quite promising. (But also something for a follow up PR)

mlir/test/Dialect/MQTOpt/Transforms/elide-permutations.mlir
burgholzer: Ah, that's indeed a bit cumbersome. Generally, I would hope that we can at some point find better ways to write these tests to increase the iteration speed.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
taminob: Done, see #1157

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
taminob: I'll do this once the other PR (#1157) is merged.

burgholzer:
> I noticed (again) that there are also the `iswap` and `iswapdg` gates. Does it make sense to also eliminate these? Can the factor `i` also be represented as some kind of rotation?

Hm. I never really thought about this to be honest.
It doesn't sound completely unreasonble to expect that there is a decomposition of the iSWAP gate into a SWAP gate and local modifications. One would have to find that though.

mlir/lib/Dialect/MQTOpt/Transforms/ElidePermutationsPattern.cpp
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h"
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h"
+
+#include <cassert>
+#include <llvm/ADT/STLExtras.h>
+#include <mlir/IR/MLIRContext.h>
+#include <mlir/IR/PatternMatch.h>
+#include <mlir/IR/Value.h>
```
burgholzer:
```suggestion
```

burgholzer (approved):
Nice 👍🏻

PR #1158, opened 2025-08-27
:sparkles: Implement (controlled) swap reconstruction MLIR pass

taminob:
## Description

Add new MLIR pass for swap reconstruction. This pass checks for three consecutive CNOTs that are equivalent to a SWAP and replaces them with a SWAP operation.
In combination with elide permutations (#1151) this can be used to eliminate CNOTs.

Additionally, the pass adds two self-negating CNOTs if only two of the required CNOTs are found in order to have 2 CNOTs -> CNOT + SWAP which in combination with the elide permutations pass will still be better than before. (not yet implemented)

Related to #1122 

## Checklist:

### DIFF
--- a/mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
+++ b/mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
@@ -65,10 +65,30 @@ def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
 }
 
 def SwapReconstruction : Pass<"swap-reconstruction", "mlir::ModuleOp"> {
-  let summary = "This pass searches for CNOTs that are equivalent to a SWAP and converts them into a SWAP gate.";
+  let summary = "This pass searches for CNOTs that can be merged into a SWAP.";
   let description = [{
-    Multiple CNOTs that are equivalent to a SWAP are merge and converted into a SWAP gate.
-    Additionally, it may insert two self-cancelling CNOTs to allow for the insertion of a SWAP.
+    Three CNOTs that are equivalent to a SWAP are directly replaced with a (potentially controlled) SWAP gate.
+
+    For swap reconstructions with just two CNOTs (utilizing an insertion of two self-cancelling CNOTs), only non-controlled SWAP gates are inserted.
+
+    Examples of swap reconstructions:
+
+    ```
+         ┌───┐         ┌───┐
+    ──■──┤ X ├    ──■──┤ X ├──■────■──    ──╳────■──
+    ┌─┴─┐└─┬─┘ => ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐ =>   |  ┌─┴─┐
+    ┤ X ├──■──    ┤ X ├──■──┤ X ├┤ X ├    ──╳──┤ X ├
+    └───┘         └───┘     └───┘└───┘         └───┘
+    ```
+
+    ```
+    ──■────■────■──    ──■──
+      |  ┌─┴─┐  |        |
+    ──■──┤ X ├──■── => ──╳──
+    ┌─┴─┐└─┬─┘┌─┴─┐      |
+    ┤ X ├──■──┤ X ├    ──╳──
+    └───┘     └───┘
+    ```
   }];
 }
 
--- a/mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
@@ -29,30 +29,31 @@
 
 namespace mqt::ir::opt {
 /**
- * @brief This pattern attempts to find three CNOT gates next to each other
- * which are equivalent to a SWAP. These gates will be removed and replaced by a
- * SWAP operation.
+ * @brief This pattern attempts to find CNOT patterns which can be replaced by a
+ * SWAP gate.
  *
- * Examples:
+ * Example (onlyMatchFullSwapPattern=false):
  *       ┌───┐         ┌───┐
  *  ──■──┤ X ├    ──■──┤ X ├──■────■──    ──╳────■──
  *  ┌─┴─┐└─┬─┘ => ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐ =>   |  ┌─┴─┐
  *  ┤ X ├──■──    ┤ X ├──■──┤ X ├┤ X ├    ──╳──┤ X ├
  *  └───┘         └───┘     └───┘└───┘         └───┘
  *
- *  ──■────■──    ──■────■────■────■──    ──■────■──
- *    |  ┌─┴─┐      |  ┌─┴─┐  |    |        |    |
- *  ──■──┤ X ├ => ──■──┤ X ├──■────■── => ──╳────■──
- *  ┌─┴─┐└─┬─┘    ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐      |  ┌─┴─┐
- *  ┤ X ├──■──    ┤ X ├──■──┤ X ├┤ X ├    ──╳──┤ X ├
- *  └───┘         └───┘     └───┘└───┘         └───┘
- *
+ * Example (onlyMatchFullSwapPattern=false, matchControlledSwap=true):
  *  ──□────□──    ──□────□────□────□──    ──□────□──
  *    |  ┌─┴─┐      |  ┌─┴─┐  |    |        |    |
  *  ──■──┤ X ├ => ──■──┤ X ├──■────■── => ──╳────■──
  *  ┌─┴─┐└─┬─┘    ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐      |  ┌─┴─┐
  *  ┤ X ├──■──    ┤ X ├──■──┤ X ├┤ X ├    ──╳──┤ X ├
  *  └───┘         └───┘     └───┘└───┘         └───┘
+ *
+ * Example (onlyMatchFullSwapPattern=true, matchControlledSwap=true):
+ *  ──■────■────■──    ──■──
+ *    |  ┌─┴─┐  |        |
+ *  ──■──┤ X ├──■── => ──╳──
+ *  ┌─┴─┐└─┬─┘┌─┴─┐      |
+ *  ┤ X ├──■──┤ X ├    ──╳──
+ *  └───┘     └───┘
  */
 template <bool onlyMatchFullSwapPattern, bool matchControlledSwap>
 struct SwapReconstructionPattern final : mlir::OpRewritePattern<XOp> {


mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -45,6 +45,14 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
   }];
 }
 
+def SwapReconstruction : Pass<"swap-reconstruction", "mlir::ModuleOp"> {
+  let summary = "This pass searches for CNOTs that are equivalent to a SWAP and converts them into a SWAP gate.";
```
burgholzer: This doesn't quite cover the second application pattern. Maybe one can rephrase this slightly.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -45,6 +45,14 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
...
+  let description = [{
+    Multiple CNOTs that are equivalent to a SWAP are merge and converted into a SWAP gate.
+    Additionally, it may insert two self-cancelling CNOTs to allow for the insertion of a SWAP.
```
burgholzer: Small typo in "merge" -> "merged"

But more importantly, this could use examples in the docstring that show very simple demonstrations of this pass in action. I'd like to make this a habit when we are writing new passes that we include proper documentation that renders nicely on readthedocs.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstruction.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <utility>
+
+namespace mqt::ir::opt {
+
+#define GEN_PASS_DEF_SWAPRECONSTRUCTION
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h.inc"
+
+/**
+ * @brief This pattern attempts to merge consecutive rotation gates.
+ */
```
burgholzer: This is a pass and not a pattern. Also, this is copy pasted from somewhere else.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstruction.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    // Define the second set of patterns to be used.
+    patterns.clear();
+    populateAdvancedSwapReconstructionPatterns(patterns);
+
+    // Apply patterns in an iterative and greedy manner.
+    if (mlir::failed(APPLY_PATTERNS_GREEDILY(op, std::move(patterns)))) {
+      signalPassFailure();
+    }
+  }
```
burgholzer: This does not quite make sense to me. Why is this not just a second pattern as part of the pass?

I'd argue we could take this one step further and make these patterns part of the ElidePermutations pass as this is where they are most useful. That should work quite well.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h"
+#include "mlir/Dialect/MQTOpt/Transforms/Passes.h"
+#include "mlir/IR/BuiltinAttributes.h"
```
burgholzer:
```suggestion
#include <mlir/IR/BuiltinAttributes.h>
```

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+   */
+  [[nodiscard]] static bool isReverseCNotPattern(XOp& a, XOp& b) {
+    // TODO: allow negative ctrl qubits (at least for first CNOT)?
+    auto ctrlQubitsA = a.getPosCtrlOutQubits();
+    auto ctrlQubitsB = b.getPosCtrlInQubits();
+    auto targetQubitsA = a.getOutQubits();
+    auto targetQubitsB = b.getInQubits();
+
+    return ctrlQubitsA.size() == 1 && ctrlQubitsB.size() == 1 &&
+           ctrlQubitsA == targetQubitsB && targetQubitsA == ctrlQubitsB;
```
burgholzer: It might be possible to generalize this pattern for any combination of control types. You can always represent the negative control as a positive control sandwiched by X gates, and the X gate commutes with the target of the other CNOT, which leaves you with the existing pattern optionally sandwiched by X gates on both qubits. That should then simplify further in case there target of the resulting CNOT is sandwiched by X gates (which cancel).
I believe one could hardcode the 4 options here depending on the polarity of the controls.

Independent of that, you need to assert that the total number of controls is only two here. Right now, this function could be entered with a multi-controlled gate.
Although, technically, one can even extend the pattern to those gates.
The pattern applies as long as the target in one gate is a control in the other gate and vice versa; and as long as all other controls agree on their polarity.

Hope the above makes sense. Happy to provide some more examples.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  mlir::LogicalResult
+  matchAndRewrite(XOp op, mlir::PatternRewriter& rewriter) const override {
+    // operation can only be part of a series of CNOTs that are equivalent to a
+    // SWAP if it has exactly one control qubit
+    auto ctrlQubits = op.getPosCtrlInQubits();
+    if (ctrlQubits.size() != 1) {
+      return mlir::failure();
+    }
```
burgholzer: See comment above, this isn't fully true and the pattern can be more general than that.

In case of the swap reconstruction of 3 CNOTs, I see multiple generalizations:
- the target of the first gate is the control of the second and vice versa, the third gate also has the target of the first gate and the same control. All three gates share the same other controls. The result is a controlled SWAP with all the additional shared controls as its controls.
- the first and last two gates are identical CNOTs and the middle gate is has flipped control and target plus arbitrarily many other control qubits with arbitrary polarity. The result is again a controlled SWAP with all additional controls of the middle gate.


Maybe one can think slightly harder to also come up with ways for handling negative controls in this pattern, but I am not sure. And maybe this is not worth the time.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    auto& firstCNot = op;
+    if (auto secondCNot = findCandidate(firstCNot)) {
+      if (auto thirdCNot = findCandidate(*secondCNot)) {
+        //         ┌───┐
+        // a: ──■──┤ X ├──■── a    a: ──╳── b
+        //    ┌─┴─┐└─┬─┘┌─┴─┐   =>      |
+        // b: ┤ X ├──■──┤ X ├ b    b: ──╳── a
+        //    └───┘     └───┘
```
burgholzer: Really like these visualizations in here! Those would be perfect for the docstrings in the public documentation.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+      }
+      if constexpr (advancedSwapReconstruction) {
+        //         ┌───┐              ┌───┐                        ┌───┐
+        // a: ──■──┤ X ├ a    a: ──■──┤ X ├──■────■── a    a: ──╳──┤ X ├ b
+        //    ┌─┴─┐└─┬─┘   =>    ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐   =>      |  └─┬─┘
+        // b: ┤ X ├──■── b    b: ┤ X ├──■──┤ X ├┤ X ├ b    b: ──╳────■── a
+        //    └───┘              └───┘     └───┘└───┘
+        replaceWithSwap(rewriter, firstCNot);
+        swapTargetControl(rewriter, *secondCNot);
+        return mlir::success();
```
burgholzer: This is just a second pattern in my opinion. So it should also be treated as such.
In my head, a single pattern is something that I can write down as a rewrite rule similar to the graphical implementation.

In the declarative format of MLIR patterns (or the PDLL), each of these would be separate (and probably fairly concise and short; we should really give those a try at some point).

I'd also separate them here.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+        eraseOperation(rewriter, *secondCNot);
+        eraseOperation(rewriter, firstCNot);
+        return mlir::success();
+      }
+      if constexpr (advancedSwapReconstruction) {
+        //         ┌───┐              ┌───┐                        ┌───┐
+        // a: ──■──┤ X ├ a    a: ──■──┤ X ├──■────■── a    a: ──╳──┤ X ├ b
+        //    ┌─┴─┐└─┬─┘   =>    ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐   =>      |  └─┬─┘
+        // b: ┤ X ├──■── b    b: ┤ X ├──■──┤ X ├┤ X ├ b    b: ──╳────■── a
+        //    └───┘              └───┘     └───┘└───┘
```
burgholzer: The last CNOT in this picture is not correct. There is no reason why it would suddenly swap its control and target just because some gates before it have been combined.
I also think that the qubits in the end should still be labeled as they are in the beginning. The relabeling only happens if the swap is then elided.
This also holds for the example above and is even more apparent there: if you would elide the SWAP, you would end up with a circuit that reads a on top and b on bottom again, just as the original circuit. However, the original circuit does not implement the identity; it is a SWAP after all.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+    auto newSwapLocation = op->getLoc();
+    auto newSwapInQubits = op.getAllInQubits();
+    assert(newSwapInQubits.size() == 2);
+
+    rewriter.setInsertionPointAfter(op);
+    auto newSwap = rewriter.create<SWAPOp>(
+        newSwapLocation, mlir::TypeRange{qubitType, qubitType},
+        mlir::TypeRange{}, mlir::TypeRange{}, mlir::DenseF64ArrayAttr{},
+        mlir::DenseBoolArrayAttr{}, mlir::ValueRange{}, newSwapInQubits,
+        mlir::ValueRange{}, mlir::ValueRange{});
```
burgholzer: Is there really no more efficient builder for this operation at the moment? This seems quite cumbersome.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,201 @@
...
+      assert(op.getNegCtrlInQubits().empty());
+      op->setOperands(
+          {op.getPosCtrlInQubits().front(), op.getInQubits().front()});
+    });
+  }
+
+  static mlir::PatternBenefit calculateBenefit() {
+    // prefer simple swap reconstruction
+    return advancedSwapReconstruction ? 10 : 100;
+  }
```
burgholzer: Technically, one could even just implement the advanced pattern and not even implement the 3 CNOT reconstruction as it follows from the first pattern by eliding permutations and cancelling CNOTs.
I am not 100% sure if one can cover all potential use cases I outlined above with that, but I would generally expect this to work just fine.
One could essentially throw these patterns into a single elide permutations pass, i.e.
- inverse CNOT pattern
- swap elision
- simple CNOT cancellation

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: I'll experiment with adding the ASCII/UTF-8 art examples from the implementation here in markdown code blocks :+1:

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstruction.cpp
taminob: I tried simply adding this as a second pattern, but it looks like the patterns are matched operation by operation and thus, if the pattern application starts in the middle of a three CNOT pattern, it will apply the "advanced" two-CNOT-swap-reconstruction instead of the simple one. Even adding the benefit didn't work because for the mid CNOT the simple one simply doesn't apply. I also experimented with a separate `match()` and `rewrite()`, but same issue.

This was in my opinion the most elegant way to avoid defining a completely separate pass. I'm sure this could also become a part of the `ElidePermutations` pass (which would need a rename then), but this might make it harder to understand since the other compilers also have this as separate ones. My previous experience with compiler construction is very limited, but I would've expected these passes to be on the smaller side and then repeat them instead.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: > Independent of that, you need to assert that the total number of controls is only two here.

You mean one, right? But true, I forgot a `{a,b}.getNegCtrlOutQubits().empty()` check here, otherwise it should work since there is the `ctrlQubitsA.size() == 1` check. Does this actually work if there are multiple control qubits? What would happen with these control qubits when merged?

> You can always represent the negative control as a positive control sandwiched by X gates, and the X gate commutes with the target of the other CNOT, which leaves you with the existing pattern optionally sandwiched by X gates on both qubits.

I get how you can convert the negative control into a positive one, but I don't understand the simplification and how that doesn't break the pattern.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: > the target of the first gate is the control of the second and vice versa, the third gate also has the target of the first gate and the same control. All three gates share the same other controls. The result is a controlled SWAP with all the additional shared controls as its controls.

Ah, I wasn't sure about the controlled SWAP and thus didn't implement it. I can look into that if you think that is worth it.
Especially with the negative controls, I don't think it is worth it since the idea of these compiler passes is exactly to not do this kind of stuff by hand and instead have e.g. a separate pass that will convert negative into positive controls and then another pass that will re-order and cancel negations. I think handling all of these cases in this one pattern might become error-prone - I already tripped over all kind of stuff when implementing this much simpler pass (like re-ordering (which is not necessary) the qubits due to the introduced SWAP).

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: I wanted to avoid the code duplication and thus decided to put it in the same class - for sure not the prettiest piece of code.
Maybe we should instead introduce free "helper" functions (like the `eraseOperation` which is also used 1:1 like this in the remove identities pattern) to allow the re-use of code.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: Good catch, I forgot to update these images - I had that wrong idea that I need to swap when adding a swap, but you are right that this is not necessary. Caused my quite some trouble :)

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: I didn't find anything, this is simply how the operations are defined. All the `PatternRewriter` methods require all constructor arguments (more accurately `build` arguments) and that is the simplest overload.
I can maybe add names for all the `params`, `static_params` etc. instead of the empty constructor calls.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: Yes, I also thought about this once I was done with everything... I guess this would need some testing if the compiler is able to always perform these three passes in the correct order to achieve the same result.

taminob:
@burgholzer thanks for all the effort you put into your answer and all the examples.

I think we're actually basically on the same page, I probably just misused pattern/pass and thought of a pass what you describe as a pattern. I just didn't want to mash everything into one huge pattern. If we want to make larger passes, we might also need to reconsider the "apply greedily" approach which already didn't work perfectly for the two patterns I had here in this PR.

Regarding the PDLL link you've already sent in the other PR, I also read into it and fully agree that it looks like exactly what we want (e.g. the remove identity and elide permutations patterns can be something like 3 lines each). I just wanted to discuss the next steps first with @flowerthrower on Tuesday. Last time we considered to already go on with the compiler pipeline after the elide permutations/swap reconstruction passes (or patterns, I guess) since that might already work as a PoC, but I'm open to discuss other next steps as well (we just need to keep the deadline in mind :) ).

taminob:
@burgholzer regarding the controlled swap reconstruction, I wasn't entirely sure and tried to look up the decomposition of a controlled swap.
According to [this paper](https://quantum-journal.org/papers/q-2022-03-30-676/) the decomposition would look like this:
image

So the controlled swap reconstruction definitely is possible, but I think your example above is not fully correct (or I didn't interpret it correctly).

burgholzer:
That's the optimized decomposition.
This is equivalent to the version I showed above. You can simply drop the controls on the outer CNOTs. If the controls of the middle CNOT trigger, this makes a (controlled) SWAP. If the controls do not trigger, the remaining CNOTs cancel.

Thus, both reconstructions are feasible.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstruction.cpp
taminob: This is also wrong in the other pass, but I guess I shouldn't correct it in this PR, right?

burgholzer:
> I did not really understand your comment "# or, better, because it saves one elision" for "Pos-pos control MCX replacement" and implemented it the other way for now since it is prettier/easier code-wise. Isn't both after one elision the same and one permutation elision is necessary anyway?

Before I start to look into the code in more detail, a quick comment on this one:
```mlir
mcx q[0] ctrl q[1] <controls: arbitrary other positive or negative controls>
mcx q[1] ctrl q[0] <controls>

# transforms to
swap q[0], q[1] <controls>
mcx q[0] ctrl q[1] <controls>

# or, better, because it saves one elision
mcx q[1] ctrl q[0] <controls>
swap q[1], q[0] <controls>
```

Both of the decompositions would be feasible (as they are equivalent). However, the second case requires one less SWAP elision because it's the `mcx` gate first and the `swap` gate afterwards. If one would go with the first decomposition, the very next step (as part of the elision) would be to transform it to the second decomposition. 
Although, maybe, this doesn't even make too much of a difference because of the local nature of the elision pass in the MQTOpt dialect. It's just exchanging the qubit SSA values for one gate (no matter where that gate is placed).
At that point, one could think about directly getting rid of the SWAP (in cases where there are no controls) as part of the pattern. Wouldn't that be even easier?

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -64,6 +64,34 @@ def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
   }];
 }
 
+def SwapReconstruction : Pass<"swap-reconstruction", "mlir::ModuleOp"> {
+  let summary = "This pass searches for CNOTs that can be merged into a SWAP.";
```
burgholzer, line 68: This is not fully correct or at least does not cover the exact functionality of the pass.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -64,6 +64,34 @@ def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
   }];
 }
 
+def SwapReconstruction : Pass<"swap-reconstruction", "mlir::ModuleOp"> {
```
burgholzer, line 67: Why is this still a separate pass? I thought we wanted to combine this with the permutation elision one. Which will demand a proper name for the combined pass.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -64,6 +64,34 @@ def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
...
+  let summary = "This pass searches for CNOTs that can be merged into a SWAP.";
+  let description = [{
+    Three CNOTs that are equivalent to a SWAP are directly replaced with a (potentially controlled) SWAP gate.
```
burgholzer, line 70: This is not fully correct. A CNOT, per definition, only has a single-control.
X gates with multiple controls are typically referred to as multi-controlled X gates or multi-controlled Toffoli gates.

I believe the general description here could be streamlined and clarified a little bit. I am happy to take a stab at this once the functionality is in the "right" place and moderately consolidated.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -64,6 +64,34 @@ def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
...
+    ```
+
+    ```
+    ──■────■────■──    ──■──
+      |  ┌─┴─┐  |        |
+    ──■──┤ X ├──■── => ──╳──
+    ┌─┴─┐└─┬─┘┌─┴─┐      |
+    ┤ X ├──■──┤ X ├    ──╳──
+    └───┘     └───┘
+    ```
```
burgholzer, line 91: I believe it would be helpful to also write down the other pattern where only the middle gate is controlled and the sandwiching gates are simple CNOTs.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
```diff
@@ -0,0 +1,406 @@
...
+ *  └───┘         └───┘     └───┘└───┘         └───┘
+ *
+ * Example (onlyMatchFullSwapPattern=true, matchControlledSwap=true):
+ *  ──■────■────■──    ──■──
+ *    |  ┌─┴─┐  |        |
+ *  ──■──┤ X ├──■── => ──╳──
+ *  ┌─┴─┐└─┬─┘┌─┴─┐      |
+ *  ┤ X ├──■──┤ X ├    ──╳──
+ *  └───┘     └───┘
+ */
```
burgholzer, line 57: Let's simplify the implementation for this pattern over here for now. We won't be using most of the multi-controlled patterns here for one.
Let's limit this to the patterns that involve two CNOTs with alternating controls.
These should be

## Pos-Pos

```qasm
x q[0] ctrl q[1];
x q[1] ctrl q[0];
```
to
```qasm
x q[1] ctrl q[0];
swap q[1], q[0];
```
or, as indicated by the other comment,
```qasm
x q[1] ctrl q[0];
# swap the output qubit SSA values of the CNOT
```

## Neg-Neg

```qasm
x q[0] negctrl q[1];
x q[1] negctrl q[0];
```
to (resolve negctrl q[1])
```qasm
x q[1];
x q[0] ctrl q[1];
x q[1];
x q[1] negctrl q[0];
```
to (resolve negctrl q[0])
```qasm
x q[1];
x q[0] ctrl q[1];
x q[1];
x q[0];
x q[1] ctrl q[0];
x q[0];
```
to (commute X gates through targets)
```qasm
x q[1];
x q[0];
x q[0] ctrl q[1];
x q[1] ctrl q[0];
x q[0];
x q[1];
```
to (apply pos-pos pattern)
```qasm
x q[1];
x q[0];
x q[1] ctrl q[0];
x q[0];
x q[1];
## exchange output SSA values of q[0] and q[1]
```
to (cancel X gates through target of CNOT)
```qasm
x q[0];
x q[1] ctrl q[0];
x q[0];
## exchange output SSA values of q[0] and q[1]
```
to (combine back to negative control)
```qasm
x q[1] negctrl q[0];
## exchange output SSA values of q[0] and q[1]
```
which gives you the final replacement for the original pattern match.


## Pos-Neg

Similar transformations as for the Neg-Neg pattern. 

```qasm
x q[0] ctrl q[1];
x q[1] negctrl q[0];
```
to (resolve negctrl q[0])
```qasm
x q[0] ctrl q[1];
x q[0];
x q[1] ctrl q[0];
x q[0];
```
to (commute X gate through targets)
```qasm
x q[0];
x q[0] ctrl q[1];
x q[1] ctrl q[0];
x q[0];
```
to (apply pos-pos pattern)
```qasm
x q[0];
x q[1] ctrl q[0];
swap q[0], q[1];
x q[0]; 
```
to (apply elide swap)
```qasm
x q[0];
x q[1] ctrl q[0];
x q[1]; 
## exchange output SSA values of q[0] and q[1]
```
either one is done here (valid); or one transforms further. Possible options:
```qasm
x q[1] negctrl q[0];
x q[0];
x q[1]; 
## exchange output SSA values of q[0] and q[1]
```
or 
```qasm
x q[0];
x q[1];
x q[1] ctrl q[0]; 
## exchange output SSA values of q[0] and q[1]
```

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: As I understood it, we wanted to create larger passes (e.g. one which combine all three (swap reconstruction, elide permutations, cancel inverse cnots)). I still created a pass for just the swap reconstruction in order to be able to properly test it and not have one huge blob test for the three combined patterns since it will be hard to say how exactly gates got removed/cancelled out.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: Ah, thanks for clearing that up, I guess I should update all of the comments to not use the term "CNOT" anymore.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
taminob: Since we agreed that multi-controlled swap has some value and it is already implemented - why throw it away? :)
If you want to keep this PR simple, I'd rather push the Neg-Pos and Neg-Neg cases to a new PR instead and finalize this one first.

Splitting the reconstruction into `onlyMatchFullSwapPattern` and `matchControlledSwap` was actually just done for better readability and it could also be done with just one bool which is checked in both places instead. However, this more fine-grained control made the intention clearer in my opinion.

Or are we talking at cross-purposes again and should clear this up in a quick meeting?

taminob: Regarding directly eliding the swaps in this pass, I guess we are in disagreement about the extend of a single pattern. :)
I liked exactly the combination of the two patterns and keep the cognitive load per pattern low. But I can also directly elide the swap in this pattern if that's what you prefer.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
burgholzer: Hm. I suppose the testing part makes sense. At least I am not aware of a way to test individual patterns. We should be able to reuse the same patterns as part of a larger pass, right?

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
burgholzer: That depends in my opinion. If you are really talking about a CNOT (a single-controlled X gate), then the name makes sense and is quite established.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
burgholzer: - Finalizing this PR first is definitely a good idea. Pushing the alternative patterns out to a separate PR seems reasonable to me.
- I guess the only thing I dislike about the multi-controlled swap handling is that it makes this extremely simple pass much more complicated in code because there are many different cases to be handled. If we were to use PDLL, the pattern for the simple swap reconstruction would probably be a one liner (or a handful of lines).
- My main concern for considering the elision as part of the pattern is performance. If we know already that this optimization opportunity arises and that we 100% of the time want to apply it, why should we delay the decision for that to the pattern matcher/rewriter? If we implement it right here, its cost is constant and close to zero. If the pattern rewriter has to run over the program that's at least `O(n)` with regard to the number of operations, which seems unnecessary to me.

taminob: Thank you, that clears it up a bit.

- I'll start working on that once we have this merged.
- Yes, multi-controlled handling would be (probably) pretty annoying in PDLL - that's also why I said that I'm not sure if everything becomes that simple when we switch to PDLL. But I also think that the swap reconstruction would become a lot easier (and yes, I am getting tempted to look into integrating PDLL).
- Understandable, that might indeed be a problem for larger circuits... Do you have a suggestion for a name for this pass? `SwapReconstructionAndElision`?

So the conclusion is that I'll remove everything regarding the multi-controlled stuff? I am just a little bit sad that I spent (quite some) time on implementing it. :)

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: Yes, the existing code is actually already built around this (I think) with the `populate*Pattern()` functions which already are included for the passes and where you can just call the populate functions of all the patterns you want to have in your pass to it to your pattern set to be applied.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionPattern.cpp
burgholzer: Just slight clarifications:
- Nice.
- I am not sure if it becomes any worse with PDLL than it is right now. It might not get easier, but I also see no indication that it might get harder.
- Hm. At that point, one might really simply add this pattern to the permutation elision pass and rename that pass to `SwapReconstructionAndElision`. This is still equally well testable as the separate pass setup at the moment. The new tests would just become additional tests in the existing test suite for the elision pass.
- I would not remove the multi-controlled stuff entirely; would be a waste of your effort. I'd just also postpone it to a separate PR because it is not yet 100% clear to me where to put this and how to integrate it properly, while it becomes clearer and clearer how to properly integrate the simple swap reconstruction and elision.

taminob: Okay, so I'll have a quick look into PDLL and how to integrate it into the build system. Since this PR needs basically a full rewrite I could also use it as a PoC for a PDLL pattern and do that in a separate PR. Then we can simply close this one with the current state to not lose the multi-control implementation (or at least the logic of it).

burgholzer: Sounds like a plan to me 👍🏼

PR #1159, opened 2025-08-27
♻️ Rework CI caching

burgholzer:
## Description

This PR reworks how compiler caching is performed in CI. 
One of the main reasons for our Windows MLIR CI constantly getting cache misses is that we are exceeding the maximum GitHub cache size of 10 GiB by over a factor of 5. As a result, caches are frequently evicted, which then hurts most if it hits the Windows MLIR builds.

This PR drops using the GitHub cache for any regular C++ build artifacts (except for the Windows MLIR builds). Instead, it enables `sccache` for Python CI and CD, where multiple sessions are performed in succession, which greatly benefits from local, per-run caching alone.

This should free up quite some resources in the GitHub cache. If successful, this should be rolled out to other repositories as well.

## Checklist:


burgholzer:
Okay, coming to the same conclusion here as before, after investigating quite a bit further:
Caching on Windows using either ccache or sccache is only really feasible when Ninja is used as a generator. Due to GitHub runners defaulting to `gcc` when Ninja is used as a generator, this is not really feasible as we want to be testing with MSVC. At the same time, setting MSVC as a compiler on top of using Ninja as a generator requires messing with the environment and feels quite prone to breaking. As a result, there will also be no build caching on Windows in the near future.

However, this means that this PR is generally ready and just needs a new patch release of the workflows.

PR #1164, opened 2025-08-28
✨ Support importing `IfElseOpereration`s to `MQTRef`

### LINKED ISSUE #1109: 🔁 MLIR - Import classic-controlled operations from QuantumComputation
#### Background
`QuantumComputation` supports classically controlled gates (e.g., `if c[i] then X(q)`). The current translation from `QuantumComputation` to `MQTDyn` added in #1099 does not yet support `ClassicControlledOperation`s.

This depends on #1117 which will provide a more direct `IfElse` operation in the MQT Core IR.

#### Problem
Translate classically controlled operations that refer to specific classical register indices or scalar bits into MLIR conditions, preserving intent when multiple measurements target the same logical bit.

#### Proposal
- Maintain a mapping from classical register indices to SSA `i1` values or to `memref<?xi1>` slots during import. This will probably be part of #1123 
- Emit `scf.if` on loaded `i1` values from `memref<?xi1>` or directly use SSA `i1` bits where available.
- Handle classical conditions targeting entire registers (probably by converting register values to integers)

#### Examples
- OpenQASM 3 source
  ```openqasm
  OPENQASM 3;
  bit c = measure q;
  if (c) x q;
  ```
- MLIR translation (scalar bit)
  ```mlir
  %c = mqtref.measure %q : !mqtref.qubit -> i1
  scf.if %c {
    mqtref.x %q : !mqtref.qubit
  }
  ```
- MLIR translation (bit in register)
  ```mlir
  %cond = memref.load %creg[%i] : memref<?xi1>
  scf.if %cond {
    // ... quantum ops ...
  }
  ```

- OpenQASM 3 source
  ```openqasm
  OPENQASM 3;
  qubit[2] q;
  bit[2] c = measure q;
  if (c) x q[0];
  ```
- MLIR translation (conditional on entire register; 2-bit OR-reduction)
  ```mlir
  // Assume %creg : memref<2xi1>
  %i0 = arith.constant 0 : index
  %i1 = arith.constant 1 : index
  %b0 = memref.load %creg[%i0] : memref<2xi1>
  %b1 = memref.load %creg[%i1] : memref<2xi1>
  %cond = arith.ori %b0, %b1 : i1
  scf.if %cond {
    // x q[0]
    %q0 = /* load or reference the first qubit as needed */
    mqtref.x %q0 : !mqtref.qubit
  }
  ```
  For dynamic-size registers `memref<?xi1>`, compute `%cond` via a loop (any-of reduction) over all indices.

#### Acceptance criteria
- Import handles scalar and indexed controls; tests for two measurements targeting the same bit and a conditional branch.
- Import handles conditional statements on entire classical registers (including dynamic-size registers via reduction).

The behavior here should at least faithfully cover the OpenQASM 3 spec (https://openqasm.com/grammar/index.html)

denialhaag:
## Description

This PR adds support for importing `IfElseOpereration`s to `MQTRef`.

Fixes #1109

## Checklist:


mlir/unittests/translation/test_translation.cpp
```diff
@@ -721,6 +729,63 @@ TEST_F(ImportTest, XXplusYY) {
...
+    CHECK: %[[Cnd0:.*]] = arith.cmpi eq, %[[Cnt0]], %[[Exp0]] : i1
+    CHECK: scf.if %[[Cnd0]] {
+    CHECK: mqtref.x() %[[Q0]]
+    CHECK: } else {
+    CHECK: mqtref.y() %[[Q0]]
+    CHECK: }
+  )";
+
+  ASSERT_TRUE(checkOutput(checkString, outputString));
+}
```
denialhaag: @burgholzer, this PR is still WIP, but is something like this what you had in mind here?

I'm now allocating a `memref` of the size of the classical register. If the `measure` operation is used, the result is saved in the `memref`. The value of the control bit is loaded from the `memref` when an `IfElseOperation` is encountered.

My implementation does not yet support `IfElseOperations` with control registers, but I will address this once this is clean.

burgholzer: This looks very good already and is basically what I had in mind here.

I am wondering if MLIR's builtin optimizations are clever enough to optimize away the memref.load here. But I would guess so.

No complaints from my side on this so far. @DRovara anything that sticks out to you?

Obviously, some more examples would be great. And probably some canonicalization on the IfElse operation. The example above could also be written without the negation by exchanging the then and else branches. I do not know if MLIR has builtin passes that would perform such an optimization. Might be worth checking.
```
CHECK: %[[M0:.*]] = mqtref.measure %[[Q0]]
CHECK: %[[I0:.*]] = arith.constant 0 : index
CHECK: memref.store %[[M0]], %[[Mem]][%[[I0]]] : memref<1xi1>
CHECK: %[[I1:.*]] = arith.constant 0 : index
CHECK: %[[Cnt0:.*]] = memref.load %[[Mem]][%[[I1]]] : memref<1xi1>
CHECK: %[[Exp0:.*]] = arith.constant false
CHECK: %[[Cnd0:.*]] = arith.cmpi eq, %[[Cnt0]], %[[Exp0]] : i1
CHECK: scf.if %[[Cnd0]] {
CHECK: mqtref.x() %[[Q0]]
CHECK: } else {
CHECK: mqtref.y() %[[Q0]]
CHECK: }
```
could essentially be shortened to
```
CHECK: %[[M0:.*]] = mqtref.measure %[[Q0]]
CHECK: %[[I0:.*]] = arith.constant 0 : index
CHECK: memref.store %[[M0]], %[[Mem]][%[[I0]]] : memref<1xi1>
CHECK: %[[I1:.*]] = arith.constant 0 : index
CHECK: %[[Cnt0:.*]] = memref.load %[[Mem]][%[[I1]]] : memref<1xi1>
CHECK: scf.if %[[Cnt0]] {
CHECK: mqtref.y() %[[Q0]]
CHECK: } else {
CHECK: mqtref.x() %[[Q0]]
CHECK: }
```
or potentially even to
```
CHECK: %[[M0:.*]] = mqtref.measure %[[Q0]]
CHECK: %[[I0:.*]] = arith.constant 0 : index
CHECK: memref.store %[[M0]], %[[Mem]][%[[I0]]] : memref<1xi1>
CHECK: scf.if %[[I0]] {
CHECK: mqtref.y() %[[Q0]]
CHECK: } else {
CHECK: mqtref.x() %[[Q0]]
CHECK: }
```

denialhaag: I'll have a better look at potential optimizations tomorrow, but the [`-mem2reg`](https://mlir.llvm.org/docs/Passes/#-mem2reg) pass sounds interesting to me. 🤔

denialhaag: We are now running a few built-in passes before returning the module, and the output has reduced to:

```
CHECK: %[[Exp0:.*]] = arith.constant false # This is canonicalized to the top
[...] # Register allocation and qubit extraction
CHECK: %[[M0:.*]] = mqtref.measure %[[Q0]]
CHECK: %[[Cnd0:.*]] = arith.cmpi eq, %[[M0]], %[[Exp0]] : i1
CHECK: scf.if %[[Cnd0]] {
CHECK: mqtref.x() %[[Q0]]
CHECK: } else {
CHECK: mqtref.y() %[[Q0]]
CHECK: }
```

I already like it a lot more. 😌

denialhaag: The swapping of `x` and `y` could maybe be done in the `IfElseOperation` itself. We already have a similar canonicalization pass there:

https://github.com/munich-quantum-toolkit/core/blob/263aae2aa488af65426b5de611614ff7170a6f85/src/ir/operations/IfElseOperation.cpp#L94-L98

Or do you prefer to only do this at the MLIR level? 🤔

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
```diff
@@ -169,10 +196,185 @@ void addResetOp(mlir::OpBuilder& builder, const qc::Operation& operation,
...
+  mlir::Value controlValue;
+  mlir::Value expectedValue;
+  if (controlRegister.has_value()) {
+    assert(!controlBit.has_value());
+
+    // Compute control value from control register
+    // Placeholder
+    auto indexValue = builder.create<mlir::arith::ConstantIndexOp>(loc, 0);
+    controlValue = builder.create<mlir::memref::LoadOp>(
+        loc, bits, mlir::ValueRange{indexValue});
```
denialhaag: This needs to be replaced with an `scf.for` loop computing the control value. I will do that tomorrow.

mlir/unittests/translation/test_translation.cpp
burgholzer: This looks great already! I believe I would prefer to have this at the MLIR level, because it might as well turn out that we are dropping the `QuantumComputation` abstraction at some point and entirely replace it with MLIR.
However, I would believe that the implementation should be pretty similar in size and complexity to the one we have as part of the IfElse operation.

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
```diff
@@ -285,5 +451,14 @@ mlir::OwningOpRef<mlir::ModuleOp> translateQuantumComputationToMLIR(
...
   builder.create<mlir::func::ReturnOp>(loc);
 
+  // Run passes
+  mlir::PassManager passManager(context);
+  passManager.addPass(mlir::createMem2Reg());
+  passManager.addPass(mlir::createRemoveDeadValuesPass());
+  passManager.addPass(mlir::createCanonicalizerPass());
+  if (failed(passManager.run(module))) {
+    emitError(loc) << "Failed to run passes";
+  }
```
burgholzer: Just putting this here as a question I am not sure I have an answer for: Should we be running these passes here? or should they be running somewhere else? E.g. only as part of the tests?

mlir/unittests/translation/test_translation.cpp
burgholzer, line 1: Just noting this here: we should be adding all kinds of tests here. Especially also ones with empty `then` branches, different comparison kinds, register and bit conditions.

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
denialhaag: Done.

mlir/unittests/translation/test_translation.cpp
```diff
@@ -721,6 +734,443 @@ TEST_F(ImportTest, XXplusYY) {
...
+  passManager.addPass(mlir::createCanonicalizerPass());
+  passManager.addPass(mlir::createMem2Reg());
+  if (failed(passManager.run(module.get()))) {
+    FAIL() << "Failed to run passes";
+  }
+
+  const auto outputString = getOutputString(&module);
+  const auto* checkString = R"(
+    CHECK: %[[Exp:.*]] = arith.constant 1 : i64
+    CHECK: %[[ANY:.*]] = arith.constant 0 : index
```
denialhaag: This is a dead value. Theoretically, another run of the `-remove-dead-values` would remove this. Wowever, if  understood correctly, the `-canonicalize` pass before leads to a module that `-remove-dead-values` cannot run on again. Not super sure if I'm missing something here...

mlir/unittests/translation/test_translation.cpp
```diff
@@ -721,6 +734,443 @@ TEST_F(ImportTest, XXplusYY) {
...
+    CHECK: }
+  )";
+
+  ASSERT_TRUE(checkOutput(checkString, outputString));
+}
+
+TEST_F(ImportTest, IfElseBit) {
+#if LLVM_VERSION_MAJOR < 20
+  GTEST_SKIP() << "Test skipped for LLVM 18+";
+#endif
```
denialhaag, line 1141: For some reason, the output of only this test is slightly different for LLVM 19, which is why I decided to skip it for LLVM 19.

mlir/unittests/translation/test_translation.cpp
denialhaag: Done! Let me know what you think

mlir/lib/Dialect/MQTRef/Translation/ImportQuantumComputation.cpp
denialhaag: I moved it to the tests now because different tests benefit from slightly different passes.

I think it's also clean if the output module is "complete" so that a user can make the canonicalizations they need and want. 🤔

mlir/unittests/translation/test_translation.cpp
denialhaag: Makes sense! I'm also done with the canonicalization now. 👍

mlir/unittests/translation/test_translation.cpp
```diff
@@ -721,6 +746,555 @@ TEST_F(ImportTest, XXplusYY) {
...
+    CHECK: }
+  )";
+
+  ASSERT_TRUE(checkOutput(checkString, outputString));
+}
+
+TEST_F(ImportTest, IfBitEqTrue) {
+#if LLVM_VERSION_MAJOR < 20
+  GTEST_SKIP() << "Test skipped for LLVM 19";
+#endif
```
denialhaag: The output of some tests is slightly different for LLVM 19, which is why I decided to skip it for LLVM 19.

mlir/unittests/translation/test_translation.cpp
denialhaag: Okay, so I again forgot to deallocate the `memref`. After putting it in, the passes no longer remove the `store` and `load` operations. 🤔

mlir/unittests/translation/test_translation.cpp
denialhaag: As #1180 will be merged first, I removed the skip again.

denialhaag: #1180 has been merged, and this PR has been rebased. Marking this comment as resolved.

mlir/unittests/translation/test_translation.cpp
denialhaag: This is relevant again now.

mlir/unittests/translation/test_translation.cpp
```diff
@@ -188,541 +269,653 @@ TEST_F(ImportTest, Reset0) {
...
-  qc::QuantumComputation qc(1);
-  qc.rx(0.1, 0);
-
-  auto module = translateQuantumComputationToMLIR(context.get(), qc);
+  // Run passes
+  mlir::PassManager passManager(context.get());
+  passManager.addPass(mlir::createMem2Reg());
+  passManager.addPass(mlir::createRemoveDeadValuesPass());
+  passManager.addPass(mlir::createCanonicalizerPass());
+  passManager.addPass(mlir::createMem2Reg());
```
burgholzer: I am pretty sure one of these `mem2reg` passes is redundant. I just can't remember which. I also kind of liked that this was moved to a function and that all IfElse tests simply called that function.

burgholzer (approved):
Just looked through everything in here again and I really like how this turned out 🙌

I just have one small comment on the tests, but that can also be ignored. Let's get this in once the comment is resolved 😌

Great work!

mlir/unittests/translation/test_translation.cpp
denialhaag: Sorry, I had somehow overlooked that when I worked on the tests. I now put back the function you had defined. The order of passes you found is even better because it removes an unneeded line that I couldn't get rid of.

burgholzer: Nice 🙂
At some point, we will probably want to create something similar to Catalyst's Compiler Driver, which would then be running these passes on every program.
For now, let's resolve this and get it in.

Issue #1124, opened 2025-08-15
🚀 MLIR - Provide Pre-Built MLIR/LLVM Binaries for CI/CD


burgholzer:
#### **Problem Statement**
Currently, our CI/CD pipelines require building LLVM and MLIR from source—especially for Windows builds and when targeting specific commits. This significantly increases build times and has already caused delays and friction in development workflows.

#### **Goal**
Introduce a system that provides pre-built binaries for MLIR/LLVM across platforms and versions, reducing CI build times and improving developer experience.

---

#### **Proposed Solution**

Create a GitHub Action that:
- Accepts a desired LLVM version as input:  
  - `"latest"`  
  - A specific version number (e.g., `17.0.6`)  
  - A specific commit hash
- Internally checks whether the requested version has already been built and cached.
- If not available, builds LLVM/MLIR from source and caches the result for future use.
- Makes the binaries available to CI workflows via release artifacts or another storage mechanism.

---

#### **Platform-Specific Strategies**
- **Linux**: Use the installer script already integrated in our CI for fixed versions.
- **macOS**: Leverage Homebrew for supported versions.
- **Windows & Custom Commits**: Build from source and cache the resulting binaries.

---

#### **Design Considerations**
- How and where to store pre-built binaries (e.g., GitHub Releases, external storage).
- How to manage versioning and cache invalidation.
- How to ensure compatibility across platforms and CI runners.

---

#### **Inspiration**
A similar approach is used in the Python ecosystem:  
[🔗 python-build-standalone](https://github.com/astral-sh/python-build-standalone)

---

#### **Next Steps**
- Define the scope and MVP for the GitHub Action.
- Explore caching strategies and storage options.
- Coordinate the integration of the action into existing workflows.

denialhaag:
Just a thought, but I generally had positive experiences with a Docker/Container registry, at least on GitLab.

burgholzer:
> Just a thought, but I generally had positive experiences with a Docker/Container registry, at least on GitLab.

Yeah. That could be one solution. Basically building docker images off of the standard images used by GitHub or the manylinux project, installing LLVM there, and publishing those images to some registry.

I have always been a bit cautious with using Docker in the past, because you need to have it installed and people need to know how to use it, which isn't always straightforward. However, in a CI scenario, that is much less of an issue.
Here, I am only concerned that we are diverging too much from the GitHub runner images that allow us to build up the trust that whatever we build also works for people.

That's just my two cents here though.
[...]

PR #1180, opened 2025-09-05
🐉 Require LLVM 21

### LINKED ISSUE #1176: ✨ Consolidate MLIR CI to LLVM 21 and align with upstream cadence
### Problem Statement

Maintaining CI across multiple LLVM versions (19/20) increases build time and maintenance while providing little practical value for plugin interoperability. Recent work on the Catalyst plugin (#881) and CUDA-Q exploration (#924) showed that cross-framework exchange via binary-compatible plugins is brittle and not feasible at scale. As MLIR development is not tied to LLVM releases, tracking multiple released LLVM versions does not reflect upstream practice and slows iteration. With a shared exchange format on top of IRs under consideration (e.g., Jeff: https://github.com/unitaryfoundation/jeff), broad binary compatibility across LLVM versions becomes less relevant. CI performance also suffers without proper caching (#1124).

### Proposed Solution

- Switch CI to a single, latest LLVM release (LLVM 21) and drop 19/20 from the matrix.
- Document a clear support policy: supported LLVM == latest release; optional periodic job against llvm-project main to detect breakage early.
- Keep #1124 for build caching improvements and moving to more recent versions
- Update build scripts, Dockerfiles, and docs to default to LLVM 21; provide a variable to override locally if needed.
- Add a separate issue to evaluate a shared exchange format (Jeff) for cross-framework compiler pass exchange.
- Benefits:
  - Faster CI and simpler maintenance.
  - Quicker adoption of MLIR changes aligned with upstream.
  - Reduced brittleness from binary plugin dependencies.
  - Clearer support policy for users.

denialhaag:
## Description

This PR raises the required LLVM version to 21. As a result, LLVM 21 is also used in the CI now.

Fixes #1176

## Checklist:

### DIFF
--- a/mlir/tools/quantum-opt/quantum-opt.cpp
+++ b/mlir/tools/quantum-opt/quantum-opt.cpp
@@ -16,7 +16,6 @@
 #include "mlir/Dialect/MQTRef/Transforms/Passes.h"         // IWYU pragma: keep
 
 // Suppress warnings about implicit captures of `this` in lambdas
-// Suppress warnings about implicit captures of `this` in lambdas
 #ifdef __clang__
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-this-capture"


denialhaag:
The docs still use LLVM 19 so that we can easily install it using the `apt_packages` in `.readthedocs.yaml`.

.github/scripts/setup-mlir.sh
```diff
@@ -71,6 +71,7 @@ case "$OS_NAME" in
     LLVM_PREFIX="$(brew --prefix)/opt/llvm@${LLVM_MAJOR}"
     append_env "CC=${LLVM_PREFIX}/bin/clang"
     append_env "CXX=${LLVM_PREFIX}/bin/clang++"
+    append_env "LDFLAGS=-L${LLVM_PREFIX}/lib/c++ -L${LLVM_PREFIX}/lib/unwind -lunwind"
```
burgholzer, line 74: Why is this necessary and what is the purpose of this here?

.readthedocs.yaml
```diff
@@ -35,7 +35,7 @@ build:
       - asdf install uv latest
       - asdf global uv latest
       # Build the MLIR documentation
-      - uvx cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_MQT_CORE_MLIR=ON -DMLIR_DIR=/usr/lib/llvm-19/lib/cmake/mlir -DLLVM_DIR=/usr/lib/llvm-19/lib/cmake/llvm
+      - uvx cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_MQT_CORE_MLIR=ON -DMQT_MLIR_MIN_VERSION=19.0 -DMLIR_DIR=/usr/lib/llvm-19/lib/cmake/mlir -DLLVM_DIR=/usr/lib/llvm-19/lib/cmake/llvm
```
burgholzer, line 38: This is slightly unfortunate because this will break as soon as we use a feature that is no longer compatible with LLVM 19.
It's not quite clear to me yet, how we will solve this. We will need to get LLVM in the docs build somehow. Even if we are including it in the wheel, that would not really help, because the docs build is not pulling in the wheel, but building from source. he only thing that changes then is that we do not need a separate build step in the docs.

What might solve this is when we implement #1124 because we could then pull in those prebuilt libraries. 

Overall, I would vote for keeping this here as it is and hope that it does not break too soon.

docs/mlir/index.md
```diff
@@ -13,18 +13,23 @@ Both dialects define various transformation passes.
...
+This page is a work in progress.
+The content is not yet complete and subject to change.
+Contributions are welcome.
+See the {doc}`contribution guide <contributing>` for more information.
+:::
+
+## Development
+
+Building the MLIR library requires the latest release version of LLVM.
+Our CI pipeline on GitHub continuously tests the MLIR library on the {code}`ubuntu-24.04`, {code}`ubuntu-24.04-arm`, {code}`macos-14`, {code}`windows-2022`, and {code}`windows-11-arm` runners.
```
burgholzer: I am a bit torn on this. The first sentence here is rather generic and kind of not explicit enough (one would definitely like to know which major version this is targeting and, in the future, this might even be an individual commit).

At the same time, the second sentence is extremely precise and probably prone to getting out of date with the frequent updates to GitHub. Can we make this slightly more generic?

.github/scripts/setup-mlir.sh
denialhaag: The macOS jobs are failing with a linker error without: https://github.com/munich-quantum-toolkit/core/actions/runs/17488820760/job/49673866248

Homebrew mentions this is in the `Caveats` after pouring `llvm`(see the setup step of the job linked above):

```
To use the bundled libc++ please use the following LDFLAGS:
  LDFLAGS="-L/opt/homebrew/opt/llvm/lib/c++ -L/opt/homebrew/opt/llvm/lib/unwind -lunwind"
```

That said, I'm not entirely sure if there's a better way to fix this. I could reproduce the behavior locally, and this was the only solution I found.

.readthedocs.yaml
denialhaag: I totally agree that this is all but ideal! I think I also get why the wheel won't help. Let's keep it in mind for #1124!

docs/mlir/index.md
denialhaag: Good point! Done. Let me know what you think.

.github/workflows/ci.yml
```diff
@@ -154,7 +154,7 @@ jobs:
     uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@d42342e610318ddf1fcd2371be987faa47caedf0 # v1.16
     with:
       cmake-args: -DBUILD_MQT_CORE_BENCHMARKS=ON -DBUILD_MQT_CORE_MLIR=ON -DBUILD_MQT_CORE_BINDINGS=ON
-      clang-version: 20
+      clang-version: 21
```
denialhaag, line 157: This is necessary for LLVM 21 to be installed in the linter job.

burgholzer: We should probably run the CI with `files-changed-only: false` once, because this also bumps the clang-tidy version and I would guess that there are new warnings popping up from that.

burgholzer (approved):
This LGTM now. Just had one concern about the linter version bump. Hope this does not trigger too many warnings.
I believe this also might warrant a separate changelog entry for the update of the clang-tidy version employed in the project.

.github/scripts/setup-mlir.sh
burgholzer: I suppose this makes sense 👍🏼 and it is only temporary anyway until we have #1124

.github/workflows/ci.yml
denialhaag: Good point! Done. The linter errors should be fixed now. It was just 8 errors. 😌

.github/workflows/ci.yml
```diff
@@ -154,9 +154,9 @@ jobs:
...
       build-project: true
-      files-changed-only: true
+      files-changed-only: false
```
denialhaag: Set this back to `true` before merging.

```suggestion
      files-changed-only: true
```

denialhaag:
@burgholzer, thanks for catching the comment there! 🙃

PR #1194, opened 2025-09-08
:sparkles: Integrate MLIR PDLL for declarative pattern matching

taminob:
## Description

This PR introduces the MLIR declarative language for pattern matching and rewriting into the existing build system.
As an example, the `ElidePermutations` pattern will be replaced by a PDLL implementation and extended by the functionality of #1158 (with the exclusion of multi-controlled swaps).

This has previously been discussed in #1158.

## Checklist:


burgholzer:
Pre-commit won't be hapy here until you add a line for PDLL to https://github.com/munich-quantum-toolkit/core/blob/8e492a9c075abffb2cfbb8750abfa7f62e6a484e/.license-tools-config.json#L13-L14

burgholzer:
Let me play around with this a little. I'll get back to you 😌

burgholzer:
I played around with this a little more now.
My personal conclusion at the moment: PDLL is quite cumbersome to use. At least considering the way we currently define our operations.
The value ranges make it fairly hard to write elegant code.
Furthermore, there is no syntax highlighting or LSP support out of the box for most IDEs.
Considering all of that, I don't believe PDLL is the way to go for the moment.
It was nice that we tried, and it's good to know that it exists, but I haven't felt that it considerably reduces complexity. Even in cases where I would have expected this to be the case.

@taminob if you agree with the above, would you mind continuing/resuming the work in #1158 and streamlining it based on all the lessons lernt in here.
It could be worth to cherry pick https://github.com/munich-quantum-toolkit/core/pull/1194/commits/ff7114de4efaad93edd9f9dcb6cf802c905262ae into that PR.

taminob:
This PR has been superseded by #1207 which re-implements the functionality in C++.

I reverted some of your changes since I think that it can still be used as a reference for how to use PDLL and the different variants serve as a showcase.

burgholzer:
> I reverted some of your changes since I think that it can still be used as a reference for how to use PDLL and the different variants serve as a showcase.

Technically, there would not have been a need for that as the branch contains all the commits and history anyway. Still fine though; just couldn't resist to comment. :o)

Issue #1119, opened 2025-08-14
🚸 MLIR - Define custom builders


denialhaag:
### Problem Statement

Currently, creating an operation in C++ is more cumbersome than it needs to be. It would be nice if some operations had custom builders that reduce boilerplate code.

### Proposed Solution

Wherever it makes sense, custom builders should be defined. The C++ code should be updated to use these new builders.

DRovara:
Some potential for custom builders (these are just suggestions, they are not all desperately necessary):
- Gates without controls
- Gates with only controls/targets 
- Gates without parameters

I'm also not sure if this makes sense but maybe we can predefine the output types (e.g. the QubitType for gate targets) in builders as well so we don't have to specify them when constructing new operations. After all, we know that the types will always be fixed.

burgholzer:
One of the past PRs introduced [InferTypeOpInterface](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/InferTypeOpInterface.td), which automatically added convenient builders to some of the classes in our dialects. As a result, this issue might boil down to replacing the actual locations where the builders are invoked to use the more convenient builders with the type inference. Maybe type inference can also be added to more of our types. There seems to be no need for explicitly defining builders ourselves to accomplish this.

burgholzer:
> I just came across some pre-defined tablegen traits defined by MLIR which pretty much do exactly this. Especially `SameOperandsAndResultType` will automatically do the type inference for the result that I mentioned above ([documentation](https://github.com/llvm/llvm-project/blob/main/mlir/docs/DeclarativeRewrites.md#building-operations)). I think the `NoParameter` would probably directly imply `SameOperandsAndResultType`, right? Might be relatively easy to integrate by using the third parameter of `ParamNativeOpTrait` to inject the additional trait for `NoParameter`.

I believe this should be true, i.e., `NoParameter` should imply that trait. Definitely worth exploring.

> It's another topic, but looking into these traits in general could be helpful (although the documentation is pretty sparse). E.g. `Commutative` could also allow us to use some standard optimizations of LLVM (maybe)?

I very much agree. We should leverage existing infrastructure as much as possible.
I am currently trying to find some time to investigate the basic setup of our quantum dialects a little bit more so that we could, potentially, put them on a little more stable basis.
I'll probably have some more thoughts on this after you are back from your vacation.

PR #1207, opened 2025-09-16
:sparkles: Add MLIR Swap Reconstruction and Elision

taminob:
## Description

This PR introduces a new pattern and pass for swap reconstruction from two consecutive reversed CNOT gates and its immediate elision.
It replaces #1158 which implemented a more advanced, but also much more complex, variant of the pattern which can also re-construct controlled SWAP gates and replace full three CNOT patterns equivalent to a SWAP.

Related to #1122 

## Checklist:


mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -45,14 +45,23 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
   }];
 }
 
-def ElidePermutations : Pass<"elide-permutations", "mlir::ModuleOp"> {
-  let summary = "This pass removes any permutation (i.e., SWAP gates) it encounters and modifies the order of the qubits instead.";
+def SwapReconstructionAndElision : Pass<"swap-reconstruction-and-elision", "mlir::ModuleOp"> {
+  let summary = "This pass tries to replace consecutive reversed CNOT gates with SWAP gates and removes then any permutation (i.e., SWAP gates) it encounters by modifying the order of the qubits instead.";
```
burgholzer:
```suggestion
  let summary = "This pass tries to replace consecutive CNOT gates with alternating controls and targets by SWAP gates and, then, elides SWAP gates (i.e., qubit permutations) it encounters by modifying the order of the qubits instead.";
```

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
```diff
@@ -45,14 +45,23 @@ def MergeRotationGates : Pass<"merge-rotation-gates", "mlir::ModuleOp"> {
...
   let description = [{
-    Elides permutations (i.e., SWAP gates) by removing them from the circuit and, instead, permuting the respective qubit values.
+    Reconstruct SWAP gates from two consecutive reversed CNOTs where possible and elides permutations (i.e., SWAP gates) by removing them from the circuit and, instead, permuting the respective qubit values.
```
burgholzer:
```suggestion
    Reconstruct SWAP gates from two consecutive CNOTs with alternating controls and targets where possible and elide SWAP gates (i.e., permutations) by removing them from the circuit and, instead, permuting the respective qubit values.
```
This even seems like a much better one line summary than the above one.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
```diff
@@ -0,0 +1,153 @@
...
+   *   - Remove 1st CNOT and swap output
+   */
+  mlir::LogicalResult
+  matchAndRewrite(XOp op, mlir::PatternRewriter& rewriter) const override {
+    if (auto secondCNot = findReverseCNotPattern(op)) {
+      rewriter.replaceOpWithNewOp<XOp>(
+          op, op.getPosCtrlOutQubits().getType(), op.getOutQubits().getType(),
+          mlir::TypeRange{}, mlir::DenseF64ArrayAttr{},
+          mlir::DenseBoolArrayAttr{}, mlir::ValueRange{},
+          op.getPosCtrlInQubits(), op.getInQubits(), mlir::ValueRange{});
```
burgholzer: Is there truly no better builder for an X operation at the moment?
Definitely something we should be adding as part of #1119 in that case.
Would you mind adding this as a comment to the respective issue so that we do not forget this use case?

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * @brief Find a user of the given operation for which isCandidate() is
+   * true.
+   *
+   * @param op Operation for which the users should be scanned for a candidate
+   * @param isThirdCNot If true, the candidate should have a subset of
+   * controls of op (a surrounding CNOT, 1st/3rd); if false, it should have a
+   * superset of controls of op (middle CNOT, 2nd)
+   */
```
burgholzer: This comment seems outdated and needs an update

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
```diff
@@ -0,0 +1,153 @@
...
+  [[nodiscard]] static std::optional<XOp> findReverseCNotPattern(XOp& op) {
+    for (auto* user : op->getUsers()) {
+      if (auto cnot = llvm::dyn_cast<XOp>(user)) {
+        if (isReverseCNotPattern(op, cnot)) {
+          llvm::errs() << "YES\n";
```
burgholzer: very much seems like a debug print.
```suggestion
```

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
```diff
@@ -0,0 +1,153 @@
...
+          return cnot;
+        } else {
+          llvm::errs() << "NOPE\n";
+        }
```
burgholzer:
```suggestion
        }
```
left over debugging?

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
```diff
@@ -0,0 +1,153 @@
...
+   */
+  mlir::LogicalResult
+  matchAndRewrite(XOp op, mlir::PatternRewriter& rewriter) const override {
+    if (auto secondCNot = findReverseCNotPattern(op)) {
+      rewriter.replaceOpWithNewOp<XOp>(
+          op, op.getPosCtrlOutQubits().getType(), op.getOutQubits().getType(),
+          mlir::TypeRange{}, mlir::DenseF64ArrayAttr{},
+          mlir::DenseBoolArrayAttr{}, mlir::ValueRange{},
+          op.getPosCtrlInQubits(), op.getInQubits(), mlir::ValueRange{});
+      rewriter.replaceOp(*secondCNot, secondCNot->getOperands());
```
burgholzer: This could potentially deserve a code comment explaining why this is realising the desired exchange because I believe it isn't entirely obvious.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * @brief This pattern attempts to find CNOT patterns which can be replaced by a
+ * SWAP gate and directly elides the new permutation.
+ *
+ *       ┌───┐         ┌───┐                            ┌───┐
+ *  ──■──┤ X ├    ──■──┤ X ├──■────■──    ──╳────■──    ┤ X ├
+ *  ┌─┴─┐└─┬─┘ => ┌─┴─┐└─┬─┘┌─┴─┐┌─┴─┐ =>   |  ┌─┴─┐ => └─┬─┘
+ *  ┤ X ├──■──    ┤ X ├──■──┤ X ├┤ X ├    ──╳──┤ X ├    ──■──
+ *  └───┘         └───┘     └───┘└───┘         └───┘
```
burgholzer: This would need annotations for qubit labels to be fully correct as the last transformation itself is not correct in a unitary sense if one does not exchange the qubit labels before and after the CNOT gate. Typically, in the literature this is depicted as a literal wire crossing representing the exchange (which is arguably hard to do in ASCII art).

mlir/test/Dialect/MQTOpt/Transforms/swap-reconstruction-and-elision.mlir
```diff
@@ -0,0 +1,257 @@
...
+    mqtopt.deallocQubit %q1_4
+    mqtopt.deallocQubit %q2_3
+
+    return
+  }
+}
+
+
+// -----
+// This test checks that consecutive CNOT gates which match a SWAP gate are merged correctly.
```
burgholzer: these two gates do not match a SWAP. that would need three gates.

mlir/test/Dialect/MQTOpt/Transforms/swap-reconstruction-and-elision.mlir
```diff
@@ -0,0 +1,257 @@
...
+    mqtopt.deallocQubit %q0_3
+    mqtopt.deallocQubit %q1_3
+
+    return
+  }
+}
+
+
+// -----
+// This test checks that controlled CNOT gates with differing controls are merged into a controlled SWAP gate.
```
burgholzer: This test does not do what it describes and I believe it should actually be removed.

mlir/test/Dialect/MQTOpt/Transforms/swap-reconstruction-and-elision.mlir
burgholzer, line 1: I believe this needs a couple more tests to
- show that SWAP reconstruction works properly (both from three consecutive CNOTs as well as from two). I believe that this might even highlight that we are missing an important pattern here: the one that combines three CNOTs into a SWAP and directly eliminates that SWAP by permuting the qubits. My argument here would again be efficiency: If I know already with constant overhead that this is the optimization I want to apply, why wait for the next pattern application round to hit this.
- show that the combination of the two patterns works as expected, e.g., by designing a sequence of gates where the application of one of the patterns opens up the door for applying the second pattern for a joint optimization.

Last, but not least, a question that comes from the new experience gained from the PDLL PR: there the patterns always had this "with recursion" specifier that one could add to the pattern. How is that mirrored in the regular pattern and pass structure here? Specifically, how does one indicate that these patterns may be applied directly to the results of a previous application?

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
taminob: This one is (or at least was) longer, that's why I wanted to use the shorter version for the summary. Should I swap it?

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
taminob: There actually are more for `XOp` than for `SWAPOp`, I can use one where I don't specify the result types explicitly, but for the ones where I could simply specify a single `mlir::ValueRange` for the operands, I am not sure how it will interpret it regarding target/control qubits.

So currently I am using `static void build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, ::mlir::TypeRange out_qubits, ::mlir::TypeRange pos_ctrl_out_qubits, ::mlir::TypeRange neg_ctrl_out_qubits, /*optional*/::mlir::DenseF64ArrayAttr static_params, /*optional*/::mlir::DenseBoolArrayAttr params_mask, ::mlir::ValueRange params, ::mlir::ValueRange in_qubits, ::mlir::ValueRange pos_ctrl_in_qubits, ::mlir::ValueRange neg_ctrl_in_qubits);`, I'll now try to switch to `static void build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, ::mlir::TypeRange resultTypes, /*optional*/::mlir::DenseF64ArrayAttr static_params, /*optional*/::mlir::DenseBoolArrayAttr params_mask, ::mlir::ValueRange params, ::mlir::ValueRange in_qubits, ::mlir::ValueRange pos_ctrl_in_qubits, ::mlir::ValueRange neg_ctrl_in_qubits);`.

` static void build(::mlir::OpBuilder &, ::mlir::OperationState &odsState, ::mlir::TypeRange resultTypes, ::mlir::ValueRange operands, ::llvm::ArrayRef<::mlir::NamedAttribute> attributes = {});` would be even simpler, but also less clear what's going to happen because I don't explicitly specify the segment sizes (as I did in the PDLL operation creation).

Edit: I think I'll stick with the current one because otherwise I get `error: 'mqtopt.x' op result count (2) does not match with the total size (0) specified in attribute 'resultSegmentSizes'` which means I need to specify the segment sizes manually again which is way uglier than simply using the current constructor. I'll make it a little bit more readable by simply supplying all data from `op` using its getters.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Passes.td
burgholzer: I'd prefer that, yes. But I'd not really swap the descriptions. The description typically should not repeat the brief/summary. The description should assume that it is prepended by the brief in the documentation. So there is really no need to reiterate the main summary point in the description.

mlir/lib/Dialect/MQTOpt/Transforms/SwapReconstructionAndElisionPattern.cpp
burgholzer: This is really something we need to investigate as part of #1119.
It should be entirely sufficient to provide the value for the target qubit as well as value ranges for positive and negative controls for constructing a MCX gate.
No type information should be necessary for that at all, the types are well known at compile time already.
However, I believe this is not a critical point for this PR here. It would be great to translate some of the discussion points here to the respective issue though so they are not lost once this PR is merged. Then, I suppose this can be resolved and we can move on here.

mlir/test/Dialect/MQTOpt/Transforms/swap-reconstruction-and-elision.mlir
taminob: > * show that SWAP reconstruction works properly (both from three consecutive CNOTs as well as from two). I believe that this might even highlight that we are missing an important pattern here: the one that combines three CNOTs into a SWAP and directly eliminates that SWAP by permuting the qubits. My argument here would again be efficiency: If I know already with constant overhead that this is the optimization I want to apply, why wait for the next pattern application round to hit this.

I think in our previous discussions you explicitly said that you don't want it because it's unnecessary and already covered by this + cancel inverses. I'd also rather work on the other tasks since I feel like I spent enough time on this pattern for now.

> * show that the combination of the two patterns works as expected, e.g., by designing a sequence of gates where the application of one of the patterns opens up the door for applying the second pattern for a joint optimization.

I think that should be part of our next step when we develop passes which combine the different patterns to be actually useful.

mlir/test/Dialect/MQTOpt/Transforms/swap-reconstruction-and-elision.mlir
taminob: I slightly adjusted it, but I think it's not unnecessary since it's also the only test that contains negative controls.

mlir/test/Dialect/MQTOpt/Transforms/swap-reconstruction-and-elision.mlir
burgholzer: Many of the aspects around MLIR are a moving target at the moment. We are not yet experts in all of this and it is all a learning process.
Back when we had that discussion, I didn't have in mind that this would have performance implications that we can easily avoid by a very cost-effective addition of the respective set of patterns.

I know that we have iterated on this quite a lot by now, but I'd rather get it "right" once and then apply that gathered knowledge to all future passes, then to do a half-baked solution across the board that collectively needs to be revisited and refactored almost immediately once it is done.

And I think this might be a perfectly nice opportunity to learn how to apply the benefit system for rewrite patterns. That is because both, the existing as well as the three CNOT one, would match on the same prefix of gates, with the 3 CNOT one being preferred over the 2 CNOT version.

> I think that should be part of our next step when we develop passes which combine the different patterns to be actually useful.

The two (or three) patterns are already combined into a pass here. This pass should _already_ apply rules greedily until no rule can be applied anymore; that's the whole point of the pattern applier that is part of the pass. Thus, I'd argue respective tests are already feasible now.

taminob:
@burgholzer I added the three-CNOT matching logic again, feel free to have another look (no need to hold back on comments :) ). I looked a little bit into the greedy driver logic and also found some interesting paragraphs in [its documentation](https://github.com/llvm/llvm-project/blob/main/mlir/docs/PatternRewriter.md).

If you're interested, [this paragraph](https://github.com/llvm/llvm-project/blob/main/mlir/docs/PatternRewriter.md#application-recursion) is about the `with recursion` (I verified in the PDLL parser sources that it's exactly the same). However, `setHasBoundedRewriteRecursion(false)` does not cause the recursion test cases to fail - I guess that the recursion detection has a way higher limit - do you think it is worth it to add a test case with accordingly many CNOTs? I'll need to do some digging on how many are needed.

The reason why the benefit didn't work before was that it was actually doing a bottom-up traversal for the operations and thus the two CNOT operation will be matched first of course. Fortunately, the greedy pattern rewrite driver does have some configuration options which are useful. For more complex passes, we might want to actually look into multiple `applyPatternsGreedily` loops since using `GreedyRewriteStrictness::ExistingOps` could prevent the loop in #1182, but that's for the future and more complex passes (`config.enableConstantCSE(true)` also simplifies the constant issues we had over there).

Issue #1154, opened 2025-08-26
✨ MLIR - Constant Canonicalization


li-mingbao:
### Problem Statement

This issue is related to #916.
Currently, quantum operations in the `mqtref` and the `mqtopt` dialect that require i64 or f64 values as input can either get the value through an attribute or as an operand. Most of the time the SSA value of the operand is created by a constant operation which can be replaced with an attribute of this value. Currently, this is only done for the extractQubit operation in the `constant-folding` pass.

### Proposed Solution

The pass  should be extended to other operations to check whether the SSA value is created by a constant operation. If yes, the value of the constant operation should be converted to an attribute of the appropriate type and the operand of this operation should be replaced with this attribute. The constant operation can then be erased from the IR, if it has no other user remaining.

Issue #1120, opened 2025-08-14
✨ MLIR - Conversion from llvm dialect to standard dialects


li-mingbao:
### Problem Statement

Once #1091 is merged, there will be a conversion from `QIR` in the `llvm` dialect into the `mqtref` dialect. During this conversion however only quantum operations are converted and other standard operation such as `llvm.constant` and `llvm.func` are not converted. While there are built-in transformation passes from higher level standard dialects such as `arith` or `func` to `llvm` there no conversion passes in the other direction.

This is not an issue for the conversion itself, but this could cause some issues later on during the transformation passes as they would expect an operation from a standard dialect and not from the `llvm` dialect.

### Proposed Solution

The best solution would be to create a new conversion pass that converts `llvm` operations to `std` dialects operations. Looking at the documentation of the `llvm` [dialect](https://mlir.llvm.org/docs/Dialects/LLVM/), there are probably a bunch of operations that do not need to be converted but atleast the standard operations that are likely to appear in QIR code and are required for the transformation passes need to be converted.
 These are for example `llvm.constant`, `llvm.cond_br` or `llvm.func`.

burgholzer:
Hey @li-mingbao 👋🏼

Just wanted to share a few thoughts and get your perspective on them:

From my understanding of MLIR and LLVM, LLVM sits strictly below MLIR in the compilation hierarchy. Typically, we lower from MLIR to LLVM—specifically to the LLVM dialect within MLIR. This seems to be supported by the issue description, which highlights MLIR’s strong support for lowering from its standard dialects to LLVM IR.

In that context, QIR is generally positioned as a low-level exchange format—essentially the output of the compiler toolchain that gets sent to the quantum device for execution. So the direction MLIR → QIR makes perfect sense.

What I’m less convinced about is the reverse direction: converting from QIR (LLVM-based) back to MLIR. It feels somewhat artificial to "raise" the abstraction level, and I imagine reconstructing structured control flow from raw LLVM IR could be quite complex—possibly even infeasible in some cases.

The only scenarios I can think of where this reverse conversion might be useful are:

- Treating QIR as an input format (i.e., a frontend language) for our compiler infrastructure.
- Enabling interoperability with compiler passes that operate exclusively on QIR.

So my main question is: how feasible do you think this reverse conversion is, especially beyond simple, straight-line quantum programs?

li-mingbao:
Hey @burgholzer,

[...]

PR #1211, opened 2025-09-18
♻️ Replace MLIR `QubitRegister` types with `memref`

### LINKED ISSUE #1126: 🧱 MLIR - Represent qubit registers in MQTRef as `memref<?x!mqtref.qubit>`
## Background
A bespoke “qubit register” abstraction is redundant; MLIR already provides `memref` semantics for memory-backed collections.

## Problem
Redundant abstractions increase complexity; `memref` can model extract/insert via load/store while aligning with MLIR conventions.

## Proposal:
- Use `memref<?x!mqtref.qubit>` for registers; replace bespoke extract/insert with `memref.load/store` or thin wrappers.
- Adjust printer/parser and verifiers as needed.

## Examples:
```mlir
%reg = memref.alloc(%n) : memref<?x!mqtref.qubit>
%q  = memref.load %reg[%i] : memref<?x!mqtref.qubit>
// ... use %q ...
memref.store %q, %reg[%j] : memref<?x!mqtref.qubit>
```

## Acceptance criteria:
- Types/printer/parser/tests updated; examples demonstrate alloc, load, and store of qubits.
- Migration notes added to docs if user-facing IR changes.

denialhaag:
## Description

This PR replaces the `mqtref.QubitRegister` with `memref`.

Fixes #1126 

## Checklist:


mlir/unittests/translation/test_translation.cpp
```diff
@@ -186,12 +187,14 @@ TEST_F(ImportTest, AllocationAndDeallocation) {
...
-    CHECK: %[[Mem:.*]] = memref.alloca() : memref<2xi1>
-    CHECK: "mqtref.deallocQubitRegister"(%[[Reg]]) : (!mqtref.QubitRegister) -> ()
+    CHECK: %[[Qreg:.*]] = memref.alloca() : memref<3x!mqtref.Qubit>
+    CHECK: %[[I0:.*]] = arith.constant 0 : index
+    CHECK: %[[Q0:.*]] = memref.load %[[Qreg]][%[[I0]]] : memref<3x!mqtref.Qubit>
+    CHECK: %[[I1:.*]] = arith.constant 1 : index
+    CHECK: %[[Q1:.*]] = memref.load %[[Qreg]][%[[I1]]] : memref<3x!mqtref.Qubit>
+    CHECK: %[[I2:.*]] = arith.constant 2 : index
+    CHECK: %[[Q2:.*]] = memref.load %[[Qreg]][%[[I2]]] : memref<3x!mqtref.Qubit>
+    CHECK: %[[Creg:.*]] = memref.alloca() : memref<2xi1>
```
denialhaag: @burgholzer, this PR is still very much a work in progress, but I have a quick question: Do we want to use `memref.alloca()` also for the qubit registers? 🤔

mlir/unittests/translation/test_translation.cpp
```diff
@@ -302,28 +313,23 @@ std::string getCheckStringTestCaseUnitary(const TestCaseUnitary& testCase) {
...
-    result += "CHECK: %[[Q" + std::to_string(i) +
-              ":.*]] = \"mqtref.extractQubit\"(%[[Reg]]) "
-              "<{index_attr = " +
-              std::to_string(i) +
-              " : i64}> : (!mqtref.QubitRegister) -> !mqtref.Qubit\n";
+    result += std::format(R"(
+      CHECK: %[[I{}:.*]] = arith.constant {} : index
+      CHECK: %[[Q{}:.*]] = memref.load %[[Qreg]][%[[I{}:.*]]] : memref<{}x!mqtref.Qubit>
+    )",
+                          i, i, i, i, testCase.numQubits);
```
denialhaag: Because of a comment by Copilot in a different PR, I tried how this looks using `std::format`, and I have to say that I really like it!

denialhaag: Hmm... it looks like we would have to bump the macOS deployment target. Not sure if this is worth it then. 😬

mlir/unittests/translation/test_translation.cpp
burgholzer: I am not 100% sure to be honest. On the one hand, I would be tempted to say "no", because we might want to explicitly deallocate registers.
On the other hand, I don't really see an argument for why we should be handling this any different than the classical registers.
Another point would be that languages like OpenQASM also don't need the concept of explicit deallocation of registers.
So I think I would be slightly in favor of using `alloca` without the explicit deallocation.

In the end, we can always do that now and pivot if we find a use case for the explicit deallocation.

mlir/unittests/translation/test_translation.cpp
burgholzer: That indeed looks really nice.

burgholzer: Yeah.. bumping the deployment target isn't really great as it rules out quite some older systems. One could argue that this is not extremely important for us directly. But other people might have tighter constraints here. It would be quite a significant bump, right?

denialhaag: Yeah, I think it would need to be 13.3 from 11.0...

I'll go back to not using `std::format` then!

burgholzer: Yeah. Sad, but probably the better option here.
Let's keep this in mind though and watch how the ecosystem changes in the next couple of months/years.

mlir/unittests/translation/test_translation.cpp
denialhaag: In a separate conversation, @burgholzer and I came to the conclusion that we should go with `memref.alloc` and `memref.dealloc` so that the register allocation and deallocation can be translated to QIR properly. Resolving this conversation.

docs/mlir/MQTRef.md
```diff
@@ -15,12 +15,6 @@ tocdepth: 3
 :start-after: "## Operations"
 ```
 
-## Passes
-
-```{include} Passes/MLIRMQTRefPasses.md
-
-```
-
```
denialhaag, line 23: The sole MQTRef pass we had was acting on the now-removed `ExtractOp`. I decided to remove the entire pass infrastructure for now. I hope that makes sense.

mlir/test/Conversion/mqtref-to-qir.mlir
```diff
@@ -107,57 +108,36 @@ module {
       // CHECK: ^[[main]]:
       // CHECK: %[[size:.*]] = llvm.mlir.constant(2 : i64) : i64
       // CHECK: %[[r_0:.*]] = llvm.call @__quantum__rt__qubit_allocate_array(%[[size]]) : (i64) -> !llvm.ptr
-      // CHECK: llvm.br ^[[main2:.*]]
-      // CHECK: ^[[main2]]:
```
denialhaag, line 111: I'm a bit confused which of my changes in `MQTRefToQIR.cpp` could have affected this. Do you see a problem somewhere, @li-mingbao? 🤔

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
```diff
@@ -162,17 +172,17 @@ struct ConvertMQTRefAllocQIR final : StatefulOpConversionPattern<ref::AllocOp> {
...
-          rewriter.create<LLVM::ConstantOp>(op.getLoc(), op.getSizeAttrAttr());
+    // get size
+    mlir::Value size;
+    if (op.getType().hasStaticShape()) {
+      const auto staticSize = op.getType().getShape().front();
+      getState().numQubits += staticSize;
+      size = rewriter.create<LLVM::ConstantOp>(
+          op.getLoc(), rewriter.getI64IntegerAttr(staticSize));
+    } else {
+      size = adaptor.getDynamicSizes().front();
```
denialhaag: One potential issue: I could not find a way to cast this to the underlying integer value like before. Not entirely sure what's so different.

mlir/test/Conversion/mqtref-to-qir.mlir
denialhaag: I think I found the solution in https://github.com/munich-quantum-toolkit/core/pull/1211/commits/947982199c4e74dbfc4283fa4d81dcf49d5c3d6f; I had overseen that changes in `ensureBlocks()` were necessary.

docs/mlir/MQTRef.md
burgholzer: Makes sense, yes!

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
```diff
@@ -411,12 +421,15 @@ struct MQTRefToMQTOpt final : impl::MQTRefToMQTOptBase<MQTRefToMQTOpt> {
     MQTRefToMQTOptTypeConverter typeConverter(context);
 
     target.addIllegalDialect<ref::MQTRefDialect>();
+    target.addIllegalDialect<memref::MemRefDialect>();
```
burgholzer: This seems shortsighted to me. The programs might very much still contain `memref` elements (such as classical registers) which are entirely valid.
I'd probably simply removed this line
```suggestion
```

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
```diff
@@ -95,60 +98,63 @@ class MQTRefToMQTOptTypeConverter final : public TypeConverter {
...
-    // QregType conversion
-    addConversion([ctx](ref::QubitRegisterType /*type*/) -> Type {
-      return opt::QubitRegisterType::get(ctx);
-    });
   }
 };
 
-struct ConvertMQTRefAlloc final : StatefulOpConversionPattern<ref::AllocOp> {
-  using StatefulOpConversionPattern<ref::AllocOp>::StatefulOpConversionPattern;
+struct ConvertMemRefAlloc final : StatefulOpConversionPattern<memref::AllocOp> {
```
burgholzer: I'll just bring this up here although it is a bit of a bigger thing: Does it really make too much sense to still keep the registers in MQTOpt here?
Wouldn't this be quite a bit simpler if we also switched this to `memref` in the Opt dialect for now?
I understand that these registers will probably go away in the future; but until we have that figured out (and I am still not sure about some of the details), it might make sense to handle registers consistently across our dialect.
Would you agree?

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
```diff
@@ -54,7 +58,8 @@ class MQTOptToMQTRefTypeConverter final : public TypeConverter {
 
     // QregType conversion
     addConversion([ctx](opt::QubitRegisterType /*type*/) -> Type {
-      return ref::QubitRegisterType::get(ctx);
+      auto qubitType = ref::QubitType::get(ctx);
+      return MemRefType::get({ShapedType::kDynamic}, qubitType);
```
burgholzer: This somewhat feels wrong to me. Most registers will have a statically defined shape. Is that information potentially available through the `type` parameter that is unused here? This might become irrelevant if we also decide to use memref for MQTOpt for now

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
```diff
@@ -266,6 +266,8 @@ struct MQTOptToMQTRef final : impl::MQTOptToMQTRefBase<MQTOptToMQTRef> {
 
     target.addIllegalDialect<opt::MQTOptDialect>();
     target.addLegalDialect<ref::MQTRefDialect>();
+    target.addLegalDialect<arith::ArithDialect>();
+    target.addLegalDialect<memref::MemRefDialect>();
```
burgholzer: These two additional lines here slightly confuse me. What's the default behavior here? Is any dialect not listed here automatically legal or illegal?
Because we definitely also use other dialects when translating from MQT Core IR to MLIR.

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
burgholzer: I think you would have to similarly get the `DefiningOp` for the size value and extract the value from there.
something like (untested)
```c++
size = adaptor.getDynamicSizes().front();
auto constantOp = size.getDefiningOp<LLVM::ConstantOp>();
const auto value = dyn_cast<IntegerAttr>(constantOp.getValue()).getInt();
getState().numQubits += value;
```

mlir/lib/Conversion/QIRToMQTRef/QIRToMQTRef.cpp
```diff
@@ -391,8 +391,13 @@ struct ConvertQIRCall final : StatefulOpConversionPattern<LLVM::CallOp> {
     }
     // match alloc register
     if (fnName == "__quantum__rt__qubit_allocate_array") {
-      const auto newOp = rewriter.replaceOpWithNewOp<ref::AllocOp>(
-          op, qregType, adaptor.getOperands());
+      const auto memRefType =
+          MemRefType::get({ShapedType::kDynamic}, qubitType);
```
burgholzer, line 395: The dynamic aspect also doesn't seem to quite fit here. In most cases, this will be statically known, right?

mlir/unittests/translation/test_translation.cpp
```diff
@@ -302,27 +313,27 @@ std::string getCheckStringTestCaseUnitary(const TestCaseUnitary& testCase) {
...
 
   // Add operation-specific check
   result += testCase.checkStringOperation;
   result += "\n";
 
-  // Add register deallocation
-  result += "CHECK: \"mqtref.deallocQubitRegister\"(%[[Reg]]) : "
-            "(!mqtref.QubitRegister) -> ()\n";
+  // Add dealocation
+  // Add register allocation
```
burgholzer:
```suggestion
```

mlir/test/Conversion/qir-to-mqtref.mlir
```diff
@@ -12,8 +12,9 @@
 module {
     // CHECK-LABEL: llvm.func @testConvertAllocRegister()
     llvm.func @testConvertAllocRegister() attributes {passthrough = ["entry_point", ["output_labeling_schema", "schema_id"], ["qir_profiles", "base_profile"], ["required_num_qubits", "2"], ["required_num_results", "0"], ["qir_major_version", "1"], ["qir_minor_version", "0"], ["dynamic_qubit_management", "true"], ["dynamic_result_management", "false"]]}  {
-        // CHECK: %[[size:.*]] = llvm.mlir.constant(14 : i64) : i64
-        // CHECK: %[[r_0:.*]] = "mqtref.allocQubitRegister"(%[[size]]) : (i64) -> !mqtref.QubitRegister
+        // CHECK: %[[Int:.*]] = llvm.mlir.constant(14 : i64) : i64
```
burgholzer, line 15: Just to emphasize a previous point: I would have expected this to be a single line memref alloc with a statically defined size.

mlir/test/Conversion/mqtopt-to-mqtref.mlir
```diff
@@ -13,7 +13,8 @@
 module {
     // CHECK-LABEL: func.func @testConvertAllocOpAttribute()
     func.func @testConvertAllocOpAttribute() {
-        // CHECK: %[[r_0:.*]] = "mqtref.allocQubitRegister"() <{size_attr = 2 : i64}>
+        // CHECK: %[[I2:.*]] = arith.constant 2 : index
```
burgholzer: Similar to the QIR conversion, I would have expected this to be a statically defined memref.

docs/mlir/index.md
```diff
@@ -28,13 +28,42 @@ Contributions are welcome.
...
+The approach of using MLIR-native `memref` operations allows us to stay more flexible.
+We had initially defined an MQTRef-specific `QubitRegister` type, but this restricted us without adding benefits.
+By using `memref<?x!mqtref.Qubit>` operands to rpresent quantum registers, development can be more agnostic to implementation specifics.
+
 ## Classical Result Semantics
 
 The `measure` operations of the MQTRef and MQTOpt dialects return classical results as `i1` values.
-If an input program defines a classical register, a `memref<?xi1>` operation of appropriate size is allocated and measurement results are stored into it.
-Similarly, if the input program contains a classically controlled operation, the necessary `i1` values are loaded from the `memref<?xi1>` operation.
+If an input program defines a classical register, a `memref<?xi1>` operand of appropriate size is allocated and measurement results are stored into it.
```
burgholzer: I'd think that `operation` is actually correct here

docs/mlir/index.md
```diff
@@ -28,13 +28,42 @@ Contributions are welcome.
 See the {doc}`contribution guide <../contributing>` for more information.
 :::
 
+## Quantum Registers
```
burgholzer: I think it could make sense to combine both register sections into a single section that elaborates a little bit more on the design decisions. Running this through some LLM should already give quite some good results.
I'd also try to write this with a little less "Innensicht" because the current text very much assumes familiarity with the history of our MLIR efforts, which hardly any (if any) external person reading this documentation will have.
This part of the documentation should be informative and provide reasoning as well as examples for the key design decisions across our MLIR efforts.
(maybe the above is already a good LLM prompt ;-) )

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
```diff
@@ -96,83 +106,73 @@ class MQTRefToMQTOptTypeConverter final : public TypeConverter {
...
-
-      // update the optQreg
-      optQreg = optInsertOp.getOutQreg();
+      storeOp->moveBefore(op);
 
       // erase the refQubit entry from the maps
       getState().qubitMap.erase(refQubit);
-      getState().qubitDataMap.erase(refQubit);
+      getState().qubitIndexMap.erase(refQubit);
     }
```
denialhaag, line 169: There is still something wrong here. When a `StoreOp` is created, the MQTRef `DeallocOp` is not replaced correctly. If no qubit is loaded (and hence no qubit needs to be stored), the `DeallocOp` is replaced correctly. The tests are all passing because I have deleted the two failing tests for now.

mlir/lib/Conversion/QIRToMQTRef/QIRToMQTRef.cpp
denialhaag: Yes, that's a very good point! It's fixed now.

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
denialhaag: I have removed them again. I think I had put them to test something.

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
denialhaag: Yes, the `type` parameter can be used to do this properly. It's fixed now.

mlir/test/Conversion/mqtopt-to-mqtref.mlir
denialhaag: Fixed. 👍

mlir/test/Conversion/qir-to-mqtref.mlir
denialhaag: I'm not sure about this one. 🤔 If I understood correctly, the QIR register is allocated using an operand. I didn't change anything about that. Should I try casting to the underlying value, or do we want to leave the behavior as is?

burgholzer: Hm. Sorry for that. Must have missed that somehow.
Just out of interest: can you try running canonicalization on this program? If canonicalization is clever enough it would probably do the conversion from dynamic to statically sized memref here automatically.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
denialhaag: I do agree that it makes sense to replace the `QubitRegister` also in the MQTOpt dialect! Done. 😌

mlir/test/Conversion/qir-to-mqtref.mlir
denialhaag: No no, not at all! I was genuinely wondering if there was something that could be improved.

I tried out a few passes now, but none of them did the job. In the [list of built-in passes](https://mlir.llvm.org/docs/Passes/), I also couldn't find anything that sounds like what we want. 😕

docs/mlir/index.md
denialhaag: `operand` is gone again. Not sure why I always stumble over that.

docs/mlir/index.md
denialhaag: The LLM did a pretty good job. Let me know what you think!

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
denialhaag: I finally figured it out in https://github.com/munich-quantum-toolkit/core/pull/1211/commits/6dee61269de0bd45d2942374a6c5514f10c84e11; the arith-to-LLVM conversion apparently inserts two `UnrealizedConversionCastOp`. 🤠

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
denialhaag: And finally, I was also able to fix this in https://github.com/munich-quantum-toolkit/core/pull/1211/commits/e6a9b9ccce06aaaf84f0ba4d0241ef6977d2e645. I'm very confused why it's necessary to make the `StoreOp` explicitly legal.

denialhaag: Is it maybe because the `MemRefDialect` is now a dependent dialect?

https://github.com/munich-quantum-toolkit/core/blob/c4541556b475a5086d70b2655230e96e5c0c16e9/mlir/include/mlir/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.td#L24-L28

Does it then make sense to mark the dialect as legal altogether? The operations we explicitly do want to replace are already marked as illegal using `addDynamicallyLegalOp`.

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
```diff
@@ -52,43 +83,45 @@ class MQTOptToMQTRefTypeConverter final : public TypeConverter {
...
-                        : IntegerAttr{};
+    const auto& qubitType = ref::QubitType::get(rewriter.getContext());
+    const auto& memRefType =
+        MemRefType::get(op.getType().getShape(), qubitType);
 
-    // replace the opt alloc operation with a ref alloc operation
-    rewriter.replaceOpWithNewOp<ref::AllocOp>(op, qregType, op.getSize(),
-                                              sizeAttr);
+    rewriter.replaceOpWithNewOp<memref::AllocOp>(op, memRefType,
+                                                 adaptor.getDynamicSizes());
```
burgholzer, line 111: Crazy thought probably: are these conversion patterns at all necessary here? There is a pattern for replacing the opt qubit type with the ref one. Is that alone already sufficient to facilitate the entire conversion? for the memrefs?

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
```diff
@@ -267,10 +289,19 @@ struct MQTOptToMQTRef final : impl::MQTOptToMQTRefBase<MQTOptToMQTRef> {
...
 
-    patterns.add<ConvertMQTOptAlloc, ConvertMQTOptDealloc,
+    target.addDynamicallyLegalOp<memref::AllocOp>(
+        [&](memref::AllocOp op) { return !isQubitType(op); });
+    target.addDynamicallyLegalOp<memref::DeallocOp>(
+        [&](memref::DeallocOp op) { return !isQubitType(op); });
+    target.addDynamicallyLegalOp<memref::LoadOp>(
+        [&](memref::LoadOp op) { return !isQubitType(op); });
+    target.addDynamicallyLegalOp<memref::StoreOp>(
+        [&](memref::StoreOp op) { return !isQubitType(op); });
```
burgholzer, line 299: This is slightly related to the comment above: Is this actually needed? Given how the opt dialect is marked as illegal, I would have assumed that the respective type is illegal as well and so, transitively any memref using the type would be illegal.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
burgholzer: I think it needs to be marked as legal because it is explicitly introduced as part of the conversion, but I am also not 100% sure about that.
This is also on of the only memref conversion pattern were non-trivial code is being added.
I was wondering: do we actually need the store operation at all? Given how registers will probably be going away from MQTOpt anyway, we might as well simplify things here. But maybe I am overlooking something.
Generally, the same comments as for the other conversion direction hold.

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
```diff
@@ -162,19 +195,36 @@ struct ConvertMQTRefAllocQIR final : StatefulOpConversionPattern<ref::AllocOp> {
...
+        llvm::errs() << "Dynamic size could not be resolved\n";
+        return failure();
+      }
+      auto sizeValueIdx = unrealizedConvOp0->getOperand(0);
+      auto unrealizedConvOp1 =
+          sizeValueIdx.getDefiningOp<UnrealizedConversionCastOp>();
+      if (!unrealizedConvOp1) {
+        llvm::errs() << "Dynamic size could not be resolved\n";
+        return failure();
+      }
```
burgholzer: This does seem strange to me and not quite what I would have expected. Are we really doing things correctly here?

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
```diff
@@ -887,11 +954,19 @@ struct MQTRefToQIR final : impl::MQTRefToQIRBase<MQTRefToQIR> {
...
     target.addIllegalDialect<ref::MQTRefDialect>();
-    mqtPatterns.add<ConvertMQTRefAllocQIR>(typeConverter, ctx, &state);
-    mqtPatterns.add<ConvertMQTRefDeallocQIR>(typeConverter, ctx);
+
+    target.addDynamicallyLegalOp<memref::AllocOp>(
+        [&](memref::AllocOp op) { return !isQubitType(op); });
+    target.addDynamicallyLegalOp<memref::DeallocOp>(
+        [&](memref::DeallocOp op) { return !isQubitType(op); });
+    target.addDynamicallyLegalOp<memref::LoadOp>(
+        [&](memref::LoadOp op) { return !isQubitType(op); });
```
burgholzer, line 960: Again, I am wondering whether this is needed.
Maybe one needs to define the memref dialect as legal. I'd suppose memrefs of mqtref qubits would still be illegal then.

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
denialhaag: I also thought so, but I tried commenting out `patterns.add<ConvertMQTOptMemRefAlloc>(...)`, and the tests immediately start failing. I double-checked one more time now, but it seems necessary. 😕 

For the MQTRef-to-MQTOpt conversion, we definitely need the patterns because we update the state there.

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
denialhaag: I also tried that again just now, but it seems necessary. The type doesn't seem enough. 😕

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
denialhaag: I tried making the `MemRefDialect` legal just now, but the operations still needed to be explicitly marked illegal.

It's absolutely possible that I'm missing a configuration that could make it work, but I couldn't find another way to make it work in the last days.

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
denialhaag: No, I also found that very confusing. If I understood everything correctly, this happens already in this step, which we don't have much control over:

https://github.com/munich-quantum-toolkit/core/blob/8acee909439768e3b487a458fee0c517df7b59ae/mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp#L933-L946

@li-mingbao needed to do something similar in the same conversion:

https://github.com/munich-quantum-toolkit/core/blob/8acee909439768e3b487a458fee0c517df7b59ae/mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp#L676-L684

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
burgholzer: Hm. Alright. That makes most comments of the review obsolete I suppose, but that's totally fine. We have a working solution here that we can iterate on in subsequent PRs 😌 feel free to resolve here.

mlir/lib/Conversion/MQTOptToMQTRef/MQTOptToMQTRef.cpp
burgholzer: Unfortunate 😢 would have been too nice.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
denialhaag: The only place where they seemed necessary was the round-trip pass, where the function returns the quantum register.

To be perfectly honest, I never really understood why we had the `InsertOp` in the first place. 😬

I think it should be very straightforward to remove the `StoreOp` everywhere (besides the round-trip pass, maybe). It would certainly make the MQTRef-to-MQTOpt conversion a lot simpler. 🤔

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
burgholzer: Just reading https://gricad-gitlab.univ-grenoble-alpes.fr/violetf/llvm-project/-/blob/8b264613260cb5c881b3b5634189bdceae3f93d8/mlir/docs/ConversionToLLVMDialect.md#index-type, this might be something that we can avoid by running some additional conversion passes. There seems to be some kind of conversion from index to LLVM.

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
burgholzer: Nah. that might just be required here. It's a bit odd but as long as it is working for the moment, I am fine with it. It's not the most complicated code that we have here and we can always iterate on that in a separate PR.
Feel free to resolve this conversation here.

mlir/test/Conversion/qir-to-mqtref.mlir
burgholzer: hm. probably something to consider in the future in a more detailed fashion.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
burgholzer: In that case, let's remove them. The roundtrip pass is also fairly uninteresting by now as we are very close to actually supporting proper passes on the MLIR side. It might still be used as part of the Catalyst plugin release; although I'd argue that even there, it's only remotely reasonable.
Maybe @flowerthrower can jump in here with a quick assessment.

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
denialhaag: It looks like the conversion can be configured: https://mlir.llvm.org/docs/Passes/#options-16

mlir/test/Conversion/qir-to-mqtref.mlir
denialhaag: Maybe this can also be considered as part of https://github.com/munich-quantum-toolkit/core/issues/1154. 🤔

burgholzer: Yeah. Potentially makes sense.
Maybe worth adding a short comment over there before resolving here.

denialhaag: Done! 👍

burgholzer:
Okay, it took me longer than I'd like to admit, but I finally managed to debug what is going on in the QIR conversion and why this doesn't really work as reliably as it should at the moment.
The reason why the reconciliation pass does nothing on the intermediate program is that the `memref.alloc` operation is still in the program at that point and its operand is an `index`, so the cast cannot be eliminated at that point.
This has sparked a fairly general thought: right now, we are first running the standard MLIR to LLVM conversion passes and only afterwards run the MQTRef -> QIR passes.
I think it might be more clever to switch this around so that the MQTRef operations are being replaced first (that way no unrealized conversions are in place). After the MQTRef operations are removed, all the LLVM conversions can happen after that.

@denialhaag would you mind playing around with that a little bit?

li-mingbao:
Hey,
I am not sure if this is relevant but I wanted to mention why the conversion from the standard dialect to LLVM happens first.
Right now when you translate from `QuantumComputation` to `MQTRef` the input consists of the standard dialects e.g. `arith`, `func` etc.
When you convert something from `QIR` to `MQTRef` the input consists of `LLVM` dialect operations to represent functions and constants and these operations are not converted yet (see issue #1120).
This causes different inputs when you want to convert from `MQTRef` to `QIR`.
That is why I decided to to convert the standard dialects first to make the operands always be from the `LLVM` dialect as this makes it easier to get the defining operation and extract the values if necessary during the `MQTRef` conversion.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
flowerthrower: (Really) quick assessment: it should be possible to adjust the round-trip to work without the register insert.

mlir/lib/Conversion/MQTRefToQIR/MQTRefToQIR.cpp
```diff
@@ -162,19 +197,35 @@ struct ConvertMQTRefAllocQIR final : StatefulOpConversionPattern<ref::AllocOp> {
...
+        return failure();
+      }
+      auto integerAttr = dyn_cast<IntegerAttr>(constantOp.getValue());
+      if (!integerAttr) {
+        llvm::errs() << "Dynamic size could not be resolved. Size cannot be "
+                        "cast to IntegerAttr.\n";
+        return failure();
+      }
+      size = integerAttr.getInt();
+      sizeValue = adaptor.getDynamicSizes().front();
```
denialhaag, line 222: I ended up using `op.getDynamicSizes().front()` for computing the size and `adaptor.getDynamicSizes().front()` as an input for the `LLVM::CallOp`.

The only reason for not creating a `LLVM::ConstantOp` from the size is that I didn't like an inconsistency in the tests. Because we are using the adaptor for getting the index from the `LoadOp`, one of the tests would have looked like this:

```
// CHECK: %[[size:.*]] = llvm.mlir.constant(1 : i64) : i64
// CHECK: %[[index:.*]] = llvm.mlir.constant(0 : index) : i64
```

In principle, we could of course also explicitly create a `LLVM::ConstantOp` for the latter case, but it seems like unnecessary overhead to first extract the underlying integer when we don't need it.

Let me know what you think, @burgholzer!

burgholzer: I think this looks like a solid solution for now. Let's go with it.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
burgholzer: Let's tackle this in a follow-up and rather try to get this PR merged sooner than later. At least the tests are passing at the moment.
Worth opening an issue for that so that we do not forget?

burgholzer (changes requested):
I believe this is pretty much ready to go in, which should allow us to get started with #1125 (@DRovara).
So I hope this will be the last review requesting changes. I would like to keep the coverage for our MLIR activities as high as we can and this PR, at the moment, has some room for improvement in that regard.
The missing lines in
- `MQTOptToMQTRef.cpp` as well as `MQTRefToMQTOpt.cpp` should be easily covered by introducing some tests using classical registers, measurements and classic-controlled operations
- For the `ToQuantumComputationPattern.cpp` at least the `DeallocOp` related missing lines should be coverable fairly easily.
- For the `MQTRefToQIR.cpp` file it should be easy to also cover the missing memref conversions/checks by using a program as input that uses classical registers.

I think it is fairly hard to cover the various failure cases and would be fine with ignoring them.
Something that could make things slightly easier with regard to coverage would be to unify some of these `isQubitType` memref checks which are repeated across multiple passes right now. But that can also be done in a follow up.

mlir/lib/Conversion/MQTRefToMQTOpt/MQTRefToMQTOpt.cpp
denialhaag: Makes sense! I created https://github.com/munich-quantum-toolkit/core/issues/1220. 👍

denialhaag:
I improved the tests a bit now, but I don't think we can cover such a failure:

```c++
if (!isQubitType(op)) {
  return failure();
}
```

The patterns are not entered for non-qubit memrefs because they are not marked illegal. Not sure if it's better to just delete these statements. 🤔 

Regarding `ToQuantumComputationPattern.cpp`, the `DeallocOp` is a bit misplaced there because it doesn't make much, if any, sense to ever deallocate the quantum register as long as it's one of the function's return values. Or am I missing something there? 🤔 

https://github.com/munich-quantum-toolkit/core/blob/05817185b8027b2723dfef5e358737037ca8af0c/mlir/lib/Dialect/MQTOpt/Transforms/FromQuantumComputationPattern.cpp#L180-L201

mlir/test/Conversion/mqtopt-to-mqtref.mlir
```diff
@@ -8,6 +8,25 @@
...
+        // CHECK: memref.dealloc %[[Memref]] : memref<1xi1>
+
+        %i0 = arith.constant 0 : index
+        %memref = memref.alloc() : memref<1xi1>
+        %0 = memref.load %memref[%i0] : memref<1xi1>
+        memref.dealloc %memref : memref<1xi1>
+
+        return
+    }
+}
```
burgholzer, line 30: Could be worth to also add a `memref.store` here to cover the full spectrum.

mlir/test/Conversion/mqtref-to-qir.mlir
```diff
@@ -8,6 +8,24 @@
...
+        // CHECK-NOT: llvm.call @__quantum__rt__qubit_release_array
+
+        %i0 = arith.constant 0 : index
+        %memref = memref.alloc() : memref<1xi1>
+        %0 = memref.load %memref[%i0] : memref<1xi1>
+        memref.dealloc %memref : memref<1xi1>
+
+        return
+    }
+}
```
burgholzer, line 27: A `memref.store` would probably also fit well here.

burgholzer (commented):
Two small comments that are fairly inconsequential probably.

burgholzer:
> I improved the tests a bit now, but I don't think we can cover a failure such as the one below. The patterns are not entered for non-qubit memrefs because they are not marked illegal. Not sure if it's better to just delete these statements. 🤔
> 
> ```c++
> if (!isQubitType(op)) {
>   return failure();
> }
> ```
>

Hm. The patterns could be used in a context where the operations are not legal, so I'd rather keep them in place and accept that they cannot be covered.

> Regarding `ToQuantumComputationPattern.cpp`, the `DeallocOp` is a bit misplaced there because it doesn't make much, if any, sense to ever deallocate the quantum register as long as it's one of the function's return values. Or am I missing something there? 🤔
> 
> https://github.com/munich-quantum-toolkit/core/blob/05817185b8027b2723dfef5e358737037ca8af0c/mlir/lib/Dialect/MQTOpt/Transforms/FromQuantumComputationPattern.cpp#L180-L201

Yeah. that's a bit odd to be fair. So are the conversions in general 🙃 let's remove the DeallocOp from there.

mlir/test/Conversion/mqtopt-to-mqtref.mlir
denialhaag: Good point! Done.

mlir/test/Conversion/mqtref-to-qir.mlir
denialhaag: Done.

denialhaag:
> > ...
> 
> Hm. The patterns could be used in a context where the operations are not legal, so I'd rather keep them in place and accept that they cannot be covered.

Okay, makes sense!

> > ...
> 
> Yeah. that's a bit odd to be fair. So are the conversions in general 🙃 let's remove the DeallocOp from there.

Done!

PR #1223, opened 2025-09-25
✨ Add a NA FoMaC Recovering Structure of the QDMI Device

ystade:
## Description

Thir PR adds a specific FoMaC for NA inheriting from the generic existing FoMaC. The core part is the recovery of the `Device` structure defined in the Generator of the NA QDMI Device.

Additionally, it includes some few fixes in the NA QDMI Device implementation that were revealed in tests of this PR. It also reactivates the option to rebuild `mqt.core` on `uv sync` in the `pyproject.toml`.

## Checklist:


bindings/na/fomac/fomac.cpp
ystade, line 1: I put that here since it is related to the Python bindings: I am not sure whether the location of the NA FoMaC implementation is the best as it is right now. I came to the current setup by asking myself, how I would like it on the Python side. Here, I can now write
```python
from mqt.core.na.fomac import devices
devices()
```
Which I think is a natural way to get the NA FoMaC. This rationale resulted in the current file structure also on the C++ side.

include/mqt-core/na/device/Generator.hpp
ystade, line 1: This file contains some bug-fixes that were uncovered in testing this PR.

include/mqt-core/na/fomac/Device.hpp
```diff
@@ -0,0 +1,157 @@
...
+#include <nlohmann/json.hpp>
+#include <vector>
+
+namespace na {
+
+/**
+ * @brief Class representing the FoMaC library with neutral atom extensions.
+ * @see qdmi::FoMaC
+ */
+class FoMaC : public qdmi::FoMaC {
```
ystade: Do we like the chosen namespaces?

include/mqt-core/fomac/FoMaC.hpp
ystade, line 1: Also a fix that was detected in this PR during testing.

pyproject.toml
```diff
@@ -297,6 +299,7 @@ repair-wheel-command = "delvewheel repair -w {dest_dir} {wheel} --namespace-pkg
 
 [tool.uv]
 required-version = ">=0.5.20"
+reinstall-package = ["mqt-core"]
```
ystade, line 302: Reverting this change from a previous PR was needed, as `uv sync` would not rebuild the project upon changes in the C++ bindings—even with the lates `uv` version.

pyproject.toml
```diff
@@ -336,6 +339,7 @@ test = [
   "pandas[output-formatting]>=2.2.3; python_version >= '3.13'",
   "qiskit[qasm3-import]>=1.0.0",
   "numpy>=2.1; python_version >= '3.13'",
+  "deepdiff>=8.6.0",
```
ystade:
```suggestion
```

src/na/device/CMakeLists.txt
```diff
@@ -104,7 +104,7 @@ if(NOT TARGET ${TARGET_NAME})
 
   # Generate prefixed QDMI headers
   generate_prefixed_qdmi_headers(${QDMI_PREFIX})
-  file(GLOB_RECURSE QDMI_HDRS ${CMAKE_CURRENT_BINARY_DIR}/include/mqt_na_qdmi/**.hpp)
+  file(GLOB_RECURSE QDMI_HDRS ${CMAKE_CURRENT_BINARY_DIR}/include/mqt_na_qdmi/**.h)
```
ystade, line 107: Actually, I do not know why it worked before... However, now it looks correct and works.

src/na/device/Generator.cpp
```diff
@@ -253,6 +256,34 @@ auto writeSites(const Device& device, std::ostream& os) -> void {
                  << "Sites.emplace_back(var.back().get())";
             }
           }
+          // this generator (same as the device implementation) only supports
```
ystade, line 260: The local multi-qubit operations were forgotten to be exported via QDMI in a previous PR.

src/na/fomac/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+  # add sources to target
+  target_sources(${TARGET_NAME} PRIVATE Device.cpp)
+
+  # add headers using file sets
+  target_sources(${TARGET_NAME} PUBLIC FILE_SET HEADERS BASE_DIRS ${MQT_CORE_INCLUDE_BUILD_DIR}
+                                       FILES ${MQT_CORE_INCLUDE_BUILD_DIR}/na/fomac/Device.hpp)
+
+  # Link nlohmann_json, spdlog
+  target_link_libraries(${TARGET_NAME} PUBLIC MQT::CoreQDMIFoMaC nlohmann_json::nlohmann_json)
```
ystade: Maybe you comment on whether this is correct to link `nlohmann_json` publicly. To me it does not make to much sense but I got problems when linking it privately.

bindings/na/fomac/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    ${SOURCES}
+    MODULE_NAME
+    fomac
+    INSTALL_DIR
+    ./na
+    LINK_LIBS
+    MQT::CoreNAFoMaC)
+
+  # Set C++ standard
+  target_compile_features(${TARGET_NAME} PRIVATE cxx_std_20)
```
burgholzer: I believe this is not necessary anymore since a couple of PRs back as the `add_mqt-python_bidning` macro automatically adds that.

```suggestion
```

bindings/na/fomac/fomac.cpp
burgholzer: I would argue that this makes sense. especially for the dedicated FoMaC libraries for the technologies.
There still is a generic FoMaC in `mqt.core.qdmi.fomac` though. Would it make sense to move this to `mqt.core.fomac`? somehow one of these has `qdmi` in it, one does not. An inconsistency we are willing to accept?

bindings/na/fomac/fomac.cpp
```diff
@@ -0,0 +1,100 @@
...
+  auto lattice = py::class_<na::Device::Lattice>(device, "Lattice");
+
+  auto vector = py::class_<na::Device::Vector>(lattice, "Vector");
+  vector.def_readonly("x", &na::Device::Vector::x);
+  vector.def_readonly("y", &na::Device::Vector::y);
+  vector.def("__repr__", [](const na::Device::Vector& v) {
+    return "<Vector x=" + std::to_string(v.x) + " y=" + std::to_string(v.y) +
+           ">";
+  });
+  vector.def(py::self == py::self); // NOLINT(misc-redundant-expression)
```
burgholzer: Truly minor: Instead of replicating these `// NOLINT(misc-redundant-expression)` statements I think it might be prettier to simply do a `// NOLINTBEGIN(misc-redundant-expression)` in the beginning of the file as well as a matching end statement.

bindings/na/fomac/fomac.cpp
```diff
@@ -0,0 +1,100 @@
...
+  });
+  size.def(py::self == py::self); // NOLINT(misc-redundant-expression)
+  size.def(py::self != py::self); // NOLINT(misc-redundant-expression)
+
+  region.def_readonly("origin", &na::Device::Region::origin);
+  region.def_readonly("size", &na::Device::Region::size);
+  region.def("__repr__", [](const na::Device::Region& r) {
+    return "<Region origin=" +
+           py::repr(py::cast(r.origin)).cast<std::string>() +
+           " size=" + py::repr(py::cast(r.size)).cast<std::string>() + ">";
```
burgholzer: Minor and not particularly important for performance, but wouldn't it be cleaner to simply have the formatting as a loose function in an anonymous namespace and call it in both cases? Avoids the Python operations.

include/mqt-core/na/device/Generator.hpp
```diff
@@ -387,6 +395,8 @@ auto writeJSONSchema(const std::string& path) -> void;
  * @param device is the protobuf representation of the device.
  * @param os is the output stream to write the header file to.
  * @throws std::runtime_error if the file cannot be opened or written to.
+ * @warning This implementation only supports multi-qubit gates up to two
```
burgholzer:
```suggestion
 * @note This implementation only supports multi-qubit gates up to two
```
I believe this is just fine as a note. No need to scream at people ;o)

include/mqt-core/na/device/Generator.hpp
```diff
@@ -396,6 +406,8 @@ auto writeHeader(const Device& device, std::ostream& os) -> void;
  * @param device is the protobuf representation of the device.
  * @param path is the path to write the header file to.
  * @throws std::runtime_error if the file cannot be opened or written to.
+ * @warning This implementation only supports multi-qubit gates up to two
```
burgholzer:
```suggestion
 * @note This implementation only supports multi-qubit gates up to two
```

include/mqt-core/na/device/Generator.hpp
burgholzer: Good to have these ironed out 👏🏼

include/mqt-core/na/fomac/Device.hpp
burgholzer: This directly relates to the Python package structure, the repository structure, and potentially a couple more elements.
Similarly to the other comment, I like this in principle but just wanted to raise the concern that one contains an explicit mention of `qdmi` while the other does not. I am entirely fine with that. It just needs to be a purposeful decision.

include/mqt-core/na/fomac/Device.hpp
```diff
@@ -0,0 +1,157 @@
...
+public:
+  /**
+   * @brief Class representing a quantum device with neutral atom extensions.
+   * @see qdmi::FoMaC::Device
+   * @note Since it inherits from @ref na::Device, Device objects can be
+   * converted to `nlohmann::json` objects.
```
burgholzer, line 32: This is precisely the reason why `nlohmann_json` needs to linked publicly.
As soon as one of the public function signatures of a library contains a type from another library, the other library needs to be publicly linked for the symbols to be available.

Side note: I really like this 👍🏼

include/mqt-core/na/fomac/Device.hpp
```diff
@@ -0,0 +1,157 @@
...
+     * @brief Initializes the shuttling units from the underlying QDMI device.
+     */
+    auto initShuttlingUnitsFromDevice() -> void;
+
+  public:
+    /**
+     * @brief Constructs a Device object from a qdmi::FoMaC::Device object.
+     * @param device The qdmi::FoMaC::Device object to wrap.
+     */
+    explicit Device(const qdmi::FoMaC::Device& device);
```
burgholzer: Just thinking out loudly here:
This requires the existence of a generic fomac instance for a device handle before being able to wrap it in the NA specific fomac.
Given how this is the only way to get a device handle from the client side at the moment, that is fine in principle.
It's kind of also nice to get the devices from the generic fomac and then get the more detailed information from the specific fomac.

Don't know if there is anything to comment on here. Just wanted to share my thoughts.

include/mqt-core/na/fomac/Device.hpp
```diff
@@ -0,0 +1,157 @@
...
+      nlohmann::to_json(nlohmannJsonJ,
+                        static_cast<const na::Device&>(nlohmannJsonT));
+    }
+  };
+
+  /// @brief Deleted default constructor to prevent instantiation.
+  FoMaC() = delete;
+
+  /// @see QDMI_SESSION_PROPERTY_DEVICES
+  [[nodiscard]] static auto getDevices() -> std::vector<Device>;
```
burgholzer, line 154: This is more of a side comment to mentally note down somewhere.
At some point (beyond this PR), we probably will want to have more public query functions for the FoMaC libraries. Such as requesting a particular device (matched, e.g., by name), or, potentially, even to add a device based on a static description (which would be entirely possible for the NA JSON based format at the moment).

The first problem can already be solved by manually writing the filtering code. So providing it as part of the library would be pure convenience, which still is a good goal.

The second one is probably not possible at the moment and will need some engineering.

Could it make sense to create tracking issues for these kinds of things? DO they even make sense?

json/na/device.json
burgholzer, line 1: Any particular reason for the changes here? (probably overlooked something)

python/mqt/core/na/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+# Copyright (c) 2025 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""NA module."""
```
burgholzer: A slightly more elaborate description would probably be adequate here as this also ends up in the API documentation.

python/mqt/core/na/fomac.pyi
```diff
@@ -0,0 +1,105 @@
...
+    The list of trap positions in the device.
+    """
+    t1: int
+    """
+    The T1 time of the device.
+    """
+    t2: int
+    """
+    The T2 time of the device.
+    """
```
burgholzer, line 101: Just stumbled over this for a second.
Why are T1 and T2 times global device properties here when in QDMI they are tied to the sites?
Does it make sense to explicitly distinguish between these?
Conceptionally, the "device" itself does not have a T1 and T2 time, but it's atoms have these properties. Furthermore, this will most likely be averaged across sites, which would probably be worth reporting in the docstring.

This comment does not only apply to the local scope here.

src/na/device/CMakeLists.txt
burgholzer: Most likely because the include paths are correctly set up anyway. Potentially, the installed distribution was slightly messed up before.

src/na/device/Generator.cpp
```diff
@@ -253,6 +257,35 @@ auto writeSites(const Device& device, std::ostream& os) -> void {
...
+              for (const auto& [i2, x2, y2] :
+                   sites | std::views::take(sites.size() - 1)) {
+                if (x2 >= operation.region.origin.x &&
+                    x2 <=
+                        operation.region.origin.x +
+                            static_cast<int64_t>(operation.region.size.width) &&
+                    y2 >= operation.region.origin.y &&
+                    y2 <= operation.region.origin.y +
+                              static_cast<int64_t>(
+                                  operation.region.size.height)) {
```
burgholzer: This is just my inner monk again (or maybe my math background), sorry for that, but I feel it is unintuitive how these chained ifs are written. Mathematically, one would write these as `origin.x <= x <= origin.x + width` and I think it looks much nicer and intuitive to replicate that structure in the code as opposed to `x >= origin.x && x <= origin.x + width`.

src/na/fomac/CMakeLists.txt
burgholzer: Hope it makes sense considering the other comment I placed in the code.

src/na/fomac/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+  # set c++ standard
+  target_compile_features(${TARGET_NAME} PRIVATE cxx_std_20)
+
+  # set versioning information
+  set_target_properties(
+    ${TARGET_NAME}
+    PROPERTIES VERSION ${PROJECT_VERSION}
+               SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
+               EXPORT_NAME CoreNAFoMaC)
```
burgholzer: This should not be necessary anymore given recent changes

```suggestion
```

src/na/fomac/Device.cpp
```diff
@@ -0,0 +1,328 @@
...
+  durationUnit.scaleFactor = getDurationScaleFactor().value_or(1.0);
+}
+auto FoMaC::Device::initDecoherenceTimesFromDevice() -> void {
+  // find first non-zone site as reference
+  const auto& sites = getSites();
+  auto referenceSite = *std::ranges::find_if(
+      sites, [](const auto& site) { return !site.isZone().value_or(false); });
+  decoherenceTimes.t1 = *referenceSite.getT1();
+  decoherenceTimes.t2 = *referenceSite.getT2();
+}
```
burgholzer, line 93: Related to another comment: the global T1 and T2 times should probably be averaged over all non-zone sites.

src/na/fomac/Device.cpp
```diff
@@ -0,0 +1,328 @@
...
+          std::views::filter(
+              [](const qdmi::FoMaC::Device::Operation& op) -> bool {
+                return op.isZoned() && !op.getQubitsNum().has_value();
+              }),
+      [&shuttlingOpTuples](const qdmi::FoMaC::Device::Operation& op) {
+        // extract the int from, e.g., `load<0>`, `move<1>`, `store<2>`
+        const auto name = op.getName();
+        const auto start = name.find('<');
+        const auto end = name.find('>');
+        const auto id = static_cast<size_t>(
```
burgholzer, line 294: The fact that we are doing string parsing here slightly points towards something in QDMI not yet being exposed properly, doesn't it?

src/na/fomac/Device.cpp
```diff
@@ -0,0 +1,328 @@
...
+}
+auto FoMaC::getDevices() -> std::vector<Device> {
+  const auto& qdmiDevices = qdmi::FoMaC::getDevices();
+  std::vector<Device> devices;
+  devices.reserve(qdmiDevices.size());
+  std::ranges::transform(
+      qdmiDevices, std::back_inserter(devices),
+      [](const qdmi::FoMaC::Device& dev) -> Device { return Device(dev); });
+  return devices;
+}
```
burgholzer, line 335: This does not yet look correct to me. Not every device will be a neutral atom device and I would argue that this function should only return the Devices that actually provide all the necessary information.
This will require some extra logic presumably.
Given how this PR is already in a really good shape, I'd be fine with creating a tracking issue for this, merging the PR here and directly continuing the resolution of the tracking issue.

test/na/fomac/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+#
+# Licensed under the MIT License
+
+if(TARGET MQT::CoreNAFoMaC)
+  # Set test target name
+  set(TARGET_NAME ${MQT_CORE_TARGET_NAME}-na-fomac-test)
+  # Add the test executable
+  package_add_test(${TARGET_NAME} MQT::CoreNAFoMaC test_device.cpp)
+  # Link further libraries
+  target_link_libraries(${TARGET_NAME} PRIVATE nlohmann_json::nlohmann_json)
```
burgholzer: This should not be necessary as the Fomac library already publicly links against the library.

```suggestion
```

test/na/fomac/test_device.cpp
```diff
@@ -0,0 +1,69 @@
...
+    const auto& v1 = lattice["latticeVector1"];
+    const auto& v2 = lattice["latticeVector2"];
+    if (v1["x"] > v2["x"] || (v1["x"] == v2["x"] && v1["y"] > v2["y"])) {
+      std::swap(lattice["latticeVector1"], lattice["latticeVector2"]);
+    }
+  }
+}
+} // namespace
+TEST(TestNAFoMaC, TrapsJSONRoundTrip) {
+  // NOLINTNEXTLINE(misc-include-cleaner)
```
burgholzer: Might be worth extracting that to a global ignore? there are quite a few ignores here..

test/na/fomac/test_fomac.cpp
```diff
@@ -0,0 +1,69 @@
...
+  std::ifstream file(NA_DEVICE_JSON);
+  ASSERT_TRUE(file.is_open()) << "Failed to open json file: " NA_DEVICE_JSON;
+  // Parse the JSON file
+  try {
+    // NOLINTNEXTLINE(misc-include-cleaner)
+    fomacDevice = nlohmann::json::parse(file);
+  } catch (const nlohmann::json::parse_error& e) {
+    GTEST_FAIL() << "JSON parsing error: " << e.what();
+  }
+  nlohmann::json qdmiDevice = FoMaC::getDevices().front();
```
burgholzer, line 45: This seems slightly fragile to me for the future. How do you know that this is always going to be the first device?
This is probably a good motivation for the additions I was proposing in the other file about querying a particular device.

test/qir/runtime/CMakeLists.txt
```diff
@@ -13,7 +13,10 @@ macro(ADD_QIR_CIRCUIT target_name circuit_path)
     get_filename_component(circuit_name ${circuit_path} NAME_WE)
     add_custom_command(
       OUTPUT ${circuit_name}.o
-      COMMAND clang -c ${circuit_path} -o ${circuit_name}.o
+      COMMAND
+        clang
+        "$<$<BOOL:${CMAKE_OSX_DEPLOYMENT_TARGET}>:-mmacos-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET}>"
+        -c ${circuit_path} -o ${circuit_name}.o
```
burgholzer, line 19: Just because I now see this here: why are we explicitly calling `clang` here for this? This seems fragile to me. Why is that not a regular CMake target?
(sorry for the off-topic comment here)

pyproject.toml
burgholzer: @denialhaag for awareness.
might be worth investigating if there is some way to indicate to uv which files it should consider for evaluating whether it needs to rebuild. Maybe there already is something like that.
This setting at the moment is slightly unfortunate as it will literally rebuild all the time, even if it is just for a docs change.

bindings/na/fomac/fomac.cpp
ystade: Moving the generic `fomac` one level up is one option. Another option would be to have a high-level `qdmi` dir ( on the same level as `mlir`) containing _everything_ QDMI related, incl. all device specific fomacs.

include/mqt-core/na/fomac/Device.hpp
ystade: Ok, I'll adapt to whatever comes out of the discussion above.

On a completely different matter here: I realized that it is sufficient to define the Device class _not_ inside a FoMaC class. This thought was motivated by the question whether one wants to have a `getDevices` function specialized to NA devices or whether the one and only workflow should be to retrieve the devices via the `get Devices` function from the generic FoMaC and then call the constructor of the NA Device with the generic Device as argument. 

So, I am asking two questions here:
1. Should I remove `getDevices` from the NA FoMaC?
2. Should I remove the entire FoMaC class here?

include/mqt-core/na/fomac/Device.hpp
ystade: Ok, I think I covered the related reasoning above now, should probably put it here... I'll resolve here, let's continue above.

json/na/device.json
ystade: Yes, the bug fix above in the Generator changed its behavior. I want to have 10 sites next to each other, according to the fixed behavior this corresponds to a width/height of 9 instead of 10 having a spearation of 1.

python/mqt/core/na/fomac.pyi
ystade: That is kinda specific to our NA QDMI Device implementation which only features a global T1 and T2 time (as usual for NA platforms in the literature).

src/na/fomac/Device.cpp
ystade: Not for our NA QDMI Device implementation as mentioned above. I'll resolve here and leave the rest of the discussion for the other thread.

src/na/fomac/Device.cpp
ystade: I would say that holds for the interpretation of all quantum operations. Their interpretation entirely is based on their `string` name. To resolve this, we would need something like `OpType` as an operation property (and for this particular case an additional property representing the shutting unit id. Maybe there is something more general for the latter...

test/qir/runtime/CMakeLists.txt
ystade: 1. the `.ll` files can only be processed with clang, not with `gcc`.
2. cmake is build all around the infrastructure for compiling C++ files. I actually never tried to feed the `.ll` files as source files to cmake. Maybe this works but could potentially be cumbersome.

python/mqt/core/na/fomac.pyi
burgholzer: Assume some neutral hardware provider comes around with their NA QDMI Device implementation publicly. We might want to use that in our compiler stack. It cannot be guaranteed and is not enforced by the interface that the T1 and T2 times need to be the same for all sites.

I would argue that the FoMaC layer we are building here should be as independent from implementation details as possible

src/na/fomac/Device.cpp
ystade: Or just removing it as suggested above...

src/na/fomac/Device.cpp
burgholzer: > I would say that holds for the interpretation of all quantum operations. Their interpretation entirely is based on their string name.

That is true and it also ties into some of the discussions for the MLIR dialect revamp we are currently having, where the goal is to explicitly provide means for getting a definition of an operation.

Keeping the string parsing here for now seems reasonable. Let's just keep the general discussion in mind as we move along.

bindings/na/fomac/fomac.cpp
burgholzer: Hm. One way or another we will be replicating some structures throughout the library. Either just like now, where FoMaC subdirectories will be in multiple places or with `na`, `sc`, etc. subdirectories being replicated in the global `qdmi` directory.
None of that is particularly nice, but probably unavoidable at some point.
I kind of like the current structure for the `mqt.core.na.fomac` and believe I would be slightly in favor of going with `mqt.core.fomac`.
In the end, FoMaC and QDMI are supposed to be separate components of the stack.

include/mqt-core/na/fomac/Device.hpp
burgholzer: Hm. I don't have an immediate answer for both of these questions.
How would one get the NA version of a device after a potential refactor? Or how would one know that a generic device is actually an NA device?
Given the other use case of asking for a particular device, I think this is well suited in a dedicated NA FoMaC library that is basically a filtered view on the underlying devices with the respective convenience function that might only apply to neutral atom systems.

src/na/fomac/Device.cpp
burgholzer: See above. I do think there is value in having this. But let's discuss in the other thread and base a solution on top of that.

test/qir/runtime/CMakeLists.txt
burgholzer: Yeah, makes sense. Overlooked that.
Maybe it makes sense to guard this behind the availability of clang as a compiler on the target system?
Anyway, sorry for the noise. Feel free to keep this as it is and resolve. I just stumbled over it and that got me thinking.

include/mqt-core/na/fomac/Device.hpp
ystade: The point of the filtered view is a very good point. (however, it should then also do the filtering as you pointed out below, something for another Issue/PR).

Consequently, I leave both options for now. Independently, we should think about what happens if one tries to instantiate a NA FoMaC Device from a generic FoMaC device that is not NA. Should it throw an error, should there be a way to check that in advance before trying to instantiate the NA device?

burgholzer: Depending on how quickly in succession to this PR you plan to work on that, it would probably make sense to create a tracking issue for that. I think it does make sense to resolve this soon as this is an important part of the general API we are building here.

I think if one directly tries, it should fail with a dedicated error almost directly.
I also think that there should be a way to check if a given device can be downcasted/specialized. This would, probably do an internal check if all the required properties are available that are later on queried. This function would then also be called internally when trying to specialize a device.

include/mqt-core/na/fomac/Device.hpp
ystade: #1226

include/mqt-core/na/fomac/Device.hpp
ystade: #1227

test/na/fomac/test_fomac.cpp
ystade: I save that for later as part of #1227.

ystade:
@burgholzer I hope I have addressed all concerns to your satisfaction. The remaining tasks are tracked by #1226 and #1227.

test/fomac/test_fomac.cpp
burgholzer, line 1: Suppose this was just an oversight that this file has some clang-tidy errors now.

burgholzer (approved):
Besides the clang-tidy warnings and the docs build failure, this looks great to me!

Changelog entries and upgrade guide notes are still in order here.
This changes where the generic FoMaC lives and adds a few new things.

PR #1232, opened 2025-09-28
✨ MLIR | Add `PlacementPass`

MatthiasReumann:
## Description

This pull request splits the routing pass into a placement and routing pass.

### Changes 

The new placement pass replaces `alloc`s and `dealloc`s with static qubits, generates and applies the initial placement, and handles the extension of results and arguments of `scf::For` and `scf::If`. 

As a consequence the routing pass becomes much more simpler. Since the placement pass doesn't have to worry about SWAPs, the `HardwareIndexPool` is replaced by a simple `std::deque<std::size_t>`. Following the same argument, the awkward `checkZeroUse` check can be avoided entirely. 

To reduce code duplication, I've also moved the `Layout` (previously `LayoutState`) to a separate `Layout.h` file and use it throughout all transpilation passes. Likewise for constants and the `isEntryPoint` helper function there is now a `Common.h` file.

### Initial Placement (Help appreciated 🙌🏼)

I'm still very much unsatisfied with the initial placement solution. Even though the interface is quite clear (_"return program <-> hardware mapping"_) the current solution still feels too disintegrated to allow LightSABRE-like logic. Maybe it's possible to give the `InitialPlacer` a copy of the `func.func` (or something along the lines of that). 

Ultimately, in the near future, we probably have to implement an initial placement strategy that requires the underlying computation (such as the subgraph isomorphism problem) to address these issues properly. 

---

Moreover:  How do we want to specify, or record, the initial placement vector for equivalence checking down the line? 

## Checklist:

### DIFF
--- a/mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/PlacementPass.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/PlacementPass.cpp
@@ -39,7 +39,6 @@
 #include <mlir/Support/WalkResult.h>
 #include <numeric>
 #include <random>
-#include <vector>
 
 #define DEBUG_TYPE "placement-sc"
 
@@ -437,7 +436,7 @@ struct PlacementPassSC final : impl::PlacementPassSCBase<PlacementPassSC> {
 private:
   [[nodiscard]] std::unique_ptr<InitialPlacer>
   getPlacer(const Architecture& arch) const {
-    switch (strategy) {
+    switch (static_cast<PlacementStrategy>(strategy)) {
     case PlacementStrategy::Identity:
       LLVM_DEBUG({ llvm::dbgs() << "getPlacer: identity placement\n"; });
       return std::make_unique<IdentityPlacer>(arch.nqubits());


mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Layout.h
```diff
@@ -0,0 +1,144 @@
...
+private:
+  struct QubitInfo {
+    QubitIndex hardwareIdx;
+    QubitIndex programIdx;
+  };
+
+  /**
+   * @brief Maps an SSA value to its `QubitInfo`.
+   */
+  DenseMap<Value, QubitInfo> valueToMapping_;
```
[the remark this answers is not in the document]
MatthiasReumann: Provided by `#include <mlir/Support/LLVM.h>`

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Stack.h
```diff
@@ -18,7 +20,7 @@ namespace mqt::ir::opt {
  * @brief Manages the routing state stack with clear semantics for accessing
  * current and parent states.
  */
-template <class Item> class [[nodiscard]] RoutingStack {
+template <class Item> class [[nodiscard]] TranspilationStack {
```
burgholzer: Maybe `LayoutStack`?
The docstring probably also needs an update.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Common.cpp
```diff
@@ -0,0 +1,31 @@
...
+bool isEntryPoint(mlir::func::FuncOp op) {
+  const auto passthroughAttr =
+      op->getAttrOfType<mlir::ArrayAttr>(PASSTHROUGH_ATTR);
+  if (!passthroughAttr) {
+    return false;
+  }
+
+  return llvm::any_of(passthroughAttr, [](const mlir::Attribute attr) {
+    return isa<mlir::StringAttr>(attr) &&
+           cast<mlir::StringAttr>(attr) == ENTRY_POINT_ATTR;
```
burgholzer, line 41: If the two attribute names aren't used elsewhere, they might as well only be defined in the CPP file here.

burgholzer (changes requested):
Overall, this looks great already! I like how much it simplifies some parts.

At the same time, I don't feel quite satisfied with the layout pass itself either. Right now, the strategy to apply is entirely fixed. Wouldn't this now be a good opportunity to introduce a parameter for choosing the strategy?

As for more complicated layout passes: I don't think there is a need to pass a copy of the `func.func` or similar. Each of the passes is free to traverse the entire program on its own into whatever depth it wants. So we should be able to walk the program in the layout stage similar to how we walk it in the routing stage.
Exemplary variants of this could be the traversal and collection of an interaction graph, which could then be used for the sub-graph isomorphism problem or other heuristics.
We should also be able to combine a layout and a routing pass in this way (hopefully).

As for storing the initial layout: I am really not quite sure how to best do this in MLIR. In OpenQASM (or Qiskit) one would simply store a mapping between register (qubits) before and register (qubits) after. In MQT Core we translate that to a mapping from program qubit indices to device qubit indices. I don't have a clear picture yet of how one would translate this to the concepts used in MLIR. Eventually, one would want to take a program before applying the layout and routing stage as well as one after and check them for equivalence. The main task is to (somehow) match the program qubits before the L+R to the device qubits after.
Side note: this will also become important when we adjust our Qiskit importer to directly translate to MLIR at some point.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Common.cpp
MatthiasReumann: Done.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Stack.h
MatthiasReumann: Done.

MatthiasReumann:
Thanks for the initial review @burgholzer! 🙏🏻

> Wouldn't this now be a good opportunity to introduce a parameter for choosing the strategy?

Absolutely! I was still in the mindset of the routing pass. That is, add options when we specify devices with QDMI. However, it makes totally sense to add a `strategy` pass option here. Done!

>  I don't think there is a need to pass a copy of the func.func or similar. Each of the passes is free to traverse the entire program on its own into whatever depth it wants. 

The reason why I thought about copying `func.func` was trying multiple random initial layouts at once (what LightSABRE does as far as I know). LLVM provides [multithreading utilities](https://llvm.org/doxygen/Parallel_8h.html) so I think we should be able to do this on a single pass level. But now I realize that these threads would probably be read-only in the layout stage - so copying isn't really necessary. It is only the final pass that requires the `PatternRewriter` and eventually rewrites the IR. 

> Eventually, one would want to take a program before applying the layout and routing stage as well as one after and check them for equivalence. The main task is to (somehow) match the program qubits before the L+R to the device qubits after.

A straight-forward solution would be to add the hardware index retrieved when replacing `alloc`s to an attribute for each `func.func`. The first `nqubits` indices in this array would exactly be the initial placement (`program <-> hardware)`[^1]. In our implementation, the indices after that depend on the `std::queue`. This way, given a `before.mlir` and an `after.mlir` with the before mentioned attribute set, it should be reproducible how the `alloc`'s have been mapped to hardware qubits. 

What's your take on this? Keep in mind that I'm not all too familiar with equivalence checking (as of now). 

[^1]: Note that we can't just add the initial layout as an attribute because we allow for arbitrary sequences of `alloc`s and `dealloc`s. This is a consequence of allowing multiple "circuits" inside a single function.

burgholzer:
> Thanks for the initial review @burgholzer! 🙏🏻
> 
> > Wouldn't this now be a good opportunity to introduce a parameter for choosing the strategy?
> 
> Absolutely! I was still in the mindset of the routing pass. That is, add options when we specify devices with QDMI. However, it makes totally sense to add a `strategy` pass option here. Done!

Nice. this LGTM now. I just pushed a couple of code cleanups.
Let's see what clang-tidy has to say about the switch statement now.
I also added a changelog entry for this.

> > I don't think there is a need to pass a copy of the func.func or similar. Each of the passes is free to traverse the entire program on its own into whatever depth it wants.
> 
> The reason why I thought about copying `func.func` was trying multiple random initial layouts at once (what LightSABRE does as far as I know). LLVM provides [multithreading utilities](https://llvm.org/doxygen/Parallel_8h.html) so I think we should be able to do this on a single pass level. But now I realize that these threads would probably be read-only in the layout stage - so copying isn't really necessary. It is only the final pass that requires the `PatternRewriter` and eventually rewrites the IR.

Yeah. That's how I would imagine it to work. IIRC, MLIR has a kind of transaction-like commit system where you can start a rewrite and it only gets committed once a dedicated function is called. So it might also be feasible to do "actual" routing in the layout stage (similar how SABRE does this). This is to be explored in a separate PR though.

> > Eventually, one would want to take a program before applying the layout and routing stage as well as one after and check them for equivalence. The main task is to (somehow) match the program qubits before the L+R to the device qubits after.
> 
> A straight-forward solution would be to add the hardware index retrieved when replacing `alloc`s to an attribute for each `func.func`. The first `nqubits` indices in this array would exactly be the initial placement (`program <-> hardware)`[1](#user-content-fn-1-e4d576304ae160695d2313fffe570421). In our implementation, the indices after that depend on the `std::queue`. This way, given a `before.mlir` and an `after.mlir` with the before mentioned attribute set, it should be reproducible how the `alloc`'s have been mapped to hardware qubits.
> 
> What's your take on this? Keep in mind that I'm not all too familiar with equivalence checking (as of now).
> 
> ## Footnotes
> 1. Note that we can't just add the initial layout as an attribute because we allow for arbitrary sequences of `alloc`s and `dealloc`s. This is a consequence of allowing multiple "circuits" inside a single function. [↩](#user-content-fnref-1-e4d576304ae160695d2313fffe570421)

This sounds reasonable in principle 👍🏼 
But I believe this can be explored in a side project at some point. Maybe when we try to port equivalence checking to this framework at some point. I believe I don't want to hold this PR up any longer than it needs to be.
So, if you agree, let's get this merged and add any extras in a follow-up PR.

PR #1299, opened 2025-11-10
:recycle: Preserve Tuple Structure and Improve Site Type Clarity of the MQT NA Default QDMI Device

marcelwa:
### Problem

The MQT NA Default QDMI Device implementation has some confusing design decisions:

**1. Tuple structure for operation sites**

The QDMI spec says operation sites should be "a list of tuples" representing valid site combinations. Instead, we were using `reinterpret_cast` to flatten `vector<pair<Site, Site>>` into `vector<Site>`. This relied on memory layout assumptions and forced downstream code to manually reconstruct pairs with stride-2 iteration.

**2. Zones and qubits mixed together**

"Sites" meant both zones (operational regions like indices 0-2) and qubits (actual qubit locations like indices 3-102). Users had to constantly check `site.is_zone()` and manually filter everywhere.

### Solution

**Fixed the tuple structure:** Replaced the flat vector with `std::variant<vector<Site>, vector<pair<Site, Site>>>` and removed the `reinterpret_cast` hack. Added `getSitePairs()` method so you can naturally iterate over pairs. The type system now enforces the structure.

**Added semantic filtering:** Added `getRegularSites()` and `getZones()` methods that filter by type, so you don't need manual filtering. The QDMI C API is unchanged. The filtering happens at the higher-level FoMaC layer.

### Example

**Before:**
```python
# Manual pair reconstruction
sites = operation.sites()
for i in range(0, len(sites), 2):
    site1, site2 = sites[i], sites[i+1]

# Manual filtering
qubits = [s for s in device.sites() if not s.is_zone()]
```

**After:**
```python
# Natural iteration
for site1, site2 in operation.site_pairs():
    # work with pairs

# Clear filtering
qubits = device.regular_sites()
zones = device.zones()
```

## Checklist:


src/na/device/Device.cpp
```diff
@@ -385,38 +384,35 @@ MQT_NA_QDMI_Operation_impl_d::MQT_NA_QDMI_Operation_impl_d(
...
+          std::ranges::sort(sites);
+        } else if constexpr (std::is_same_v<
+                                 T, std::vector<std::pair<MQT_NA_QDMI_Site,
+                                                          MQT_NA_QDMI_Site>>>) {
+          // Two-qubit: ensure each pair is ordered, then sort pairs
+          std::ranges::for_each(sites, [](auto& p) {
+            if (p.first > p.second) {
+              std::swap(p.first, p.second);
+            }
+          });
```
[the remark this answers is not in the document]
marcelwa: Done.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -430,6 +430,21 @@ class FoMaC {
...
           -> std::optional<std::vector<Site>>;
+      /**
+       * @brief Get valid site pairs for two-qubit operations.
+       * @details Returns the site combinations as pairs, preserving the tuple
+       * structure as specified by QDMI. This method should be used for local
+       * two-qubit operations to get the valid site combinations.
+       * @param sites Optional sites for query.
+       * @param params Optional parameters for query.
+       * @return Optional vector of site pairs if this is a 2-qubit operation,
+       * std::nullopt otherwise.
```
ystade: Slight concretisations:
```suggestion
       * @details For local 2-qubit operations, this function interprets the
       * returned list of sites by QDMI as site pairs according to the QDMI
       * specification. Hence, this function facilitates easier iteration over
       * supported site pairs.
       * @param sites is an optional vector of sites for the query.
       * @param params is an optional vector of parameters for the query.
       * @return Optional vector of site pairs if this is a local 2-qubit operation,
       * std::nullopt otherwise.
```

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -512,6 +527,23 @@ class FoMaC {
     [[nodiscard]] auto getQubitsNum() const -> size_t;
     /// @see QDMI_DEVICE_PROPERTY_SITES
     [[nodiscard]] auto getSites() const -> std::vector<Site>;
+    /**
+     * @brief Get only qubit sites (non-zone sites).
+     * @details Returns sites where isZone() is false. These represent actual
+     * physical qubit locations on the device lattice.
+     * @return Vector of qubit sites (filtered from all sites)
```
ystade:
```suggestion
     * @details Filters all sites and only returns regular sites, i.e., where `isZone()` yields
     * `false`. These represent actual potential physical qubit locations on the device lattice.
     * @returns vector of regular sites
```

include/mqt-core/fomac/FoMaC.hpp
ystade: Slight concretisations:
```suggestion
       * @details For local 2-qubit operations, this function interprets the
       * returned list of sites by QDMI as site pairs according to the QDMI
       * specification. Hence, this function facilitates easier iteration over
       * supported site pairs.
       * @param sites is an optional vector of sites for the query.
       * @param params is an optional vector of parameters for the query.
       * @return Optional vector of site pairs if this is a local 2-qubit operation,
       * std::nullopt otherwise.
```

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -512,6 +527,23 @@ class FoMaC {
...
+     * @see QDMI_DEVICE_PROPERTY_SITES
+     */
+    [[nodiscard]] auto getQubits() const -> std::vector<Site>;
+    /**
+     * @brief Get only zone sites.
+     * @details Returns sites where isZone() is true. These represent
+     * operational regions where global operations are performed, not individual
+     * qubit locations.
+     * @return Vector of zone sites (filtered from all sites)
```
ystade:
```suggestion
     * @details Filters all sites and only returns zone sites, i.e., where `isZone()` yields `true`. 
     * These represent a zone, i.e., an extent where zoned operations can be performed, not individual
     * qubit locations.
     * @returns a vector of zone sites
```

include/mqt-core/na/device/Device.hpp
```diff
@@ -317,8 +318,23 @@ struct MQT_NA_QDMI_Operation_impl_d {
...
+  /**
+   * @brief Storage for supported sites with different structures.
+   * @details Uses std::variant to preserve the tuple structure of operation
+   * sites:
+   * - Single-qubit operations: vector<Site>
+   * - Two-qubit operations: vector<pair<Site, Site>>
+   * This maintains type safety and QDMI specification compliance which states
+   * that operation sites should be "a list of tuples" for multi-qubit
+   * operations.
+   */
```
ystade, line 331:
```suggestion
  /**
   * @brief Storage for individual sites and site pairs.
   * @details Uses std::variant to preserve the tuple structure of the operation
   * sites:
   * - Single-qubit and zoned operations: vector<Site>
   * - Local two-qubit operations: vector<pair<Site, Site>>
   * This maintains type safety and QDMI specification compliance, which states
   * that operation sites should be "a list of tuples" for local multi-qubit
   * operations.
   */
```

python/mqt/core/fomac.pyi
```diff
@@ -81,6 +81,10 @@ class Device:
             self, sites: Iterable[Device.Site] = ..., params: Iterable[float] = ...
         ) -> Iterable[Device.Site] | None:
             """Returns the list of sites the operation can be performed on."""
+        def site_pairs(
+            self, sites: Iterable[Device.Site] = ..., params: Iterable[float] = ...
+        ) -> Iterable[tuple[Device.Site, Device.Site]] | None:
+            """Returns the list of site pairs the 2-qubit operation can be performed on."""
```
ystade:
```suggestion
            """Returns the list of site pairs the local 2-qubit operation can be performed on."""
```

python/mqt/core/fomac.pyi
```diff
@@ -99,7 +103,11 @@ class Device:
     def qubits_num(self) -> int:
         """Returns the number of qubits available on the device."""
     def sites(self) -> Iterable[Site]:
-        """Returns the list of sites available on the device."""
+        """Returns the list of all sites (zones and qubits) available on the device."""
```
ystade:
```suggestion
        """Returns the list of all sites (zone and regular sites) available on the device."""
```

python/mqt/core/fomac.pyi
```diff
@@ -99,7 +103,11 @@ class Device:
...
+    def qubits(self) -> Iterable[Site]:
+        """Returns the list of qubit sites (non-zone sites) available on the device."""
```
ystade:
```suggestion
        """Returns the list of regular sites (without all zone sites) available on the device."""
```

python/mqt/core/fomac.pyi
```diff
@@ -99,7 +103,11 @@ class Device:
...
+    def zones(self) -> Iterable[Site]:
+        """Returns the list of zone sites available on the device."""
```
ystade:
```suggestion
        """Returns the list of zone sites (without regular sites) available on the device."""
```

src/fomac/FoMaC.cpp
```diff
@@ -319,6 +319,35 @@ auto FoMaC::Device::Operation::getSites(const std::vector<Site>& sites,
...
+  const auto& qdmiSites = queryProperty<std::optional<std::vector<QDMI_Site>>>(
+      QDMI_OPERATION_PROPERTY_SITES, sites, params);
+
+  if (!qdmiSites.has_value()) {
+    return std::nullopt;
+  }
+
+  if (qdmiSites->size() % 2 != 0) {
+    return std::nullopt; // Invalid: odd number of sites
+  }
```
ystade: The semantics is probably even as wanted, but I stumbled over the comment: For global/zoned 2-qubit gates this is actually valid, i.e., the number of zones the operation is supported on may be even.

src/fomac/FoMaC.cpp
```diff
@@ -351,6 +380,24 @@ auto FoMaC::Device::getSites() const -> std::vector<Site> {
                          });
   return sites;
 }
+auto FoMaC::Device::getQubits() const -> std::vector<Site> {
+  const auto& allSites = getSites();
+  std::vector<Site> qubits;
+  qubits.reserve(allSites.size());
+  std::ranges::copy_if(allSites, std::back_inserter(qubits), [](const Site& s) {
+    return !s.isZone().value_or(false);
+  });
```
ystade: Lukas would probably prefer a for loop, but I love the use of `::ranges::`. ❤️

src/na/fomac/Device.cpp
```diff
@@ -116,10 +116,11 @@ auto FoMaC::Device::initDurationUnitFromDevice() -> bool {
   return true;
 }
 auto FoMaC::Device::initDecoherenceTimesFromDevice() -> bool {
-  // find first non-zone site as reference
-  auto regularSites = getSites() | std::views::filter([](const auto& site) {
-                        return !site.isZone().value_or(false);
-                      });
+  const auto regularSites = getQubits();
+  if (regularSites.empty()) {
+    SPDLOG_INFO("Device has no regular qubit sites with decoherence data");
```
ystade:
```suggestion
    SPDLOG_INFO("Device has no regular sites with decoherence data");
```

src/na/fomac/Device.cpp
```diff
@@ -116,10 +116,11 @@ auto FoMaC::Device::initDurationUnitFromDevice() -> bool {
   return true;
 }
 auto FoMaC::Device::initDecoherenceTimesFromDevice() -> bool {
-  // find first non-zone site as reference
-  auto regularSites = getSites() | std::views::filter([](const auto& site) {
-                        return !site.isZone().value_or(false);
-                      });
+  const auto regularSites = getQubits();
```
ystade:
```suggestion
  const auto regularSites = getRegularSites();
```
For consistency witth the notions used in the QDMI spec (as far as I recall) this method should probably be renamed. I must have overlooked this above.

ystade (dismissed):
Hi @marcelwa , this PR looks very clean to me. Thanks for the changes. I have only two minor comments
1. I would like to fine tune some docstrings as you see inline. Probably it suffices to just accept the suggestions.
2. I am a bit concerend whether global/zoned 2-qubit operations are handled correctly. Even though they should be covered by tests, I am a bit unsure as you can see in my comments.

src/fomac/FoMaC.cpp
burgholzer: Ranges are fine, but I still think this is way harder to read than a "traditional" range-based for loop. ;-)

marcelwa: Agree to disagree 😂 STL algorithms and ranges provide meaning and intent to the code. Not even to speak of the possibility of execution policies.

marcelwa: Agree to disagree 😂 STL algorithms and ranges provide meaning and intent to the code. Not even to speak of the possibility of execution policies.

burgholzer: Fair enough. Maybe it's just me being old school here or hating the syntax and look and feel of ranges.. I'll get used to it eventually.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 - ✨ Add support for bridge gates for the neutral atom hybrid mapper ([#1293]) ([**@lsschmid**])
 
+### Changed
+
+- ♻️ Preserve Tuple Structure and Improve Site Type Clarity of the MQT NA Default QDMI Device ([#1299]) ([**@marcelwa**])
```
```suggestion
- ♻️ Preserve tuple structure and improve site type clarity of the MQT NA Default QDMI Device ([#1299]) ([**@marcelwa**])
```

src/fomac/FoMaC.cpp
marcelwa: Hmm, that's true, I suppose, but then this function shouldn't be called, right? You could use `.getSites()` instead. My understanding is that `.getSitePairs()` is intended for the cases where the operation is not defined on zones but qubits. Am I mistaken?

src/na/fomac/Device.cpp
marcelwa: Done.

src/fomac/FoMaC.cpp
ystade: Yeah, this makes sense. Then, for better semantics of the function, I would suggest to add the check whether the respective operation is zoned to line 326 and also report `std::nullopt` in the case it is.

marcelwa: Done.

bindings/fomac/fomac.cpp
```diff
@@ -95,7 +95,12 @@ PYBIND11_MODULE(MQT_CORE_MODULE_NAME, m, py::mod_gil_not_used()) {
                 "params"_a = std::vector<double>{});
   operation.def("sites", &fomac::FoMaC::Device::Operation::getSites,
                 "sites"_a = std::vector<fomac::FoMaC::Device::Site>{},
-                "params"_a = std::vector<double>{});
+                "params"_a = std::vector<double>{},
+                "Get operation sites as a flat list");
+  operation.def("site_pairs", &fomac::FoMaC::Device::Operation::getSitePairs,
+                "sites"_a = std::vector<fomac::FoMaC::Device::Site>{},
+                "params"_a = std::vector<double>{},
+                "Get operation site pairs for 2-qubit operations");
```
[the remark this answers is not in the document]
burgholzer: @marcelwa this seems like a reasonable suggestion

include/mqt-core/na/device/Device.hpp
```diff
@@ -317,8 +318,23 @@ struct MQT_NA_QDMI_Operation_impl_d {
...
+   * sites:
+   * - Single-qubit and zoned operations: vector<Site>
+   * - Local two-qubit operations: vector<pair<Site, Site>>
+   * This maintains type safety and QDMI specification compliance, which states
+   * that operation sites should be "a list of tuples" for local multi-qubit
+   * operations.
+   */
+  using SitesStorage =
+      std::variant<std::vector<MQT_NA_QDMI_Site>,
+                   std::vector<std::pair<MQT_NA_QDMI_Site, MQT_NA_QDMI_Site>>>;
```
burgholzer, line 334: This is totally fine for now and for the currently supported features. However, as the docstring already states, the return type for local operations is generally a _tuple_ and _not a pair_. There have been demonstrations of NA platforms with local three-qubit CCZ operations.
I am happy to postpone this discussion to when it becomes necessary. I just wanted to make sure that we are all aware of this. Probably more geared towards @ystade than @marcelwa

python/mqt/core/fomac.pyi
```diff
@@ -99,7 +103,11 @@ class Device:
     def qubits_num(self) -> int:
         """Returns the number of qubits available on the device."""
     def sites(self) -> Iterable[Site]:
-        """Returns the list of sites available on the device."""
+        """Returns the list of all sites (zone and regular sites) available on the device."""
+    def regular_sites(self) -> Iterable[Site]:
+        """Returns the list of regular sites (without zone sites) available on the device."""
+    def zones(self) -> Iterable[Site]:
+        """Returns the list of zone sites (without regular sites) available on the device."""
```
burgholzer: Real nitpick, but the wording in these docstrings is slightly different than the wording in the other related docstrings. The difference is subtle, but I think I prefer the wording here.

src/fomac/FoMaC.cpp
```diff
@@ -319,6 +319,36 @@ auto FoMaC::Device::Operation::getSites(const std::vector<Site>& sites,
                          });
   return returnedSites;
 }
+auto FoMaC::Device::Operation::getSitePairs(
+    const std::vector<Site>& sites, const std::vector<double>& params) const
+    -> std::optional<std::vector<std::pair<Site, Site>>> {
+  const auto qubitsNum = getQubitsNum(sites, params);
+  if (!qubitsNum.has_value() || *qubitsNum != 2 ||
+      isZoned(sites, params).value_or(false)) {
```
burgholzer: This might be just me, but it puzzles me a little bit that the result of `isZoned` is an optional here. In my mind, an operation is zoned or it is not; that's a very binary thing. I don't particularly see the value of having a third state (`std::nullopt`) that effectively carries the same meaning as `false`.
This might hold for some more of the functions here.

src/fomac/FoMaC.cpp
```diff
@@ -319,6 +319,36 @@ auto FoMaC::Device::Operation::getSites(const std::vector<Site>& sites,
...
+
+  std::vector<std::pair<Site, Site>> pairs;
+  pairs.reserve(qdmiSites->size() / 2);
+
+  for (size_t i = 0; i < qdmiSites->size(); i += 2) {
+    pairs.emplace_back(Site{Token{}, device_, (*qdmiSites)[i]},
+                       Site{Token{}, device_, (*qdmiSites)[i + 1]});
+  }
+
+  return pairs;
```
burgholzer, line 345: Is there a way for us to avoid the amount of copying here?
Ideally, we would just reuse the same memory and just reinterpret it differently.
`std::pair` is a zero-cost abstraction after all, so `std::vector<Site>` and `std::vector<std::pair<Site, Site>>` should be "memory-compatible".

src/fomac/FoMaC.cpp
```diff
@@ -351,6 +381,24 @@ auto FoMaC::Device::getSites() const -> std::vector<Site> {
...
+  regularSites.reserve(allSites.size());
+  std::ranges::copy_if(
+      allSites, std::back_inserter(regularSites),
+      [](const Site& s) { return !s.isZone().value_or(false); });
+  return regularSites;
+}
+auto FoMaC::Device::getZones() const -> std::vector<Site> {
+  const auto& allSites = getSites();
+  std::vector<Site> zones;
+  zones.reserve(allSites.size());
```
burgholzer: I think this reserve does not make too much sense here.
In general, the order of magnitude for the number of zones might be way smaller than the number of sites.
There could be 6000 atoms and those would probably still be clustered into 3 zones.

src/fomac/FoMaC.cpp
```diff
@@ -351,6 +381,24 @@ auto FoMaC::Device::getSites() const -> std::vector<Site> {
                          });
   return sites;
 }
+auto FoMaC::Device::getRegularSites() const -> std::vector<Site> {
+  const auto& allSites = getSites();
```
burgholzer: This might just be me being pedantic, but we are doing a lot of copying here, which might not be particularly desirable as devices grow larger.
Would it be better to define an optional filter function for `getSites`?
Essentially, I am just looking for ways to avoid unnecessary copies here. Could also be accomplished by moving instead of copying. the underlying full vector is being thrown away after the function call.

src/na/device/Device.cpp
```diff
@@ -521,8 +535,44 @@ auto MQT_NA_QDMI_Operation_impl_d::queryProperty(
...
+                return QDMI_ERROR_INVALIDARGUMENT;
+              }
+              // Copy pairs as flat array for C API
+              // NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+              auto* siteArray = static_cast<MQT_NA_QDMI_Site*>(value);
+              size_t idx = 0;
+              for (const auto& [site1, site2] : storedSites) {
+                siteArray[idx++] = site1;
+                siteArray[idx++] = site2;
+              }
```
burgholzer: Related to another comment: I am pretty sure we could simply reuse the list macro here because the underlying `.data()` memory layout is the same between the list of sites and the list of pairs of sites.

src/na/fomac/Device.cpp
```diff
@@ -509,6 +499,20 @@ auto FoMaC::Device::initOperationsFromDevice() -> bool {
...
+          return false;
+        }
+
+        std::vector<fomac::FoMaC::Device::Site> allSites;
+        allSites.reserve(sitePairsOpt->size() * 2);
+        for (const auto& [site1, site2] : *sitePairsOpt) {
+          allSites.push_back(site1);
+          allSites.push_back(site2);
+        }
+        const auto pairRegion = calculateExtentFromSites(allSites);
```
burgholzer: This seems like another bunch of copies that could be avoided if `calculateExtentFromSites` was given another signature.

test/python/na/test_na_fomac.py
```diff
@@ -197,12 +197,53 @@ def test_operations(device_tuple: tuple[Device, Mapping[str, Any]]) -> None:
...
                 sites = operation.sites()
                 assert sites is not None
                 sites = list(sites)
+                assert len(sites) == len(site_pairs_list) * 2
                 assert calculate_extent_from_sites(sites) == (
                     device_dict["localMultiQubitOperations"][0]["region"]["origin"]["x"],
                     device_dict["localMultiQubitOperations"][0]["region"]["origin"]["y"],
                     device_dict["localMultiQubitOperations"][0]["region"]["size"]["width"],
                     device_dict["localMultiQubitOperations"][0]["region"]["size"]["height"],
                 )
```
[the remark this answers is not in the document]
burgholzer: @marcelwa this also seems like a reasonable suggestion to me

python/mqt/core/fomac.pyi
marcelwa: Which "related docstrings" do you mean here exactly? In the C++ bindings code? I.e.,
```c++
  device.def("sites", &fomac::FoMaC::Device::getSites,
             "Get all sites (zones and qubits)");
  device.def("regular_sites", &fomac::FoMaC::Device::getRegularSites,
             "Get only qubit sites (non-zone sites)");
  device.def("zones", &fomac::FoMaC::Device::getZones, "Get only zone sites");
  ```

Comment (burgholzer) [reply python/mqt/core/fomac.pyi]: Yeah. Basically all occurrences where this triple (sites, regular sites, and zones) appears throughout the PR.

File: src/fomac/FoMaC.cpp (hunk 18 above)
Comment (marcelwa) [reply src/fomac/FoMaC.cpp]: I understand your concern. However, that function is not new. It existed before:
```c++
auto FoMaC::Device::Operation::isZoned(const std::vector<Site>& sites,
                                       const std::vector<double>& params) const
    -> std::optional<bool> {
  return queryProperty<std::optional<bool>>(QDMI_OPERATION_PROPERTY_ISZONED,
                                            sites, params);
}
```
I'll adapt it.

burgholzer: Yeah, I am aware. I might not have reviewed the initial PR adding this functionality in as much detail as possible.

include/mqt-core/na/device/Device.hpp
ystade: I would address this issue when it becomes relevant. These demonstrations were very experimental so far. And then, I'd also would suggest using an array not a tuple.

src/fomac/FoMaC.cpp
marcelwa: Assuming that `std::vector` starts with a capacity of 0 and grows exponentially, a `.reserve()` with even just 3 zones would reduce the need for re-allocation by 1 or 2 (depending on implementation). I'd keep it :man_shrugging:

src/fomac/FoMaC.cpp
marcelwa: I tried to make it a bit neater by deleting from `allSites` instead of copying the desired items over.

bindings/fomac/fomac.cpp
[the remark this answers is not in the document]
marcelwa: Done.

test/python/na/test_na_fomac.py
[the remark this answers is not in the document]
marcelwa: Done.

python/mqt/core/fomac.pyi
marcelwa: Done.

src/fomac/FoMaC.cpp
marcelwa: Understood. Done.

src/fomac/FoMaC.cpp
marcelwa: Should be better now.

src/na/device/Device.cpp
marcelwa: Needed a small workaround using `std::span`, but it should be a bit nicer now, nevertheless.

src/na/fomac/Device.cpp
marcelwa: I added an overload.

src/fomac/FoMaC.cpp
burgholzer: I am not quite sure I follow that logic.
Assuming there are 100 sites and only two zones (as in the default example), the current strategy will allocate space for 100 sites, while the default allocation strategy will only allocate a maximum of 3 (first one, then two).
It's very hard to imagine a scenario where the current strategy would yield less allocations given how disproportional the number of sites and zones are.

marcelwa: I think I understand your point now. In that case, I agree with you. Either way, the code is now gone due to refactoring related to your other comments.

src/na/device/Device.cpp
```diff
@@ -521,8 +536,34 @@ auto MQT_NA_QDMI_Operation_impl_d::queryProperty(
...
+                reinterpret_cast<const MQT_NA_QDMI_Site*>(storedSites.data()),
+                storedSites.size() * 2);
+            ADD_LIST_PROPERTY(QDMI_OPERATION_PROPERTY_SITES, MQT_NA_QDMI_Site,
+                              flatView, prop, size, value, sizeRet)
+          }
+          // more cases go here if needed in the future
+          return QDMI_ERROR_NOTSUPPORTED;
+        },
+        supportedSites_);
+  }
```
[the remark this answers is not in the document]
marcelwa: Done.

src/fomac/FoMaC.cpp
```diff
@@ -319,6 +321,33 @@
...
+
+  std::vector<std::pair<Site, Site>> pairs;
+  pairs.reserve(sitesVec.size() / 2);
+
+  for (size_t i = 0; i < sitesVec.size(); i += 2) {
+    pairs.emplace_back(std::move(sitesVec[i]), std::move(sitesVec[i + 1]));
+  }
+
+  return pairs;
+}
```
marcelwa: Done.

src/fomac/FoMaC.cpp
```diff
@@ -319,6 +321,33 @@ auto FoMaC::Device::Operation::getSites(const std::vector<Site>& sites,
...
+
+  std::vector<std::pair<Site, Site>> pairs;
+  pairs.reserve(sitesVec.size() / 2);
+
+  for (size_t i = 0; i < sitesVec.size(); i += 2) {
+    pairs.emplace_back(sitesVec[i], sitesVec[i + 1]);
+  }
+
+  return pairs;
+}
```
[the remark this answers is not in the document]
marcelwa: Done.

src/fomac/FoMaC.cpp
```diff
@@ -351,6 +380,20 @@ auto FoMaC::Device::getSites() const -> std::vector<Site> {
...
+  allSites.erase(newEnd.begin(), allSites.end());
+  return allSites;
+}
+auto FoMaC::Device::getZones() const -> std::vector<Site> {
+  auto allSites = getSites();
+  const auto newEnd = std::ranges::remove_if(
+      allSites, [](const Site& s) { return !s.isZone(); });
+  allSites.erase(newEnd.begin(), allSites.end());
+  return allSites;
+}
```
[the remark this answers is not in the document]
marcelwa: Done.

src/na/device/Device.cpp
```diff
@@ -385,38 +386,37 @@ MQT_NA_QDMI_Operation_impl_d::MQT_NA_QDMI_Operation_impl_d(
...
+            if (std::less<>{}(p.second, p.first)) {
+              std::swap(p.first, p.second);
+            }
+          });
+          std::ranges::sort(sites);
+        }
+        // more cases go here if needed in the future
+      },
+      supportedSites_);
 }
```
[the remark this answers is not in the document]
marcelwa: Done.

PR #1301, opened 2025-11-11
✨Implement Unitized Routing Passes

MatthiasReumann:
## Description

This pull request implements an "unit"-ized approach to route quantum-classical programs.  Furthermore, it solves the issue that the def-use and textual (program) operation order is not _that_ compatible. 

### Unit

An Unit divides a quantum-classical program into routable sections:

```
┌─────┐
│unit |
└─────┘
---------------- divider (`scf.for`)
┃ ┌─────┐
┃ |unit | // loop-body
┃ └─────┘ <-- `restore=true` 
┌─────┐
│unit |
└─────┘
---------------- divider (`scf.if`)
┃ ┌─────┐
┃ |unit | // then-body
┃ └─────┘<-- `restore=true` 
┣
┃ ┌─────┐
┃ │unit | // else-body
┃ └─────┘<-- `restore=true` 
┌─────┐
│unit |
└─────┘
```

Each unit is routed separately. This PR introduces two modes of traversal for such units:
- `SequentialUnit`: Traverses the ops in the unit sequentially (Naive Routing)
- `LayeredUnit`: Traverses the ops layer-by-layer (A* Routing) 

## Changes
- Split naive (`--route-naive-sc`) from A* routing pass (`--route-astar-sc`): This makes sense from a logical as well as a pass option perspective. The pass options for the A* routing pass belong to the A* routing pass. If we were to implement another method, let's say SABRE, its options would again reside in a separate pass.  
- Implement `SequentialUnit` and `LayeredUnit`.
- Refactor A* and Naive routing passes to use the respective unit implementation.
- Fix SSA Dominance Issues by implementing the reordering strategy discussed below.
- Use `SequentialUnit` also for verification 

## Checklist:

[^1]: Switching the custom stack for recursive function calls can probably also be applied to the placement and verification pass. Something for a future PR.

### DIFF
--- a/mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Architecture.h
+++ b/mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Architecture.h
@@ -64,13 +64,6 @@ public:
   }
 
   /**
-   * @brief Collect the shortest path between @p u and @p v.
-   * @returns The path from the destination (v) to source (u) qubit.
-   */
-  [[nodiscard]] llvm::SmallVector<std::size_t>
-  shortestPathBetween(uint32_t u, uint32_t v) const;
-
-  /**
    * @brief Collect the shortest SWAP sequence to make @p u and @p v adjacent.
    * @returns The SWAP sequence from the destination (v) to source (u) qubit.
    */
--- a/mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Architecture.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Architecture.cpp
@@ -10,9 +10,11 @@
 
 #include "mlir/Dialect/MQTOpt/Transforms/Transpilation/Architecture.h"
 
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h"
 #include "mlir/Dialect/MQTOpt/Transforms/Transpilation/Common.h"
 #include "mlir/Dialect/MQTOpt/Transforms/Transpilation/Layout.h"
 
+#include <cassert>
 #include <cstddef>
 #include <cstdint>
 #include <llvm/ADT/SmallVector.h>
@@ -23,27 +25,6 @@
 #include <utility>
 
 namespace mqt::ir::opt {
-llvm::SmallVector<std::size_t>
-Architecture::shortestPathBetween(uint32_t u, uint32_t v) const {
-  if (u == v) {
-    return {};
-  }
-
-  if (prev_[u][v] == UINT64_MAX) {
-    throw std::domain_error("No path between qubits " + std::to_string(u) +
-                            " and " + std::to_string(v));
-  }
-
-  llvm::SmallVector<std::size_t> path;
-  uint32_t curr = v;
-  path.push_back(curr);
-  while (curr != u) {
-    curr = prev_[u][curr];
-    path.push_back(curr);
-  }
-
-  return path;
-}
 
 llvm::SmallVector<std::pair<uint32_t, uint32_t>>
 Architecture::shortestSWAPsBetween(uint32_t u, uint32_t v) const {


MatthiasReumann:
## The Problem

> [!NOTE]
> I will keep updating this comment while I'm working on the problem. 

The problem is the following. Given a circuit representation of a quantum program, there is no canonical way of representing these as an MLIR program. For example,

```MLIR
module {
  %q0_0 = mqtopt.qubit 0
  %q1_0 = mqtopt.qubit 1
  %q2_0 = mqtopt.qubit 2
  %q3_0 = mqtopt.qubit 3
  %q4_0 = mqtopt.qubit 4
  %q4_0 = mqtopt.qubit 5

  %q2_1, %q1_1 = mqtopt.x %q2_0 ctrl %q1_0
  %q5_1, %q4_1 = mqtopt.x %q5_0 ctrl %q4_0 
  %q1_2, %q0_1 = mqtopt.x %q1_1 ctrl %q0_0 
  %q3_1, %q2_2 = mqtopt.x %q3_0 ctrl %q2_1
}
```

and 

```MLIR
module {
  %q0_0 = mqtopt.qubit 0
  %q1_0 = mqtopt.qubit 1
  %q2_0 = mqtopt.qubit 2
  %q3_0 = mqtopt.qubit 3
  %q4_0 = mqtopt.qubit 4
  %q4_0 = mqtopt.qubit 5

  %q2_1, %q1_1 = mqtopt.x %q2_0 ctrl %q1_0
  %q1_2, %q0_1 = mqtopt.x %q1_1 ctrl %q0_0  
  %q3_1, %q2_2 = mqtopt.x %q3_0 ctrl %q2_1
  %q5_1, %q4_1 = mqtopt.x %q5_0 ctrl %q4_0 // moved below.
}
```

represent the same circuit:

```

                       
0: ───────■────
        ┌─┴─┐          
1: ──■──┤ X ├──
   ┌─┴─┐└───┘
2: ┤ X ├──■────
   └───┘┌─┴─┐
3: ─────┤ X ├──
        └───┘  
4: ──■───O─────
   ┌─┴─┐     
5: ┤ X ├─O──────
   └───┘     
```

From a circuit perspective it is pretty easy to see that the `CNOT`s  acting on `(1,2)` and `(4, 5)` can be executed in parallel. It is also pretty "easy" to find these gates by traversing the def-use chain in MLIR. The hard part is to decide where to insert the SWAPs in the textual order:

- **Circuit perspective**: Insert the SWAPs before the first `Operation*` in the current layer while traversing the def-use chain. Consequently the first operation must be search: (`op->isBeforeInBlock`). However, the output qubits of the `(4, 5)` gate (marked with `O`) above might be below the gates acting on `(0,1)` and `(2,3)` (which is the case in the second MLIR code snippet) leading to an SSA Dominance Error. Not that great.
- **Find and collect**: Collect all `Operation*` for a given layer. Iterate the IR top-down until all those operations have been found and then continue with the next layer. Unfortunately, this leads to the same SSA dominance problem discussed above. 
- **"Advanced" Scheduling** (Current implementation): Traverse the def-use chain, collect all quantum ops and the layers along the way. Iterate over these ops and reorder them based on the scheduled order.  This works well for quantum programs. However: Once you add classical components to it (such as `scf.for` and `scf.if`) synchronization must be added. For example, only schedule a `scf.for` whenever all its inputs have been seen in the def-use traversal. From my experience in the past  week, it is really hard to wrap one's head around the details of this. 

    It is also fragile when using with classical components. For example, it requires including the `--canonicalize` pass to provide a valid IR whenever the program contains `scf.for`s (Hoist `arith.constant` ops because of SSA Dominance). 

    Regarding runtime: Reordering is pretty fast. The operation chain in MLIR is implemented as a DLL, so this is `~O(1)`. Preliminary benchmarks show that the current implementation is as fast as the one currently in the `main` branch. 

All of this is pretty frustrating, given that the "find layers and route" idea is so simple. More so because finding these layers is relatively easy and really fast in MLIR, thanks to the def-use chain. The "Advanced Scheduling" approach works, for now, but I'm not sure if I like going forward with this. There _has_ to be a better, more elegant, way.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Common.h
```diff
@@ -80,4 +80,35 @@ using QubitIndexPair = std::pair<QubitIndex, QubitIndex>;
...
+ * @param newValue The new value to use.
+ * @param region The region in which to perform replacements.
+ * @param exceptOp Operation to exclude from replacements.
+ * @param rewriter The pattern rewriter.
+ */
+void replaceAllUsesInRegionAndChildrenExcept(mlir::Value oldValue,
+                                             mlir::Value newValue,
+                                             mlir::Region* region,
+                                             mlir::Operation* exceptOp,
+                                             mlir::PatternRewriter& rewriter);
```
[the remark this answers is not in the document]
MatthiasReumann: This should also be fixed by including `<mlir/IR/PatternMatch.h>`, right?

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/NaiveRoutingPass.cpp
```diff
@@ -0,0 +1,174 @@
...
+              .Case<ResetOp>([&](ResetOp op) { remap(op, unit.layout()); })
+              .Case<MeasureOp>([&](MeasureOp op) { remap(op, unit.layout()); })
+              .Case<scf::YieldOp>([&](scf::YieldOp op) {
+                if (unit.restore()) {
+                  rewriter.setInsertionPointAfter(op->getPrevNode());
+                  insertSWAPs(op.getLoc(),
+                              llvm::to_vector(llvm::reverse(history)),
+                              unit.layout(), rewriter);
+                }
+              });
```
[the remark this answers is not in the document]
MatthiasReumann: This can also be solved by using ` rewriter.setInsertionPoint(op)` instead. `rewriter.setInsertionPoint` sets the insertion point right before the given op.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Layout.h
```diff
@@ -256,13 +265,91 @@ class [[nodiscard]] Layout : public ThinLayout {
...
+ * @param op The 'scf.if' op.
+ * @param layout The current layout.
+ */
+inline void remapIfResults(mlir::scf::IfOp op, Layout& layout) {
+  const auto nqubits = layout.getNumQubits();
+  const auto results = op->getResults().take_front(nqubits);
+  for (const auto [in, out] : llvm::zip(layout.getHardwareQubits(), results)) {
+    layout.remapQubitValue(in, out);
+  }
+}
```
burgholzer: These all feel like they could be part of the `Layout` class itself as member functions. Wouldn't that make this nicer?

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/QubitIndex.h
burgholzer, line 1: This feels a bit overkill to have in its own header, don't you think?
Especially since we won't be changing this any time soon I suppose.
Going for 64bits wouldn't make sense and 16bits are not enough for far-term use cases.
I am not against adding this, but it feels slightly strange to me.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Router.h
```diff
@@ -13,63 +13,38 @@
...
+  route(QubitIndexPair gate, const ThinLayout& layout,
+        const Architecture& arch) {
+    mlir::SmallVector<QubitIndexPair, 64> swaps;
+    const auto hw0 = layout.getHardwareIndex(gate.first);
+    const auto hw1 = layout.getHardwareIndex(gate.second);
+    const auto path = arch.shortestPathBetween(hw0, hw1);
+    assert(path.size() >= 2 && "NaiveRouter::route: invalid shortest path");
+    for (std::size_t i = 0; i < path.size() - 2; ++i) {
+      swaps.emplace_back(path[i], path[i + 1]);
     }
```
burgholzer: Just an optimization: Can we avoid creating two vectors here?
The current code generates the path as a vector only for it to group it up into pairs afterwards.
Certainly not a bottleneck, but feels like we could be doing better here.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Router.h
```diff
@@ -176,72 +148,66 @@ struct AStarHeuristicRouter final : RouterBase {
...
+  /// @brief Expand frontier with all neighbouring SWAPs in the current front.
+  void expand(MinQueue& frontier, const Node& parent,
+              mlir::SmallVector<mlir::ArrayRef<QubitIndexPair>> window,
               const Architecture& arch) const {
-    llvm::SmallDenseSet<QubitIndexPair, 64> swaps{};
-    for (const QubitIndexPair gate : layers.front()) {
+    llvm::SmallDenseSet<QubitIndexPair, 64> expansionSet{};
+
+    /// Currently: Don't revert last SWAP.
+    /// TODO: Idea? Don't revert "front" (independent) SWAPs?
```
burgholzer, line 184: Just for documentation purposes: As discussed privately, this is definitely a reasonable idea that should be tackled in a follow-up PR at some point. For here, this is fine and you can resolve this.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Unit.h
```diff
@@ -0,0 +1,42 @@
...
+
+#pragma once
+
+#include "mlir/Dialect/MQTOpt/Transforms/Transpilation/Layout.h"
+
+#include <utility>
+
+namespace mqt::ir::opt {
+
+/// @brief An Unit divides a quantum-classical program into routable sections.
```
burgholzer:
```suggestion
/// @brief A Unit divides a quantum-classical program into routable sections.
```
Might be worth checking through the PR for proper spelling of that.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/AStarRoutingPass.cpp
```diff
@@ -0,0 +1,203 @@
...
+                  remap(op, unit.layout());
+                })
+                .Case<ResetOp>([&](ResetOp op) { remap(op, unit.layout()); })
+                .Case<MeasureOp>(
+                    [&](MeasureOp op) { remap(op, unit.layout()); })
+                .Case<scf::YieldOp>([&](scf::YieldOp op) {
+                  if (unit.restore()) {
+                    rewriter.setInsertionPoint(op);
+                    insertSWAPs(op.getLoc(),
+                                llvm::to_vector(llvm::reverse(history)),
```
burgholzer: I assume that `to_vector` materialises an actual vector here, right?
If so, could we also just use a range for this to avoid copying? same goes for the naive router

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/Common.cpp
```diff
@@ -91,4 +94,81 @@ bool isTwoQubitGate(UnitaryInterface op) {
...
+
+    rewriter.setInsertionPointAfter(swap);
+    replaceAllUsesInRegionAndChildrenExcept(in0, out1, swap->getParentRegion(),
+                                            swap, rewriter);
+    replaceAllUsesInRegionAndChildrenExcept(in1, out0, swap->getParentRegion(),
+                                            swap, rewriter);
+
+    layout.swap(in0, in1);
+    layout.remapQubitValue(in0, out0);
+    layout.remapQubitValue(in1, out1);
```
burgholzer: couldn't this be handled via the `layout.remap(swap)` method that will potentially be introduced?

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Unit.h
MatthiasReumann: Good catch, thanks! I thought something was off with the `an`  here 👀

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Layout.h
MatthiasReumann: Good idea.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/QubitIndex.h
MatthiasReumann: Yeah, I thought so too. 

However, I initially didn't know where to put it else. I can't define it in `Common.h` since `Common.h` requires `Layout.h` and `Layout.h` requires `Common.h (QubitIndex)` (cyclic). 

Nonetheless, I've moved it to `Architecture.h` for now - which somewhat makes sense. In theory, the implementation of the Architecture could specify the required integer size for qubit indices.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/AStarRoutingPass.cpp
MatthiasReumann: Yes. Even though this requires a template to be fully compatible with any range, I think it's worth it 👍🏻

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Router.h
MatthiasReumann: Would it make sense if the `Architecture` returns pairs? I am not entirely sure 🤔

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/LayeredUnit.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+private:
+  /// @brief Maps operations to to-be-released iterators.
+  mlir::DenseMap<mlir::Operation*, mlir::SmallVector<Wire, 0>> onHold;
+  /// @brief Maps operations to ref counts.
+  mlir::DenseMap<mlir::Operation*, std::size_t> refCount;
+};
+
+mlir::SmallVector<Wire, 2> skipTwoQubitBlock(mlir::ArrayRef<Wire> wires,
+                                             Layer& opLayer) {
```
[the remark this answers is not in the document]
MatthiasReumann: We can safely assume that all 3+-qubit gates have been decomposed into one and two-qubit gates. Otherwise the input IR for the routing pass is invalid.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Common.h
```diff
@@ -80,4 +71,77 @@ using QubitIndexPair = std::pair<QubitIndex, QubitIndex>;
...
+ * @brief Given a layout, validate if the two-qubit unitary op is executable on
+ * the targeted architecture.
+ *
+ * @param op The two-qubit unitary.
+ * @param layout The current layout.
+ * @param arch The targeted architecture.
+ */
+[[nodiscard]] bool isExecutable(UnitaryInterface op, const Layout& layout,
+                                const Architecture& arch);
+
```
burgholzer: This seems like a clear candidate for a member function in the Architecture class, don't you think?

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Router.h
burgholzer: It could make sense to have an additional utility function in the architecture class that would directly return pairs.

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/AStarRoutingPass.cpp
```diff
@@ -0,0 +1,200 @@
...
+            }
+
+            /// Forward layout.
+            TypeSwitch<Operation*>(curr)
+                .Case<UnitaryInterface>([&](UnitaryInterface op) {
+                  if (auto swap = dyn_cast<SWAPOp>(op.getOperation())) {
+                    const auto in0 = swap.getInQubits()[0];
+                    const auto in1 = swap.getInQubits()[1];
+                    history.push_back({unit.layout().lookupHardwareIndex(in0),
+                                       unit.layout().lookupHardwareIndex(in1)});
```
burgholzer:
```suggestion
history.emplace_back(unit.layout().lookupHardwareIndex(in0),
                                       unit.layout().lookupHardwareIndex(in1));
```

mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/NaiveRoutingPass.cpp
```diff
@@ -0,0 +1,172 @@
...
+                      });
+                    }
+                  }
+                }
+
+                if (auto swap = dyn_cast<SWAPOp>(op.getOperation())) {
+                  const auto in0 = swap.getInQubits()[0];
+                  const auto in1 = swap.getInQubits()[1];
+                  history.push_back({unit.layout().lookupHardwareIndex(in0),
+                                     unit.layout().lookupHardwareIndex(in1)});
```
burgholzer:
```suggestion
history.emplace_back(unit.layout().lookupHardwareIndex(in0),
unit.layout().lookupHardwareIndex(in1));
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Changed
 
+- ♻️ Group dynamic circuit operations into scheduling units for MLIR routing ([#1301]) ([**@MatthiasReumann**])
```
```suggestion
- ♻️ Group circuit operations into scheduling units for MLIR routing ([#1301]) ([**@MatthiasReumann**])
```
Not particularly sure if "dynamic" really carried any meaning here.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Common.h
MatthiasReumann: Unfortunately, this would lead to a cyclic import as of now (Layout requires Architecture for the `QubitIndex` type).

My initial rationale was that the implementation of the `Architecture` class should be independent of MLIR ops, layouts, and so on.

MatthiasReumann: Maybe we should just drop the QubitIndex alias... 🤔

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Router.h
MatthiasReumann: Done.

mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/Common.h
burgholzer: Yeah. Let's just drop that alias. It isn't really helping much, is it?

burgholzer: Technically, you could also make this a member function of the `Layout` class 🤷🏼

MatthiasReumann: Done. I've added the `isExecutable` function to the Architecture class.

burgholzer (approved):
I am happy, the Rabbit is happy, CI is happy.
Let's go. Nice work 🙏🏼

Issue #1132, opened 2025-08-15
📦 MLIR - Enable by default and package MLIR/LLVM in Releases


burgholzer:
## Background
At the moment, none of the MLIR activities in the MQT Core project are included in release artifacts. The `mlir` subdirectory isn’t built by default; a dedicated CMake option must be set to enable our MLIR infrastructure. Historically this was intentional because:
- It was unclear whether we’d pursue this direction seriously or keep it as an experiment.
- There wasn’t much end-to-end functionality yet.
- There’s uncertainty around how to package LLVM/MLIR and distribute it efficiently.

The situation has changed:
- We intend to continue and invest in MLIR-based compilation.
- We recently merged initial support to import MQT Core IR into our MLIR dialect, conversions between internal dialects, initial optimization passes, and QIR export is imminent. At that point a minimal pipeline like “import Qiskit/OpenQASM → MLIR → QIR” is conceivable.
- The remaining blocker is sustainable packaging and distribution, especially for our Python wheels.

## Goal
Ship a minimal, usable MLIR toolchain by default in MQT Core releases and expose key functionality via Python, while keeping base wheels lean and managing PyPI storage constraints. Achieve this with a clear packaging strategy for LLVM/MLIR, robust CI/CD, and strict binary size budgets.

## Problem
- MLIR isn’t built or shipped by default; users can’t access it out of the box.
- Packaging LLVM/MLIR into Python wheels is complex and risks ballooning size (current base wheels are ~2–3 MiB; PyPI has a 10 GiB per-project cap; we already publish ~50 wheels per release).
- CI/CD must integrate MLIR builds (cibuildwheel), ensure correct dependency bundling (auditwheel/delocate), and maintain reliability across platforms.

## Proposal:
Stage the rollout and adopt a packaging strategy that balances usability and size.

1) Rollout plan
- Stage 0 (pre-releases):
  - Keep MLIR OFF by default in the top-level build, but produce experimental “MLIR-enabled” wheels as a separate companion package for pre-releases.
  - Gather feedback on usability, size, and platform coverage.
- Stage 1 (general availability):
  - Turn MLIR ON by default for source builds and developer configs.
  - Publish a companion binary wheel that adds MLIR/LLVM runtime and Python bindings, installable via an extra (e.g., `pip install mqt-core[mlir]`).
  - Keep base wheels lean; the extra pulls in the companion wheel only for users who need MLIR.
- Stage 2 (optional):
[...]

PR #1310, opened 2025-11-20
✨ Add `WireIterator`

MatthiasReumann:
## Description

This pull request adds a bi-directional `WireIterator` that traverses the def-use chain of a single qubit value.

## Checklist:


.github/workflows/reusable-mlir-tests.yml
```diff
@@ -103,9 +103,12 @@ jobs:
       - name: Build MLIR lit target
         run: cmake --build build --config ${{ matrix.coverage && 'Debug' || 'Release' }} --target mqt-core-mlir-lit-test-build-only
 
-      - name: Build MLIR unittests
+      - name: Build MLIR translation unittests
         run: cmake --build build --config ${{ matrix.coverage && 'Debug' || 'Release' }} --target mqt-core-mlir-translation-test
 
+      - name: Build MLIR wireiterator unittests
+        run: cmake --build build --config ${{ matrix.coverage && 'Debug' || 'Release' }} --target mqt-core-mlir-wireiterator-test
+
```
MatthiasReumann, line 88: This works but we should probably think about creating a single executable for MLIR unit tests.

MatthiasReumann: FYI: I've added a standalone `wireiterator` test here because it doesn't fit into the "translation" category.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,310 @@
...
+#include <llvm/Support/Debug.h>
+#include <llvm/Support/ErrorHandling.h>
+#include <mlir/Analysis/SliceAnalysis.h>
+#include <mlir/Dialect/SCF/IR/SCF.h>
+#include <mlir/IR/Operation.h>
+#include <mlir/IR/Value.h>
+#include <mlir/Support/LLVM.h>
+
+namespace mqt::ir::opt {
+using namespace mlir;
```
taminob: Having a `using namespace` in a header is bad practice because it forces all files that include this file to use this statement. Also, it can make things hard to figure out because include order becomes important.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * It does not visit operations within nested regions. These include the loop
+ * body of the scf::ForOp and the THEN and ELSE branches of the scf::IfOp. From
+ * the iterator's perspective these act like regular gates. As a consequence,
+ * an input qubit is mapped to the respective output qubit. For example, finding
+ * which result of an scf::IfOp corresponds to a qubit passed into one of its
+ * regions.
+ */
+class WireIterator {
+  /// @returns a view of all input qubits.
```
taminob, line 42: I'm not entirely sure about the coding guidelines (@burgholzer the link in the [CONTRIBUTING.md](https://github.com/munich-quantum-toolkit/core/blob/main/.github/CONTRIBUTING.md) is broken), but I think a lot of common style guides prefer to have the public interface first and then below the protected and private member functions with finally the member variables at the end.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,310 @@
...
+  [[nodiscard]] static auto getAllInQubits(UnitaryInterface op) {
+    return llvm::concat<Value>(op.getInQubits(), op.getPosCtrlInQubits(),
+                               op.getNegCtrlInQubits());
+  }
+
+  /// @returns a view of all output qubits.
+  [[nodiscard]] static auto getAllOutQubits(UnitaryInterface op) {
+    return llvm::concat<Value>(op.getOutQubits(), op.getPosCtrlOutQubits(),
+                               op.getNegCtrlOutQubits());
+  }
```
taminob, line 52: I really like these helpers, I hope this will become unnecessary in the new `UnitaryInterface` :)

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,310 @@
...
+   *
+   * @note That we don't use the interface method here because
+   * it creates temporary std::vectors instead of using views.
+   */
+  [[nodiscard]] static Value findInput(UnitaryInterface op, Value out) {
+    const auto outs = getAllOutQubits(op);
+    const auto it = llvm::find(outs, out);
+    assert(it != outs.end() && "output qubit not found in operation");
+    const auto index = std::distance(outs.begin(), it);
+    return *(std::next(getAllInQubits(op).begin(), index));
```
taminob: I think this is fine, but I had to think for a moment to make sure this is correct regarding lifetimes. It might be easier to read if you move the `getAllInQubits(op)` out of this statement (similar to the `outs`).

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,310 @@
...
+  using value_type = Operation*;
+
+  explicit WireIterator() = default;
+  explicit WireIterator(Value q, Region* region)
+      : currOp(q.getDefiningOp()), q(q), region(region) {}
+
+  Operation* operator*() const {
+    assert(!sentinel && "Dereferencing sentinel iterator");
+    assert(currOp && "Dereferencing null operation");
+    return currOp;
```
taminob, line 204: Maybe a `value()` function might also be valuable to get the qubit that the wire is running on? At least I'll need to then determine the position of the qubit in the qubits of the operation and would need to keep track of the qubit externally without that information.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,310 @@
...
+            report_fatal_error("unknown op in def-use chain: " +
+                               op->getName().getStringRef());
+          });
+    }
+
+    /// Find the next operation.
+    /// If it is a sentinel there are no more ops.
+    if (sentinel) {
+      return;
+    }
```
taminob, line 270: I don't really get why the sentinel is checked here again? Won't this also be covered by the `q.use_empty()` check?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Added.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Done.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Pretty sure they will be. 

@burgholzer Maybe this is of interest for you. These helpers create concatenated views instead of creating many temporary `std::vector`s.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: That makes so much sense. I was always wondering why people would prefer this. Thanks! Changed.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Essentially this is what I am doing too. I just put the private _static_ firsts. Maybe @burgholzer can weigh in?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: The sentinel check is here to return after we found an `scf::YieldOp` (or any other terminator of a given region, for that matter). 

The `q.use_empty()` check is here to deal with static qubits which won't be deallocated and hence don't have a nice definition of "sentinel". Specifically, in entry point functions the quantum flow of a static qubit just stops at a given unitary.

taminob: You are right, thanks for your explanation!

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
taminob: Thanks! Sorry that I think about that just now, but maybe we should rather name it `qubit()` instead of `value()` because in e.g. `std::optional` the dereference operator and the `value()` function both return the same result which could become confusing here if it doesn't return the same type/value... What do you think?

MatthiasReumann: Yeah. I think this makes sense. Good catch!

MatthiasReumann: Done.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
burgholzer: The contributing link has been fixed in the meantime.

We don't explicitly dictate the order of the individual entities of a class. I am fine with either solution here. If that turns out to follow some best practice, even better.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+   */
+  [[nodiscard]] mlir::Region* getRegion() {
+    return region != nullptr ? region : q.getParentRegion();
+  }
+
+  mlir::Operation* currOp{};
+  mlir::Value q;
+  mlir::Region* region{};
+  bool sentinel{false};
+};
```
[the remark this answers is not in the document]
MatthiasReumann: Unfortunately `q.getParentRegion()` is not a const method.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: I don't think it's a best practice necessarily, but I've seen this in a bunch of LLVM files.

mlir/unittests/dialect/test_wireiterator.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+      rec++;
+      checkOperationEqual(*rec,
+                          "%out_qubits_7 = mqtopt.h() %arg1 : !mqtopt.Qubit");
+
+      rec++;
+      checkOperationEqual(*rec, "scf.yield %out_qubits_7, %out_qubits_8 : "
+                                "!mqtopt.Qubit, !mqtopt.Qubit");
+    }
+  }
```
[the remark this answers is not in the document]
MatthiasReumann: Even though trivial, I thought testing the post increment operators also is a good idea. What do you say, Mr. 🐇?

.github/workflows/reusable-mlir-tests.yml
burgholzer: I think this will become less relevant, when we resolve #1132 because at that point, we will always be running the full (google)test suite and all test executables anyway.
However, you can combine both lines here and simplify this. You can pass multiple `--target <...>` parameters to a single CMake invocation. That should keep the changes here to a minimum.

mlir/unittests/dialect/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+set(testname "mqt-core-mlir-wireiterator-test")
+file(GLOB_RECURSE WIREITERATOR_TEST_SOURCES *.cpp)
+get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
+
+if(NOT TARGET ${testname})
+  # create an executable in which the tests will be stored
+  add_executable(${testname} ${WIREITERATOR_TEST_SOURCES})
+  # link the Google test infrastructure and a default main function to the test executable.
+  target_link_libraries(${testname} PRIVATE GTest::gmock GTest::gtest_main ${dialect_libs})
```
burgholzer:
```suggestion
  target_link_libraries(${testname} PRIVATE GTest::gtest_main ${dialect_libs})
```
At least I have not seen any use of any mocking utilities.

mlir/unittests/dialect/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# Copyright (c) 2025 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+set(testname "mqt-core-mlir-wireiterator-test")
+file(GLOB_RECURSE WIREITERATOR_TEST_SOURCES *.cpp)
+get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
```
burgholzer: I hope I am not opening Pandora's box here, but could it be worth to be a little more selective about the libraries we declare as link libraries here?
I would assume that it may help overall compile times if we only explicitly list the libraries that we depend on here. Also when it comes to recompilation etc.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ## [Unreleased]
 
+### Added
+
+- ✨ Add WireIterator ([#1310]) ([**@MatthiasReumann**])
```
burgholzer: Can we make this a little bit more self-explanatory?
Changelogs are meant to be read by humans and, from the current entry, I would have close to no idea what this change actually implements.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+ * operations in a given region. It respects the semantics of the respective
+ * quantum operation including control flow constructs (scf::ForOp and
+ * scf::IfOp).
+ *
+ * It does not visit operations within nested regions. These include the loop
+ * body of the scf::ForOp and the THEN and ELSE branches of the scf::IfOp. From
+ * the iterator's perspective these act like regular gates. As a consequence,
+ * an input qubit is mapped to the respective output qubit. For example, finding
+ * which result of an scf::IfOp corresponds to a qubit passed into one of its
+ * regions.
```
burgholzer: > As a consequence,
> an input qubit is mapped to the respective output qubit. For example, finding
> which result of an scf::IfOp corresponds to a qubit passed into one of its
> regions.

This is likely just me, but these sentences somehow do not make too much sense to me. Could you try rephrasing that somehow to make the intent clearer?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
burgholzer: Yeah. The new interface will offer these functions.
@denialhaag `llvm::concat` might be something to consider for avoiding the explicit construction of a dedicated SmallVector in the respective implementations for the control modifier. I think we still copy there.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+   */
+  [[nodiscard]] static mlir::Value findInput(UnitaryInterface op,
+                                             mlir::Value out) {
+    const auto ins = getAllInQubits(op);
+    const auto outs = getAllOutQubits(op);
+    const auto it = llvm::find(outs, out);
+    assert(it != outs.end() && "output qubit not found in operation");
+    const auto index = std::distance(outs.begin(), it);
+    return *(std::next(ins.begin(), index));
+  }
```
burgholzer, line 84: @denialhaag also looks like really good inspiration for the unitary interface implementations.

Overall, these functions can be assumed to be present in the dialect rewrite and can be assumed to be implemented performantly.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+      mlir::Operation* curr = it.qubit().getDefiningOp();
+      if (curr == prev || curr == nullptr) {
+        break;
+      }
+      prev = *it;
+      --it;
+    }
+
+    return it.qubit();
+  }
```
burgholzer, line 174: I assume these are all here to stay and there is nothing that we could do in the dialect rewrite that would simplify any of this, right?
These generally seem to be useful helpers for control flow handling though. Might it be worth collecting them in a more common place where they could more easily be reused?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * @brief Find corresponding result value from input qubit value (Forward).
+   *
+   * @details Recursively traverses the IR "downwards" until the respective
+   * yield is found. Requires that each branch takes and returns the same
+   * (possibly modified) qubits. Hence, we can just traverse the then-branch.
+   */
+  [[nodiscard]] static mlir::Value findResult(mlir::scf::IfOp op,
+                                              mlir::Value q) {
+    WireIterator it(q, &op.getThenRegion());
```
burgholzer: Obviously a micro-optimization, but assuming there is an efficient way of determining the size of each of the regions (then and else), wouldn't it be a clever heuristic to traverse the shorter of the regions?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+   */
+  [[nodiscard]] static mlir::Operation* getUserInRegion(mlir::Value v,
+                                                        mlir::Region* region) {
+    for (mlir::Operation* user : v.getUsers()) {
+      if (user->getParentRegion() == region) {
+        return user;
+      }
+    }
+    return nullptr;
+  }
```
burgholzer, line 190: Does such a function not exist in MLIR already? I vaguely remember seeing something like this.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+    return tmp;
+  }
+
+  bool operator==(const WireIterator& other) const {
+    return other.q == q && other.currOp == currOp && other.sentinel == sentinel;
+  }
+
+  bool operator==([[maybe_unused]] std::default_sentinel_t s) const {
+    return sentinel;
+  }
```
burgholzer, line 237: Just thinking out loud here, but could it be valuable to implement the starship operator here `<=>`?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+    /// If it is a sentinel there are no more ops.
+    if (sentinel) {
+      return;
+    }
+
+    /// If there are no more uses, set `sentinel` to true.
+    if (q.use_empty()) {
+      sentinel = true;
+      return;
+    }
```
burgholzer, line 276: This would be an ill-formed program, right?
any qubit that was allocated needs to be deallocated at some point.
Edit: ah, I assume for static qubits there is no deallocation and they just end in the void, which would trigger this condition here.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+
+    /// Otherwise, search the user in the targeted region.
+    currOp = getUserInRegion(q, getRegion());
```
burgholzer, line 279: At this point, it is pretty much guaranteed that `q` has `q.hasOneUse() == true`, right? due to the linear typing of qubit values in the Opt dialect.
Couldn't you simply use `currOp = *q.getUsers().begin()` here? I am probably overlooking something here.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,337 @@
...
+        .Case<DeallocQubitOp>([&](DeallocQubitOp op) { q = op.getQubit(); })
+        .Case<mlir::scf::ForOp>(
+            [&](mlir::scf::ForOp op) { q = findInitArg(op, q); })
+        .Case<mlir::scf::IfOp>(
+            [&](mlir::scf::IfOp op) { q = findValue(op, q); })
+        .Case<AllocQubitOp, QubitOp>([&](auto) { /* hold (no-op) */ })
+        .Default([&](mlir::Operation* op) {
+          report_fatal_error("unknown op in def-use chain: " +
+                             op->getName().getStringRef());
+        });
```
burgholzer, line 320: How come that the backwards path is so much shorter and cleaner than the forward pass?

mlir/unittests/dialect/CMakeLists.txt
MatthiasReumann: Good catch, thanks!

mlir/unittests/dialect/CMakeLists.txt
MatthiasReumann: Makes a lot of sense. Done!

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Yeah, I agree. All I wanted so way with that is "takes n qubits, produces n qubits, doesn't matter what's in the respective nested region". 

I've rephrased this bit.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: The forward pass has to handle the past-the-end iterator logic that C++ requires. That adds a bit of complexity. Moreover, since there is no `mqtopt.release` (or similar) that defines the end of a static qubit, `advanceForward` requires the 
```cpp
if (q.use_empty()) {
  sentinel = true;
  return;
}
```

check.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Note: `q.hasOneUse()` returns true whenever `q` has _exactly_ one use (as opposed "more than one").  

At this point, it may be the case that the current value of `q` (for example produced by a `UnitaryInterface` op) is used in the then and else-branch (´getNumUses() == 2`). If so, there is no user in the active region, and hence, `currOp == nullptr`. The current operation is set to this `scf.if`. 

To retrieve the corresponding result in one of the branches (by finding the `YieldOp`), we use a new iterator with the then-branch as targeted region. Since the qubit might be used in the then and else-branch (For example, the `%out_qubits_4` in the unit test), we need to make sure that we iterate over the correct user. Hence, the `getUserInRegion` check instead of `currOp = *q.getUsers().begin()`.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: You probably saw it in another of my pull requests, I've copied the function to the `WireIterator`. As mentioned above we might want to collect these somewhere.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: My gut feeling said no, but I wasn't 100% sure either. For the `bidirectional_iterator` it definitely isn't required. 

Both GPT and Claude agreed that it wouldn't bring any benefit for the exact same reason as stated above: It fulfills the requirements of the `bidirectional_iterator`.

MatthiasReumann: Let's see what Mr. 🐇 has to say.

@coderabbitai Do you think implementing the starship operator for the `WireIterator` would bring any benefit?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Good idea.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,341 @@
...
+#pragma once
+
+#include "mlir/Dialect/MQTOpt/IR/MQTOptDialect.h"
+
+#include <cstddef>
+#include <iterator>
+#include <llvm/ADT/STLExtras.h>
+#include <llvm/ADT/TypeSwitch.h>
+#include <llvm/Support/Debug.h>
+#include <llvm/Support/ErrorHandling.h>
```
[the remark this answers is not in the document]
MatthiasReumann: Seems to included in `<iterator>`.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
```diff
@@ -0,0 +1,340 @@
...
+   */
+  [[nodiscard]] static mlir::Value findResult(mlir::scf::IfOp op,
+                                              mlir::Value q) {
+    /// Use the branch with fewer ops.
+    /// Note: LLVM doesn't guarantee that range_size is in O(1).
+    /// Might effect performance.
+    const auto szThen = llvm::range_size(op.getThenRegion().getOps());
+    const auto szElse = llvm::range_size(op.getElseRegion().getOps());
+    mlir::Region& region =
+        szElse >= szThen ? op.getThenRegion() : op.getElseRegion();
```
burgholzer, line 121: Can the else branch of an if be empty and would this still work in that case?
I would assume that an `scf::IfOp` does not necessarily need an else case.
Might be worth a test for validation purposes?

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
burgholzer: Thanks for the explanation. That makes a lot more sense now 👍🏻

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
burgholzer: Thanks also here for the explanation 🙌
Makes sense.

burgholzer (approved):
Thanks @MatthiasReumann for the additional round of changes. This looks good to me now. Feel free to resolve the open comments.
That should leave one last question that popped up during the review. Once resolved, this can go in.

mlir/include/mlir/Dialect/MQTOpt/IR/WireIterator.h
MatthiasReumann: Whenever a `scf.if` produces values (i.e. yields values) it requires an else branch. In the opt (flux) dialect this will always be the case. 

From the [MLIR documentation](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfif-scfifop) of the SCF Dialect
> In case the scf.if produces results, the “else” region must also have exactly 1 block.

burgholzer: Ah, perfect 👌🏻 thanks for checking!

PR #1337, opened 2025-11-27
🔧 Streamline DX

denialhaag:
## Description

This PR includes some changes to streamline the DX across repositories. 

## Checklist:

- [x] ~~I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.~~

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -88,10 +88,10 @@ repos:
     hooks:
       - id: ty-check
         name: ty check
-        entry: uv run --only-dev ty check
-        language: system
-        pass_filenames: false
-        always_run: true
+        entry: python -c "import subprocess; import sys; subprocess.run(['uv', 'sync', '--no-install-project']); sys.exit(subprocess.run(['uv', 'run', '--no-sync', 'ty', 'check']).returncode)"
+        language: unsupported
+        require_serial: true
+        types_or: [python, pyi, jupyter]
 
   # Also run Black on examples in the documentation
   - repo: https://github.com/adamchainz/blacken-docs


.pre-commit-config.yaml
```diff
@@ -96,7 +88,7 @@ repos:
     hooks:
       - id: ty-check
         name: ty check
-        entry: uvx ty check .
+        entry: uvx ty==0.0.1a27 check .
```
denialhaag: This is necessary for `prek run -a` not to fail for me locally. I don't know if there's a better solution, as Renovate won't be updating this. 🤔

burgholzer: I think I found a good solution for this (see https://github.com/munich-quantum-toolkit/core/pull/1337/commits/c2f58e9e37842f5d8a2d06481010f18fc80421ce). Could you try that locally?

denialhaag: That works, and it seems a lot better! Thanks!

burgholzer: Nice! I also created a PR in the workflows repo to make use of this simpler command https://github.com/munich-quantum-toolkit/workflows/pull/257 and tagged you for a review.

pyproject.toml
```diff
@@ -348,6 +348,5 @@ dev = [
   {include-group = "test"},
   "lit>=18.1.8",
   "nox>=2025.11.12",
-  "ruff>=0.14.6",
-  "ty>=0.0.1a27",
+  "ty==0.0.1a27",
```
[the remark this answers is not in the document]
burgholzer: The dependency is not unused.
It is used as part of the pre-commit checks, particularly the `ty-check` hook.
The explicit pin here ensures that everyone is running the same version of `ty` and that we do not get hit by unexpected surprises due to updates in `ty`, which is still in alpha and changing rapidly.

PR #1328, opened 2025-11-24
✨ New QDMI Device to represent SC Devices

ystade:
## Description

This PR adds a—for the moment—minimalistic SC QDMI device that only features a coupling map. The changes extend the QDMI device framework by relocating the Device class to the qdmi::na namespace, adding a new superconducting (SC) device backend with session and job management APIs, introducing JSON-based device configuration and code generation utilities, and providing a CLI tool and comprehensive test coverage for the SC device.

## Checklist:

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

PR #1355, opened 2025-12-03
:sparkles: QDMI authentication via FoMaC

### LINKED ISSUE #1332: Unify the Directory Structure for QDMI Devices
As pointed out by @burgholzer in https://github.com/munich-quantum-toolkit/core/issues/1328#issuecomment-3570503942, especially the NA device lives currently out of place compared to other QDMI devices. Consequently, the NA device should be moved to the same location as the others. While doing that, one could also think about proper namespacing the devices.

### LINKED ISSUE #1350: ✨ Authentication via QDMI-Qiskit Backend
### Problem Statement

Currently, the QDMI-Qiskit Backend added in #1243 doesn't support authentication. Clients need a way to provide credentials for individual devices to authenticate.

### Proposed Solution

According to @burgholzer, these need to be constructor parameters for the backends and also need to be passed to the provider.

marcelwa:
## Description

This PR aims to introduce session authentication via QDMI to FoMaC. I didn't see a clear path to implement this without any API changes.

In essence, in C++ land, `FoMaC` has been renamed to `Session` (as agreed upon by @burgholzer) and is no longer a singleton; it must be instantiated to handle sessions. The implementation sets session parameters during construction. The Python bindings directly follow the new C++ semantics. Furthermore, the extensions have been propagated to `backend.py` and `provider.py` for Qiskit integration.

On the driver side, `Driver::addDynamicDeviceLibrary` has been extended with a `qdmi::DeviceSessionConfig` parameter to set `QDMI_DEVICE_SESSION_PARAMETER`s upon loading a library. This function has also been exposed to Python through `mqt.core.qdmi.fomac`, which is an inconsistency, but the `Driver` interface is currently not available in the bindings. We could clean this up later.

Known limitations:
- Authentication applies to all devices (#1362).
- Current QDMI devices don't communicate the forms of authentication they require or support (#1361).

All this will be addressed in future PRs.

Fixes #1350 
Fixes #1332

## Checklist:

### DIFF
--- a/include/mqt-core/qdmi/Common.hpp
+++ b/include/mqt-core/qdmi/Common.hpp
@@ -51,6 +51,8 @@ namespace qdmi {
     }                                                                          \
   }
 
+// STRNCPY wrapper: strncpy_s on Windows (auto null-terminates),
+// strncpy on other platforms (requires manual null-termination - see usage).
 #ifdef _WIN32
 #define STRNCPY(dest, src, size)                                               \
   strncpy_s(static_cast<char*>(dest), size, src, size);
@@ -67,6 +69,7 @@ namespace qdmi {
           return QDMI_ERROR_INVALIDARGUMENT;                                   \
         }                                                                      \
         STRNCPY(value, prop_value, size);                                      \
+        /* Ensure null-termination: strncpy doesn't guarantee it on non-Win */ \
         /* NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) */  \
         static_cast<char*>(value)[size - 1] = '\0';                            \
       }                                                                        \


include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -141,30 +143,64 @@ concept maybe_optional_value_or_string_or_vector =
...
+  /// Custom parameter 2 (driver-defined)
+  CUSTOM2 = QDMI_SESSION_PARAMETER_CUSTOM2,
+  /// Custom parameter 3 (driver-defined)
+  CUSTOM3 = QDMI_SESSION_PARAMETER_CUSTOM3,
+  /// Custom parameter 4 (driver-defined)
+  CUSTOM4 = QDMI_SESSION_PARAMETER_CUSTOM4,
+  /// Custom parameter 5 (driver-defined)
+  CUSTOM5 = QDMI_SESSION_PARAMETER_CUSTOM5
+};
+
```
burgholzer, line 176: Why do we need to re-expose this enum when there already is a perfectly fine enum? Doesn't this create an additional level of abstraction that we probably do not want?

python/mqt/core/fomac.pyi
```diff
@@ -34,6 +37,73 @@ class ProgramFormat(Enum):
...
+    """Custom parameter 1 (driver-defined)"""
+    CUSTOM2 = ...
+    """Custom parameter 2 (driver-defined)"""
+    CUSTOM3 = ...
+    """Custom parameter 3 (driver-defined)"""
+    CUSTOM4 = ...
+    """Custom parameter 4 (driver-defined)"""
+    CUSTOM5 = ...
+    """Custom parameter 5 (driver-defined)"""
+
```
burgholzer: For consistent design, I would nest this enum in the `Session` class

python/mqt/core/fomac.pyi
```diff
@@ -34,6 +37,73 @@ class ProgramFormat(Enum):
...
+class Session:
+    """A FoMaC session for managing QDMI devices.
+
+    Allows creating isolated sessions with independent authentication settings.
+    """
+
+    def __init__(self) -> None:
+        """Create a new FoMaC session."""
+
+    def set_session_parameter(self, param: SessionParameter, value: str) -> None:
```
burgholzer: This can be called `set_parameter` since we are in a session already, anyway.

python/mqt/core/fomac.pyi
```diff
@@ -205,5 +275,39 @@ class Device:
...
+        >>> import mqt.core.fomac as fomac
+        >>> fomac.set_session_parameter(fomac.SessionParameter.TOKEN, "my_token")
+        >>> devices = fomac.devices()
+
+        # Or use Session class for isolated sessions:
+        >>> session = fomac.Session()
+        >>> session.set_session_parameter(fomac.SessionParameter.TOKEN, "token")
+        >>> session_devices = session.get_devices()
+    """
+
```
burgholzer: I would not consider this necessary.
In my opinion, backwards compatibility is not an issue here. We can happily break things. MQT Core is the only project that currently uses this functionality.
So I would probably also remove the default `devices()`

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -674,12 +709,13 @@ class FoMaC {
 
 private:
   QDMI_Session session_ = nullptr;
+  mutable std::mutex mutex_;
+  bool initialized_ = false;
+  std::unordered_map<SessionParameter, std::string> pendingParameters_;
```
burgholzer: In order to stay consistent with how we handled other aspects of QDMI in the C++ interface, I would argue that the session parameters should either be individual member functions or they should be keyword arguments in the constructor of a session. I would have a slight preference for the kwarg approach as it is very close to what we would expect to expose to Python/Qiskit.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -696,14 +732,65 @@ class FoMaC {
   }
 
 public:
+  /**
+   * @brief Constructs a new FoMaC session.
+   * @details Creates and allocates a new QDMI session. The session is not
+   * initialized until the first call to getDevices() or after setting
+   * authentication parameters.
+   */
+  FoMaC();
```
burgholzer: I am entirely fine with calling this a session as well. Breaking changes are fine and I think this is a good one.
The namespace can still be called `fomac` and our understanding of that is then that it provides the C++ and Python abstractions over QDMI.

burgholzer (changes requested):
This is a very quick review with the goal to dump some thoughts on the general direction here.
I'll try to summarize them again here:
- Breaking changes are entirely fine here
- I would call the entity `Session` also in C++. Makes the intent clearer
- I would use keyword arguments for the constructor of the session (all optional; all defaulted to `std::nullopt`). That might safe exposing the actual enum.
- We can drop the default/global `get_devices` in my opinion. In a not too distant future, we may need to provide credentials for some of the devices that we include.

I think that's it from the very direct feedback.
There is one aspect, that we won't be solving as part of this PR.
And that is https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/161
Right now, only one set of authentication credentials can be set.
However, each device might potentially require different authentication criteria. The current way we generally implement drivers and, potentially, the way the client interface is structured don't allow us to set individual parameters per device.
This is something to think about in the not-too-distant future, but not as part of this PR.

bindings/fomac/fomac.cpp
```diff
@@ -183,7 +200,17 @@ PYBIND11_MODULE(MQT_CORE_MODULE_NAME, m, py::mod_gil_not_used()) {
...
+  auto fomac = py::class_<fomac::FoMaC>(m, "Session");
+  fomac.def(py::init<>());
+  fomac.def("set_session_parameter", &fomac::FoMaC::setSessionParameter,
+            "param"_a, "value"_a);
+  fomac.def("get_devices", &fomac::FoMaC::getDevices);
+
+  // Module-level convenience functions (use default session)
+  m.def("set_session_parameter", &fomac::setSessionParameter, "param"_a,
+        "value"_a);
+  m.def("devices", &fomac::getDevices);
```
ystade: Not sure whether these function should have a more explicit name like, e.g., `get_default_devices`

include/mqt-core/fomac/FoMaC.hpp
ystade: Mh, I am slightly in favor of the individual memebr functions way as this corresponds to the rest of the current look and feel of the fomac implementation.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -674,12 +709,13 @@ class FoMaC {
 
 private:
   QDMI_Session session_ = nullptr;
+  mutable std::mutex mutex_;
+  bool initialized_ = false;
+  std::unordered_map<SessionParameter, std::string> pendingParameters_;
+
+  /// @brief Ensures the session is initialized, applying pending parameters
+  auto ensureInitialized() -> void;
```
ystade: I would prefer a function that aligns closer with the workflow of QDMI, i.e., a function `initialize` that may also throw an exception if the session has been already initialized.

include/mqt-core/fomac/FoMaC.hpp
ystade: Mh, as discussed with Lukas yesterday, also the entire suite including QDMI and around QDMI could be referred to as QDMI including the C interface, a C++ API, and a Python API. Then, I would suggest to put all of this into a namespace `qdmi`.

And I am quite fine with dropping the slightly akward name `FoMaC` entirely.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -696,14 +732,65 @@ class FoMaC {
...
+
+  // Delete copy constructors and assignment operators
   FoMaC(const FoMaC&) = delete;
   FoMaC& operator=(const FoMaC&) = delete;
-  FoMaC(FoMaC&&) = default;
-  FoMaC& operator=(FoMaC&&) = default;
+
+  // Allow move semantics
+  FoMaC(FoMaC&&) noexcept;
+  FoMaC& operator=(FoMaC&&) noexcept;
```
ystade: This LGTM, I just wanted to note something because it popped up in antoher PR: If the instance of the singleton is stored in a pointer, also move semantics should not be allowed because the static `getInstance` methods then expects the instance at a certain memory location. For this imeplementation this should just be fine.

python/mqt/core/fomac.pyi
ystade: Applies also to the C++ side.

python/mqt/core/fomac.pyi
ystade: Fine with me

ystade (commented):
Hi @marcelwa , this is a thoughtful and forward-looking refactoring of the FoMaC. To summarize my comments:
- I am totally fine with dropping the name `FoMaC` and call it `Session` (in a `qdmi` namespace)
- I would prefer to align the session handling closer to the original QDMI workflow, i.e., having individual memeber functions to set the indidivual parameters and then calling a function `initialize`. Only afterwards, one can for example retrieve the list of devices.

include/mqt-core/fomac/FoMaC.hpp
burgholzer: That is true. Couldn't we do a compromise here and actually provide both, arguments to the constructor and getters/setters as member functions?
One would still need to explicitly call `initialize`.

include/mqt-core/fomac/FoMaC.hpp
burgholzer: I am also fine with putting this under a `qdmi` namespace entirely. This is simply the `C++` API for QDMI. And, on top of it, we build a Python API as well as a Qiskit API.
All still QDMI.
The question is, whether this might be a bit too much for this PR here.
Since we have a deadline upcoming, we should try to keep the scope of this as small as possible and rather move quickly.
We can polish/rename/refactor later.
@marcelwa @ystade does that sound reasonable?

include/mqt-core/fomac/FoMaC.hpp
burgholzer: Since `FoMaC` here is no longer a singleton, this should not be a concern.

include/mqt-core/fomac/FoMaC.hpp
ystade: I would be happy with either _one_ alternative as having two ways setting the parameters. In general, this sounds pretty much like a problem that is usually solved with the Builder Pattern. So in conclusion, I rather accept the kwarg solution than having both.

include/mqt-core/fomac/FoMaC.hpp
ystade: Absolutely reasonable. For anything not covered we should open tracking issues.

include/mqt-core/fomac/FoMaC.hpp
ystade: True that.

include/mqt-core/fomac/FoMaC.hpp
burgholzer: The question to me is: from a practical standpoint, would there ever be a case where a user might sequentially set authentication information because it is not known in its entirety apriori?
But now that I think of it, the kwargs solution is only convenient in Python, where the order of the function parameters doesn't matter as long as you use named arguments.
For C++, a function that takes 8 parameters which are all optional is quite cumbersome to use if you, e.g., want to only set the 8th parameter.

How bad would it be to expose a different API for C++ and Python? 🙃

ystade: For the C++ side, we could define a struct that carries all options as optionals. Then the respective paramters can be set and the resulting object can be passe to the constructor of the Session. THis translates quite well to a kwarg list on the Python side, as I did in the zoned neutral atom compiler in QMAP.

ystade: For reference: https://github.com/munich-quantum-toolkit/qmap/blob/01e3c465ab69bf5053dd57907ad7fa34506bef09/bindings/na/zoned/zoned.cpp#L61-L89

burgholzer: I like that approach 👍🏻

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -177,16 +181,36 @@ inline auto throwIfError(int result, const std::string& msg) -> void {
...
+  /// URL to authentication server
+  std::optional<std::string> authUrl;
+  /// Username for authentication
+  std::optional<std::string> username;
+  /// Password for authentication
+  std::optional<std::string> password;
+  /// Project ID for session
+  std::optional<std::string> projectId;
+};
+
```
burgholzer, line 176: I am pretty sure that we will need to also expose the custom config parameters here as they might be necessary for certain use cases. This naturally extends throughout the entire PR.

One particular problem that we have at the moment: The client interface does not allow to set a BASEURL because this is a DEVICE_SESSION_PARAMETER, not a SESSION_PARAMETER.
So, at the moment, there is no (direct) way that you could be setting the baseURL for a QDMI device.
At least not directly via QDMI.
In the case that we have to deal with (not as part of this PR), I am pretty sure that we can get around this by making use of environment variables, which isn't necessarily pretty, but should work.

src/fomac/FoMaC.cpp
```diff
@@ -776,14 +810,105 @@ auto FoMaC::Job::getSparseProbabilities() const
...
+Session& Session::operator=(Session&& other) noexcept {
+  if (this != &other) {
+    if (session_ != nullptr) {
+      QDMI_session_free(session_);
+    }
+    session_ = other.session_;
+    other.session_ = nullptr;
+  }
+  return *this;
+}
```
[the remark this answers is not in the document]
burgholzer: Let's not care too much about thread safety as part of this PR. This will be investigated more deeply in the future. There already is a tracking issue for that.

bindings/fomac/fomac.cpp
```diff
@@ -26,23 +27,52 @@ namespace py = pybind11;
...
+            config.password = password;
+            config.projectId = projectId;
+            return fomac::Session{config};
+          }),
+      "token"_a = defaultConfig.token, "auth_file"_a = defaultConfig.authFile,
+      "auth_url"_a = defaultConfig.authUrl,
+      "username"_a = defaultConfig.username,
+      "password"_a = defaultConfig.password,
+      "project_id"_a = defaultConfig.projectId);
+
```
burgholzer: Wouldn't it be simpler here to remove `defaultConfig` and simply use `std::nullopt` for the default values?

python/mqt/core/fomac.pyi
```diff
@@ -34,6 +34,66 @@ class ProgramFormat(Enum):
...
+            >>> # Session with file-based authentication
+            >>> session = Session(auth_file="/path/to/auth.json")
+            >>> devices = session.get_devices()
+            >>>
+            >>> # Session with multiple parameters
+            >>> session = Session(
+            ...     auth_url="https://auth.example.com", username="user", password="pass", project_id="project-123"
+            ... )
+            >>> devices = session.get_devices()
+        """
```
burgholzer, line 103: Not an expert here, but shouldn't code examples be in Markdown fences (```)?

burgholzer (changes requested):
Thanks Marcel for all the work on this. This is looking like a great addition.
I have less than a handful of comments left. We should get this in soon.
The biggest comment is probably on the Custom parameters which should probably also be exposed.
Code Rabbit also still seems to have some comments that seem valid and valuable.

Let's use our internal retreat to maybe come up with better solutions for how we could be setting options for individual devices from the C++/Client side.

At the moment, I can't really think of a better way to test the authentication without a device that actually implements all these. We'll do better once such a device is publicly available.

burgholzer:
The conflicts here stem from the merge of #1328.
I hope they are quick enough to resolve.

bindings/fomac/fomac.cpp
marcelwa: I based this off @ystade's example he posted here: https://github.com/munich-quantum-toolkit/core/pull/1355#discussion_r2588318278

burgholzer: Yeah, I think it makes a little more there, where there is such a diverse set of defaults.
Here, everything is optional and a string.

python/mqt/core/fomac.pyi
marcelwa: The `>>>` syntax is preferred by many for its proper integration with `doctest`. In a nutshell, `doctest` can execute this as if it were plain Python code, whereas Markdown fences get ignored as static text. It may not be a significant consideration for this project, but I'd be happy to accommodate your preference here.

burgholzer: Ah. I did not know that 💡 Then let's keep the `>>>` please.

src/fomac/FoMaC.cpp
```diff
@@ -776,14 +810,78 @@ auto FoMaC::Job::getSparseProbabilities() const
...
+  }
+  // Validate URL format for authUrl
+  if (config.authUrl) {
+    // Adapted from: https://uibakery.io/regex-library/url
+    static const std::regex URL_PATTERN(
+        R"(^https?://(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_\+.~#?&/=]*)$)");
+    if (!std::regex_match(*config.authUrl, URL_PATTERN)) {
+      throw std::runtime_error("Invalid URL format: " + *config.authUrl);
+    }
+  }
```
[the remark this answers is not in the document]
marcelwa: Good point. What do you think, @burgholzer? Is it desired to test with local IPs?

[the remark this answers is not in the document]
burgholzer: I wouldn't be overly strict here and relax the regex to accommodate the above use cases.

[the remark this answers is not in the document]
marcelwa: Alright :+1:

python/mqt/core/plugins/qiskit/provider.py
```diff
@@ -42,12 +42,58 @@ class QDMIProvider:
...
+
+        >>> provider = QDMIProvider(token="my_token")
+        >>> # or with username and password
+        >>> provider = QDMIProvider(username="user", password="pass")
     """
 
-    def __init__(self) -> None:
-        """Initialize the QDMI provider."""
+    def __init__(
+        self,
```
burgholzer, line 54:
```suggestion
        self,
        *,
```
Truly minor, but i think it could make sense to turn these into keyword-only arguments. Similar to how this is handled across other parts of this PR.

burgholzer (approved):
Except for the clang-tidy errors, this looks good to me and ready to merge. 🙌🏼

bindings/fomac/fomac.cpp
```diff
@@ -92,98 +133,135 @@ PYBIND11_MODULE(MQT_CORE_MODULE_NAME, m, py::mod_gil_not_used()) {
...
+        qdmi::Driver::get().addDynamicDeviceLibrary(libraryPath, prefix,
+                                                    config);
+      },
+      "library_path"_a, "prefix"_a, "base_url"_a = std::nullopt,
+      "token"_a = std::nullopt, "auth_file"_a = std::nullopt,
+      "auth_url"_a = std::nullopt, "username"_a = std::nullopt,
+      "password"_a = std::nullopt, "custom1"_a = std::nullopt,
+      "custom2"_a = std::nullopt, "custom3"_a = std::nullopt,
+      "custom4"_a = std::nullopt, "custom5"_a = std::nullopt);
+#endif // _WIN32
```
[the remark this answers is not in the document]
burgholzer: While this might change in the future, we do not currently add docstrings to the actual pybind11 bindings code, but rather to the corresponding stub file.
This may change if we decide to adopt nano bind and its automatic stub generation at some point.

test/fomac/test_fomac.cpp
```diff
@@ -720,14 +724,250 @@ TEST_F(SimulatorJobTest, getSparseProbabilitiesReturnsValidProbabilities) {
...
+    tmpFile << "test_token_content";
+  }
+
+  SessionConfig config2;
+  config2.authFile = tmpPath.string();
+  EXPECT_NO_THROW({ const Session session(config2); });
+
+  // Clean up
+  std::filesystem::remove(tmpPath);
+}
```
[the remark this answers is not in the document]
burgholzer: That's a risk we are willing to take here. We generally like to keep the platform specific code to a minimum.

include/mqt-core/qdmi/Driver.hpp
```diff
@@ -22,6 +23,36 @@
...
+  std::optional<std::string> custom1;
+  /// Custom configuration parameter 2
+  std::optional<std::string> custom2;
+  /// Custom configuration parameter 3
+  std::optional<std::string> custom3;
+  /// Custom configuration parameter 4
+  std::optional<std::string> custom4;
+  /// Custom configuration parameter 5
+  std::optional<std::string> custom5;
+};
```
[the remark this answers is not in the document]
burgholzer: This is a limitation by QDMI at the moment. There is no way for a QDMI device (and hence the other levels of abstraction) to communicate the semantics of supported/required combinations of parameters.
This is tracked in https://github.com/munich-quantum-toolkit/core/issues/1361

src/qdmi/Driver.cpp
```diff
@@ -349,20 +461,11 @@ Driver::~Driver() {
...
-    // Re-throw other exception
-    throw;
-  }
-  return true;
+                                     const std::string& prefix,
+                                     const DeviceSessionConfig& config)
+    -> void {
+  devices_.emplace_back(std::make_unique<QDMI_Device_impl_d>(
+      std::make_unique<DynamicDeviceLibrary>(libName, prefix), config));
 }
```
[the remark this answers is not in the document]
burgholzer: The method has been removed.

include/mqt-core/qdmi/na/Device.hpp
```diff
@@ -70,7 +70,9 @@ class Device final {
   Device();
 
   /// @brief The singleton instance.
-  static std::atomic<Device*> instance;
+  static std::atomic<Device*> instance_;
+  /// @brief Reference count for the singleton instance.
+  static std::atomic<size_t> refCount_;
```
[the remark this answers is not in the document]
burgholzer: Only threads that have called `initialize` themselves may call `finalize`. Thus, the race condition above should never occur when the library is correctly used.

include/mqt-core/qdmi/na/Device.hpp
```diff
@@ -85,14 +87,20 @@ class Device final {
...
   /**
-   * @brief Destroys the singleton instance.
-   * @details After this call, `get()` must not be called until a new
-   * `initialize()` call.
+   * @brief Decrements the reference count and destroys the singleton instance
+   * if this is the last active reference.
+   * @details This must be called exactly once per call to `initialize()`.
+   * After calling `finalize()`, `get()` must not be called until `initialize()`
+   * is called again.
    */
```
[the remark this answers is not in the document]
burgholzer: No, enough is enough.

include/mqt-core/qdmi/sc/Device.hpp
```diff
@@ -53,7 +53,9 @@ class Device final {
   Device();
 
   /// @brief The singleton instance.
-  static std::atomic<Device*> instance;
+  static std::atomic<Device*> instance_;
+  /// @brief Reference count for the singleton instance.
+  static std::atomic<size_t> refCount_;
```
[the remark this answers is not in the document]
burgholzer: No, enough is enough.

burgholzer (approved):
Alright, I looked a little bit more deeply into this the last couple of hours and pushed a bunch of commits that resulted from that investigation.
I tried to keep them as self-contained as possible and added explicit changelog items wherever I saw fit.
I am happy with this for now and hope CI will be happy as well.
Tagging this for auto-merge 🚀

src/qdmi/sc/Device.cpp
```diff
@@ -94,10 +94,16 @@
       return QDMI_SUCCESS;                                                     \
     }                                                                          \
   }
+
+#define IS_INVALID_ARGUMENT(prop, prefix)                                      \
+  ((prop) >= prefix##_MAX && (prop) != prefix##_CUSTOM1 &&                     \
+   (prop) != prefix##_CUSTOM2 && (prop) != prefix##_CUSTOM3 &&                 \
+   (prop) != prefix##_CUSTOM4 && (prop) != prefix##_CUSTOM5)
 // NOLINTEND(bugprone-macro-parentheses)
```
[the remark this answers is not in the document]
burgholzer: This is fine for now. We will upstream these macros to the QDMI repository so that they can be reused in depending projects like here without having to define them over and over again.

PR #1379, opened 2025-12-08
♻️ Apply CRTP to Unit class

MatthiasReumann:
## Description

This pull requests updates the unit implementations such that the base class `Unit` is an actual interface. For that purpose, I implemented the Curiously Recurring Template Pattern ([CRTP](https://en.cppreference.com/w/cpp/language/crtp.html)) to avoid virtual tables. This pattern is commonly used in the LLVM code base so it is also a good fit here, I think. 

I've also removed the `restore` flag from the Unit because the routing pass already knows when to restore and when not to. It didn't really serve any purpose in the unit itself. 

## Rationale

While I was writing the respective section for my thesis, I noticed that the `Unit` class is semantically an interface - but the implementation isn't. Hence, this PR. I think units could be useful elsewhere also, with potentially different strategies, so refactoring the class to an interface makes sense imo. 

## Checklist:

### DIFF
--- a/mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/NaiveRoutingPass.cpp
+++ b/mlir/lib/Dialect/MQTOpt/Transforms/Transpilation/sc/NaiveRoutingPass.cpp
@@ -141,7 +141,7 @@ private:
               .Case<ResetOp>([&](ResetOp op) { unit.layout().remap(op); })
               .Case<MeasureOp>([&](MeasureOp op) { unit.layout().remap(op); })
               .Case<scf::YieldOp>([&](scf::YieldOp op) {
-                rewriter.setInsertionPointAfter(op->getPrevNode());
+                rewriter.setInsertionPoint(op);
                 insertSWAPs(op.getLoc(), llvm::reverse(history), unit.layout(),
                             rewriter);
               });


mlir/include/mlir/Dialect/MQTOpt/Transforms/Transpilation/LayeredUnit.h
```diff
@@ -65,6 +60,14 @@ class LayeredUnit : public Unit {
 #endif
 
 private:
+  friend class Unit<LayeredUnit>;
```
burgholzer, line 63: Is this really the preferred solution of simply making these functions `protected` instead of `private`?

MatthiasReumann: Unfortunately, the only options are to use a friendship or make the `<name>Impl` methods public because `Unit` can't access protected members of the sub classes.

burgholzer: Ah. Then it's a no brainer to keep this as is. I would have hoped that `protected` solves this.
Feel free to resolve this and hit merge if you are happy 😌

PR #1412, opened 2025-12-30
⚡ Improve uv build caching by removing unconditional `reinstall-package` and configuring dedicated `cache-keys`

burgholzer:
## Description

This PR removes the unconditional `reinstall-package = ["mqt-core"]` configuration from the `uv` settings.
The purpose of this setting was to ensure that mqt-core is rebuilt, when the `bindings` directory changed, in order to avoid having a stale version of mqt-core installed.
However, this had the unintended consequence that the package was _always_ rebuilt and `uv` could never cache the built wheel.
This is extremely apparent in our CI, where this adds very negatively to our CI runtimes.
Our Python test suite runs 5 versions of Python (at the moment 3.10, 3.11, 3.12, 3.13, and 3.14) both with minimum dependencies and latest dependencies. On main, this produces 10 separate builds.

Simply dropping the setting was not an option because of the above use case.
But it turns out that `uv` is offering more fine granular controls over the caching mechanism since literally forever and we just haven't found it.
This PR, instead of the unconditional `reinstall-package`, sets `tool.uv.cache-keys` which means that the package will be rebuilt whenever
- `pyproject.toml` changes (uv does this by default), or
- any commit or tag information changes (which matches changes caused by setuptools-scm), or
- any `.cpp` file in the `bindings` folder changes.

This reduces the number of builds from 10 down to 3 (3.10, 3.11, 3.12+ ABI3), which should drastically decrease our CI times.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Changed
 
+- ⚡ Improve uv build caching by removing unconditional `reinstall-package` and configuring dedicated `cache-keys` ([#1412]) ([**@burgholzer**])
 - 👨‍💻📦 Build `spdlog` as a shared library on project installs ([#1411]) ([**@burgholzer**])
 - ♻️🏁 Remove Windows-specific restrictions for dynamic QDMI device library handling ([#1406]) ([**@burgholzer**])
 - ♻️ Migrate Python bindings from `pybind11` to `nanobind` ([#1383]) ([**@denialhaag**], [**@burgholzer**])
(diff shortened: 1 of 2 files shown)


burgholzer:
Comparison of CI runtimes:

| Runner | [Before](https://github.com/munich-quantum-toolkit/core/actions/runs/20606865974) | [After](https://github.com/munich-quantum-toolkit/core/actions/runs/20608008733?pr=1412) |
|---|---|---|
| ubuntu-24.04 | 10m30s | 8m32s |
| ubuntu-24.04-arm | 9m2s | 7m14s |
| macos-15-intel | 22m16s | 10m34s |
| macos-14 | 7m13s | 3m48s |
| windows-2022 | 43m4s | 16m1s |

While the speedup is not so drastic on Linux systems, it is impressive for macOS and Windows.
Both macOS jobs are now roughly twice as fast. The Windows job is almost three times as fast.
Overall, this reduces the critical path of our CI (Windows Python testing) from 43min down to 16min. ⚡

PR #1426, opened 2026-01-07
:sparkles: Add unitary matrix to QCO dialect

taminob:
## Description

This PR implements the feature of the unitary matrix definition for `UnitaryOpInterface` which has been removed from the original dialect rewrite PR in favor of this follow-up.

It aims to allow access to the matrix definition of the standard gates via a fast, stack-allocated matrix returned by the function `getUnitaryMatrixDefinition()` in the trait.
A dynamic, heap-based matrix (size `Eigen::Dynamic`) is returned by the `UnitaryOpInterface`. This requires a copy at the moment and further evaluations are required if all matrices should be moved to the heap or if there is a way to make use of the draft function `getFastUnitaryMatrix()` which uses a template parameter, but is currently also only accessible via a trait and thus needs a cast to the correct type first.

The changes require the newly introduced `Eigen` library for its linear algebra type.
For now, it uses the unsupported kronecker product feature which needs to be discussed before merging. "Unsupported" features in Eigen are community-contributed and are not an indication of how mature the code is. However, the maintainers do not provide any API guarantees.

## Checklist:

### DIFF
--- a/mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
+++ b/mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
@@ -50,8 +50,6 @@
 // Interfaces
 //===----------------------------------------------------------------------===//
 
-#include "mlir/Dialect/QCO/IR/QCOInterfaces.h.inc" // IWYU pragma: export
-
 namespace mlir::qco {
 
 /**
@@ -135,6 +133,8 @@ public:
 };
 
 } // namespace mlir::qco
+
+#include "mlir/Dialect/QCO/IR/QCOInterfaces.h.inc" // IWYU pragma: export
 
 //===----------------------------------------------------------------------===//
 // Operations


CMakeLists.txt
```diff
@@ -118,6 +119,18 @@ if(APPLE AND CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
...
+  FetchContent_Declare(
+    Eigen
+    GIT_REPOSITORY https://gitlab.com/libeigen/eigen.git
+    GIT_TAG 3.4.1
+    GIT_SHALLOW TRUE)
+  FetchContent_MakeAvailable(Eigen)
+  if(WIN32 AND ${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL ARM64)
+    message(STATUS "Enabling non-optimal vectorization in Eigen to avoid alignment issues")
+    add_compile_definitions(EIGEN_DONT_ALIGN_STATICALLY)
+  endif()
```
burgholzer: This should be moved to the `cmake/ExternalDependencies` file like all other external dependencies.

Also: the latest release of Eigen is 5.0.1 which is two major versions newer than what is being used here. Please use the latest version.
Also try to follow the remaining external dependencies and add a commit SHA instead of a tag; with the appropriate comment.

Lastly: do we really need the `EIGEN_DONT_ALIGN_STATICALLY` here? could you remove it temporarily so that we see the CI failure (if any).

mlir/unittests/Dialect/QCO/IR/CMakeLists.txt
burgholzer, line 1: would be good to follow the test structure that Damian established in #1436

mlir/unittests/dialect/mqtopt/CMakeLists.txt
burgholzer, line 1: Must be a leftover from some kind of previous merge. `MQTOpt` no longer exists and neither does the wire iterator.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+#include <llvm/Support/ErrorHandling.h>
+#include <mlir/IR/BuiltinTypes.h>
+#include <mlir/IR/Region.h>
+#include <numbers>
+
+namespace mlir::utils {
+
+inline Eigen::Matrix<std::complex<double>, 1, 1> getMatrixGPhase(double theta) {
+  using namespace std::complex_literals;
+  return Eigen::Matrix<std::complex<double>, 1, 1>{std::exp(1i * theta)};
```
burgholzer:
```suggestion
  return {std::exp(1i * theta)};
```
This should "just work" and safe a lot of the boilerplate code.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+}
+
+inline Eigen::Matrix2cd getMatrixId() {
+  using namespace std::complex_literals;
+  return Eigen::Matrix2cd{{1.0 + 0i, 0.0 + 0i},  // row 0
+                          {0.0 + 0i, 1.0 + 0i}}; // row 1
+}
```
burgholzer: any matrix that does not have parameters should be statically defined. There is absolutely no reason to redefine these matrices on every call.
something like

```suggestion
inline Eigen::Matrix2cd getMatrixId() {
  using namespace std::complex_literals;
  static auto mat = Eigen::Matrix2cd{{1.0 + 0i, 0.0 + 0i},  // row 0
                          {0.0 + 0i, 1.0 + 0i}}; // row 1
  return mat;                        
}
```
should do and avoid repeated allocations entirely.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+  using namespace std::complex_literals;
+  return Eigen::Matrix2cd{{1.0 + 0i, 0.0 + 0i},  // row 0
+                          {0.0 + 0i, 0.0 - 1i}}; // row 1
+}
+
+inline Eigen::Matrix2cd getMatrixT() {
+  using namespace std::complex_literals;
+  const std::complex<double> m00 = 1.0 + 0i;
+  const std::complex<double> m01 = 0.0 + 0i;
+  const std::complex<double> m11 = std::exp(1i * std::numbers::pi / 4.0);
```
burgholzer: Expressions like these can be replaced with

```suggestion
  const std::complex<double> m11 = std::polar(1., std::numbers::pi / 4);
```
which is ought to be faster than `exp`.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+
+inline Eigen::Matrix2cd getMatrixZ() {
+  using namespace std::complex_literals;
+  return Eigen::Matrix2cd{{1.0 + 0i, 0.0 + 0i},   // row 0
+                          {0.0 + 0i, -1.0 + 0i}}; // row 1
+}
+
+inline Eigen::Matrix2cd getMatrixH() {
+  using namespace std::complex_literals;
+  const std::complex<double> m00 = 1.0 / std::sqrt(2) + 0i;
```
burgholzer: There is `std::numbers::sqrt2` and I would recommend using it.

burgholzer: Most of the computations of matrix elements here can actually be defined as `constexpr` so that they are definitely evaluated at compile time.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+
+inline Eigen::Matrix<std::complex<double>, 1, 1> getMatrixGPhase(double theta) {
+  using namespace std::complex_literals;
+  return Eigen::Matrix<std::complex<double>, 1, 1>{std::exp(1i * theta)};
+}
+
+inline Eigen::Matrix2cd getMatrixId() {
+  using namespace std::complex_literals;
+  return Eigen::Matrix2cd{{1.0 + 0i, 0.0 + 0i},  // row 0
+                          {0.0 + 0i, 1.0 + 0i}}; // row 1
```
burgholzer: I think it's fine to drop redundant zeros, such as `+0i` or `0.0`.
So `0.0+0i` can simply be `0`.
For `x!=0`, `x + 0i` can simply be `x` and `0.0 + x * 1i` can simply be `x * 1i`.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+}
+
+inline Eigen::Matrix2cd getMatrixR(double theta, double phi) {
+  using namespace std::complex_literals;
+  const std::complex<double> m00 = std::cos(theta / 2.0) + 0i;
+  const std::complex<double> m01 =
+      -1i * std::exp(-1i * phi) * std::sin(theta / 2.0);
+  const std::complex<double> m10 =
+      -1i * std::exp(1i * phi) * std::sin(theta / 2.0);
+  const std::complex<double> m11 = std::cos(theta / 2.0) + 0i;
```
burgholzer: This is just m00 again.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+                          {m0, mc, -ms, m0}, // row 1
+                          {m0, -ms, mc, m0}, // row 2
+                          {ms, m0, m0, mc}}; // row 3
+}
+
+inline Eigen::Matrix4cd getMatrixRZX(double theta) {
+  using namespace std::complex_literals;
+  const std::complex<double> m0 = 0.0 + 0i;
+  const std::complex<double> mc = std::cos(theta / 2.0) + 0i;
+  const std::complex<double> ms = -1i * std::sin(theta / 2.0);
```
burgholzer:
```suggestion
  const std::complex<double> ms = 1i * std::sin(theta / 2.0);
```
this should be the correct way

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+}
+
+inline Eigen::MatrixXcd getMatrixCtrl(size_t numControls,
+                                      Eigen::MatrixXcd targetMatrix) {
+  // get dimensions of target matrix
+  const auto targetDim = targetMatrix.cols();
+  assert(targetMatrix.cols() == targetMatrix.rows());
+
+  // define dimensions and type of output matrix
+  const auto dim = static_cast<int64_t>((1 << numControls) * targetDim);
```
burgholzer:
```suggestion
  const auto dim = static_cast<int64_t>((1ULL << numControls) * targetDim);
```
Avoids a warning about an implicit cast on Windows

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
```diff
@@ -0,0 +1,290 @@
...
+
+  // initialize result with identity
+  Eigen::MatrixXcd matrix = Eigen::MatrixXcd::Identity(dim, dim);
+
+  // Apply target matrix
+  matrix.bottomRightCorner(targetDim, targetDim) = targetMatrix;
+
+  return matrix;
+}
```
burgholzer: That's a fairly convenient way to define the matrix 👍🏼 

Should this function be specialized to also accept 1*1, 2*2, and 4*4 matrices and return the respective 2*2, 4*4, or 8*8 matrix instead of the generic `Eigen::MatrixXcd` type?

I am not quite sure how much of Eigen is `constexpr` capable, but technically the matrix of `Ctrl(X)` should be computable at compile time. Just a thought.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
```diff
@@ -28,6 +28,7 @@ def UnitaryOpInterface : OpInterface<"UnitaryOpInterface"> {
 
     let cppNamespace = "::mlir::qco";
 
+    // TODO: fix const correctness?
```
burgholzer: Const correctness is not that much of a thing in MLIR, see https://mlir.llvm.org/docs/Rationale/UsageOfConst/
So I would probably also remove the `const` from the extra class method here.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
```diff
@@ -107,7 +108,42 @@ def UnitaryOpInterface : OpInterface<"UnitaryOpInterface"> {
...
+
+    let extraClassDeclaration = [{
+      template<typename MatrixType>
+      MatrixType getFastUnitaryMatrix() const {
+        // removing const since it's safe to say UnitaryOpInterface::getFastUnitaryMatrixVariant()
+        // and OpType::getUnitaryMatrixDefinition() will not modify the state of this
+        auto&& matrix = const_cast<UnitaryOpInterface*>(this)->getFastUnitaryMatrixVariant();
+        return std::get<MatrixType>(matrix);
+      }
+    }];
```
burgholzer: Do we really need both versions here?
Can't we simply go with the variant one and return `Eigen::MatrixXcd` in the general case (such as for CtrlOp)?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -238,11 +238,18 @@ def ResetOp : QCOOp<"reset", [Idempotent, SameOperandsAndResultType]> {
 // Traits
 //===----------------------------------------------------------------------===//
 
-class TargetAndParameterArityTrait<int T, int P>
-        : ParamNativeOpTrait<"TargetAndParameterArityTrait", !strconcat(!cast<string>(T), ",", !cast<string>(P))> {
+class MatrixDefinitionLambda<code MatrixSize, code MatrixDefinitionBody> {
+  code MatrixType = "Eigen::Matrix<std::complex<double>, " # MatrixSize # ", " # MatrixSize # ">";
+  code Code = !cond(!empty(MatrixDefinitionBody): "nullptr", true: "[](UnitaryOpInterface op) -> " # MatrixType # " { " # MatrixDefinitionBody # " }");
+}
+class FixedSizeMatrixDefinitionLambda<int T, code MatrixDefinitionBody> : MatrixDefinitionLambda<"1 << " # !cast<string>(T), MatrixDefinitionBody>;
```
```suggestion
class FixedSizeMatrixDefinitionLambda<size_t T, code MatrixDefinitionBody> : MatrixDefinitionLambda<"1 << " # !cast<string>(T), MatrixDefinitionBody>;
```
to be consistent with most of the other code.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -252,11 +259,22 @@ def TwoTargetZeroParameter : TargetAndParameterArityTrait<2, 0>;
...
+class TwoTargetZeroParameterUnitaryMatrix<code MatrixDefinitionBody> : TargetAndParameterArityTrait<2, 0, FixedSizeMatrixDefinitionLambda<2, MatrixDefinitionBody>>;
+class TwoTargetOneParameterUnitaryMatrix<code MatrixDefinitionBody> : TargetAndParameterArityTrait<2, 1, FixedSizeMatrixDefinitionLambda<2, MatrixDefinitionBody>>;
+class TwoTargetTwoParameterUnitaryMatrix<code MatrixDefinitionBody> : TargetAndParameterArityTrait<2, 2, FixedSizeMatrixDefinitionLambda<2, MatrixDefinitionBody>>;
+
 //===----------------------------------------------------------------------===//
 // Unitary Operations
 //===----------------------------------------------------------------------===//
 
-def GPhaseOp : QCOOp<"gphase", traits = [UnitaryOpInterface, ZeroTargetOneParameter, MemoryEffects<[MemWrite]>]> {
+def GPhaseOp : QCOOp<"gphase", traits = [UnitaryOpInterface, ZeroTargetOneParameterUnitaryMatrix<[{ return utils::getMatrixGPhase(tryGetParameterAsDouble(op, 0).value()); }]>, MemoryEffects<[MemWrite]>]> {
```
burgholzer: Don't we have a UnitaryInterface function for `tryGetParameterAsDouble(op, 0).value()`?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -238,11 +238,18 @@ def ResetOp : QCOOp<"reset", [Idempotent, SameOperandsAndResultType]> {
...
-class TargetAndParameterArityTrait<int T, int P>
-        : ParamNativeOpTrait<"TargetAndParameterArityTrait", !strconcat(!cast<string>(T), ",", !cast<string>(P))> {
+class MatrixDefinitionLambda<code MatrixSize, code MatrixDefinitionBody> {
+  code MatrixType = "Eigen::Matrix<std::complex<double>, " # MatrixSize # ", " # MatrixSize # ">";
+  code Code = !cond(!empty(MatrixDefinitionBody): "nullptr", true: "[](UnitaryOpInterface op) -> " # MatrixType # " { " # MatrixDefinitionBody # " }");
+}
+class FixedSizeMatrixDefinitionLambda<int T, code MatrixDefinitionBody> : MatrixDefinitionLambda<"1 << " # !cast<string>(T), MatrixDefinitionBody>;
+
+class TargetAndParameterArityTrait<int T, int P, MatrixDefinitionLambda MatrixDefinition = FixedSizeMatrixDefinitionLambda<T, [{}]>>
+        : ParamNativeOpTrait<"TargetAndParameterArityTrait", !cast<string>(T) # ", " # !cast<string>(P) # ", " # MatrixDefinition.MatrixType # ", " # MatrixDefinition.Code> {
```
burgholzer: This is picking up on a comment from the MatrixUtils.h file: Maybe we could be doing something differently here for gates without parameters, where the matrix is essentially constexpr and only needs to be constructed and stored once.
Do you think some kind of optimization along those lines would be possible?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -252,11 +259,22 @@ def TwoTargetZeroParameter : TargetAndParameterArityTrait<2, 0>;
 def TwoTargetOneParameter : TargetAndParameterArityTrait<2, 1>;
 def TwoTargetTwoParameter : TargetAndParameterArityTrait<2, 2>;
 
+class DynamicTargetZeroParameterUnitaryMatrix<code MatrixDefinitionBody> : TargetAndParameterArityTrait<0, 0, MatrixDefinitionLambda<"Eigen::Dynamic", MatrixDefinitionBody>>;
```
burgholzer: I think I would personally drop all the `Matrix` suffixes from the names of these trait classes. Makes the names more compact and still carries the same meaning.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
burgholzer, line 1: This file should be included in the include file sets of the targets that use it in CMake.
Or it should be paired with a `.cpp` file and built into a proper target that is not header-only.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -51,6 +56,13 @@
 
 namespace mlir::qco {
 
+/**
+ * @brief Retrieve C++ type of static mlir::Value.
+ * @details The returned float attribute can be used to get the value of the
+ *          given parameter as a C++ type.
+ */
+[[nodiscard]] inline std::optional<double>
+tryGetParameterAsDouble(UnitaryOpInterface op, size_t i);
```
burgholzer: This docstring does not match the function.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -60,8 +72,15 @@ namespace mlir::qco {
...
  * @tparam P The parameter arity.
+ * @tparam MatrixDefinition A function returning the matrix definition of the
+ *                          operation. The operation will be provided as the
+ *                          only argument of the function. If the operation does
+ *                          not have a matrix definition, set this value to
+ *                          nullptr.
  */
-template <size_t T, size_t P> class TargetAndParameterArityTrait {
+template <size_t T, size_t P, typename UnitaryMatrixType,
+          UnitaryMatrixType (*MatrixDefinition)(UnitaryOpInterface)>
```
burgholzer:
```
          UnitaryMatrixType (*MatrixDefinition)(UnitaryOpInterface)>
```
could this be a `llvm::function_ref`?

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -116,6 +140,13 @@ template <size_t T, size_t P> class TargetAndParameterArityTrait {
       return dyn_cast<FloatAttr>(constantOp.getValue());
     }
 
+    [[nodiscard]] UnitaryMatrixType getUnitaryMatrixDefinition() const
+      requires(MatrixDefinition != nullptr)
+    {
+      const auto* op = this->getConstOperation();
+      return MatrixDefinition(llvm::dyn_cast<UnitaryOpInterface>(op));
```
burgholzer: Based on https://mlir.llvm.org/docs/Rationale/UsageOfConst/, I would not bother too much with `const` and the respective casting.
Makes the code here a bit simpler.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -136,16 +167,74 @@ template <size_t T, size_t P> class TargetAndParameterArityTrait {
...
+      // operator provides a const overload, getOperation() does not
+      return *concrete;
+    }
   };
 };
 
 } // namespace mlir::qco
 
-#include "mlir/Dialect/QCO/IR/QCOInterfaces.h.inc" // IWYU pragma: export
+// #include "mlir/Dialect/QCO/IR/QCOInterfaces.h.inc" // IWYU pragma: export
```
burgholzer: If this is not needed here, it might as well be removed.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -136,16 +167,74 @@ template <size_t T, size_t P> class TargetAndParameterArityTrait {
...
+  using DummyArityType =
+      TargetAndParameterArityTrait<0, 0, Eigen::MatrixXcd, nullptr>;
+  const auto param = op.getParameter(i);
+  const auto floatAttr =
+      DummyArityType::Impl<arith::ConstantOp>::getStaticParameter(param);
+  if (!floatAttr) {
+    return std::nullopt;
+  }
+  return floatAttr.getValueAsDouble();
+}
```
burgholzer: I noted this elsewhere already, but I really thought we had that functionality somewhere already. I haven't checked in detail, but this seems overly complex to me.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: Actually, Eigen marks its constructors as `explicit` which makes the explicit constructor call necessary :)

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: Not sure how much this will actually improve the performance since a copy is made anyway on return. Might be slightly more efficient than the current in-place construction (via copy-elision), but I don't think it's a significant improvement.
Changing the return value to a reference could be possible for these static matrices, but that won't work for the trait or the interface and is thus probably worthless.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: I used that before, I think there was a CI that didn't support the numbers library yet. But I can try again - the use of a constant instead is definitely a good idea

taminob: This entire file was actually just compied from the dialect rewrite PR - I'll look into improving performance here :)

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: I'll look into the more recent version you linked (although it's not actually two major versions, just one minor version - they changed their versioning scheme. Release 5.0 is basically 3.5). The version that is currently used, does not allow for a `Eigen::Matrix` to be `constexpr`, unfortunately.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
taminob: Didn't you want a stack-only version for improved performance? The variant-version is basically just a helper and in my opinion less convenient than directly accessing the correct type via this `getFastUnitaryMatrix()` template function.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: This is tablegen and not C++, so I don't think this change is possible since size_t isn't supported by tablegen

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: Do you mean `UnitaryOpInterface::getParameter(int i)`? Because that will return a `mlir::Value` which needs to be put into `getStaticParameter` (currently a static method in the arity trait class) - that is also what this function will do internally.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: It should be possible to have a different return type for the `MatrixDefinition` depending on if the matrix has parameters or not. However, I'm not sure if there is any actual benefit worth the additional complexity, considering that the matrices will probably almost always be accessed via `UnitaryOpInterface` anyway, right? Do we have a use-case where a static matrix is accessed on its actual op type and it is a static matrix? Because then one could also simply use the functions in `MatrixUtils.h` directly...

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: Are you sure that it is still well-understood if it is simply called `DefinitionBody`? Without the context of this PR, I probably wouldn't know what that is supposed to be if I read about this in the code...

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: I'll check it out, but it can't be a `std::function` because it is a template parameter

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: Ah, thanks for the link - I read in the MLIR documentation somewhere that the `const` should be generated if applicable. It's a bit annoying, because e.g. `FuncOp::getOps<OpType>` returned a `const mlir::Operation*` in a test case for me, requiring a workaround via `FuncOp::getOperations()`.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: `getStaticParameter` exists twice (once for `QC` and once here in this file) as part of the trait. This here is done multiple times in the different StandardGates source files for the rewrite patterns ("MergeSubsequentXXX"), but I wanted a function directly doing it without depending on the trait for the tablegen code which does not have simple access to its own trait.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
burgholzer: Is copy-elision guaranteed here? If it is, then there might not be much of an advantage indeed. This could be something to very quickly confirm in CompilerExplorer (https://godbolt.org/) before opting for one or the other solution.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
burgholzer: I see. I suppose you could at least remove the template arguments in the return call though to not repeat it twice. 

Small side note: please do not resolve conversations yourself prematurely. I would have overlooked this, if I had not checked the resolved comments as well 😌

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
burgholzer: Should be available everywhere since we are using C++20. we might need to bump our minimum OSX deployment version. That's the only thing where I could imagine some problems.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
burgholzer: Yeah. All I am trying to say here is that I'd prefer a single solution and not two. and at that point, one can drop the "fast" from the name.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: fair enough. does tablegen support `uint`?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: Yeah, probably.
In any case, the code pattern here of querying an optional but then unconditionally dereferencing it is not ideal. This should either contain a check for the availability or not even use an optional for quicker computation.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: > considering that the matrices will probably almost always be accessed via UnitaryOpInterface anyway, right?

I am not so sure about that. Assuming that you are writing a single-qubit gate fusion/synthesis pass, you might actually check for the single-target trait (or something similar) and call specialized methods on it.
Similarly for the two-qubit synthesis, I would expect that one would rather call specialized 1- and 2-qubit methods than the generic methods once it has found 1- and 2-qubit gates.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: I was not clear enough. I was only talking about `DynamicTargetZeroParameterUnitaryMatrix` -> `DynamicTargetZeroParameterUnitary`

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: I think I might have encountered similar needs for workarounds due to `const`. The better solution in almost all those cases was to remove the `const`, which seems to fit well with the general recommendation from the link.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
taminob: I thought that it may be more convenient to also have an easy-to-use method. But then I'll remove the non-template function and will simply use its definition if `Eigen::MatrixXcd` is given as a template argument in the "fast" version.

Do you mind the "variant" method which should actually never be needed by a user? Unfortunately, MLIR tablegen does not allow for private/protected methods and introducing a "private" interface which is added to every operation seems a bit bloaty/brittle for one function.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: The pattern with `DummyArityType` still very much seems like a hack to me and not an ideal solution. There has to be a cleaner way to achieve this.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: No, just `bits<N>` for fixed-width integers

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
burgholzer: Can't you simply put `protected`/`private` into the `let extraClassDeclaration = [{...}]` block?
I am pretty sure that this allows you to define arbitrary C++ code.

Alternatively, can't you simply inline the definition of the variant function? Or would you expect to use it somewhere else?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: I'm not sure yet about error handling in these functions - so far, `llvm::reportFatalUsageError()` is used (in existing code and this PR), but I think this will just abort the entire compilation process which might not be what one wants to do, just because e.g. a qubit index is out-of-bounds or a parameter is not defined as a constant.
`std::optional::value()` will throw an exception if it does not have any value (in constrast to the dereference operator) which should be rare enough, but at least can be handled by the user. Otherwise, we can just wrap every matrix into an `std::optional`.

Do you have any preference on how to approach this issue?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: Let's keep using `int` then. Sorry for the noise.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: I'd probably still just use the `UnitaryOpInterface` with the template method for accessing the 1/2 qubit methods - also, if one wants to actually access it via the trait, one needs to know beforehand if it has parameters or not since the return value will change the trait class (different template parameter). So I still don't understand the use-case, or did I misunderstand you?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: Ah, that's reasonable - I'll rename it :+1:

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: Doesn't seem to be possible - the allowed types for a template parameter are pretty restrictive

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: LLVM itself does not generally use exceptions. One even has to enable the respective flag manually. So I would rather not use exceptions and rely on them in our MLIR code.
I am actually striving for ways to turn them off entirely for the MLIR part as that would result in faster code generally.

`llvm::reportFatalUsageError()` just aborts, yes.
If a qubit index is out of bounds, I'd argue that this is justified and in most cases will point to an error that cannot be recovered from at runtime.
I would argue that the main usage pattern should be to first use a UnitaryInterface function to query whether the operation has a statically known matrix, and, if so, query the matrix unconditionally (without `std::optional`). Something like that is in the Dialect Redesign RFC if I remember correctly.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: If I didn't miss anything, it should be guaranteed since C++17 (see https://cppreference.net/cpp/language/copy_elision.html at "Prvalue semantics"). The only advantage of your suggestion is that it would skip the repeated construction of the complex numbers - however, I'm not sure if that'll actually cause any measurable difference.

Godbolt does confirm copy-elision as far as I see, but I can't tell you which of the two versions is more performant... My guess would actually be the one without the static variable because no thread-safety checks are required: https://godbolt.org/z/b1bjhaWcE

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: Maybe what I am asking for is entirely covered by the template method already.
I guess we will see as this PR progresses.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: 👍🏼 just searched a little bit on the internet and it seems that the template parameter (like it is) is the preferred solution for this anyway.
Maybe briefly looking at https://iifx.dev/en/articles/457550510/boosting-c-performance-from-std-function-to-template-parameters could be helpful. The `Callable&& callback` pattern paired with `std::forward` seems like a reasonable approach.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: The clean way would be to have it in a separate "utils" file (already exists) and re-use the same code for `QC` and `QCO`. Then I could directly access the utils without the need for a concrete type - for backwards-compatibility and to keep this PR small, I'll keep the `getStaticParameter` functions in the `QC` and `QCO` trait definitions here

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
taminob: I tried a lot of hacks and workarounds here, unfortunately, it is impossible to access the conrete type in the `extraClassDeclaration` section - and I need the conrete type to access the definition somehow.
Also, it is impossible to have a template function as a interface method (I tried...) because they are stored as function pointers in models which does not work with templates. Some inheritance and `if`s with dynamic casts etc. could work, but that would also ruin performance.

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
burgholzer: Thanks for the link(s).
Looked at the godbolt code a little bit. Also let it run with clang: https://godbolt.org/z/hadK9Pn6r
The clang output seems to suggest the direct construction leads to shorter code. But I could also misinterpret things here.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: > I would argue that the main usage pattern should be to first use a UnitaryInterface function to query whether the operation has a statically known matrix, and, if so, query the matrix unconditionally (without `std::optional`). Something like that is in the Dialect Redesign RFC if I remember correctly.

Interface-wise that seems like a fast solution, but I think that implementation-wise it might actually be slower than always returning an `optional` because for the "hasStaticMatrix" we'd first need to retrieve the static parameter and then do it again for the matrix definition (involving `getDefiningOp()` and `dyn_cast` which are probably more expensive than an additional bool in the optional).

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
burgholzer: Hm. I might have to look at this myself at some point and play with the code.
I feel this has to be possible somehow.
In the meantime, let's just keep the interface method that no one is expected to use around.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: Backwards-compatibility is not really a concern here to be honest since none of the new features are released.
Might it make sense to implement your suggestion in a separate, small PR; get that merged; then use it here?

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: I think your link is about something completely different. In this PR, a new template instance is actually created for each function because not the "callable" type is part of the template parameters but the "callable" itself (function pointer). So basically, each operation has its own, custom trait.

In an earlier version, I tried to avoid this by having a separate trait `UnitaryMatrixTrait`, but I wasn't able to make it work with two native traits in a single tablegen `def` and I wanted to avoid having to specify for each operation the number of targets twice. Or do you think that's the lesser evil?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: That is a fair point 👍🏼 let's stick with the optional approach then for now.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: If you agree with moving it out of the trait classes, I can create a new PR - I just thought that you mentioned somewhere that you don't like free utility functions so I wanted to avoid introducing them on my own if not necessary.

burgholzer: I am not fully opposed against utility functions. Especially if they safe code duplication all around the library.
I just dislike putting various functions into a single "utils.hpp" without any separation of concerns. I'd much rather have a "Support"/"Utils" folder that contains multiple headers. I think LLVM typically goes with "Support".

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: Yeah, you are right.
I think the current solution will unnecessarily bloat the binary size due to the many template instantiations, which are also not good for compilation time.
While specifying the number of qubits twice is not particularly pretty, I think that would be a reasonable trade-off.

Different thought: each operation has its own CPP file by now. Couldn't the definitions of these functions for getting the matrix simply be part of that CPP file? Then we wouldn't need the matrix utility file at all and each matrix definition would be "in the right place". These do not really _need_ to be part of any kind of trait, right?

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: > I see. I suppose you could at least remove the template arguments in the return call though to not repeat it twice.

Eigen is unable to perform template deduction for the 1x1 matrix, so I think we're stuck with this. I will change the return type to `auto` to avoid the duplication (which should be fine for this short function in my opinion).

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
taminob: If you find some way, that would be awesome - but I did some digging in the LLVM and generated code and couldn't come up with a way without a helper method (either in that interface or a second interface that hides these implementation details, although I didn't explicitly try out that approach).

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
taminob: Can this be resolved?

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: Since we have the concrete instance of the operation in the interface, we can also simply add the method declaration in `extraClassDeclaration` and then add the implementation in a corresponding source file (which seems to already exist for most (maybe all) operations). Would simplify the code generation by a lot, but would also require touching multiple files for adding new operations (although we'll probably rarely add new gates, right?). Should I rewrite the PR to use that approach?
It's definitely easier to understand than the current solution, although it's less cool and doesn't use the capabilities of tablegen as nicely :)

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
burgholzer: Unfortunate, but I kind of like the solution with `auto` to avoid the duplication 👍🏼 
resolving here.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: I think we have source code files for every single operation already, so this feels like an easy choice.
I get your point about the nice capabilities of tablegen ;o)
I don't think that having to touch multiple files isn't too bad. As you said, we'll rarely add new gates anyway.
And the whole purpose of the dedicated Operation files is to place code that is related to the individual operations there.
Overall, this seems like a good way forward.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: Done. I moved it further up in order to be able to use the interface in the trait (although that's no longer required, I still think it might be a good idea).

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: For me, using loops instead of the `->` operator on the iterators actually resolved the issue (although I'm not sure why)

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: I implemented a draft for this, do you want to check if that's better? I guess it definitely is cleaner than the previous approach and also offers more flexibility (only wrap in `std::optional` if it is not a static matrix).

Do we also need this feature for the `QC` dialect? Because then it could actually be benefitial to have the `MatrixUtils.h` file (although I don't want to move it back again for each operation :( )

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
taminob: The function has been removed

CMakeLists.txt
taminob: Done, let's see if the Windows ARM CI fails

mlir/unittests/Dialect/QCO/IR/CMakeLists.txt
taminob: Done

mlir/unittests/dialect/mqtopt/CMakeLists.txt
taminob: Removed

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: I use `std::polar` now whenever possible (with some help by ChatGPT), but please feel free to also check if I missed any mistake in the translation

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: While moving the definitions to the separate source files, I also re-structured the functions for (hopefully) the best performance :+1:

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: Done, I hope I didn't miss any occurrence

mlir/include/mlir/Dialect/Utils/MatrixUtils.h
taminob: Fixed in `ROp.cpp`

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
burgholzer: I'll check.
I don't think we particularly need the matrix features for the `QC` dialect for now. It's fine to keep them isolated to the QCO dialect for now.

burgholzer: Okay. Just briefly looked through it and this seems to look very clean.
I need to play with the code a little bit once the PR is in a pretty much final state, but this is definitely going in the right direction and the code rabbit comments should point you in the right direction already.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -363,3 +366,29 @@ void CtrlOp::getCanonicalizationPatterns(RewritePatternSet& results,
...
+    return std::nullopt;
+  }
+
+  // get dimensions of target matrix
+  const auto targetDim = targetMatrix->cols();
+  assert(targetMatrix->cols() == targetMatrix->rows());
+
+  // define dimensions and type of output matrix
+  const auto dim = static_cast<int64_t>((1ULL << getNumControls()) * targetDim);
+
```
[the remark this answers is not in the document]
taminob: I'll add an `assert(getNumControls() < sizeof(unsigned long long) * 8)` with the next push

cmake/ExternalDependencies.cmake
```diff
@@ -22,6 +22,15 @@ if(BUILD_MQT_CORE_BINDINGS)
...
 endif()
 
+FetchContent_Declare(
+  Eigen
+  GIT_REPOSITORY https://gitlab.com/libeigen/eigen.git
+  GIT_TAG 5.0.1
+  GIT_SHALLOW TRUE
+  FIND_PACKAGE_ARGS 3.4.0 # minimum version for locally installed package
+)
+list(APPEND FETCH_PACKAGES Eigen)
```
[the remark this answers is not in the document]
taminob: @burgholzer do you have any thoughts on this? I think it should be fine without explicitly documenting this (since we'll probably just forget about updating the comment) and we can increase the minimum version when required. I think many distros will probably still have Eigen 3.4 in their repositories.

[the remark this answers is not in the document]
burgholzer: Eigen is fetched anyway if it is not found in a suitable version. We typically use rather tight lower bounds for our C++ dependencies. I'd bump this to the version we are actually using.
In contrast to Python, we have no CI testing with the lower bounds of our C++ dependencies.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -21,13 +21,15 @@
 #pragma clang diagnostic pop
 #endif
 
+#include <Eigen/Core>
```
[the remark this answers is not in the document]
burgholzer: The team is aligned and we are fine with using this part of Eigen.

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/U2Op.cpp
```diff
@@ -37,7 +37,7 @@ struct ReplaceU2WithH final : OpRewritePattern<U2Op> {
     const auto phi = valueToDouble(op.getPhi());
     const auto lambda = valueToDouble(op.getLambda());
     if (!phi || std::abs(*phi) > TOLERANCE || !lambda ||
-        std::abs(*lambda - std::numbers::pi) > TOLERANCE) {
+        std::abs(*lambda - llvm::numbers::pi) > TOLERANCE) {
```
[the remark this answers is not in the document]
burgholzer: @coderabbitai please check the current state of the PR again and see whether it is still inconsistent.

mlir/unittests/Dialect/QCO/IR/test_unitary_op_interface.cpp
```diff
@@ -0,0 +1,120 @@
...
+  [[nodiscard]] static std::string toString(ModuleOp moduleOp) {
+    std::string buffer;
+    llvm::raw_string_ostream serializeStream{buffer};
+    moduleOp->print(serializeStream);
+    return serializeStream.str();
+  }
+
+private:
+  std::unique_ptr<MLIRContext> context;
+};
```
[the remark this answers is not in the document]
burgholzer: The tests themselves could probably be improved quite substantially. However, this is out of scope for the current PR supposedly and the functionality will be easier to cover based on the unitary synthesis tests.

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/POp.cpp
```diff
@@ -12,11 +12,14 @@
 #include "mlir/Dialect/QCO/QCOUtils.h"
 #include "mlir/Dialect/Utils/Utils.h"
 
+#include <Eigen/Core>
+#include <complex>
```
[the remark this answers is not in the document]
burgholzer: @coderabbitai please check again whether this is still inconsistent.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -128,16 +144,30 @@ template <size_t T, size_t P> class TargetAndParameterArityTrait {
...
 //===----------------------------------------------------------------------===//
 
 #define GET_OP_CLASSES
 #include "mlir/Dialect/QCO/IR/QCOOps.h.inc" // IWYU pragma: export
+
+namespace mlir::qco {
+
+[[nodiscard]] inline UnitaryOpInterface getControlledOp(UnitaryOpInterface op) {
+  return llvm::cast<CtrlOp>(op).getBodyUnitary();
+}
```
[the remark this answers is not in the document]
burgholzer: @coderabbitai please check if this is still up to date

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/U2Op.cpp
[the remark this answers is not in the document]
burgholzer: This is from an outdated diff, so we should be all good based on your analysis.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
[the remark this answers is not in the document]
burgholzer: @taminob maybe you can judge better if @coderabbitai is right with its assessment above. If so, can the function be removed?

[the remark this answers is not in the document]
taminob: This function does not exist anymore in the latest diff, I already removed it because it was unused :)

burgholzer (approved):
Okay, I think this should also be practically merge ready.
The main implementation now looks rather clean.
To be fair, the tests are fairly arbitrary and I wish they were a little more rigorous (e.g., by comparing the created matrices against the ones we create from our DD package in https://github.com/munich-quantum-toolkit/core/blob/main/src/dd/GateMatrixDefinitions.cpp), but I suspect this is too much of an ask here.
That would be a nice follow-up/replacement though. @denialhaag maybe you can take a look at that in a subsequent PR.

taminob:
Oops, looks like there is no changelog entry for this PR - I can add it in my decomposition PR :)

PR #1430, opened 2026-01-08
:bug: Rename CtrlOp region to `region` to avoid hiding trait method `getBody()`

taminob:
## Description

Unfortunately, this became a little bit more cumbersome to figure out than expected.
The automatically generated getter for the parameter "body" was hiding one of the trait interface methods. These methods do have a slightly different signature (`getBody()` returns a `Block*` and `getBodyRegion()` accepts an index parameter) which can lead to confusion and an inconsistent API.

An alternative to this proposal could be simply the name "region" because `getRegion()` of the `OneRegion` trait has the same signature and thus the override would probably not be noticed. However, there would be two identical implementations for this method and this could make debugging harder.

The downside to the current solution is that there are now three functions to access the region: `getRegion()` (by `OneRegion`), `getBodyRegion()` (by `SingleBlock`) and `getCtrlBodyRegion()` (by getter).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
-- ✨ Add initial infrastructure for new QC and QCO MLIR dialects ([#1264], [#1402]) ([**@burgholzer**], [**@denialhaag**])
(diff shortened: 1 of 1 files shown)


burgholzer (approved):
This LGTM and I wouldn't worry too much about the different ways to get the region.
However, the proposal of simply calling it `$region` seems enticing.
This would simply be an override of the respective method, right?
Could we try this here in the PR?

PR #1443, opened 2026-01-12
♻️ Refactor getStaticParameter to valueToDouble

taminob:
## Description

This PR reduces the current code duplication of `getStaticParameter` which is used in both `QC` and `QCO`.
It additionally adds the conversion to a standard C++ double type since there is no other use of the function in the code base so far.

An alternative to the currently returned `std::optional` would be a `llvm::FailureOr<double>` which is simply a subclass of `std::optional`. However, it is in my opinion slightly more annoying to use since it requires `llvm::succeeded` and `llvm::failed` to check the result instead of the implicit conversion to bool available in `std::optional`. This is meant to make it slightly safer to use, but makes the code also more verbose and less convenient which is why I decided against using it. `std::optional` is an established standard C++ tool and people usually know how to handle it correctly which cannot be said for `llvm::FailureOr<double>`.

Required for #1426

## Checklist:

### DIFF
--- a/mlir/unittests/Dialect/Utils/test_utils.cpp
+++ b/mlir/unittests/Dialect/Utils/test_utils.cpp
@@ -28,27 +28,28 @@ class UtilsTest : public ::testing::Test {
 protected:
   MLIRContext context;
   std::unique_ptr<OpBuilder> builder;
+  std::unique_ptr<Location> loc;
 
   void SetUp() override {
     context.loadDialect<arith::ArithDialect>();
 
     builder = std::make_unique<OpBuilder>(&context);
+    loc = std::make_unique<Location>(builder->getUnknownLoc());
   }
 
   arith::AddFOp createAddition(double a, double b) {
-    auto firstOperand = builder->create<arith::ConstantOp>(
-        builder->getUnknownLoc(), builder->getF64FloatAttr(a));
-    auto secondOperand = builder->create<arith::ConstantOp>(
-        builder->getUnknownLoc(), builder->getF64FloatAttr(b));
-    return builder->create<arith::AddFOp>(builder->getUnknownLoc(),
-                                          firstOperand, secondOperand);
+    auto firstOperand =
+        builder->create<arith::ConstantOp>(*loc, builder->getF64FloatAttr(a));
+    auto secondOperand =
+        builder->create<arith::ConstantOp>(*loc, builder->getF64FloatAttr(b));
+    return builder->create<arith::AddFOp>(*loc, firstOperand, secondOperand);
   }
 };
 
 TEST_F(UtilsTest, valueToDouble) {
   constexpr double expectedValue = 1.234;
   auto op = builder->create<arith::ConstantOp>(
-      builder->getUnknownLoc(), builder->getF64FloatAttr(expectedValue));
+      *loc, builder->getF64FloatAttr(expectedValue));
   ASSERT_TRUE(op);
 
   auto value = op.getResult();
@@ -60,7 +61,7 @@ TEST_F(UtilsTest, valueToDouble) {
 TEST_F(UtilsTest, valueToDoubleCastFromInteger) {
   constexpr int expectedValue = 42;
   auto op = builder->create<arith::ConstantOp>(
-      builder->getUnknownLoc(), builder->getI32IntegerAttr(expectedValue));
+      *loc, builder->getI32IntegerAttr(expectedValue));
   ASSERT_TRUE(op);
 
   auto value = op.getResult();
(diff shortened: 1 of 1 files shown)


mlir/unittests/Dialect/Utils/test_utils.cpp
```diff
@@ -0,0 +1,86 @@
...
+    for (auto&& constantOp : funcOp.getOps<arith::ConstantOp>()) {
+      auto value = constantOp.getResult();
+      auto stdValue = utils::valueToDouble(value);
+      EXPECT_FALSE(stdValue.has_value());
+      return;
+    }
+    FAIL() << "No arith::ConstantOp found in function!";
+  }
+  FAIL() << "No func::FuncOp found in module!";
+}
```
burgholzer, line 109: In my opinion, this should actually succeed. Int is implicitly convertible to double and I would certainly expect this to work for "anything" that can be converted to double.
Should only take a small modification to the main utility function and now that there is only one place where it is defined, this should be quite easy 🥳

mlir/unittests/Dialect/Utils/test_utils.cpp
```diff
@@ -0,0 +1,86 @@
...
+    for (auto&& addOp : funcOp.getOps<arith::AddFOp>()) {
+      auto value = addOp.getResult();
+      auto stdValue = utils::valueToDouble(value);
+      EXPECT_FALSE(stdValue.has_value());
+      return;
+    }
+    FAIL() << "No arith::AddFOp found in function!";
+  }
+  FAIL() << "No func::FuncOp found in module!";
+}
```
burgholzer, line 132: Maybe this is just me, but wouldn't it be easier to simply write these tests programatically, i.e., to construct the corresponding `arith` ops in the C++ code instead of involving a parser here and using `func.func`.
I would guess this reduces the dependencies of this test target quite a bit.

mlir/unittests/Dialect/Utils/test_utils.cpp
taminob: I adapted it to also work with integers, although I avoided the topics of bool -> double and string -> double since that would be too dangerous in my opinion to do by default. Had to do some reading regarding signless integers in MLIR, but I think it makes sense to just assume a signed value for these integers on conversion since our values should usually be small enough to fit a signed `int64_t`.

mlir/unittests/Dialect/Utils/test_utils.cpp
taminob: You are right - thanks for the pointer :)
These workarounds with the loops are just hacky workarounds anyway because of the `const` issues I mentioned in the other PR.

mlir/unittests/Dialect/Utils/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+target_link_libraries(
+  mqt-core-mlir-dialect-utils-test
+  PRIVATE GTest::gtest_main
+          MLIRFuncDialect
+          MLIRArithDialect
+          MLIRIR
+          MLIRParser
+          MLIRSupport
+          LLVMSupport)
```
burgholzer:
```suggestion
          MLIRArithDialect
          MLIRIR
          MLIRSupport
          LLVMSupport)
```
Not sure if this is the only thing that can be removed from here.

mlir/unittests/Dialect/Utils/test_utils.cpp
```diff
@@ -0,0 +1,132 @@
...
+    builder = std::make_unique<OpBuilder>(&context);
+  }
+
+  arith::AddFOp createAddition(double a, double b) {
+    auto firstOperand = builder->create<arith::ConstantOp>(
+        builder->getUnknownLoc(), builder->getF64FloatAttr(a));
+    auto secondOperand = builder->create<arith::ConstantOp>(
+        builder->getUnknownLoc(), builder->getF64FloatAttr(b));
+    return builder->create<arith::AddFOp>(builder->getUnknownLoc(),
+                                          firstOperand, secondOperand);
```
burgholzer: Obviously only a minor improvement, but you could add a `loc` member variable to the test class that get's initialized to `builder->getUnknownLoc()`.
Safes a couple dozen characters throughout this file 🤷🏼‍♂️

mlir/unittests/Dialect/Utils/test_utils.cpp
burgholzer: Yeah. that totally makes sense. the current tests are quite exhaustive and nice already 👍🏼  feel free to resolve this.

mlir/unittests/Dialect/Utils/test_utils.cpp
burgholzer: The tests look way nicer now! Feel free to resolve this.

burgholzer (approved):
One more round of small adjustments. Then let's get this in 👍🏼

mlir/unittests/Dialect/Utils/test_utils.cpp
taminob: Done - `Location` doesn't have a default constructor unfortunately, so I had to wrap it into a `unique_ptr`

PR #1470, opened 2026-01-18
✨ Allow OpenQASM as mqt-cc input

taminob:
## Description

For the evaluation of my thesis, I will need to compile OpenQASM files using optimizations in the `QCO` dialect.
The easiest solution was for me to directly add this feature to the `mqt-cc` tool.
Since I think this might be a valuable addition in general, I wanted to propose to merge this feature to main.

Additionally, I added proper handling of the CLI argument parsing since currently errors are just ignored without any user output.

## Checklist:

### DIFF
--- a/mlir/tools/mqt-cc/mqt-cc.cpp
+++ b/mlir/tools/mqt-cc/mqt-cc.cpp
@@ -118,17 +118,17 @@ static OwningOpRef<ModuleOp> loadMLIRFile(StringRef filename,
 /**
  * @brief Write the module to the output file
  */
-static LogicalResult writeOutput(ModuleOp module, StringRef filename) {
+static mlir::LogicalResult writeOutput(ModuleOp module, StringRef filename) {
   std::string errorMessage;
   const auto output = openOutputFile(filename, &errorMessage);
   if (!output) {
     errs() << errorMessage << "\n";
-    return failure();
+    return mlir::failure();
   }
 
   module.print(output->os());
   output->keep();
-  return success();
+  return mlir::success();
 }
 
 int main(int argc, char** argv) {


mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -45,6 +47,10 @@ const cl::opt<std::string> OUTPUT_FILENAME("o", cl::desc("Output filename"),
                                            cl::value_desc("filename"),
                                            cl::init("-"));
 
+const cl::opt<bool> QASM_INPUT("qasm-input",
+                               cl::desc("Input files are read as OpenQASM"),
+                               cl::init(false));
```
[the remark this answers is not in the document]
burgholzer: @taminob I think @coderabbitai has a point here. I was also wondering, whether we actually need a dedicated flag for this or if we could just try to import as MLIR and when that fails try to import as OpenQASM. 
Should make for a slightly better user experience.

mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -69,14 +75,32 @@ const cl::opt<bool>
...
+  auto file = mlir::openInputFile(filename, &errorMessage);
+  if (!file) {
+    errs() << errorMessage << "\n";
+    return nullptr;
+  }
+
+  // Parse the input QASM
+  qc::QuantumComputation qc = qasm3::Importer::importf(filename.str());
+  return mlir::translateQuantumComputationToQC(context, qc);
+}
```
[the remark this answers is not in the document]
burgholzer: @taminob This also seems like a valid point from @coderabbitai
Especially the part about the error handling.

mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -32,7 +35,6 @@
 #include <utility>
 
 using namespace llvm;
-using namespace mlir;
```
burgholzer, line 35: I'd prefer to keep the using directive to save a couple dozen useless namespace qualifiers throughout this file.

burgholzer (changes requested):
This is great and a welcome addition. Some nitpicks in the comments. This should not be long until it can go in.
You might want to add the PR number to the change log.

mlir/tools/mqt-cc/mqt-cc.cpp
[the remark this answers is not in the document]
burgholzer: @coderabbitai considering file type hints is a very good suggestion.
It should be enough to check for ".mlir" and ".qasm"

mlir/tools/mqt-cc/mqt-cc.cpp
taminob: The issue here is that we get a name conflict between the namespaces `::qc` and `::mlir::qc` which caused some look-up issues. I guess I can make it for the (I think) few cases where it's an issue explicit and otherwise re-introduce the `using namespace`.

mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -108,39 +134,47 @@ int main(int argc, char** argv) {
...
+  mlir::OwningOpRef<mlir::ModuleOp> module;
+  if (INPUT_FILENAME.getValue().ends_with(".qasm")) {
+    module = loadQASMFile(INPUT_FILENAME, &context);
+  } else {
+    module = loadMLIRFile(INPUT_FILENAME, &context);
+  }
   if (!module) {
-    errs() << "Failed to load input file: " << INPUT_FILENAME << "\n";
     return 1;
   }
```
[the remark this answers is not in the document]
burgholzer: `.qasm2` and `.qasm3` are not established file extensions in the quantum domain. People simply use `.qasm`

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 ### Added
 
 - ✨ Add initial infrastructure for new QC and QCO MLIR dialects ([#1264], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465]) ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**])
+- ✨ Add OpenQASM support to mqt-cc CLI tool ([#1470]) ([**@taminob**])
```
```suggestion
- ✨ Add initial infrastructure for new QC and QCO MLIR dialects ([#1264], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470]) ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**])
```
I would argue this is still part of the entire effort of building a solid infrastructure here.

mlir/tools/mqt-cc/mqt-cc.cpp
burgholzer: I'd prefer that, yeah. Should be a fairly limited number of places where the `qc` produces a conflict.

mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -8,10 +8,15 @@
  * Licensed under the MIT License
  */
 
+#include "ir/QuantumComputation.hpp"
```
burgholzer, line 11:
```suggestion
```
As a consequence of the suggestion below

mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -69,31 +73,53 @@ const cl::opt<bool>
...
 } // namespace
 
+/**
+ * @brief Load and parse a .qasm file
+ */
+static mlir::OwningOpRef<mlir::ModuleOp>
+loadQASMFile(StringRef filename, mlir::MLIRContext* context) {
+  try {
+    // Parse the input QASM using MQT-Core
+    const qc::QuantumComputation qc = qasm3::Importer::importf(filename.str());
```
burgholzer:
```suggestion
    // Parse the input OpenQASM file
    const auto qc = qasm3::Importer::importf(filename.str());
```

mlir/tools/mqt-cc/mqt-cc.cpp
```diff
@@ -108,39 +134,47 @@ int main(int argc, char** argv) {
   const InitLLVM y(argc, argv);
 
   // Parse command-line options
-  cl::ParseCommandLineOptions(argc, argv, "MQT Core Compiler Driver\n");
+  if (!cl::ParseCommandLineOptions(argc, argv, "MQT Core Compiler Driver\n",
+                                   &errs())) {
+    // error message is printed to stderr via errs()
+    return -1;
+  }
```
burgholzer: Is this change still necessary/helpful?

taminob: Actually, not really since this is the default behavior anyway - I really dislike the design of the LLVM command-line parsing with all its global variables and that it just exits on its own... But simplifies things here, I guess :)

mlir/tools/mqt-cc/mqt-cc.cpp
taminob: I tried this before, but that will cause issues with incomplete types since it's only forward-declared in the Importer header file. I also explicitly removed the use of `auto` to make the include cleaner aware of this issue :)

mlir/tools/mqt-cc/mqt-cc.cpp
taminob: I could do this, but I think this is better than a `NOLINT` for the include, right?

mlir/tools/mqt-cc/mqt-cc.cpp
burgholzer: Ah yeah, that makes sense. Should have tried this in my IDE first.
In that case, disregard the other comment.

mlir/tools/mqt-cc/mqt-cc.cpp
burgholzer: See above

mlir/tools/mqt-cc/mqt-cc.cpp
burgholzer: Might as well remove this then.
CLI parsing is hard (apparently). There are hardly any good libraries for that out there.

burgholzer (approved):
Let's go 🚀 Let's see which of the PRs makes it in first #1426, the other one will need to have its conflicts on the changelog resolved.

PR #1474, opened 2026-01-21
:sparkles: Implement getInputQubits()/getOutputQubits()

taminob:
## Description

Introduce `getInputQubits()` and `getOutputQubits()` which is especially required for a smooth interaction with `CtrlOp` through `UnitaryOpInterface`.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
-- ✨ Add initial infrastructure for new QC and QCO MLIR dialects ([#1264], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1475]) ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**])
+- ✨ Add initial infrastructure for new QC and QCO MLIR dialects ([#1264], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475]) ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**])
 
 ### Changed
 
@@ -327,6 +327,7 @@ _📚 Refer to the [GitHub Release Notes](https://github.com/munich-quantum-tool
 
 [#1481]: https://github.com/munich-quantum-toolkit/core/pull/1481
 [#1475]: https://github.com/munich-quantum-toolkit/core/pull/1475
+[#1474]: https://github.com/munich-quantum-toolkit/core/pull/1474
 [#1472]: https://github.com/munich-quantum-toolkit/core/pull/1472
 [#1471]: https://github.com/munich-quantum-toolkit/core/pull/1471
 [#1470]: https://github.com/munich-quantum-toolkit/core/pull/1470


mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -165,6 +165,14 @@ Value CtrlOp::getInputQubit(const size_t i) {
   llvm::reportFatalUsageError("Invalid qubit index");
 }
 
+ValueRange CtrlOp::getInputQubits() {
+  // TODO: needs to materialize into internal storage; I wasn't able to find a
+  // way to convert concat_range -> ValueRange; however: operations cannot
+  // define new data members
```
taminob: @burgholzer I wasn't really able to find a nice solution for this issue here. If you have some time, maybe you can give your opinion on what to do - I see the following options:
- change signature to `void getInputQubits(SmallVector<Value>& storage)` for all operations
- change signature to `SmallVector<Value> getInputQubits()` for all operations
- change signature to `std::variant<ValueRange, SmallVector<Value>> getInputQubits()` for all operations (nasty API, but might be best easy-to-implement/performance ratio solution?)
- change signature to one of the above just for `CtrlOp` (and probably other modifiers in the future) and use more advanced implementation in `UnitaryOpInterface` to materialize input (although I'm not sure yet where/how since interfaces also can't have data members... might also require some global cache like the next option)
- use (thread-safe) global cache as storage for `CtrlOp`s and their input qubits (nasty, but easy to implement and nice API :) )

MatthiasReumann: Thinking out loud: Shouldn't it be possible to define extra members via [Extra Declarations](https://mlir.llvm.org/docs/DefiningDialects/Operations/#extra-declarations)?

taminob: I tried that, but MLIR has a static assertion to ensure no additional data members are declared - I think the operations all need the same memory layout for the template polymorphism they do with the interfaces

MatthiasReumann: Ah, interesting. Good to know 👍🏻

ystade: @coderabbitai what is your opinion on the above?

For context: the issue is that the interface defines a function `getInputQubits` that currently returns a `ValueRange` which is as reference to a container. Here in this scenario, two `ValueRange`s need to be concatenated making it not a valid `ValidRange` anymore. I know that this problem also exists in pure C++ and that C++26 introduces a [`concat_view`](https://en.cppreference.com/w/cpp/ranges/concat_view.html) for that.

ystade: @taminob @burgholzer @coderabbitai I would follow the suggestion of @coderabbitai. I had actually the same in my mind but wanted to know @coderabittai 's opinion first.

MatthiasReumann: @taminob If time allows it, could you maybe benchmark / profile the `SmallVector` vs `ConcatRange` approach? I remember that building and returning these vectors had a significant runtime impact on the routing pass which calls these methods very, _very_, often. Otherwise, I'll investigate that when I re-implement the `WireIterator`.

taminob: > @taminob If time allows it, could you maybe benchmark / profile the `SmallVector` vs `ConcatRange` approach? I remember that building and returning these vectors had a significant runtime impact on the routing pass which calls these methods very, _very_, often. Otherwise, I'll investigate that when I re-implement the `WireIterator`.

The issue with using the concat range is that it requires the types that are being concatenated as template arguments, making it a rather long/complicated return type. Another solution (or maybe that is also what you meant) I just thought about is that I could concat for everything, except in `CtrlOp`, with an empty `ValueRange` to get the same type for all functions.

Unfortunately, I currently don't have the time to properly benchmark the different approaches since I need to focus on finalizing my thesis.

MatthiasReumann: > The issue with using the concat range is that it requires the types that are being concatenated as template arguments, making it a rather long/complicated return type. 

For the `WireIterator` I used `auto` as the return type to avoid that issue. Probably not the cleanest solution, but works fine regardless. See [here](https://github.com/munich-quantum-toolkit/core/pull/1310/changes#diff-a0155bfcca60556d4ec8130b7f4a4a36c8ea3b1cf406de3ec2017fa0078f0a68R43). Plus, in the MLIR code base they sometimes also uses it (Line 248 [here](https://mlir.llvm.org/doxygen/TransformInterfaces_8h_source.html)).

> Unfortunately, I currently don't have the time to properly benchmark the different approaches since I need to focus on finalizing my thesis.

No worries. Good luck! 🫰🏼

taminob: > For the `WireIterator` I used `auto` as the return type to avoid that issue. Probably not the cleanest solution, but works fine regardless. See [here](https://github.com/munich-quantum-toolkit/core/pull/1310/changes#diff-a0155bfcca60556d4ec8130b7f4a4a36c8ea3b1cf406de3ec2017fa0078f0a68R43). Plus, in the MLIR code base they sometimes also uses it (Line 248 [here](https://mlir.llvm.org/doxygen/TransformInterfaces_8h_source.html)).

It is actually pretty clean, but unfortunately that doesn't work for the interface because it needs to know the return type before-hand because the implementation is in a separate compile unit (header/source file).

MatthiasReumann: Ah got it. Unfortunate. ~Wouldn't it then be an option to use `auto` and provide a default implementation of the method in the tablegen definition? (Even though having C++ code in tablegen is kind of ugly)~

Edit: My bad, this is of course just the ctrl op.

taminob: > Ah got it. Unfortunate. Wouldn't it then be an option to use `auto` and provide a default implementation of the method in the tablegen definition? (Even though having C++ code in tablegen is kind of ugly)

I don't think that'll work either because tablegen works internally with function pointers which cannot have an unspecified return type - that's also why it's not possible to have template interface methods (if you're curious, you can checkout #1426 for a workaround).

MatthiasReumann: Interesting, thanks! Really unfortunate. Consequently, I would still require helper methods (using concat views) whenever I don't want the "overhead" of building temporary vectors. 

One last question then I won't bother you anymore (I promise!): Given that all the above doesn't work and the assumption that we want to avoid temporary vectors, wouldn't an alternative solution be to use static methods or simply functions, e.g.:
```c++
ValueRange getInputQubits(HOp op) {
    ...
}

auto getInputQubits(CtrlOp op) {
    ...
}
```
Essentially, the same as the WireIterator had. Definitely not as elegant as methods, but an option (?). Any opinions on that?

burgholzer: I think the main problem here is that we want a uniform interface. We do not want to distinguish between a CtrlOp and a regular gate. The respective function should return the same type of value.
I haven't thought very deeply about this right now, but I feel the `SmallVector` is probably the only real option for that.
At least if we do not find a better way to get a ValueRange out of the CtrlOp.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
```diff
@@ -104,15 +104,31 @@ def UnitaryOpInterface : OpInterface<"UnitaryOpInterface"> {
...
-            "ValueRange", "getInputQubits", (ins)
+            "std::invoke_result_t<decltype(llvm::concat<Value, ValueRange, ValueRange>), ValueRange, ValueRange>",
+            "getInputQubits", (ins),
+            [{
+              if constexpr (std::is_same_v<decltype($_op.getInputQubits()), ValueRange>) {
+                return llvm::concat<Value>($_op.getInputQubits(), ValueRange{});
+              } else {
+                return $_op.getInputQubits();
+              }
+            }]
```
taminob: After some trial and error, I got this working proposal - although I'm not 100% sure if this actually has better performance than using a `llvm::SmallVector` (but it probably should be). Any thoughts on this?

I can't use the return type of `llvm::concat` directly because it's in the `detail` namespace and thus not really suitable for a public interface.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -160,22 +160,42 @@ Value CtrlOp::getInputQubit(const size_t i) {
     return getControlsIn()[i];
   }
   if (numControls <= i && i < getNumQubits()) {
-    return getBodyUnitary().getInputQubit(i - numControls);
+    return getTargetsIn()[i - numControls];
```
taminob, line 163: I think this is actually a bug in the current implementation because it doesn't make sense to return the "internal" argument qubit which has a completely different scope and is defined by the `CtrlOp` - at least to me, this behavior was unintuitive and caused me some headache because figuring out which `mlir::Value` is which qubit can be a pain in gdb. I'm just not sure how to actually properly expose the mapping "input qubit" -> "body unitary input"/"block argument" in a clean interface, but I think that's a topic for a separate issue.

This is basically a follow-up for #1436.

burgholzer: Yeah. This was definitely an oversight even before #1436. The new implementation makes much more sense.

mlir/include/mlir/Dialect/QCO/IR/QCOInterfaces.td
```diff
@@ -102,10 +102,20 @@ def UnitaryOpInterface : OpInterface<"UnitaryOpInterface"> {
...
+        >,
         InterfaceMethod<
             "Returns the i-th output qubit (targets + controls combined).",
             "Value", "getOutputQubit", (ins "size_t":$i)
         >,
+        InterfaceMethod<
+            "Returns a range of all output qubits (targets + controls combined).",
+            "ResultRange",
+            "getOutputQubits", (ins)
+        >,
```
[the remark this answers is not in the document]
taminob: I think this is fine since the same wording in both descriptions implies the same order

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
burgholzer: Just to update everyone here: I found a way cleaner solution for this that seems to work quite well.

PR #1479, opened 2026-01-27
✨ Add conversions between Jeff and QCO

### LINKED ISSUE #1546: ✨ Add initial support for Jeff
As we have decided to support Jeff, we should add conversions between Jeff and QCO. For now, these conversions will cover only a subset of the operations Jeff supports, but we can extend them as our dialects grow.

denialhaag:
## Description

This PR adds conversions for the [Jeff dialect](https://github.com/PennyLaneAI/jeff-mlir). 

Fixes #1546

## Checklist:

### DIFF
--- a/mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
+++ b/mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
@@ -24,9 +24,11 @@ def JeffToQCO : Pass<"jeff-to-qco"> {
       - Only specific `PPROp`s are currently supported
       - `Qureg_Op`s are currently not supported
       - `SCF_Op`s are currently not supported
+      - Support for multiple functions is currently limited
    }];
 
   let dependentDialects = [
+     "mlir::arith::ArithDialect",
      "mlir::jeff::JeffDialect",
      "mlir::qco::QCODialect",
   ];


cmake/ExternalDependencies.cmake
```diff
@@ -33,6 +33,21 @@ if(BUILD_MQT_CORE_MLIR)
...
+    GIT_TAG main
+    GIT_SUBMODULES "" GIT_SUBMODULES_RECURSE FALSE)
+  FetchContent_MakeAvailable(jeff_mlir)
+
+  # Include jeff-mlir headers
+  include_directories(${jeff_mlir_SOURCE_DIR}/include)
+  include_directories(${jeff_mlir_BINARY_DIR}/include)
+
+  # Restore C++ standard changed by jeff-mlir
+  set(CMAKE_CXX_STANDARD 20)
```
denialhaag: The build would currently fail without this workaround, as `CMAKE_CXX_STANDARD` is set to `17` in `jeff-mlir`: https://github.com/PennyLaneAI/jeff-mlir/blob/ec59bc0440f45b659ff8b680337ffdbb3c9d6495/CMakeLists.txt#L5

CMakeLists.txt
```diff
@@ -145,16 +145,12 @@ endif()
...
-  # copy generated MLIR documentation
-  add_custom_command(
-    TARGET mlir-doc
-    POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_BINARY_DIR}/docs/
-            ${CMAKE_CURRENT_SOURCE_DIR}/docs/mlir/
-    COMMAND ${CMAKE_COMMAND} -D DOCS_DIR:PATH=${CMAKE_CURRENT_SOURCE_DIR}/docs/mlir -P
-            ${CMAKE_CURRENT_SOURCE_DIR}/cmake/CleanMLIRDocs.cmake
-    COMMENT "Copying and cleaning up generated MLIR documentation"
-    VERBATIM)
```
denialhaag, line 157: This currently makes the configuration fail. I don't exactly understand what's going on, but it looks like the `mlir-doc` target is also being created by `jeff-mlir`, leading to interference.

Another problem seems to be that calling `mlir-doc` also triggers the generation of the Jeff documentation in the `_deps` directory.

Maybe it's possible to add flags to `jeff-mlir` to toggle the documentation generation. 🤔

cmake/ExternalDependencies.cmake
burgholzer: I will try to submit a PR to fix that. We probably want a similar fix in the MQT, where we similarly set the CXX standard to 20.

denialhaag: Thanks for fixing it in `jeff-mlir`! I have created #1480 to fix this issue here. 😌

CMakeLists.txt
burgholzer: I haven't quite managed to work around this for now. Will have to reinvestigate at a later point.

mlir/unittests/Conversion/JeffRoundTrip/test_jeff_round_trip.cpp
```diff
@@ -0,0 +1,909 @@
...
+INSTANTIATE_TEST_SUITE_P(QCGPhaseOpTest, JeffRoundTripTest,
+                         testing::Values(JeffRoundTripTestCase{
+                             "GlobalPhase", MQT_NAMED_BUILDER(qc::globalPhase),
+                             MQT_NAMED_BUILDER(qc::globalPhase)}));
+/// @}
+
+/// \name JeffRoundTrip/Operations/StandardGates/HOp.cpp
+/// @{
+INSTANTIATE_TEST_SUITE_P(
+    QCHOpTest, JeffRoundTripTest,
```
denialhaag: @burgholzer, I just wanted to confirm that I'm applying the new test infrastructure correctly. Two questions:

1. Is `JeffRoundTrip/Operations/StandardGates/HOp.cpp` correct? If it's meant to refer to a file, that file, of course, wouldn't exist.
2. Is the instantiation name meant to be `QCHopTest` here? In principle, this does make sense to me, as the test starts with a `qc.h` operation.

I have adapted this from here:

https://github.com/munich-quantum-toolkit/core/blob/d9209b73485c61b6c336cdcd76d165732ad2cb6b/mlir/unittests/Conversion/QCToQCO/test_qc_to_qco.cpp#L176-L179

burgholzer: I think this is generally a consequence of an earlier design in the test PR where I had all of these split into separate files. That led to crazy long compile times though, so I combined everything back into one.
I still wanted to somehow have separators between the individual tests. What you linked is the end result of that.
It's definitely not perfect, so we can happily adjust it. I just want a way to separate the individual test groups and have something to easily search for in the file.
Best case, we would not even do round trip tests here, but would have the jeff Builder and verify both conversions similarly to how we verify QC<->QCO.
Probably the round trip tests do not hurt though and I think the current structure should be fine.
Feel free to tweak the names of the tests. (Shouldn't this be QCO for example, and shouldn't it be HOp?)

Does any of that make sense?

denialhaag: > I think this is generally a consequence of an earlier design in the test PR where I had all of these split into separate files. That led to crazy long compile times though, so I combined everything back into one.
> I still wanted to somehow have separators between the individual tests. What you linked is the end result of that.

Okay, that makes perfect sense to me!

> Best case, we would not even do round trip tests here, but would have the jeff Builder and verify both conversions similarly to how we verify QC<->QCO.

Agreed. I'd see the current round-trip-only tests as a temporary solution until we have a Jeff builder.

> Shouldn't this be QCO for example, and shouldn't it be HOp?

At this very moment, the conversion is still between QC and Jeff, but I'll change this tomorrow! I wanted to put the testing on a more solid footing first. If I'm not mistaken, updating the tests after the change should be very straightforward now (basically, search and replace QC with QCO).

cmake/ExternalDependencies.cmake
```diff
@@ -33,6 +33,20 @@ if(BUILD_MQT_CORE_MLIR)
...
+  # Fetch jeff-mlir
+  set(BUILD_JEFF_MLIR_TRANSLATION
+      OFF
+      CACHE BOOL "Disable building the translation submodule of jeff-mlir" FORCE)
+  FetchContent_Declare(
+    jeff_mlir
+    GIT_REPOSITORY https://github.com/PennyLaneAI/jeff-mlir.git
+    GIT_TAG deserialization
+    GIT_SUBMODULES "" GIT_SUBMODULES_RECURSE FALSE)
+  FetchContent_MakeAvailable(jeff_mlir)
```
denialhaag: @burgholzer, for some reason, things do not work if we use `list(APPEND FETCH_PACKAGES jeff_mlir)` here. This is a shot in the dark, but maybe this is because we are "setting up MLIR twice" (see also https://github.com/munich-quantum-toolkit/core/pull/1479#discussion_r2733221979 and https://github.com/PennyLaneAI/jeff-mlir/pull/5#discussion_r2852382487). 🤔

burgholzer: Yeah, as commented there, maybe some/all of the MLIR setup code needs to be hidden behind `PROJECT_IS_TOP_LEVEL` so that it is not done twice.

denialhaag: The changes in https://github.com/munich-quantum-toolkit/core/pull/1479/changes/24b9e135539c91d1bfd55fa71aaa2742f097c0a9 should have fixed the problem. If I understood correctly, it's just important to include `SetupMLIR` before `cmake/ExternalDependencies.cmake`. 🤔

CMakeLists.txt
denialhaag: The problem should be fixed; see https://github.com/munich-quantum-toolkit/core/pull/1479#discussion_r2852404827 for some context.

cmake/ExternalDependencies.cmake
burgholzer: Ah, that makes sense 👍🏼 I think I have seen something similar with the Catalyst plugin.
Nice fix 👍🏼

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+    ConversionTarget target(*context);
+    RewritePatternSet patterns(context);
+    QCOToJeffTypeConverter typeConverter(context);
+
+    LoweringState state;
+
+    // Configure conversion target
+    target.addIllegalDialect<QCODialect, arith::ArithDialect>();
+    target.addLegalDialect<jeff::JeffDialect>();
+
```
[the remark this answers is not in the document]
denialhaag: The `arith` dialect is no longer marked as illegal. The conversion between `arith` and Jeff will be expanded on in follow-ups.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -0,0 +1,1017 @@
...
+             ConvertJeffQubitFreeZeroOpToQCO, ConvertJeffQubitMeasureOpToQCO,
+             ConvertJeffQubitMeasureNDOpToQCO, ConvertJeffQubitResetOpToQCO,
+             ConvertJeffGPhaseOpToQCO, ConvertJeffIOpToQCO, ConvertJeffXOpToQCO,
+             ConvertJeffYOpToQCO, ConvertJeffZOpToQCO, ConvertJeffHOpToQCO,
+             ConvertJeffSOpToQCO, ConvertJeffTOpToQCO, ConvertJeffRxOpToQCO,
+             ConvertJeffRyOpToQCO, ConvertJeffRzOpToQCO, ConvertJeffR1OpToQCO,
+             ConvertJeffUOpToQCO, ConvertJeffSwapOpToQCO,
+             ConvertJeffCustomOpToQCO, ConvertJeffPPROpToQCO,
+             ConvertJeffIntConst1OpToArith, ConvertJeffIntConst64OpToArith,
+             ConvertJeffFloatConst64OpToArith>(typeConverter, context);
```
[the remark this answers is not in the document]
denialhaag: This is a known shortcoming of this first version. Any QCO module can be assumed to have a single function with no inputs and that returns an integer.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+    return failure();
+  }
+  returnOp->erase();
+
+  // Add trivial return operation
+  builder.setInsertionPointToEnd(block);
+  func::ReturnOp::create(builder, loc);
+
+  // Fix return type
+  main.setType(FunctionType::get(ctx, {}, {}));
```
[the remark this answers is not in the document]
denialhaag: This is a known shortcoming of this first version. Any QCO module can be assumed to have a single function with no inputs and that returns an integer.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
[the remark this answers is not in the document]
denialhaag: @coderabbitai, to be clear, this will change in the foreseeable future. For now, however, this assumption holds.

cmake/ExternalDependencies.cmake
```diff
@@ -33,6 +33,18 @@ if(BUILD_MQT_CORE_MLIR)
...
+  set(BUILD_JEFF_MLIR_TRANSLATION
+      OFF
+      CACHE BOOL "Disable building the translation submodule of jeff-mlir")
+  FetchContent_Declare(
+    jeff-mlir
+    GIT_REPOSITORY https://github.com/PennyLaneAI/jeff-mlir.git
+    GIT_TAG 7960f4e83e9bbc48c6d6bf203ebe5250945d9de3
+    GIT_SUBMODULES "" GIT_SUBMODULES_RECURSE FALSE)
+  list(APPEND FETCH_PACKAGES jeff-mlir)
+  set(CMAKE_CXX_STANDARD 20)
```
burgholzer: If this is necessary at the moment, then we should propose a change to jeff-mlir that does something similar to what we do here these days in that we only set the standard if it is not already set or set to something older.

Also: is there a particular reason for the submodules code here? the repository should not have submodules anymore, right?

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/UOp.cpp
```diff
@@ -87,6 +87,24 @@ struct ReplaceUWithRY final : OpRewritePattern<UOp> {
...
+    const auto theta = valueToDouble(op.getTheta());
+    if (!theta || std::abs(*theta - (std::numbers::pi / 2.0)) > TOLERANCE) {
+      return failure();
+    }
+    rewriter.replaceOpWithNewOp<U2Op>(op, op.getInputQubit(0), op.getPhi(),
+                                      op.getLambda());
+    return success();
+  }
+};
+
```
burgholzer, line 107: Just noting this here for the sake of it: Maybe we could try to pull changes like these into individual PRs a little bit more in the future to avoid PRs becoming too big and to make sure that nice additions like these are not kept behind features that are still under review.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+  Value target;
+  if (!state.inModifier()) {
+    target = adaptor.getQubitIn();
+  } else {
+    target = state.targetsIn[0];
+  }
+
+  auto jeffOp =
+      rewriter.create<JeffOpType>(op.getLoc(), target,
+                                  /*in_ctrl_qubits=*/state.controlsIn,
```
burgholzer: `rewriter.create<JeffOpType>` -> `JeffOpType::create(rewriter,...)`
LLVM 22 will even issue warnings that the former is deprecated. Applies to the entire PR

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+  // Remove return operation
+  auto* returnOp = block->getTerminator();
+  if (!llvm::isa<func::ReturnOp>(returnOp)) {
+    return failure();
+  }
+  returnOp->erase();
+
+  // Add trivial return operation
+  builder.setInsertionPointToEnd(block);
+  func::ReturnOp::create(builder, loc);
```
burgholzer: Kind of feels like this could also be part of a rewrite rule. This is essentially replacing the return in the function. the pattern could match on FuncOps that have the passthorugh entry point attribute

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+
+  llvm::SmallVector<llvm::StringRef> stringRefs;
+  stringRefs.reserve(state.strings.size());
+  for (const auto& str : state.strings) {
+    stringRefs.push_back(str);
+  }
+  module->setAttr("jeff.strings", builder.getStrArrayAttr(stringRefs));
+
+  module->setAttr("jeff.tool", builder.getStringAttr("mqt-cc"));
+  module->setAttr("jeff.toolVersion", builder.getStringAttr("4.0.0"));
```
burgholzer: Should this be fed from a CMake define of the actual version?

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+  matchAndRewrite(qco::GPhaseOp op, OpAdaptor /*adaptor*/,
+                  ConversionPatternRewriter& rewriter) const override {
+    auto& state = getState();
+
+    auto jeffOp =
+        rewriter.create<jeff::GPhaseOp>(op.getLoc(), op.getParameter(0),
+                                        /*in_ctrl_qubits=*/state.controlsIn,
+                                        /*num_ctrls=*/state.controlsIn.size(),
+                                        /*is_adjoint=*/state.inInvOp,
+                                        /*power=*/1);
```
burgholzer: It's slighlty annoying that `power=1` has to be specified in all these cases. Would it be worth to add a better builder or default value in jeff-mlir?

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+      return rewriter.notifyMatchFailure(
+          op, "Nested control operations are not supported");
+    }
+
+    if (state.inInvOp) {
+      return rewriter.notifyMatchFailure(
+          op,
+          "Control operations inside inversion operations are not supported");
+    }
+
```
burgholzer, line 1203: Can't we rule both of these out by design because of our canonicalizations? but,  in the end, these do not cost much. might as well keep them

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1472 @@
...
+    return llvm::TypeSwitch<Type, LogicalResult>(op.getType())
+        .Case<FloatType>([&](auto type) -> LogicalResult {
+          auto floatAttr = llvm::dyn_cast<FloatAttr>(value);
+          if (!floatAttr) {
+            return rewriter.notifyMatchFailure(op, "Expected float attribute");
+          }
+          switch (type.getWidth()) {
+          case 64:
+            rewriter.replaceOpWithNewOp<jeff::FloatConst64Op>(op, floatAttr);
+            return success();
```
burgholzer, line 1405: I may be wrong, but I think there is also a 32 bit float in jeff. same goes below for the integers. Aren't a couple of cases missing here?

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -0,0 +1,1025 @@
...
+static void createModified(
+    JeffOpType& op, ConversionPatternRewriter& rewriter,
+    const llvm::SmallVector<Value>& controls,
+    const llvm::SmallVector<Value>& targets,
+    llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> lambda) {
+  auto loc = op.getLoc();
+  if (op.getNumCtrls() != 0) {
+    qco::CtrlOp ctrlOp;
+    if (!op.getIsAdjoint()) {
+      ctrlOp = rewriter.create<qco::CtrlOp>(loc, controls, targets, lambda);
```
burgholzer: Pointing this out here for the rest of this file. `<OP>::create` functions shall be used

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -0,0 +1,1025 @@
...
+  // Add return operation
+  builder.setInsertionPointToStart(block);
+  auto constOp =
+      arith::ConstantOp::create(builder, loc, builder.getI64IntegerAttr(0));
+
+  builder.setInsertionPointToEnd(block);
+  func::ReturnOp::create(builder, loc, constOp.getResult());
+
+  // Fix return type
+  main.setType(FunctionType::get(ctx, {}, {builder.getI64Type()}));
```
burgholzer: Similar to the other direction, it feels like this could actually be a proper rewrite rule as part of the conversion

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -0,0 +1,1025 @@
...
+struct ConvertJeffGPhaseOpToQCO final : OpConversionPattern<jeff::GPhaseOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(jeff::GPhaseOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter& rewriter) const override {
+    if (op.getPower() != 1) {
+      return rewriter.notifyMatchFailure(
+          op, "Operations with power != 1 are not yet supported");
+    }
```
burgholzer, line 604: for the gphase operation this should be fairly straight forward to support. it's just a multiplication of the angle with the power

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -0,0 +1,1025 @@
...
+    : OpConversionPattern<jeff::FloatConst64Op> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(jeff::FloatConst64Op op, OpAdaptor /*adaptor*/,
+                  ConversionPatternRewriter& rewriter) const override {
+    rewriter.replaceOpWithNewOp<arith::ConstantOp>(op, op.getValAttr());
+    return success();
+  }
+};
```
burgholzer, line 983: Do these translations automatically figure out the right bitwidth or am I overlooking things?

Furthermore, aren't there quite some of the arithmetic operations missing from Jeff? Or is the primary goal here to facilitate the roundtrip of our test suite and nothing more? I am fine with that. as long as it is a decision on purpose.

mlir/unittests/Conversion/JeffRoundTrip/test_jeff_round_trip.cpp
```diff
@@ -0,0 +1,613 @@
...
+using namespace mlir;
+
+struct JeffRoundTripTestCase {
+  std::string name;
+  mqt::test::NamedBuilder<qco::QCOProgramBuilder> programBuilder;
+  mqt::test::NamedBuilder<qco::QCOProgramBuilder> referenceBuilder;
+
+  friend std::ostream& operator<<(std::ostream& os,
+                                  const JeffRoundTripTestCase& info);
+};
```
burgholzer, line 44: Maybe that's too naive, but is a 1-1 roundtrip feasible to achieve here? because in that case, there isn't realy a need for two builders here, right? you just construct a program once and check that the final program is equivalent to the one we started with. Would simplify some bits and pieces here.

I would guess the answer is yes, as long as we start with QC/QCO (our dialects).

cmake/ExternalDependencies.cmake
denialhaag: > If this is necessary at the moment, then we should propose a change to jeff-mlir that does something similar to what we do here these days in that we only set the standard if it is not already set or set to something older.

It is necessary even though we have merged https://github.com/PennyLaneAI/jeff-mlir/pull/2. I'm a bit confused about what the issue is, frankly.

> Also: is there a particular reason for the submodules code here? the repository should not have submodules anymore, right?

That is indeed no longer necessary. I'm removing it in one of the commits I'm pushing soon. 👍

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/UOp.cpp
denialhaag: Totally agreed!

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: Yes, I have only implemented the bare minimum in this PR to get the round-trip tests to work. I want to do this properly by adding `arith`-to-Jeff and Jeff-to-`arith` conversion passes in the `jeff-mlir` repository, maybe even next week.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: > Do these translations automatically figure out the right bitwidth or am I overlooking things?

Yes, that works automatically because `getValAttry()` already returns, in this case, a `F64Attr`: https://github.com/PennyLaneAI/jeff-mlir/blob/main/include/jeff/IR/JeffOps.td#L1646-L1662

> Furthermore, aren't there quite some of the arithmetic operations missing from Jeff? Or is the primary goal here to facilitate the roundtrip of our test suite and nothing more? I am fine with that. as long as it is a decision on purpose.

See https://github.com/munich-quantum-toolkit/core/pull/1479#discussion_r2899938760.

mlir/unittests/Conversion/JeffRoundTrip/test_jeff_round_trip.cpp
denialhaag: Yes, it was surprisingly easy to get the QCO-to-Jeff-to-QCO round-trip tests working here. I like your idea from Wednesday to add a serialization-deserialization step in the middle to ensure the Jeff-MLIR module is valid, but I'd propose doing that in a follow-up. As long as we trust our implementation to not make errors that cancel each other out (e.g., consistently mixing up X and Y gates), the round-trip tests should be fine. I'm curious how things will look when we have programs that utilize SCF and comprise multiple functions, but we'll see!

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: Yes, these can be ruled out in principle. I'm not sure how we want to deal with this in general. I added these so that a user would be notified if they didn't run the canonicalization before (maybe it would make sense to add a note to the error message), but I don't know what the philosophy should be. Is it okay to just assume that a program has been canonicalized before it enters any pass?

burgholzer: hm. good question.
The [MLIR docs](https://mlir.llvm.org/docs/Canonicalization/) say
> Pass pipelines should not rely on the canonicalizer pass for correctness. They should work correctly with all instances of the canonicalization pass removed.

I think that is pretty clear.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: I don't know why I keep forgetting this... Done.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: Done.

cmake/ExternalDependencies.cmake
burgholzer: Okay, that was confusing at first as well, but I think this is due to us only having a similar code block in `mlir/CMakeLists.txt` and not in the top-level CMake file. That way, the jeff-mlir one comes first and might cause some problems. I think moving the few lines of code
```cmake
if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 20)
  set(CMAKE_CXX_STANDARD 20 CACHE STRING "C++ standard to conform to")
endif()
```
to the top level file and before the `cmake/ExternalDependencies` call, should fix this.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
burgholzer: alright, sounds reasonable to me!

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
burgholzer: Nice 👍🏼 feel free to resolve this then

mlir/unittests/Conversion/JeffRoundTrip/test_jeff_round_trip.cpp
burgholzer: Sounds like a plan. we can always make this more elaborate in follow ups 👍🏼

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: Given that, do you think it's fair to raise an error telling the user to run the canonicalization? I guess this isn't exactly in the spirit of the statement, but it at least ensures nothing breaks unexpectedly. If we cannot ask the user to run the canonicalization, the conversion would be quite a bit uglier. 🤔

burgholzer: I think it is fine to just issue a warning. These are conversions that we have control over. The pipelines we set up will always ensure canonicalization is run.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1478 @@
...
+    : StatefulOpConversionPattern<qco::BarrierOp> {
+  using StatefulOpConversionPattern::StatefulOpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(qco::BarrierOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter& rewriter) const override {
+    auto& state = getState();
+
+    createCustomOp(op, rewriter, state, adaptor.getQubitsIn(), {}, false,
+                   "barrier");
```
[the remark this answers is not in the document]
denialhaag: This should be fixed now, right? My initial assumption was wrong; barrier operations can be controlled and inverted. We have canonicalization patterns that fix this. I have now changed everything to support controlled and inverted barrier operations.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1482 @@
...
+  llvm::SmallVector<Value> controlsIn;
+  llvm::SmallVector<Value> controlsOut;
+  llvm::SmallVector<Value> targetsIn;
+  llvm::SmallVector<Value> targetsOut;
+
+  [[nodiscard]] bool inModifier() const { return inCtrlOp || inInvOp; }
+
+  // Module information
+  llvm::SmallVector<std::string> strings;
+};
```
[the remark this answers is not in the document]
denialhaag: The implementation explicitly assumes that `qco.ctrl { qco.inv { ... } }` without any further nesting. The code added in this PR raises an error if this is not given.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: Agreed! I have created https://github.com/PennyLaneAI/jeff-mlir/issues/9 to keep track of this.

cmake/ExternalDependencies.cmake
denialhaag: That fixed the problem; thanks!

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: Good point! Done in https://github.com/munich-quantum-toolkit/core/pull/1479/commits/06f4ed94e569423d11e7b268f294eaec272efe9c.

For some reason, it is necessary to use `obj.QCOToJeff` instead of just `QCOToJeff` in `target_compile_definitions()`. I don't know if there's a better way. :thinking:

Edit: Windows does not like `obj.QCOToJeff`. I have fixed this in see https://github.com/munich-quantum-toolkit/core/pull/1479/commits/55eb725a92fa841ea389e5118855a88fda32d991.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: I have added a short note to the error messages in https://github.com/munich-quantum-toolkit/core/pull/1479/commits/d84c5cf41daa1b618b95cd9875a754497920be86. 👍

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: If I'm not mistaken, this could be a bit of a problem. The `arith` dialect can only multiply two floats or two integers. In Jeff, the power [is defined to be `UInt8`](https://github.com/unitaryfoundation/jeff/blob/c2faa47a0ccf58e482a0e7c7a42c9dd64293422d/impl/capnp/jeff.capnp#L520). This means that we need to create a `arith::SIToFPOp` to cast from integer to float. However, Jeff does not have a counterpart for this operation, so we cannot complete the round trip. I'm pretty sure this will require a change to the Jeff spec. :thinking:

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1478 @@
...
+    return failure();
+  }
+  returnOp->erase();
+
+  // Add trivial return operation
+  builder.setInsertionPointToEnd(block);
+  func::ReturnOp::create(builder, loc);
+
+  // Fix return type
+  main.setType(FunctionType::get(ctx, {}, {}));
```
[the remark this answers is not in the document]
denialhaag: It's not exactly a good idea to remove the constant because it could have other uses. Instead of manually checking this, we can leave this to the built-in `-remove-dead-values` pass.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag: Good point! Done.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: Done.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1500 @@
...
+        ConvertQCOTdgOpToJeff, ConvertQCOSXOpToJeff, ConvertQCOSXdgOpToJeff,
+        ConvertQCORXOpToJeff, ConvertQCORYOpToJeff, ConvertQCORZOpToJeff,
+        ConvertQCOPOpToJeff, ConvertQCOROpToJeff, ConvertQCOU2OpToJeff,
+        ConvertQCOUOpToJeff, ConvertQCOSWAPOpToJeff, ConvertQCOiSWAPOpToJeff,
+        ConvertQCOECROpToJeff, ConvertQCODCXOpToJeff, ConvertQCORXXOpToJeff,
+        ConvertQCORYYOpToJeff, ConvertQCORZXOpToJeff, ConvertQCORZZOpToJeff,
+        ConvertQCOXXMinusYYOpToJeff, ConvertQCOXXPlusYYOpToJeff,
+        ConvertQCOBarrierOpToJeff, ConvertQCOCtrlOpToJeff,
+        ConvertQCOInvOpToJeff, ConvertQCOYieldOpToJeff, ConvertQCOMainToJeff,
+        ConvertArithConstOpToJeff>(typeConverter, context, &state);
```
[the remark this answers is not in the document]
denialhaag: Such functions are not supported yet in our infrastructure. You are correct that this will have to be updated in the future, but it is fine as is for now.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1506 @@
...
+    if (!passthrough) {
+      return failure();
+    }
+
+    if (!llvm::any_of(passthrough, [](Attribute attr) {
+          const auto strAttr = llvm::dyn_cast<StringAttr>(attr);
+          return strAttr && strAttr.getValue() == "entry_point";
+        })) {
+      return failure();
+    }
```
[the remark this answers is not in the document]
denialhaag: This is fine for now. There are no other functions with a `passthrough` attribute, and the `passthrough` attribute can always be removed entirely.

mlir/lib/Conversion/QCOToJeff/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+  QCOToJeffIncGen
+  LINK_LIBS
+  MLIRDialect
+  MLIRJeff
+  MLIRQCODialect
+  MLIRTransforms
+  DISABLE_INSTALL)
+
+target_include_directories(QCOToJeff PRIVATE ${jeff-mlir_SOURCE_DIR}/include
+                                             ${jeff-mlir_BINARY_DIR}/include)
```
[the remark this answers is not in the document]
denialhaag: `BUILD_JEFF_MLIR_TRANSLATION` is disabled, and the build passes. `${jeff-mlir_SOURCE_DIR}` and `${jeff-mlir_BINARY_DIR}` exist regardless.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1506 @@
...
+    // Update function signature and remove passthrough attribute
+    rewriter.startOpModification(op);
+    op.setType(FunctionType::get(rewriter.getContext(), {}, {}));
+    op->removeAttr("passthrough");
+    rewriter.finalizeOpModification(op);
+
+    // Replace return operation
+    rewriter.setInsertionPointToEnd(block);
+    func::ReturnOp::create(rewriter, returnOp->getLoc());
+    rewriter.eraseOp(returnOp);
```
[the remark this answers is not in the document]
denialhaag: Your suggestion is not valid; `replaceOpWithNewOp()` does not accept a location.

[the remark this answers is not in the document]
denialhaag: Omitting the location leads to the operation being placed incorrectly. If there's another way of achieving this, I'd be all ears. Otherwise, I'd say we can keep the replacement as is.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -0,0 +1,1506 @@
...
+ *
+ * @par Example:
+ * ```mlir
+ * %q_out = qco.r(%theta, %phi) %q_in : !qco.qubit -> !qco.qubit
+ * ```
+ * is converted to
+ * ```mlir
+ * %q_out = jeff.custom "r"(%theta, %phi) {is_adjoint = true, num_ctrls = 0 :
+ * i8, power = 1 : i8} %q_in : !jeff.qubit
+ * ```
```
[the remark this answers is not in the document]
denialhaag: This should be resolved, right?

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
burgholzer: Okay, I definitely did not want to open a whole can of worms with this. So let's try to close that quickly again.
Long-term, this might need some casting infrastructure in Jeff, or clearly defined promotion rules between types, at which point casts could either be translated or removed entirely.
Let's keep this as-is for now and live with the limitation.
Maybe it is worth to start a list somewhere (in the documentation) of known limitations.
These could go into the docstrings for the individual conversions. Might help to remind us a bit of the things that we cannot do (yet).

mlir/lib/Conversion/QCOToJeff/CMakeLists.txt
burgholzer, line 24: Feel free to resolve this here for now, but this is definitely something we should get into jeff-mlir. The corresponding target should properly define its include directories so that including it in the `LINK_LIBS` above is sufficient here.

denialhaag: Agreed! We can improve this once https://github.com/PennyLaneAI/jeff-mlir/issues/8 is closed!

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: Makes sense! I have started a list in https://github.com/munich-quantum-toolkit/core/pull/1479/commits/caedf613696077dfab6ec036a040df9fb18a088e. I will add these limitations to the respective docstrings when I update our implementation after https://github.com/PennyLaneAI/jeff-mlir/pull/10 is merged!

mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
```diff
@@ -0,0 +1,33 @@
...
+      Note that this pass is still in development as QCO and Jeff do not have full feature parity yet.
+
+      Known limitations:
+
+      - Gates with a non-trivial power are currently not supported
+      - Only specific `CustomOp`s are currently supported
+      - Only specific `PPROp`s are currently supported
+      - `Qureg_Op`s are currently not supported
+      - `SCF_Op`s are currently not supported
+   }];
```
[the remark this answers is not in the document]
denialhaag: Done in my most recent commit.

PR #1481, opened 2026-01-29
⬆️ Update `nanobind` to 2.11.0

denialhaag:
## Description

This PR updates `nanobind` to 2.11.0.

## Checklist:

- [x] ~I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.~

### DIFF
--- a/bindings/ir/register_ir.cpp
+++ b/bindings/ir/register_ir.cpp
@@ -23,7 +23,7 @@ void registerQuantumComputation(const nb::module_& m);
 
 NB_MODULE(MQT_CORE_MODULE_NAME, m) {
   m.doc() =
-      R"pb(MQT Core IR  - The MQT Core Intermediate Representation (IR) module.)pb";
+      R"pb(MQT Core IR - The MQT Core Intermediate Representation (IR) module.)pb";
 
   registerPermutation(m);
 
--- a/python/mqt/core/ir/__init__.pyi
+++ b/python/mqt/core/ir/__init__.pyi
@@ -6,7 +6,7 @@
 #
 # Licensed under the MIT License
 
-"""MQT Core IR  - The MQT Core Intermediate Representation (IR) module."""
+"""MQT Core IR - The MQT Core Intermediate Representation (IR) module."""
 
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -8,7 +8,7 @@
 
 [build-system]
 requires = [
-  "nanobind>=2.10.2",
+  "nanobind==2.11.0",
```
burgholzer: Why the exact pin here?
We could/should at least accept patch version updates, i.e.,
```suggestion
  "nanobind~=2.11.0",
```

python/mqt/core/na/__init__.pyi
```diff
@@ -6,4 +6,9 @@
 #
 # Licensed under the MIT License
 
+"""MQT Core NA - The MQT Core neutral atom module.
+
+This module contains all neutral atom related functionality of MQT Core.
+"""
```
burgholzer: Is this really the only top-level unit file where we have such a docstring?
I am pretty sure that we had these before we moved to nanobind for every module (ir, etc.)

burgholzer (approved):
Two clarifications, otherwise this looks great!

pyproject.toml
denialhaag: I'm not sure whether stub generation could change between two patch versions (it did change between two minor versions now). If so, I feel like an exact pin makes sense, since developers will necessarily need to use the exact same version when generating the stubs. 🤔

burgholzer: As commented over in the SyReC repository, I believe patch updates to be okay. Patch updates are only going to make the stubs more correct (hopefully). So it shouldn't be too bad if such an update comes out. We can always update the stubs mid-release if that happens.

python/mqt/core/na/__init__.pyi
denialhaag: I must have forgotten to copy them over; I have added them back now. 😌

pyproject.toml
denialhaag: Makes sense! Done.

PR #1506, opened 2026-02-09
✨ Add QCO If operation

li-mingbao:
## Description

This PR adds the `qco.if` operation to the QCO dialect. This operation is similar to the `scf.if` operation but it takes an additional list of qubits as input that are passed down to the branches as block arguments. This is used in order to support linar typing for qubit values.

Example: 
```
 %q2 = qco.if %condition qubits(%arg0 = %q0) {
      %q1 = qco.h %arg0 : !qco.qubit -> !qco.qubit
      qco.yield %q1
    } else qubits(%arg0 = %q0) {
      qco.yield %arg0
    } : {i1, !qco.qubit} -> {!qco.qubit}
``` 

## Checklist:

### DIFF
--- a/mlir/unittests/programs/qco_programs.cpp
+++ b/mlir/unittests/programs/qco_programs.cpp
@@ -2027,11 +2027,11 @@ void nestedTrueIf(QCOProgramBuilder& b) {
 }
 
 void nestedFalseIf(QCOProgramBuilder& b) {
-  auto q = b.allocQubitRegister(2);
+  auto q = b.allocQubitRegister(1);
   auto q0 = b.h(q[0]);
   auto [measuredQubit, measureResult] = b.measure(q0);
   b.qcoIf(
-      measureResult, q[1],
+      measureResult, measuredQubit,
       [&](mlir::ValueRange qubits) {
         auto innerQubit = b.x(qubits[0]);
         return llvm::SmallVector<mlir::Value>{innerQubit};


mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -0,0 +1,271 @@
...
+      SmallVector<Location>(qubits.size(), odsState.location));
+  odsBuilder.setInsertionPointToStart(&elseBlock);
+  qco::YieldOp::create(odsBuilder, odsState.location,
+                       elseBuilder(elseBlock.getArguments()));
+}
+
+Block* IfOp::thenBlock() { return &getThenRegion().back(); }
+YieldOp IfOp::thenYield() { return dyn_cast<YieldOp>(&thenBlock()->back()); }
+Block* IfOp::elseBlock() { return &getElseRegion().back(); }
+YieldOp IfOp::elseYield() { return dyn_cast<YieldOp>(&elseBlock()->back()); }
```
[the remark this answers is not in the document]
li-mingbao: The else region and block is guaranteed to exist and the operation itself is guaranteed to return a value. This also holds true for all the other comments you posted like for the verifier().

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -0,0 +1,278 @@
...
+  }
+  SmallPtrSet<Value, 4> uniqueQubitsIn;
+  for (const auto& qubit : inputQubits) {
+    if (!uniqueQubitsIn.insert(qubit).second) {
+      return emitOpError("Input qubits must be unique.");
+    }
+  }
+
+  return success();
+}
```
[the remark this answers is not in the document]
li-mingbao: It can be assumed that the last operation is always a yield operation.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,69 @@ def InvOp : QCOOp<"inv", traits =
...
+                          SizedRegion<1>:$elseRegion);
+
+    let assemblyFormat = [{
+        $condition
+         custom<IfOpAliasing>($thenRegion, $elseRegion, $qubits)
+         attr-dict `:`
+        `{` type($condition) `,` type($qubits) `}`
+        `->`
+        `{` type($results) `}`
+    }];
```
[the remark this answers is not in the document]
li-mingbao: @burgholzer @denialhaag 
Any preferences here if we want to keep the `i1` in the list for completeness sake or if we just want to print the qubit types?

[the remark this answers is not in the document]
denialhaag: I'm pretty sure we consistently print all types. For example, the `MeasureOp` is defined to always return `I1`, but we still print it.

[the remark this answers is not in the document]
burgholzer: How does `scf` handle this? I tend to disagree with @coderabbitai here and agree with @denialhaag that we probably want to print all the types.
However, I think the least surprising thing to do is to follow what scf is doing in MLIR.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,69 @@ def InvOp : QCOOp<"inv", traits =
...
+def IfOp : QCOOp<"if", traits =
+            [
+                DeclareOpInterfaceMethods<RegionBranchOpInterface, [
+                "getNumRegionInvocations", "getRegionInvocationBounds",
+                "getEntrySuccessorRegions"]>,
+                SingleBlock,
+                RecursiveMemoryEffects
+            ]> {
+
+    let summary = "If then else operation for qubit types";
```
denialhaag:
```suggestion
    let summary = "If-then-else operation for qubit types";
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1054,6 +1054,54 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * ```mlir
+   * %q2 = qco.if %condition qubits(%q0 : !qco.qubit) -> !qco.qubit {
+   *   ^bb0(%arg0 : !qco.qubit) :
+   *      %q1 = qco.h %arg0 : !qco.qubit -> !qco.qubit
+   *      qco.yield %q1
+   * } else {
+   *   ^bb0(%arg0 : !qco.qubit) :
+   *      qco.yield %arg0
+   * }
+   * ```
```
denialhaag, line 1110: This needs to be updated to the new assembly format.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -672,6 +672,57 @@ QCOProgramBuilder& QCOProgramBuilder::dealloc(Value qubit) {
...
+    llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> elseBody) {
+  checkFinalized();
+
+  auto ifOp = IfOp::create(*this, condition, qubits);
+  // Create the then and else block
+  auto& thenBlock = ifOp->getRegion(0).emplaceBlock();
+  auto& elseBlock = ifOp->getRegion(1).emplaceBlock();
+
+  // Create the blockarguments and add them as valid qubits
+  for (const auto& qubitType : qubits.getTypes()) {
```
denialhaag: I'm pretty sure that the `const` is not necessary here, as `Type` is an IR type: https://mlir.llvm.org/docs/Rationale/UsageOfConst/

This also applies to other lines of this PR, but I will not spam the same comment everywhere now. 🤠

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1054,6 +1054,54 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   */
+  ValueRange
+  qcoIf(Value condition, ValueRange qubits,
+        llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> thenBody,
+        llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> elseBody);
```
denialhaag: I'd say the `elseBody` argument shouldn't be required, since it won't be uncommon not to have an `else` block. 🤔

This could also be solved through overloading.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
denialhaag, line 1: I'd be slightly in favor of calling the directory `SCF` instead of `Scf`.

mlir/unittests/Dialect/QCO/IR/SCF/CMakeLists.txt
denialhaag, line 1: Same comment on the directory name here.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -0,0 +1,278 @@
...
+///      print(%cmp)
+///   }
+///
+///  becomes
+///
+///   scf.if %cmp {
+///      print(true)
+///   }
+///
+struct ConditionPropagation : public OpRewritePattern<IfOp> {
```
denialhaag: Maybe we can use our usual docstrings using `@brief` and `@detail` here.

Edit: Ahh, this docstring is taken as is from the original implementation in the SCF dialect. I'm not entirely sure how we want to handle this. 🤔

li-mingbao: Yeah besides the comments, there are also some smaller codestyle differences e.g. how they create operations.
I am not sure if it is worth to change these in every single instance of code that is directly copied without any modifications, especially when we want to copy larger chunks of code like in the proposed `qtensor` dialect for example.
Maybe @burgholzer has an opinion regarding this.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1054,6 +1054,51 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+  // SCF operations
+  //===--------------------------------------------------------------------===//
+
+  /**
+   * @brief Construct an if operation for qubits with linear typing
+   *
+   * @details
+   * Constructs an if operation that takes a bool Value and a range of qubit
+   * Values that are used in the then/else region of this operation. The qubit
+   * values are passed down as blockarguments to each region.
```
burgholzer:
```suggestion
   * values are passed down as block arguments to each region.
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1054,6 +1054,51 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * } else qubits(%arg0 = %q0) {
+   *      qco.yield %arg0
+   * } : {i1, !qco.qubit} -> {!qco.qubit}
+   * ```
+   */
+  ValueRange
+  qcoIf(Value condition, ValueRange qubits,
+        llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> thenBody,
+        std::optional<llvm::function_ref<llvm::SmallVector<Value>(ValueRange)>>
+            elseBody = std::nullopt);
```
burgholzer: If I am not mistaken, there is no need for making this optional and the function ref could simply default to nullptr. I have not checked though.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
burgholzer: When I was reading just this docstring, I assumed that I can only pass in `qco.qubit` values. However, the interface is more flexible than that, right? At least it takes a generic ValueRange. So this could also be used to pass down any other linear types, right? (Like registers)

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,70 @@ def InvOp : QCOOp<"inv", traits =
     let hasVerifier = 1;
 }
 
+//===----------------------------------------------------------------------===//
+// Scf operations
```
```suggestion
// SCF operations
```

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,70 @@ def InvOp : QCOOp<"inv", traits =
...
+            [
+                DeclareOpInterfaceMethods<RegionBranchOpInterface, [
+                "getNumRegionInvocations", "getRegionInvocationBounds",
+                "getEntrySuccessorRegions"]>,
+                SingleBlock,
+                SingleBlockImplicitTerminator<"::mlir::qco::YieldOp">,
+                RecursiveMemoryEffects
+            ]> {
+
+    let summary = "If-then-else operation for qubit types";
```
```suggestion
    let summary = "If-then-else operation for linear (qubit) types";
```

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,70 @@ def InvOp : QCOOp<"inv", traits =
...
+    let description = [{
+        The `qco.if` operation is an if-then-else construct similar to the existing scf.if operation.
```
```suggestion
        The `qco.if` operation is an if-then-else construct similar to the standard scf.if operation.
```

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,70 @@ def InvOp : QCOOp<"inv", traits =
...
+
+        Example:
+        ```mlir
+        %result = qco.if %condition qubits(%arg0 = %q0) {
+            %q1 = qco.h %arg0 : !qco.qubit -> !qco.qubit
+            qco.yield %q1
+         } else qubits(%arg0 = %q0) {
+            qco.yield %arg0
+        } : {i1, !qco.qubit} -> {!qco.qubit}
+        ```
```
burgholzer, line 1239: Does `SCF.if` also print empty else branches in its assembly format?
Not that it is too important, but technically one can infer the empty else branch operations when parsing the IR (if it's not there, it is just a yield operation that forwards the block arguments. Might be nice syntactic sugar.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1209,4 +1210,70 @@ def InvOp : QCOOp<"inv", traits =
...
+    }];
+
+    let arguments = (ins I1:$condition,  Variadic<QubitType>:$qubits);
+    let results = (outs Variadic<QubitType>:$results);
```
burgholzer, line 1243: Ah, it is less flexible then I assumed in the builder. We will probably have to extend/change this once registers are in QCO again, right?
Or do we plan to keep this as "strict" and always mandate extraction from registers before an SCF.if?
This sounds like it could create some overhead down the line if I think about a for loop nested in an if operation.

mlir/unittests/Dialect/QCO/IR/SCF/CMakeLists.txt
burgholzer, line 1: I have not particularly reviewed the test setup here, as it will need some changes anyway once the refined test setup PR is merged.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -0,0 +1,262 @@
...
+      SmallVector<Location>(qubits.size(), odsState.location));
+  odsBuilder.setInsertionPointToStart(&elseBlock);
+  qco::YieldOp::create(odsBuilder, odsState.location,
+                       elseBuilder(elseBlock.getArguments()));
+}
+
+Block* IfOp::thenBlock() { return &getThenRegion().back(); }
+YieldOp IfOp::thenYield() { return dyn_cast<YieldOp>(&thenBlock()->back()); }
+Block* IfOp::elseBlock() { return &getElseRegion().back(); }
+YieldOp IfOp::elseYield() { return dyn_cast<YieldOp>(&elseBlock()->back()); }
```
burgholzer: These are so simple that you may inline them in the tablegen file.
I believe you can also use a plain `cast` over the `dyn_cast` here as it is guaranteed to be a yield.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -672,6 +673,66 @@ QCOProgramBuilder& QCOProgramBuilder::dealloc(Value qubit) {
...
+    std::optional<llvm::function_ref<llvm::SmallVector<Value>(ValueRange)>>
+        elseBody) {
+  checkFinalized();
+
+  auto ifOp = IfOp::create(*this, condition, qubits);
+  // Create the then and else block
+  auto& thenBlock = ifOp->getRegion(0).emplaceBlock();
+  auto& elseBlock = ifOp->getRegion(1).emplaceBlock();
+
+  // Create the blockarguments and add them as valid qubits
```
burgholzer:
```suggestion
  // Create the block arguments and add them as valid qubits
```

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -672,6 +673,66 @@ QCOProgramBuilder& QCOProgramBuilder::dealloc(Value qubit) {
...
+  const auto& ifResults = ifOp->getResults();
+  for (auto [input, output] : llvm::zip_equal(qubits, ifResults)) {
+    updateQubitTracking(input, output);
+  }
+
+  // Remove the inner qubits as valid qubits
+  for (auto [thenOut, elseOut] : llvm::zip_equal(thenResult, elseResult)) {
+    validQubits.erase(thenOut);
+    validQubits.erase(elseOut);
+  }
```
burgholzer, line 741: Is the qubit tracking really correct here? To me, this seems to be missing an association between the input qubits and the block arguments in each branch.
If I remember correctly, you had something related to that in the bigger PR for adding the conversions.
But maybe I am missing something here.

Edit: Just an observation, but this is feeling like it has some overlap with the WireIterator that @MatthiasReumann is implementing. Maybe you can chime in here real quick.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -113,6 +115,52 @@ static void printTargetAliasing(OpAsmPrinter& printer, Operation* /*op*/,
...
+
+  // Parse the else region
+  if (parseTargetAliasing(parser, elseRegion, operands)) {
+    return failure();
+  }
+
+  // Remove duplicate operands
+  llvm::DenseSet<llvm::StringRef> seen;
+  llvm::erase_if(operands,
+                 [&](const auto& op) { return !seen.insert(op.name).second; });
```
burgholzer, line 165: What's the purpose here? Asked differently: which scenario does this tackle?

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+// Copied from
+// https://github.com/llvm/llvm-project/blob/llvmorg-21.1.7/mlir/lib/Dialect/SCF/IR/SCF.cpp
```
burgholzer:
```suggestion
// Adjusted from
// https://github.com/llvm/llvm-project/blob/llvmorg-21.1.8/mlir/lib/Dialect/SCF/IR/SCF.cpp
```
Seems like nothing changed in 21.1.8 in this file.

However, it may be worth to consider https://github.com/llvm/llvm-project/blob/llvmorg-22.1.0-rc3/mlir/lib/Dialect/SCF/IR/SCF.cpp or even https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/SCF/IR/SCF.cpp
There have been quite a few commits to the file. (Maybe also worth checking the tablegen file)

This is one of the parts that gets slightly annoying when we start doing our own thing with the dialects; we don't get upstream changes and improvements for free anymore. This is not the end of the world, but it is something to look out for.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer: Hm. I am a bit torn on this. 
On the one hand, I would like to maintain a consistent code style throughout the project, which would make me lean towards changing quite a few things here.
At the same time, I want to keep it as simple as possible for us to feed in upstream changes. The fewer changes we make here, the easier it is to simply apply a patch, making it easier to maintain.

I have a slight preference for the first option. How about you?

burgholzer (changes requested):
Thanks @li-mingbao for your work on this.
It's quite interesting to me how we are slowly but surely converging to how SCF ops in the QCO dialect should look like.
I gave this an initial review now. Most comments should be easy to address or are simple questions for understanding.
The update to the latest version of the SCF code from upstream may be a bit more work, but hopefully not too bad.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
li-mingbao: I kept it for just qubits for now since we dont have registers yet.
This can be changed later with the modifier operations like `ctrl` and `inv` as the they need to changed too.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
li-mingbao: Since both regions use the same initial inputs here and are parsed twice, this just prevents that they are added twice to the operand list of this operation and only unique inputs are added.

burgholzer: Ah 👍🏼 Gotcha

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
li-mingbao: The `scf.if` operation does not print the else block if it does not return any values. In this case here, since we always return something the else block should be printed.

burgholzer: Even if it is entirely clear how that else block will look like? Very much feels like redundant information in the IR.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: In this case here, I add a new block argument to each region for each input qubit and then add the new block argument as new valid qubit to the set of valid qubits. Inside the regions the new temporary qubits are used for tracking and are deleted afterwards. 
The initial input qubits are just updated to the result of the if operation.

The initial qubits cant be directly updated with the block arguments as they are used in both regions and therefore are used twice.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
li-mingbao: I initially looked at one of the newer releases but there were some compatibility issues, as the implementation of the other classes are also modified compared to the llvm 21.x version that we are currently using. Thats why I used the 21.1.7 version of the implementation for now but I will look into it again.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
burgholzer: Yeah. I figured as much. However, this still seems incomplete to me to some degree. At least for now, I do not see anything that would prevent one from using the outside qubit values in one of the blocks, which would constitute an invalid IR.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer: I see. It's fine to stick with 21.1.7/8 then for now.
We will highly likely update to LLVM 22 once it is out. At that point, this file
will need updates.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: If you mean that you could use the QCO builder to create an invalid IR where one of the input qubits is used directly inside one of the blocks, then this is not possible. 
The input qubits are validated and consumed at the end with the output qubits.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
li-mingbao: I checked again the prints from the `scf.if` operation and as long as the else region should exist, it is printed.
I also checked it with the LLM and it is recommended to print it even though it is trivial to make it obvious that there is an else block and to avoid implicit regions.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
li-mingbao: Replacing the docstrings and some smaller codestyle differences should be fine. For now I just replaced the docstrings and the creation of the operation which were the main differences that I saw.

Maintaining it will be a problem anyway imo as we need to change some parts of the adjusted code when there is a new release.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
burgholzer: Okay. then I might just be overly cautious. Let's leave this as is for now.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: Fair enough 👍🏼 convinced 😌

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer: 👍🏼 sounds like a plan. thanks

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
burgholzer, line 1: I believe you could add one separate test here (via `TEST_F`) that tests the `IfOp` builder that is currently uncovered by tests. Would that make sense?

mlir/unittests/programs/qco_programs.cpp
```diff
@@ -1948,4 +1947,106 @@ void invCtrlSandwich(QCOProgramBuilder& b) {
...
 }
 
+void simpleIf(QCOProgramBuilder& b) {
+  auto q = b.allocQubitRegister(1);
+  auto [measuredQubit, measureResult] = b.measure(q[0]);
+  b.qcoIf(measureResult, measuredQubit, [&](mlir::ValueRange qubits) {
+    auto innerQubit = b.h(qubits[0]);
+    return llvm::SmallVector<mlir::Value>{innerQubit};
+  });
+}
```
burgholzer, line 1958: to make this a "useful" program, one would probably apply an `h` gate before the measurement and an `x` gate inside of the loop. That would be the very simple demonstration of a reset operation. This has some implications on the other programs supposedly.

mlir/unittests/programs/qco_programs.cpp
```diff
@@ -1948,4 +1947,106 @@ void invCtrlSandwich(QCOProgramBuilder& b) {
...
+      measureResult, measuredQubit,
+      [&](mlir::ValueRange qubits) {
+        auto innerQubit = b.h(qubits[0]);
+        return llvm::SmallVector<mlir::Value>{innerQubit};
+      },
+      [&](mlir::ValueRange qubits) {
+        auto innerQubit = b.x(qubits[0]);
+        return llvm::SmallVector<mlir::Value>{innerQubit};
+      });
+}
```
burgholzer, line 1985: These two are a bit redundant as the first one also already has an `else` branch.
I'd propose to
- remove the else branch from the original two qubit example
- turn the `IfElse` program into one that uses two qubits.

If you want to make it a little more realistic, you could apply an X in one and a Z in the other branch.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
li-mingbao: I added a test that creates a circuit without the QCO builders and compare it against one that is created with the QCOProgramBuilder.

mlir/unittests/programs/qco_programs.cpp
li-mingbao: I adjusted the tests, though I am not sure if I understood you correctly.
Feel free to correct me.

mlir/unittests/programs/qco_programs.cpp
```diff
@@ -1948,4 +1947,106 @@ void invCtrlSandwich(QCOProgramBuilder& b) {
...
+    return llvm::SmallVector<mlir::Value>{innerQubit0, innerQubit1};
+  });
+}
+
+void ifElse(QCOProgramBuilder& b) {
+  auto q = b.allocQubitRegister(2);
+  auto q0 = b.h(q[0]);
+  auto [measuredQubit, measureResult] = b.measure(q0);
+  b.qcoIf(
+      measureResult, {q[1]},
```
burgholzer:
```suggestion
  auto q = b.allocQubitRegister(1);
  auto q0 = b.h(q[0]);
  auto [measuredQubit, measureResult] = b.measure(q0);
  b.qcoIf(
      measureResult, {measuredQubit},
```
Not what I originally what I intended, but I think this should be good enough.

burgholzer (approved):
Nice! Let's get this in. I'll quickly apply the one suggestion to the test.

PR #1508, opened 2026-02-10
More comprehensive test infrastructure

taminob:
## Description

This PR introduces a more comprehensive testing infrastructure and fixes multiple minor issues that were discovered during implementation, mainly regarding modifier operations.

## Checklist:

### DIFF
--- a/mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
@@ -15,6 +15,7 @@
 #include <cstdint>
 #include <llvm/ADT/STLExtras.h>
 #include <llvm/ADT/STLFunctionalExtras.h>
+#include <llvm/ADT/SmallPtrSet.h>
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/Support/Casting.h>
 #include <llvm/Support/ErrorHandling.h>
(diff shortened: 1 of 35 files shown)


mlir/include/mlir/Support/IRVerification.h
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+#pragma once
+
+#include <mlir/IR/BuiltinOps.h>
+
+/// Compare two MLIR modules for structural equivalence, allowing permutations
+/// of speculatable operations.
+bool areModulesEquivalentWithPermutations(mlir::ModuleOp lhs,
+                                          mlir::ModuleOp rhs);
```
taminob, line 18:
```suggestion
[[nodiscard]] bool areModulesEquivalentWithPermutations(mlir::ModuleOp lhs,
                                          mlir::ModuleOp rhs);
```

mlir/lib/Dialect/QC/IR/Modifiers/CtrlOp.cpp
```diff
@@ -35,72 +35,67 @@ struct MergeNestedCtrl final : OpRewritePattern<CtrlOp> {
...
-
   LogicalResult matchAndRewrite(CtrlOp op,
                                 PatternRewriter& rewriter) const override {
-    // Require at least one positive control
-    // Trivial case is handled by RemoveTrivialCtrl
-    if (op.getNumControls() == 0) {
-      return failure();
+    auto bodyUnitary = op.getBodyUnitary().getOperation();
+    // Inline ops from empty control modifiers, IdOp and BarrierOp
+    if (op.getNumControls() == 0 || llvm::isa<IdOp, BarrierOp>(bodyUnitary)) {
```
taminob, line 70: I noticed this issue while working on the decomposition: the patterns will also be executed on operations inside modifiers, so doing the identity removal directly here is a good way of avoiding violating the assumtion that a `CtrlOp` will always contain exactly one unitary operation. However, I don't think this is sufficient because we would also need to add every possible pattern here that removes operations (e.g. rotation gates with angle of `0`). In the decomposition, I avoided that issue by having a check for `op.getParentOp()` and do not perform any decompositions inside a control block - but this would be needed for every pattern that potentially removes operations.

Since that is kind of hard to enforce, I'd maybe suggest to relax our definition of `CtrlOp` and that it is possible to not have any operation in the body, which will result in the removal of the `CtrlOp` in the canonicalization here.

mlir/lib/Dialect/QC/IR/Modifiers/CtrlOp.cpp
```diff
@@ -109,7 +104,12 @@ struct CtrlInlineGPhase final : OpRewritePattern<CtrlOp> {
 } // namespace
 
 UnitaryOpInterface CtrlOp::getBodyUnitary() {
-  return llvm::dyn_cast<UnitaryOpInterface>(&getBody()->front());
+  // In principle, the body region should only contain exactly two operations,
+  // the actual unitary operation and a yield operation. However, the region may
+  // also contain constants and arithmetic operations, e.g., created as part of
+  // canonicalization. Thus, the only safe way to access the unitary operation
+  // is to get the second operation from the back of the region.
+  return llvm::dyn_cast<UnitaryOpInterface>(*(++getBody()->rbegin()));
```
taminob, line 112: I think we should at least add an `assert` here for the `dyn_cast` result, because otherwise we could also just use a `llvm::cast`. While working on the decompositions, I already added an `assert` here and it saved me some debugging time

taminob (commented):
@burgholzer I hope you don't mind that I created this PR already - I just wanted to add two minor comments about my experience with the control modifier, because I think I had similar troubles while working on the decomposition PR

burgholzer:
Closing this again as this is nowhere near finished or complete and I want to conserve CI resources here.
There's always Discord and links to lines in the PR for discussions.

mlir/lib/Dialect/QC/IR/Modifiers/CtrlOp.cpp
burgholzer: This is guaranteed by the verifiers. No need for an assert. Maybe we can simply use llvm::cast though.

mlir/lib/Dialect/QC/IR/Modifiers/CtrlOp.cpp
burgholzer: Allowing only a yield and canonicalizing such operations by eliminating them seems reasonable. I'll take that into account.

PR #1510, opened 2026-02-13
Add `WireIterator` for the QCO Dialect

MatthiasReumann:
## Description

This pull request migrates the `WireIterator` to the new QCO dialect. Unit-Tests ensure the correctness of the implementation. 

> [!NOTE]
>  Because we haven't properly defined yet how we want to handle structured control flow and other pull requests tackling this issue (such as #1506) are still open, the current `WireIterator` implementation only supports quantum operations. 

## Checklist:

### DIFF
--- a/mlir/lib/Dialect/QCO/Utils/CMakeLists.txt
+++ b/mlir/lib/Dialect/QCO/Utils/CMakeLists.txt
@@ -18,7 +18,7 @@ add_mlir_dialect_library(
   MLIRQCOInterfacesIncGen
   LINK_LIBS
   PUBLIC
-  Eigen3::Eigen
+  MLIRQCODialect
   DISABLE_INSTALL)
 
 mqt_mlir_target_use_project_options(MLIRQCOUtils)


mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
```diff
@@ -0,0 +1,152 @@
...
+          qubit_ = op.getInputForOutput(qubit_);
+        })
+        .Case<qco::MeasureOp>(
+            [&](qco::MeasureOp op) { qubit_ = op.getQubitIn(); })
+        .Case<qco::ResetOp>([&](qco::ResetOp op) { qubit_ = op.getQubitIn(); })
+        .Default([&](mlir::Operation* op) {
+          report_fatal_error("unknown op in def-use chain: " +
+                             op->getName().getStringRef());
+        });
+  }
```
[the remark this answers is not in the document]
MatthiasReumann: @coderabbitai Would you say you should be able to do something like this? 

```cpp
for (; it != std::default_sentinel; --it) {
    llvm::dbgs() << **it << '\n'; /// Keep for debugging purposes.
 }
```

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
```diff
@@ -0,0 +1,169 @@
...
+
+namespace mlir::qco {
+
+/**
+ * @brief A bidirectional_iterator traversing the def-use chain of a qubit wire.
+ *
+ * The iterator follows the flow of a qubit through a sequence of quantum
+ * operations while respecting the semantics of the respective operation.
+ **/
+class [[nodiscard]] WireIterator {
```
burgholzer, line 24: I have never seen `[[nodiscard]]` being used on a class (only functions). What does this do exactly? Ensure that whenever a WireIterator is returned, it must be used?

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
```diff
@@ -0,0 +1,169 @@
...
+  using difference_type = std::ptrdiff_t;
+  using value_type = mlir::Operation*;
+
+  WireIterator() : op_(nullptr), qubit_(nullptr), isSentinel_(false) {}
+  explicit WireIterator(mlir::Value qubit)
+      : op_(qubit.getDefiningOp()), qubit_(qubit), isSentinel_(false) {}
+
+  /// @returns the qubit the iterator points to.
+  [[nodiscard]] mlir::Value qubit() const {
+    // A deallocation doesn't have an OpResult.
```
burgholzer: Neither does a `YieldOp`, but I suppose this is not a problem here?

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
```diff
@@ -0,0 +1,169 @@
...
+  [[nodiscard]] mlir::Operation* operator*() const { return operation(); }
+
+  WireIterator& operator++() {
+    forward();
+    return *this;
+  }
+
+  WireIterator operator++(int) {
+    auto tmp = *this;
+    ++*this;
```
burgholzer:
```suggestion
    operator++();
```
Just a suggestion, but might look a little nicer.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  WireIterator& operator--() {
+    backward();
+    return *this;
+  }
+
+  WireIterator operator--(int) {
+    auto tmp = *this;
+    --*this;
```
burgholzer:
```suggestion
    operator--;
```
Same as above. Purely optional though.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
burgholzer, line 1: There are two upcoming changes that will probably imply changes here: the introduction of registers, as well as the handling of SCF operations. Just that we are aware of this.

mlir/unittests/Dialect/QCO/Utils/test_wireiterator.cpp
```diff
@@ -0,0 +1,135 @@
...
+  builder.initialize();
+  const auto q00 = isDynamic ? builder.allocQubit() : builder.staticQubit(0);
+  const auto q10 = isDynamic ? builder.allocQubit() : builder.staticQubit(1);
+  const auto q01 = builder.h(q00);
+  const auto [q02, q11] = builder.cx(q01, q10);
+  const auto [q03, c0] = builder.measure(q02);
+  const auto q04 = builder.reset(q03);
+  builder.dealloc(q04);
+  builder.dealloc(q11);
+  [[maybe_unused]] auto module = builder.finalize();
```
burgholzer, line 56:
```suggestion
  std::ignore = builder.finalize();
```
is more idiomatic I suppose.

mlir/unittests/Dialect/QCO/Utils/test_wireiterator.cpp
burgholzer, line 1: I am not yet 100% sure how this best fits into the test infrastructure of #1513, but I am pretty sure we will find a place for it.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add bidirectional iterator that traverses the def-use chain of a qubit value for the QCO MLIR dialect ([#1510]) ([**@MatthiasReumann**])
```
burgholzer: I would have a slight tendency to also put this into the larger changelog entry below as it constitutes a fundamental aspect of the QCO dialect in my opinion.
At some point, we have to get this out in a release and start with finer-grained entries again.

burgholzer (changes requested):
Hey @MatthiasReumann 👋🏼 
Thanks for the quick action on this! Great to see. I only have some minor comments here.
I was wondering whether we could already use the wire iterator somewhere, e.g., the QCO builder, so that it doesn't stand as alone as it does right now. But it is not clear to me whether that is even reasonable. A short comment would be appreciated.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
MatthiasReumann: > Ensure that whenever a WireIterator is returned, it must be used?

Yes, exactly. Or explicitly ignored:

```c++
std::ignore = WireIterator(q);
```

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
MatthiasReumann: True! Until we support SCF this isn't a problem.

mlir/unittests/Dialect/QCO/Utils/test_wireiterator.cpp
MatthiasReumann: I don't know why I keep forgetting `std::ignore`. That's the second time you need to remind me of that 🫠 
Thanks!

CHANGELOG.md
MatthiasReumann: Moved it to the "initial infrastructure" changelog

mlir/unittests/Dialect/QCO/Utils/test_wireiterator.cpp
MatthiasReumann: Ah. Now I remember why I didn't do that... If you use `std::ignore` this leads to a segfault.

burgholzer: AH. I now also remember why that is the case.
Since the finalize call returns an owning reference, the module goes out of scope and is destroyed immediately if it is not materialized as an actual variable.
Sorry for the noise.

MatthiasReumann: No need to apologize. I could have remembered that also 😅

MatthiasReumann:
@burgholzer Thanks for the review. I've changed the necessary bits. 

Regarding _using_ the WireIterator: I've quickly glanced over the `QCOProgramBuilder` and didn't find any useful place to utilize the WireIterator. However, as I am planning to reimplement the A* mapping pass next, it will be in use very soon 👀

burgholzer:
> @burgholzer Thanks for the review. I've changed the necessary bits.
> 
> Regarding _using_ the WireIterator: I've quickly glanced over the `QCOProgramBuilder` and didn't find any useful place to utilize the WireIterator. However, as I am planning to reimplement the A* mapping pass next, it will be in use very soon 👀

Nice, that was quick.
I will try to get in #1513 asap so that this is unblocked (as far as one can consider it blocked at the moment).

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
burgholzer, line 1: Since I had my fair share of wrestling with compile times in #1513, I was wondering if you have considered not only defining this in a header, but moving the implementation to a `.cpp` file?
I believe this could be fairly beneficial here and should be straight-forward to do as there are no templates involved.

mlir/unittests/Dialect/QCO/Utils/test_wireiterator.cpp
```diff
@@ -0,0 +1,135 @@
...
+using namespace mlir;
+
+namespace {
+class WireIteratorTest : public testing::TestWithParam<bool> {
+protected:
+  void SetUp() override {
+    DialectRegistry registry;
+    registry
+        .insert<qco::QCODialect, arith::ArithDialect, cf::ControlFlowDialect,
+                func::FuncDialect, scf::SCFDialect, LLVM::LLVMDialect>();
```
burgholzer: Similar to the other comment, I think it would be good to be a little more conservative here with the dialects being loaded.
At the moment, I suppose the only ones that are needed are teh QCO, the Func, and the Arith one. The others can be removed here, from the headers, and from the link line in CMake.

mlir/unittests/Dialect/QCO/Utils/CMakeLists.txt
burgholzer, line 1: I think it should be fine to integrate these here for now. Given how there is no direct usage of the iterator in the compiler pipeline (yet), keeping it separate is kind of the only option.

burgholzer (changes requested):
Two small requests and a comment, then this should be good to go 🚀

mlir/unittests/Dialect/QCO/Utils/test_wireiterator.cpp
MatthiasReumann: Good catch.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
MatthiasReumann: Good idea. Added a `WireIterator.cpp`. 

Could you please review the respective CMakeLists.txt? Any suggestions for improvement or is it fine?

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
```diff
@@ -0,0 +1,89 @@
...
+  void backward();
+
+  mlir::Operation* op_;
+  mlir::Value qubit_;
+  bool isSentinel_;
+};
+
+static_assert(std::bidirectional_iterator<WireIterator>);
+static_assert(std::sentinel_for<std::default_sentinel_t, WireIterator>,
+              "std::default_sentinel_t must be a sentinel for WireIterator.");
```
burgholzer: I think these static asserts can be moved to the source file. But I don't know if that safes an import.

mlir/lib/Dialect/QCO/Utils/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  MLIRQCOUtils
+  ${UTILS_CPP}
+  ADDITIONAL_HEADER_DIRS
+  ${PROJECT_SOURCE_DIR}/mlir/include/mlir/Dialect/QCO
+  DEPENDS
+  MLIRQCOOpsIncGen
+  MLIRQCOInterfacesIncGen
+  LINK_LIBS
+  PUBLIC
+  Eigen3::Eigen
```
burgholzer:
```suggestion
```
I don't think that you need to link against eigen for the utils file that you wrote, do you?
But potentially against some of the MLIR libraries 🤷🏽‍♂️ the QCODialect one for example

burgholzer (approved):
LGTM with two minor comments.

mlir/lib/Dialect/QCO/Utils/CMakeLists.txt
MatthiasReumann: Wouldn't let me compile without Eigen, but since Eigen is included in `MLIRQCODialect` including it should be the way to go.

burgholzer: Ah, yeah. Because the interfaces of QCO for unitary matrices need eigen. We may want to split this off at some point to avoid having to link against eigen in every QCO related code.

PR #1518, opened 2026-02-16
🤖 Adding AGENTS.md repository instructions for AI tooling

### LINKED ISSUE #1459: 📝 Add repository instructions for AI Tooling
> [!NOTE]
> 
> This issue is part of unitaryDESIGN 2026; note that you have to be [registered](https://unitary.design/register/) to complete this issue. Learn more about the PR review process [here](https://unitary.design/participant-guide/). 
>
> A message from unitaryDESIGN about [AI Slop](https://en.wikipedia.org/wiki/AI_slop): "While we are open to collaboration with LLMs for unitaryDESIGN, please note that fully AI-generated PRs are not acceptable. It is up to the discretion of the individual maintainers whether or not LLM-generated PRs are the right fit for their issues, and they have the right to reject those that appear fully AI-generated."
>
> Learn more about unitaryDESIGN rules and opportunities [here](https://unitary.design/rules/)!

One of the biggest challenges when using AI tools such as Claude Code, Cursor, Copilot, etc. is how to best instruct/prompt them and make them understand the repository, how to perform certain tasks, and what practices should be followed.
While there does not seem to be a real standard format for providing such instructions (yet), there are a handful of guides on how to best instruct agents/AI systems, such as
- <https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions?tool=jetbrains> (`.github/copilot-instructions.md` and `.github/prompts`, e.g., for changelog creation or release preparation)
- <https://agents.md> (`AGENTS.md`)
- <https://www.reddit.com/r/ClaudeAI/comments/1mecx5t/how_we_structure_our_claudemd_file_and_why/> and <https://www.anthropic.com/engineering/claude-code-best-practices> (`CLAUDE.md`)

From what I could gather, `AGENTS.md` seems to be emerging as a quasi standard and all the tooling that we are using consumes it.

The goal in this issue would be to curate such a file for this project, clearly emphasizing the commands that shall be executed for building, testing, listing, etc. as well as the best practices for development and the things that we value.
I personally liked <https://github.com/astral-sh/ruff/blob/main/CLAUDE.md> as an inspiration, and the suggestion from the blogpost above to run the [prompt improver](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prompt-improver) on the respective file.

The [documentation of CodeRabbit](https://docs.coderabbit.ai/reference/configuration#param-file-patterns) seems to indicate that it also natively supports AGENTS.md, which would also be great for reviews.

CodeMaverick2:
## Description - 
Summary

-   Add AGENTS.md at the repository root to provide AI coding tools with project-specific instructions
-   Covers build commands, testing workflows, linting/formatting, documentation, and code conventions
-   Every command verified against actual project configuration files (CMakeLists.txt, pyproject.toml, noxfile.py,  .pre-commit config.yaml, .clang-format, .clang-tidy, .cmake-format.yaml, .editorconfig)

Format
- AGENTS.md follows the agents.md standard — the emerging universal format consumed by Codex, Cursor, CodeRabbit, Jules, Windsurf, Aider, and more.
- CodeRabbit natively scans AGENTS.md for PR reviews.

Fixes #1459 
## Checklist:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,7 +74,7 @@
 - Update @file:CHANGELOG.md and @file:UPGRADING.md when changes are user-facing, breaking, or otherwise noteworthy.
 - Stub files (`.pyi`) in `python/mqt/core/` are **auto-generated** by nanobind's stubgen. NEVER edit `.pyi` files manually.
 - If C++ bindings are added or modified (files in `bindings/`), stubs need to be regenerated via `uvx nox -s stubs`.
-- Never modify files that start with "This file has been generated from an external template. Please do not modify it directly."
+- Never modify files that start with "This file has been generated from an external template. Please do not modify it directly." These files are managed by [the MQT templates action](https://github.com/munich-quantum-toolkit/templates) and changes will be overwritten.
 - Prefer fixing reported warnings over suppressing them (e.g., with `# noqa` comments for ruff). Only add ignore rules when necessary and document why.
 - Prefer fixing typing issues reported by `ty` before adding suppression comments (`# ty: ignore[code]`). Suppressions are sometimes necessary for incompletely typed libraries (e.g., Qiskit).
 - Ruff is configured in `pyproject.toml` with `select = ["ALL"]`. All rules are enabled by default, with a small set selectively disabled.


AGENTS.md
```diff
@@ -0,0 +1,224 @@
+# MQT Core
+
+MQT Core is the backbone of the [Munich Quantum Toolkit (MQT)](https://mqt.readthedocs.io/) — a C++20 and Python library providing core data structures and algorithms for quantum computing design automation. Key components include an intermediate representation (IR) for quantum computations, a state-of-the-art decision diagram (DD) package, a ZX-calculus engine, an OpenQASM 3.0 parser, MLIR-based quantum compilation dialects, and a quantum device management interface (QDMI).
```
denialhaag: I would maybe say that this repository only includes _implementations_ of QDMI, as QDMI itself lives here: https://github.com/Munich-Quantum-Software-Stack/QDMI

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+ctest --test-dir build -R test_dd_package
+
+# Run with verbose output on failure
+ctest --test-dir build --output-on-failure
+
+# Run a specific test binary directly for more control
+./build/test/dd/mqt-core-dd-test
+```
+
+Tests are organized by component under `test/`: `algorithms/`, `circuit_optimizer/`, `dd/`, `ir/` (includes QASM3 parser tests), `na/`, `zx/`, `qdmi/`, `qir/`, `fomac/`. Prefer running targeted tests over the full suite during development.
```
denialhaag: This should also mention that the MLIR tests are located at `mlir/unittests`.

AGENTS.md
denialhaag, line 1: I would propose that we add a line break after every sentence to make future diffs easier to read.

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+In-source builds are prevented. External dependencies (nlohmann_json, Boost.Multiprecision, Google Test, spdlog, QDMI, Eigen) are fetched automatically via CMake FetchContent if not found on the system.
+
+## Running C++ Tests
+
+```bash
+# Run the full C++ test suite
+ctest --test-dir build
+
+# Run a single test by name (regex match)
+ctest --test-dir build -R test_dd_package
```
denialhaag: There is no such test:

```
ctest --test-dir build -R test_dd_package
Test project <...>/core/build
No tests were found!!!
```

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+
+```bash
+# Full test suite (via nox, handles environment setup)
+uvx nox -s tests
+
+# Single test file (if environment is already set up via uv sync)
+uv run pytest test/python/dd/test_dd_package.py
+
+# Single test function
+uv run pytest test/python/dd/test_dd_package.py::test_function_name
```
denialhaag: I would maybe make this a valid example.

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+4. **Type checking**: ty check (Python)
+
+### Ruff configuration
+
+The ruff config in `pyproject.toml` uses `select = ["ALL"]` — **all rules are enabled by default**. Only a small set of rules are selectively disabled. Do not add new rules; only disable specific ones if necessary and document why. Line length is 120. Google-style docstrings are enforced.
+
+### C++ formatting and linting
+
+- clang-format uses LLVM base style with `PointerAlignment: Left` and C++20 standard
+- clang-tidy enforces extensive checks from bugprone, cppcoreguidelines, google, misc, modernize, performance, portability, and readability categories
```
denialhaag: It might make sense to mention the configuration files for these tools.

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+
+- Line endings: LF
+- Charset: UTF-8
+- Trailing whitespace: trimmed (except Markdown)
+- Final newline: required
+
+## Project Layout
+
+```text
+include/mqt-core/     C++ public headers (organized by component)
```
denialhaag:
```suggestion
include/mqt-core/      C++ public headers (organized by component)
```

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+## Type Stubs
+
+```bash
+# Regenerate Python type stubs from nanobind bindings
+uvx nox -s stubs
+```
+
+Stub files (`.pyi`) in `python/mqt/core/` are **auto-generated** by nanobind's stubgen. NEVER edit `.pyi` files manually — they will be overwritten. If bindings change, regenerate stubs and commit the updated `.pyi` files.
+
+## Code Style
```
denialhaag: Maybe it makes sense to combine this section with the "Linting and Formatting" section above, as the two sections have quite some overlap right now.

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+```text
+include/mqt-core/     C++ public headers (organized by component)
+src/                   C++ source files (mirrors include/ structure)
+python/mqt/core/       Python package and auto-generated .pyi stubs
+bindings/              nanobind C++ → Python bindings
+test/                  C++ tests (Google Test)
+test/python/           Python tests (pytest)
+cmake/                 CMake modules and helpers
+docs/                  Sphinx documentation source
+mlir/                  MLIR quantum compilation dialects (QC, QCO, QIR)
```
denialhaag: We do not have a dedicated QIR dialect. There exists only a QIR builder that generates programs in the LLVM dialect.

AGENTS.md
```diff
@@ -0,0 +1,224 @@
...
+Components: `ir` (intermediate representation), `dd` (decision diagrams), `zx` (ZX-calculus), `algorithms`, `circuit_optimizer`, `qasm3` (OpenQASM 3.0 parser), `na` (neutral atoms), `qdmi` (device management), `qir` (QIR runtime), `fomac` (formal methods), `mlir` (MLIR dialects).
+
+## Platform Support
+
+- **Operating systems**: Linux (Ubuntu 24.04), macOS (14, 15), Windows (2022, 2025)
+- **Architectures**: x86_64 and arm64
+- **C++ compilers**: GCC 14+, Clang 20+, MSVC (latest), AppleClang 17+
+- **CI**: GitHub Actions with reusable workflows from `munich-quantum-toolkit/workflows`
+
+## Important Notes
```
denialhaag: This section should also mention that some files are created by [our templating action](https://github.com/munich-quantum-toolkit/templates). These files must not be edited, as any changes will be overwritten. The files are generally marked by an initial comment (see, e.g., [`release-drafter.yml`](https://github.com/munich-quantum-toolkit/core/blob/main/.github/release-drafter.yml)).

denialhaag (commented):
Thanks a lot for addressing my comments! This looks good from my side now. 🙂 

I propose leaving this PR open for a bit longer so that other contributors (in particular, @burgholzer) can also review the file.

CodeMaverick2:
> Thanks again!
> 
> @burgholzer would appreciate your review whenever you get a chance

@burgholzer  Bumping this in case missed Whenever you get time

AGENTS.md
```diff
@@ -0,0 +1,245 @@
+# MQT Core
+
+MQT Core is the backbone of the [Munich Quantum Toolkit (MQT)](https://mqt.readthedocs.io/) — a C++20 and Python library providing core data structures and algorithms for quantum computing design automation.
+Key components include an intermediate representation (IR) for quantum computations, a state-of-the-art decision diagram (DD) package, a ZX-calculus engine, an OpenQASM 3.0 parser, MLIR-based quantum compilation dialects, and implementations of the [quantum device management interface (QDMI)](https://github.com/Munich-Quantum-Software-Stack/QDMI).
```
burgholzer: Out of all the buzzwords that we have in our documentation, QIR is missing.
As for the mention of MLIR, I think this should be more generic. We do not only provide compilation dialects; we provide an entire quantum compiler infrastructure (called MQT Compiler Collection or `mqt-cc`)

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Tech Stack
+
+| Area                    | Tool                      | Notes                                       |
+| ----------------------- | ------------------------- | ------------------------------------------- |
+| C++ standard            | C++20                     | Minimum required                            |
+| Build system            | CMake 3.24+               | Up to 4.2 supported                         |
+| Python                  | 3.10+                     | Supports 3.10 through 3.14                  |
```
burgholzer: This should probably mention that we also support the free-threading version of Python on 3.14.
And, to some degree, that we build Stable ABI wheels against the Python 3.12 Stable ABI (enabled via nanobind).

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+| MLIR/LLVM               | 21+                       | Required only when `BUILD_MQT_CORE_MLIR=ON` |
```
burgholzer: That option defaults to `ON`, so I am not sure how valuable the mention is.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+| Python package manager  | **uv** (>=0.5.20)         | NEVER use pip                               |
+| Python build backend    | scikit-build-core         | With nanobind for bindings                  |
+| Python linter/formatter | **ruff**                  | NEVER use pylint, black, flake8, or isort   |
+| Python type checker     | **ty**                    | NEVER use mypy                              |
+| Python bindings         | **nanobind** (~2.11.0)    | NOT pybind11                                |
```
burgholzer:
```suggestion
| Python bindings         | **nanobind**    | NOT pybind11                                |
```
Feels like this would be cumbersome to keep updated

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+| C++ formatter           | clang-format (LLVM style) |                                             |
```
burgholzer:
```suggestion
| C++ formatter           | clang-format | Config in .clang-format                                             |
```
or similar. maybe include the LLVM style in the notes column.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+| C++ linter              | clang-tidy                |                                             |
```
burgholzer: maybe worth noting where the config file is (or rather the config files; we have some in subdirectories)

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+| CMake formatter         | cmake-format              |                                             |
```
burgholzer:
```suggestion
```
Don't think this is really worth mentioning. It will be run automatically as part of the lint job.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+Key CMake options:
+
+| Option                       | Default                | Description                                                |
+| ---------------------------- | ---------------------- | ---------------------------------------------------------- |
+| `BUILD_MQT_CORE_TESTS`       | ON (if master project) | Build C++ test suite                                       |
+| `BUILD_MQT_CORE_BINDINGS`    | OFF                    | Build Python bindings via nanobind                         |
+| `BUILD_MQT_CORE_MLIR`        | ON                     | Build MLIR quantum compilation dialects (requires LLVM 21) |
+| `BUILD_MQT_CORE_SHARED_LIBS` | OFF                    | Build as shared libraries                                  |
+| `BUILD_MQT_CORE_BENCHMARKS`  | OFF                    | Build evaluation benchmarks                                |
```
burgholzer:
```suggestion
```
That part can be removed, as it is no longer really relevant.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+| `ENABLE_COVERAGE`            | OFF                    | Enable code coverage instrumentation                       |
+
+In-source builds are prevented.
```
burgholzer:
```suggestion
```
Feels superfluous. This file should be as concise as possible as every single model will read it.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+External dependencies (nlohmann_json, Boost.Multiprecision, Google Test, spdlog, QDMI, Eigen) are fetched automatically via CMake FetchContent if not found on the system.
```
burgholzer: I think this should rather be turned into a separate section on how dependencies in C++ are handled.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Run a single test by name (regex match)
+ctest --test-dir build -R mqt-core-dd-test
+
+# Run with verbose output on failure
+ctest --test-dir build --output-on-failure
+
+# Run a specific test binary directly for more control
+./build/test/dd/mqt-core-dd-test
+```
```
burgholzer: This should include the mention of running unittests with a specific label, such as the `mqt-mlir-unittests` label added in #1513

CodeMaverick2:
> Thanks for the ping. This is still in my pile of reviews. I'll try to get to it today 👍🏻

Thanks for the update! Really appreciate you fitting this into your review pile

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+Tests are organized by component under `test/`: `algorithms/`, `circuit_optimizer/`, `dd/`, `ir/` (includes QASM3 parser tests), `na/`, `zx/`, `qdmi/`, `qir/`, `fomac/`.
```
burgholzer: Is this information really useful?

AGENTS.md
burgholzer: I don't 100% know for sure, but I would guess, that this and the following sections would be better structured as multiple separate code cells, where the current comments are the regular text, always followed by the respective command to execute. This holds for all such sections below.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+Prefer running targeted tests over the full suite during development.
+
+## Building the Python Package
+
+```bash
+# Install all dev dependencies and build the package in one step
+uv sync
+
+# Build only (produces wheel)
+uv build
```
burgholzer: This does not only build a wheel. It also builds an sdist.
`uv sync` should be enough here.
I generally think it is preferable to have non-ambiguous advice on the commands to use.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+Build artifacts are placed in `build/{wheel_tag}/{build_type}/`.
+The Python package uses scikit-build-core as its build backend with nanobind for C++ bindings.
```
burgholzer: Redundant information with the table on top

```suggestion
```

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+IMPORTANT: Always use `uv` for dependency management.
+Never use `pip install`, `pip install -e .`, or any pip-based workflow.
```
burgholzer: This also feels redundant.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+uvx nox -s tests -- --cov
+
+# Test with minimum dependency versions
+uvx nox -s minimums
+
+# Test against latest Qiskit from git
+uvx nox -s qiskit
+```
+
+Python tests live in `test/python/` and use pytest with parallel execution (pytest-xdist).
```
burgholzer: feels redundant

```suggestion
```

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+# With coverage
+uvx nox -s tests -- --cov
+
+# Test with minimum dependency versions
+uvx nox -s minimums
+
+# Test against latest Qiskit from git
+uvx nox -s qiskit
+```
```
burgholzer: This should definitely mention how to only run specific python versions as the regular nox commands will typically run on all supported Python versions. Both `uv` and `nox` have settings for that.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+```bash
+# Run ALL pre-commit checks (recommended before committing)
+uvx prek run -a
+
+# Alternative via nox
+uvx nox -s lint
+
+# Python type checking only
+uv run ty check
+```
```
burgholzer: Tied to a comment above: Instead of offering alternatives, a single option shall be used.
specific `prek/pre-commit` checks can also be specified when using `nox`, e.g., the last line above is also possible as
```
uvx nox -s lint ty-check
```

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Pre-commit hook priority order
+
+1. **Validation**: merge conflicts, pyproject schema, GitHub workflows, ReadTheDocs config, capitalization, typos, license headers, uv lockfile sync
+2. **Formatting**: clang-format (C++), cmake-format (CMake), prettier (YAML/MD/JSON), ruff format (Python)
+3. **Linting**: ruff check (Python)
+4. **Type checking**: ty check (Python)
```
burgholzer: Does this add any value to the agent?
I would guess that the agent just runs the checks.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+### C++
+
+- Formatting: LLVM-based (see `.clang-format`)
+- Linting: clang-tidy (see `.clang-tidy`).
+  To include nanobind bindings in clang-tidy analysis, configure with `-DBUILD_MQT_CORE_BINDINGS=ON`.
+- Naming: `camelBack` for functions, methods, variables, parameters; `CamelCase` for classes, structs, enums; `UPPER_CASE` for global/static constants; `lower_case` for namespaces
```
burgholzer: This feels like it would get easily out of date. Wouldn't it be better to directly point to the config file and have the agent read it, if it deems it necessary?

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- Type checker: ty (not mypy)
+- Docstrings: Google style
+- Indentation: 4 spaces
+- Imports: sorted by ruff (isort rules), `from __future__ import annotations` enforced
+
+The ruff config in `pyproject.toml` uses `select = ["ALL"]` — **all rules are enabled by default**.
+Only a small set of rules are selectively disabled.
+Do not add new rules; only disable specific ones if necessary and document why.
+Line length is 120.
+Google-style docstrings are enforced.
```
burgholzer: This lists the same information twice. Would need to be simplified.
Most of these details also feel like they are not particularly important to an AI agent as they are things that are imposed by running prek.
One of the important parts here is that we configure ruff through pyproject.toml and enable `"ALL"`.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- Line width: 100
+- Formatter: cmake-format (see `.cmake-format.yaml`)
+- Indentation: 2 spaces
+
+### General
+
+- Line endings: LF
+- Charset: UTF-8
+- Trailing whitespace: trimmed (except Markdown)
+- Final newline: required
```
burgholzer: These are all enforced by the prek checks. there is no need to spell them out.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+Do not add new rules; only disable specific ones if necessary and document why.
+Line length is 120.
+Google-style docstrings are enforced.
+
+### CMake
+
+- Keywords: `UPPER_CASE`
+- Line width: 100
+- Formatter: cmake-format (see `.cmake-format.yaml`)
+- Indentation: 2 spaces
```
burgholzer: This is all enforced by prek. No AI needs to know about this. They only need to know that they are supposed to let the linter run eventually.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+## Type Stubs
+
+```bash
+# Regenerate Python type stubs from nanobind bindings
+uvx nox -s stubs
+```
+
+Stub files (`.pyi`) in `python/mqt/core/` are **auto-generated** by nanobind's stubgen.
+NEVER edit `.pyi` files manually — they will be overwritten.
+If bindings change, regenerate stubs and commit the updated `.pyi` files.
```
burgholzer: This is an example of a section that feels really useful as it contains actionable advice that is fairly hard to directly get from the repository. More of this is needed throughout the file.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+test/                   C++ tests (Google Test)
+test/python/            Python tests (pytest)
+cmake/                  CMake modules and helpers
+docs/                   Sphinx documentation source
+mlir/                   MLIR quantum compilation dialects (QC, QCO) and QIR builder
+eval/                   Benchmarking and evaluation
+json/                   Device configuration files (NA, SC)
+```
+
+Components: `ir` (intermediate representation), `dd` (decision diagrams), `zx` (ZX-calculus), `algorithms`, `circuit_optimizer`, `qasm3` (OpenQASM 3.0 parser), `na` (neutral atoms), `qdmi` (device management), `qir` (QIR runtime), `fomac` (formal methods), `mlir` (MLIR dialects).
```
burgholzer: This could benefit from more of a file tree structure, to give a more detailed overview. On the other hand, I do not know how important this really is in the agents file, as this is essentially the result of an `ls -lA`, which is cheap and always up to date.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Platform Support
+
+- **Operating systems**: Linux (Ubuntu 24.04), macOS (14, 15), Windows (2022, 2025)
+- **Architectures**: x86_64 and arm64
+- **C++ compilers**: GCC 14+, Clang 20+, MSVC (latest), AppleClang 17+
+- **CI**: GitHub Actions with reusable workflows from `munich-quantum-toolkit/workflows`.
+  Regular CI runs a subset of the matrix (Ubuntu/GCC, macOS/AppleClang, Windows/MSVC).
+  The full compiler/OS matrix is tested when a PR has the `extensive-cpp-ci` label.
```
burgholzer: This only covers the C++ aspects and leaves out the Python aspects that we test for platform support.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Important Notes
+
+- **License headers are required** on all source files: MIT license, SPDX identifier.
+  Managed by `mz-lictools`.
+  The license-tools pre-commit hook will fail if headers are missing.
+- **Capitalization matters**: use the exact capitalizations "nanobind", "CMake", "ccache", "GitHub", "NumPy", "pytest", "MQT", and "TUM".
+  A pre-commit hook (`disallow-caps`) enforces this and will reject common misspellings.
```
burgholzer: No need to mention these. The linter will either fix these or complain.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- **MLIR caveats**: MLIR is disabled automatically on macOS with GCC (ABI incompatibility) and with AppleClang < 17 (incomplete C++20 support).
```
burgholzer: This information does not feel necessary.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- **Stub files**: Never edit `.pyi` files by hand.
+  Always regenerate with `uvx nox -s stubs`.
```
burgholzer: Redundant with information above

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- **uv lockfile**: The `uv.lock` file must stay in sync.
+  The pre-commit hook `uv-lock` checks this.
+  Run `uv lock` if you change dependencies in `pyproject.toml`.
```
burgholzer: This is ensured by a prek check and does not need to be mentioned explicitly

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- **Templated files**: Some files are generated by the [templating action](https://github.com/munich-quantum-toolkit/templates) and must not be edited directly, as changes will be overwritten.
+  These files are marked by an initial comment such as `# This file has been generated from an external template. Please do not modify it directly.`
+- **Pre-commit hooks must pass** before submitting pull requests.
+  Run `uvx prek run -a` to check locally.
```
burgholzer: This should supposedly be phrased differently. The checks should not only pass before submitting a PR. They are generally supposed to pass. And running prek is not only for checking, but also for auto-fixing and formatting, which also provides automatic feedback for fixing issues.

AGENTS.md
```diff
@@ -0,0 +1,245 @@
...
+- **Test your changes**: all changes must be tested.
+  Write C++ tests with Google Test in `test/`, Python tests with pytest in `test/python/`.
```
burgholzer: This feels like something to go into the respective C++ and Python sections

AGENTS.md
burgholzer: Are you sure that you addressed all of the comments and have not simply resolved some? This one is definitely not done yet. Please carefully check.

CodeMaverick2: Sorry might have missed 1-2

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_MQT_CORE_TESTS=ON
+cmake --build build
+```
+
+For a Debug build:
+
+```bash
+cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DBUILD_MQT_CORE_TESTS=ON
+cmake --build build
+```
```
burgholzer: The `-DBUILD_MQT_CORE_TESTS=ON` flag is redundant, as this is on by default anyway.
Furthermore, the current instructions will not work with multi-config generators like MSVC, where a configuration also needs to be passed to the build stage. You can find examples of the right commands in the documentation or our CI/CD setup.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+
+Key CMake options:
+
+| Option                       | Default                | Description                                                   |
+| ---------------------------- | ---------------------- | ------------------------------------------------------------- |
+| `BUILD_MQT_CORE_TESTS`       | ON (if master project) | Build C++ test suite                                          |
+| `BUILD_MQT_CORE_BINDINGS`    | OFF                    | Build Python bindings via nanobind                            |
+| `BUILD_MQT_CORE_MLIR`        | ON                     | Build MLIR quantum compiler infrastructure (requires LLVM 21) |
+| `BUILD_MQT_CORE_SHARED_LIBS` | OFF                    | Build as shared libraries                                     |
```
burgholzer: Might be worth to add a few details to the description of this setting as it is mostly used internally as part of the Python packaging build and generally should not be used by people directly (if not for debugging)

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+| `ENABLE_COVERAGE`            | OFF                    | Enable code coverage instrumentation                          |
+
+### C++ Dependencies
+
+C++ dependencies are managed via CMake's `FetchContent` (configured in `cmake/ExternalDependencies.cmake`).
+Each dependency is first searched on the system via `find_package`; if not found, it is automatically fetched from its upstream source.
```
burgholzer: This is not true as stated here. As the above statement indicates, we are generally using FetchContent.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+Dependencies include nlohmann_json, Boost.Multiprecision, Google Test, spdlog, QDMI, and Eigen (MLIR only).
```
burgholzer: It may be worth to include links to the respective projects?
Furthermore, `Eigen` is currently the only dependency that has some extra context attached to it, while some others are also only pulled in in specific contexts.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+
+```bash
+ctest --test-dir build -L mqt-mlir-unittests
+```
+
+Run a specific test binary directly:
+
+```bash
+./build/test/dd/mqt-core-dd-test
+```
```
burgholzer: Related to the comment above: These commands will not properly work on multi-configuration generators such as MSVC. Please adjust accordingly.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+
+MLIR tests are located at `mlir/unittests/`.
+Prefer running targeted tests over the full suite during development.
+All C++ tests use Google Test. Write new tests in `test/` (or `mlir/unittests/` for MLIR).
+
```
burgholzer: Redundant information on the location of the MLIR unittests. Would be great to deduplicate.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+## Building the Python Package
+
+Install all dev dependencies and build the package:
+
+```bash
+uv sync
+```
+
+Build artifacts are placed in `build/{wheel_tag}/{build_type}/`.
```
burgholzer: Some context on what to expect for `wheel_tag` and `build_type` would probably be good here.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+uvx nox -s tests
+```
+
+Run tests for a specific Python version:
+
+```bash
+uvx nox -s tests-3.12
+```
+
+Single test file (if environment is already set up via `uv sync`):
```
burgholzer: The statement in the parenthesis here is incorrect. Please read up on the semantics of  `uv run`

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+
+```bash
+uvx nox -s minimums
+```
+
+Test against latest Qiskit from git:
+
+```bash
+uvx nox -s qiskit
+```
```
burgholzer: Both of these nox commands also take optional python version specifiers to only run with a selected version of Python. This should probably be added. Similarly, the next paragraph should probably include a sentence that it is preferable to run a single Python version during iteration and only run the full command with all Python versions for final checks.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+
+Prefer running individual test files during development.
+Write new Python tests with pytest in `test/python/`.
+
+## Linting, Formatting, and Code Style
+
+Run all pre-commit checks (auto-fixes formatting and linting issues):
```
burgholzer: The pre-commit checks do not automatically fix _all_ linting issues.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+Run a specific check by hook ID (e.g., type checking):
+
+```bash
+uvx nox -s lint -- ty-check
+```
+
+### C++
+
+- Formatting and naming conventions are enforced by clang-format and clang-tidy; see `.clang-format` and `.clang-tidy` for the full configuration.
+- To include nanobind bindings in clang-tidy analysis, configure with `-DBUILD_MQT_CORE_BINDINGS=ON`.
```
burgholzer: This should probably mention that this requires a virtual environment being activated with nanobind installed so that the build can find it. (such as the environment generated by a `uv sync` call; or, minimalistically a `uv sync --only-group build` call; potentially with `--inexact` to not overwrite an existing environment)

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+- Header guards: `#pragma once`
+
+### Python
+
+Ruff is configured in `pyproject.toml` with `select = ["ALL"]` — all rules are enabled by default, with a small set selectively disabled.
+Do not add new ignore rules without documenting why.
```
burgholzer: This section could mention that we prefer to fix warnings reported by ruff whenever easily possible over ignoring them via `noqa` statements. A similar statement holds for type checking with `ty`. We try to fix reported warnings in most case before adding a type ignore. Sometimes these ignores are necessary, because a library, such as Qiskit, might not be properly typed.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+
+```bash
+uvx nox -s docs
+```
+
+Build without serving (non-interactive):
+
+```bash
+uvx nox --non-interactive -s docs
+```
```
burgholzer: I believe the first command is never really useful from an AI perspective, as it is interactive and needs active cancellation. So I suppose the non-interactive command is the only one of the two that makes sense here.
Just to be sure: Also the non-interactive one requires doxygen.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+Regenerate Python type stubs from nanobind bindings:
+
+```bash
+uvx nox -s stubs
+```
+
+Stub files (`.pyi`) in `python/mqt/core/` are **auto-generated** by nanobind's stubgen.
+NEVER edit `.pyi` files manually — they will be overwritten.
+If bindings change, regenerate stubs and commit the updated `.pyi` files.
+
```
burgholzer: No agent should probably commit things by itself so I believe the sentence is ill-advised.
It might also be worth to quantify "if bindings change".

AGENTS.md
burgholzer: I believe this has not really been addressed. I feel like this can simply be removed.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+mlir/                   MLIR quantum compiler infrastructure and QIR builder
+eval/                   Benchmarking and evaluation
+json/                   Device configuration files (NA, SC)
+```
+
+Components: `ir` (intermediate representation), `dd` (decision diagrams), `zx` (ZX-calculus), `algorithms`, `circuit_optimizer`, `qasm3` (OpenQASM 3.0 parser), `na` (neutral atoms), `qdmi` (device management), `qir` (QIR runtime and runner), `fomac` (formal methods), `mlir` (MLIR dialects and mqt-cc compiler).
+
+## Platform Support
+
+- **Operating systems**: Linux (Ubuntu 24.04), macOS (14, 15), Windows (2022, 2025)
```
burgholzer: This is not really true. This reflects the runners we use in CI. We support a much broader and older spectrum of OSes. Check out our Python wheel tags for more detailed information on that.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+- **Architectures**: x86_64 and arm64
+- **C++ compilers**: GCC 14+, Clang 20+, MSVC (latest), AppleClang 17+
```
burgholzer: Similar things apply here. This reflects what we test with in CI. It does not properly reflect the compiler _support_  yet.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+- **Python**: 3.10–3.14 tested across all supported OS/architecture combinations. Wheels are built for Linux, macOS, and Windows on x86_64 and arm64 (Stable ABI, targeting Python 3.12+). Free-threading wheels are built for Python 3.14.
```
burgholzer: This is not entirely true as not all of the combinations are tested (or can be tested).

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+- **CI**: GitHub Actions with reusable workflows from `munich-quantum-toolkit/workflows`.
```
burgholzer: Don't particularly know if this adds any value

burgholzer (changes requested):
Thanks @CodeMaverick2 for your continued work on this.
I made it through the document once more and added inline comments wherever I saw the need for that. Hopefully, most of the comments are self-explanatory. If not, do not hesitate to ask.

Additionally, it would be great if you could run the document through something like https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prompting-tools#prompt-improver which seems like a fairly good practice to me.

For all of this, please keep in mind our contributing guidelines (https://mqt.readthedocs.io/projects/core/en/latest/contributing.html), particularly
> Respond to comments to signal that you have addressed the feedback. Do not resolve review comments yourself; the reviewer will do so once satisfied.

AGENTS.md
CodeMaverick2: Removed -DBUILD_MQT_CORE_TESTS=ON (defaults ON). Added --config Release to cmake --build. Added note explaining the flag is required for multi-config generators   (MSVC) and ignored by single-config. Consolidated to a single build example.

AGENTS.md
CodeMaverick2: Updated description to: "Build as shared libraries (used internally for Python packaging; not for direct use)"

AGENTS.md
CodeMaverick2: Removed the incorrect "first searched via find_package" sentence.

AGENTS.md
CodeMaverick2: Replaced the prose list with a table showing each dependency with a link to its upstream project and its specific condition. nanobind noted as found via        
  find_package, not fetched. Boost notes USE_SYSTEM_BOOST=ON option.

AGENTS.md
CodeMaverick2: Added -C Release to all ctest commands. Added note explaining the -C flag for multi-config generators.

AGENTS.md
CodeMaverick2: Removed standalone "MLIR tests are located at mlir/unittests/" line. Now mentioned once in the label example ("all MLIR unit tests in mlir/unittests/") and once in "write new tests

AGENTS.md
CodeMaverick2: Added: "where {wheel_tag} is a PEP 425 tag (e.g., cp312-cp312-linux_x86_64) and {build_type} is the CMake build type (e.g., Release)"

AGENTS.md
CodeMaverick2: Fixed. Now reads: "uv run automatically syncs the environment" — which reflects that uv run handles env setup itself without requiring a prior uv sync.

AGENTS.md
CodeMaverick2: Changed examples to minimums-3.12 and qiskit-3.12. Added: "During iteration, prefer running a single Python version (e.g., tests-3.12, minimums-3.12). Run the  full command across all versions for final checks."

AGENTS.md
CodeMaverick2: Changed to: "auto-fixes formatting issues and reports linting problems"

AGENTS.md
CodeMaverick2: Added: "This requires a virtual environment with nanobind installed (e.g., via uv sync or minimally uv sync --only-group build)."

AGENTS.md
CodeMaverick2: Added: "Prefer fixing reported warnings over suppressing them with # noqa comments." and "fix reported issues before adding # type: ignore comments. Type ignores are sometimes necessary for incompletely typed libraries (e.g., Qiskit)."

AGENTS.md
CodeMaverick2: Removed interactive uvx nox -s docs. Kept only uvx nox --non-interactive -s docs. Header still notes doxygen is required.

AGENTS.md
CodeMaverick2: Removed commit advice

AGENTS.md
CodeMaverick2: Replaced with factual wheel platform data from cibuildwheel config: "Linux (glibc 2.17+), macOS (11.0+), Windows" on x86_64/arm64.

AGENTS.md
CodeMaverick2: Removed compiler list entirely.

AGENTS.md
CodeMaverick2: Removed inaccurate claim. Platform Support now states factual wheel availability only.

AGENTS.md
CodeMaverick2: Removed

CodeMaverick2:
@burgholzer Addressed all the comments do take a look and give more feedback

AGENTS.md
burgholzer: This is not correct. Our wheels are not compatible with glibc 2.17+ and I am unsure where you got that information from.

AGENTS.md
burgholzer: That is also not fully correct, which I tried to emphasize with the "_all_" above.
The pre-commit checks do in fact fix some of the linter warnings automatically, but only some.

AGENTS.md
burgholzer: It's probably still worth to mention that you can also build in Debug mode by replacing `Release` with `Debug`

AGENTS.md
burgholzer: It's probably still worth to mention that you can also build in Debug mode by replacing Release with Debug

AGENTS.md
burgholzer: The table is good. I believe it would be helpful to note down the versions of the respective tools in a separate column.

AGENTS.md
burgholzer: I think this reads much better now. May I ask you to change the `3.12` to `3.14` in all the cases? We do support 3.14 for all of these tasks and I believe it makes sense to instruct the agent to use the most recently supported Python version by default.

AGENTS.md
burgholzer: `ty` uses a different command for fine-grained type ignores; see https://docs.astral.sh/ty/suppression/

AGENTS.md
burgholzer: This seems better now. However, it only talks about wheels now; i.e., only about Python.
Previously, this also mentioned more about the C++ part, which constitutes the majority of the project.

AGENTS.md
CodeMaverick2: Added "Replace Release with Debug for a debug build." after the build example

AGENTS.md
CodeMaverick2: Added a Version column to the dependency table with exact versions from cmake/ExternalDependencies.cmake. nanobind version left blank since it's detectedfrom the Python environment rather than pinned in CMake

AGENTS.md
CodeMaverick2: Added note: "Replace Release with Debug for debug builds." to the -C flag explanation.

AGENTS.md
CodeMaverick2: Changed all Python version examples to 3.14 (tests-3.14, minimums-3.14, qiskit-3.14, and the iteration advice). Stable ABI correctly remains at 3.12 since that's the ABI compatibility target.

AGENTS.md
CodeMaverick2: Updated wording to: "auto-fixes formatting and some linting issues, and reports remaining problems" — reflecting that pre-commit does auto-fix some (but not all) linting issues.

AGENTS.md
CodeMaverick2: Updated to use ty's actual suppression syntax: \# ty: ignore[code]` with a link to the ty suppression documentation. Removed the mypy-style `# type: ignore` reference.

AGENTS.md
CodeMaverick2: Removed the incorrect glibc 2.17+ claim. Platform Support now just states "Linux (x86_64/arm64)" without specifying glibc version

AGENTS.md
CodeMaverick2: Added C++ platform info: "The C++ library requires a compiler with C++20 support. It is regularly tested with GCC, Clang, AppleClang, and MSVC."  statedfactually without claiming specific minimum compiler versions.

burgholzer:
Hey @CodeMaverick2 👋🏼 
I'll get back to you on this over the weekend. We got a couple more days to work this out. 🙌🏼

AGENTS.md
```diff
@@ -0,0 +1,86 @@
...
+- Shortcut for running tests with minimal version dependency resolution `uvx nox -s minimums` or `uvx nox -s minimums-3.14`
+- Shortcut for running against the upstream Qiskit package `uvx nox -s qiskit` or `uvx nox -s qiskit-3.14`
+
+## Documentation
+
+- Source files are in `docs/`.
+- Documentation uses Sphinx with MyST (Markdown) and the furo theme.
+- Build MLIR docs via the `mlir-doc` target in the C++ build (`cmake --build build_cpp --target mlir-doc --config Release`)
+- Build docs locally via `uvx nox --non-interactive -s docs`
+- Check links via `uvx nox -s docs -- -b linkcheck`
```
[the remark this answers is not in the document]
burgholzer: None of the other items end in a punctuation mark, neither should this one.

AGENTS.md
```diff
@@ -0,0 +1,86 @@
...
+  - ZX-calculus for transforming and optimizing quantum circuits
+  - OpenQASM 3.0 parser and exporter
+  - QIR runner and runtime (based on DDs)
+  - MLIR-based quantum compiler infrastructure (MQT Compiler Collection; mqt-cc)
+  - Example QDMI device implementations, C++ and Python bindings for QDMI
+  - Utilities for neutral atom (NA) quantum computing
+
+## Tech Stack
+
+- Targets Linux (glibc 2.28+), macOS (11.0+), and Windows on x86_64 and arm64 architectures
```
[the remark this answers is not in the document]
burgholzer: The default cibuildwheel version uses `manylinux_2_28` by now, which is glibc 2.28+
So this is fine

PR #1537, opened 2026-03-03
✨ Heuristic Mapping Pass

MatthiasReumann:
## Description

This pull request reimplements the heuristic mapping pass of the old mqtopt dialect. Similarly to #1510 the mapping pass ignores structured control flow (for now). I've copied the CMake setup of #1206.

A key difference between the implementations is that here I avoid building the "fat" Layout object (which keeps track of the SSA values) and use the `WireIterator`s and the "lightweight" Layout instead. Moreover, just because I wanted to highlight how cool the `WireIterator` is, this pull request adds LightSABRE-like back and forth traversal to find a better initial layout.

Lastly, I think it should be possible to use the ideas implemented here for a unitized IR also. 

## Checklist:

### DIFF
--- a/mlir/include/mlir/Passes/Passes.td
+++ b/mlir/include/mlir/Passes/Passes.td
@@ -43,7 +43,7 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
     * `layers` is an array of layers with size `1 + nlookahead`.
     * `depth(n)` returns the distance from the node `n` to the root node.
     * `dist(i, j)` returns the distance between the qubits `i` and `j` on the target's coupling graph.
-    * `h(L, pi) := sum(dist(p[gate.first], p[gate.second]) for gate in L)`
+    * `h(L, p) := sum(dist(p[gate.first], p[gate.second]) for gate in L)`
 
     To iteratively refine the mapping, the pass performs multiple forward and backward traversals of the circuit. In
     each traversal, the pass routes the circuit and updates the dynamic-to-static mapping based on the routing decisions


mlir/lib/Compiler/CompilerPipeline.cpp
```diff
@@ -147,7 +148,8 @@ QuantumCompilerPipeline::runPipeline(ModuleOp module,
 
   // Stage 5: Optimization passes
   // TODO: Add optimization passes
-  addCleanupPasses(pm);
+  pm.addPass(qco::createHeuristicMappingPass());
+  // addCleanupPasses(pm);
```
MatthiasReumann: This is probably not the right place to put this.

MatthiasReumann: Removed it for now.

mlir/unittests/Passes/Mapping/test_mapping.cpp
```diff
@@ -0,0 +1,194 @@
...
+    bool executable = true;
+    std::ignore = module->walk([&](qc::UnitaryOpInterface op) {
+      if (op.getNumQubits() > 1) {
+        const auto i0 = mappings[op.getQubit(0)];
+        const auto i1 = mappings[op.getQubit(1)];
+        if (!arch.areAdjacent(i0, i1)) {
+          executable = false;
+          return WalkResult::interrupt();
+        }
+      }
```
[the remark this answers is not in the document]
MatthiasReumann: Here we can assume that we have a valid mapping. How about an assertion? @coderabbitai

mlir/lib/Compiler/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ add_mlir_library(
   QCToQCO
   QCOToQC
   QCToQIR
+  QcoPasses
```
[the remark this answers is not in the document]
MatthiasReumann: This dependency is actually not needed anymore. I've removed it.

mlir/include/mlir/Passes/Passes.td
```diff
@@ -0,0 +1,36 @@
...
+#ifndef QCO_PASSES
+#define QCO_PASSES
+
+include "mlir/Pass/PassBase.td"
+
+//===----------------------------------------------------------------------===//
+// Transpilation Passes
+//===----------------------------------------------------------------------===//
+
+def HeuristicMappingPass : Pass<"map", "mlir::ModuleOp"> {
```
burgholzer: Would "place-and-route" be a better name here?

mlir/include/mlir/Passes/Passes.td
```diff
@@ -0,0 +1,36 @@
...
+  }];
+  let options = [
+    Option<"nlookahead", "nlookahead", "std::size_t", "1",
+           "The number of lookahead steps.">,
+    Option<"alpha", "alpha", "float", "1.0F",
+           "The alpha factor in the cost function.">,
+    Option<"lambda", "lambda", "float", "0.5F",
+           "The lambda factor in the cost function.">,
+    Option<"repeats", "repeats", "std::size_t", "2",
+           "The number of forwards and backwards traversal to improve the initial layout.">
```
burgholzer: Hm, can we do some bike shedding on the name here?
"Iterations"?

Do I understand it correctly that repeats == 0 means to only map forward once?

mlir/include/mlir/Passes/Mapping/Architecture.h
burgholzer, line 1: I have not reviewed this in very long detail. It is mostly unchanged to the old code, right? And this might go away when moving to QDMI, right?

mlir/unittests/Passes/Mapping/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
+# Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+set(target_name mqt-core-mlir-unittest-mapping)
+add_executable(${target_name} test_heuristic_mapping.cpp)
```
burgholzer: Reading this, I think we can drop the "heuristic" from the entire framing and labeling throughout. It is unlikely that we are going to port the optimal/exakt approach to MLIR. So any mapping will be heuristic by design.

mlir/lib/Passes/Mapping/HeuristicMapping.cpp
```diff
@@ -0,0 +1,660 @@
...
+     * swap to the layout of the parent node and evaluate the cost.
+     */
+    Node(const Node& parent, IndexGate swap, ArrayRef<Layer> layers,
+         const Architecture& arch, const Parameters& params)
+        : sequence(parent.sequence), layout(parent.layout), f(0) {
+      // Apply node-specific swap to given layout.
+      layout.swap(swap.first, swap.second);
+
+      // Add swap to sequence.
+      sequence.push_back(swap);
```
burgholzer:
```suggestion
      sequence.emplace_back(swap);
```
Might not make much of a difference.

mlir/lib/Passes/Mapping/Mapping.cpp
```diff
@@ -0,0 +1,660 @@
...
+private:
+  /**
+   * @brief Collect dynamic qubits contained in the given function body.
+   * @returns a vector of SSA values produced by qco.alloc operations.
+   */
+  [[nodiscard]] static SmallVector<QubitValue>
+  collectDynamicQubits(Region& funcBody) {
+    return SmallVector<QubitValue>(map_range(
+        funcBody.getOps<AllocOp>(), [](AllocOp op) { return op.getResult(); }));
+  }
```
burgholzer, line 359: Just to point this out: this will probably also need some
Adjustments once we properly support tensors for registers. Feel free to resolve.

mlir/lib/Passes/Mapping/HeuristicMapping.cpp
```diff
@@ -0,0 +1,660 @@
...
+   * the function traverses the def-use chain of each qubit until a two-qubit
+   * gate is found. If a two-qubit gate is visited twice, it is considered ready
+   * and inserted into the layer. This process is repeated until no more
+   * two-qubit are found anymore.
+   * @returns a vector of layers.
+   */
+  template <Direction d>
+  static SmallVector<Layer> collectLayers(MutableArrayRef<WireIterator> wires) {
+    constexpr auto step = d == Direction::Forward ? 1 : -1;
+    const auto stop = [](const WireIterator& it) {
```
burgholzer: This is not named ideally, because it is the opposite of a stop condition, and should maybe be called `continue` or similar

mlir/lib/Passes/Mapping/HeuristicMapping.cpp
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    return layers;
+  }
+
+  /**
+   * @brief "Cold" routing of the given layers.
+   * @details Iterates over a sliding window of layers and uses the A* search
+   * engine to find a sequence of SWAPs that makes that layer executable.
+   * Instead of inserted these SWAPs into the IR, this function only updates the
```
burgholzer:
```suggestion
   * Instead of inserting these SWAPs into the IR, this function only updates the
```

mlir/include/mlir/Passes/Passes.td
MatthiasReumann: iterations sounds promising. 

Yes, if `iteration==0` there will be a single forward pass starting with the identity layout.

mlir/include/mlir/Passes/Mapping/Architecture.h
MatthiasReumann: Yup.

mlir/include/mlir/Passes/Passes.td
MatthiasReumann: Sounds good.

mlir/unittests/Passes/Mapping/CMakeLists.txt
MatthiasReumann: Dropped the "Heuristic"

mlir/lib/Passes/Mapping/HeuristicMapping.cpp
MatthiasReumann: Good catch. Thanks!

mlir/lib/Passes/Mapping/HeuristicMapping.cpp
MatthiasReumann: Renamed the lambda to `shouldContinue` as `continue` is a C++ keyword.

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 ### Added
 
 - ✨ Add initial infrastructure for new QC and QCO MLIR dialects
-  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521])
+  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1537])
   ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**], [**@Ectras**], [**@MatthiasReumann**])
```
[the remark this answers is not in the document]
MatthiasReumann: Any opinions on that @burgholzer ?

mlir/include/mlir/Passes/Passes.td
```diff
@@ -0,0 +1,36 @@
...
+    Option<"nlookahead", "nlookahead", "std::size_t", "1",
+           "The number of lookahead steps.">,
+    Option<"alpha", "alpha", "float", "1.0F",
+           "The alpha factor in the cost function.">,
+    Option<"lambda", "lambda", "float", "0.5F",
+           "The lambda factor in the cost function.">,
+    Option<"iterations", "iterations", "std::size_t", "2",
+           "The number of forwards and backwards traversal to improve the initial layout.">
+  ];
+}
```
burgholzer, line 63: I do not want to be too pedantic here, but given how this is basically the user facing documentation at the moment, it would probably be good for the description to spell out what "lookahead", "the cost function" and "forwards and backwards traversal" (and maybe "initial layout") mean.
This is one of the first really big passes to be added to the code base, so there is no real precedent for that kind of documentation. However, I believe it would be cool if people got a rough idea of what's going on when they look at the project documentation on RtD.
Does that make sense?

mlir/lib/Passes/Mapping/Mapping.cpp
```diff
@@ -0,0 +1,659 @@
...
+  };
+
+  using MinQueue = std::priority_queue<Node, std::vector<Node>, std::greater<>>;
+
+public:
+  using MappingPassBase::MappingPassBase;
+
+  void runOnOperation() override {
+    Parameters params(this->alpha, this->lambda, this->nlookahead);
+    // TODO: Hardcoded architecture.
```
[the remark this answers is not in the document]
burgholzer: @MatthiasReumann the suggestion from @coderabbitai seems half-reasonable here. Although I would want to avoid having to many precautious safety checks everywhere.. AI really likes to do that. I'd much rather add this to the documentation explicitly or some other less invasive measure.

mlir/lib/Passes/Mapping/Mapping.cpp
```diff
@@ -0,0 +1,659 @@
...
+        return;
+      }
+
+      const auto [ltr, rtl] = computeBidirectionalLayers(dyn);
+
+      // Use the SABRE Approach to improve the initial layout choice (here:
+      // identity): Traverse the layers from left-to-right-to-left and
+      // cold-route along the way. Repeat this procedure "iterations" times.
+
+      Layout layout = Layout::identity(arch.nqubits());
```
burgholzer, line 327: Feel free to immediately resolve this after reading (or turn it into an issue on GitHub):
In a follow-up PR, we should definitely rethink this part. Starting from the identity is likely to be far away from the optimum initial layout; so it is also unreasonable to expect a handful of iterations to converge on a good one.
More likely, either multiple random trials could be evaluated in parallel (similar to how Qiskit does this) or a better initial heuristic based on some interaction graph analysis should be used here.

mlir/unittests/Passes/Mapping/test_mapping.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  static void runHeuristicMapping(OwningOpRef<ModuleOp>& module) {
+    PassManager pm(module->getContext());
+    pm.addPass(createQCToQCO());
+    pm.addPass(qco::createMappingPass(qco::MappingPassOptions{
+        .nlookahead = 5, .alpha = 1, .lambda = 0.85, .iterations = 2}));
+    pm.addPass(createQCOToQC());
+    auto res = pm.run(*module);
+    ASSERT_FALSE(failed(res));
```
burgholzer: This is just me being overly picky again, but instead of the double negation, I believe code like this is clearer as

```suggestion
    ASSERT_TRUE(succeeded(res));
```

mlir/unittests/Passes/Mapping/test_mapping.cpp
```diff
@@ -0,0 +1,195 @@
...
+    // Register all necessary dialects
+    DialectRegistry registry;
+    registry.insert<qc::QCDialect, qco::QCODialect, arith::ArithDialect,
+                    func::FuncDialect>();
+    context = std::make_unique<MLIRContext>();
+    context->appendDialectRegistry(registry);
+    context->loadAllAvailableDialects();
+  }
+
+  static void runHeuristicMapping(OwningOpRef<ModuleOp>& module) {
```
burgholzer: It's already apparent from the syntax highlighting in the git diff, but `module` might not really be a good name here as it is a reserved word in C++20 onwards. While I think it is only reserved in a certain context, it might still be better to avoid it or to use `module_` (which might trigger clang-tidy).

CHANGELOG.md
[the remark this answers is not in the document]
burgholzer: @MatthiasReumann I think that @coderabbitai has a point here. This feels like it does not really belong to the "core infrastructure" any more, but rather feels like it builds on top of it. So I'd argue for a separate line item here.
Given how there will be follow-up PRs that improve the features being introduced here, these might again go towards the same line item, similar to how the list of PRs going towards the initial infrastructure keeps growing.

mlir/lib/Passes/Mapping/Mapping.cpp
MatthiasReumann: Yes, definitely!

I think investigating "random + parallel" could be interesting. I know that there are some utilities for multi-threading in LLVM, so it might be worthwhile (also for future efforts and other passes) to continue on that route.

mlir/unittests/Passes/Mapping/test_mapping.cpp
MatthiasReumann: Changed to `moduleOp`.

mlir/include/mlir/Passes/Passes.td
MatthiasReumann: I've updated the description, accordingly. I think there is still a lot of room for improvement. But I think this could be a good starting point.

PR #1542, opened 2026-03-04
✨ Add `qtensor` dialect to MLIR

li-mingbao:
## Description

This PR adds the `qtensor` dialect to the environment.
The `qtensor` dialect is an adjusted version of the standard `tensor` dialect that supports linear typing for tensors with qubit types.

At the moment the following operations are supported:
`extract`, `extract_slice`, `from_elements`, `insert`, `insert_slice`.

- Both `extract` operations are modified so that they also return the input tensor as result.
- The `insert` and the `from_elements` operations are just wrapper operations that are canonicalized to a standard `tensor` operation.
- The `insert_slice` is identical to the standard operation but the canonicalizations are slightly adjusted to fit the new extract operation.

In addition, a `qtensor.dealloc` operation is added that deallocates a tensor and serves as a sink operation.

Example:
```
module {
    %tensor= qtensor.alloc(3) : tensor<3x!qco.qubit>
    %q_extracted_slice:2 = qtensor.extract_slice %tensor[0] [2] [1] : tensor<3x!qco.qubit> to tensor<2x!qco.qubit>
    %q_extracted:2 = qtensor.extract %q_extracted_slice#0[%c0] : tensor<2x!qco.qubit>
    %3 = qco.h %q_extracted#0 : !qco.qubit -> !qco.qubit
    %inserted = tensor.insert %3 into %q_extracted#1[%c0] : tensor<2x!qco.qubit>
    %inserted_slice = qtensor.insert_slice %inserted into %q_extracted_slice#1[0] [2] [1] : tensor<2x!qco.qubit> into tensor<3x!qco.qubit>
    qtensor.dealloc %inserted_slice : tensor<3x!qco.qubit>
    return %c0_i64 : i64
  }
}
```

## Checklist:

### DIFF
--- a/mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
+++ b/mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
@@ -1300,10 +1300,10 @@ public:
    * @brief Finalize the program and return the constructed module
    *
    * @details
-   * Automatically deallocates all remaining valid qubits, adds a return
-   * statement with exit code 0 (indicating successful execution), and
-   * transfers ownership of the module to the caller.
-   * The builder should not be used after calling this method.
+   * Automatically deallocates all remaining valid qubits and tensors of qubits,
+   * adds a return statement with exit code 0 (indicating successful execution),
+   * and transfers ownership of the module to the caller. The builder should not
+   * be used after calling this method.
    *
    * @return OwningOpRef containing the constructed quantum program module
    */
--- a/mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
+++ b/mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
@@ -93,8 +93,8 @@ def FromElementsOp : QTensorOp<"from_elements", [
     let summary = "Create a tensor from a range of values";
     let description = [{
         The `qtensor.from_elements` operation is a wrapper operation of the `tensor.from_elements` operation.
-        This operation creates a one-dimensional qubit tensor using the operands as its input. The also causes
-        the input qubits to be consumed.
+        This operation creates a one-dimensional qubit tensor using the operands as its input.
+        This also causes the input qubits to be consumed.
 
         Example:
         ```mlir
--- a/mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
+++ b/mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
@@ -77,6 +77,7 @@ TEST_P(QCOTest, ProgramEquivalence) {
   ASSERT_TRUE(program);
   printer.record(program.get(), "Original QCO IR" + name);
   EXPECT_TRUE(verify(*program).succeeded());
+
   runCanonicalizationPasses(program.get());
   printer.record(program.get(), "Canonicalized QCO IR" + name);
   EXPECT_TRUE(verify(*program).succeeded());
--- a/mlir/include/mlir/Dialect/QTensor/IR/QTensorDialect.td
+++ b/mlir/include/mlir/Dialect/QTensor/IR/QTensorDialect.td
@@ -17,9 +17,9 @@ def QTensorDialect : Dialect {
   let summary = "The QTensor dialect for one-dimensional tensors of qubits with linear typing.";
 
   let description = [{
-      The QTensor dialect is an adjusted variant of the standard tensor dialect of MLIR that supports linear typing for one-dimensional tensors of qubits.
-      In order to support linear typing, the extract operations of this dialect are modified so that they also return the updated tensor.
-      In addition, alloc/dealloc operations are added to the dialect to support the bulk allocation and deallocation of qubit tensors with linear types.
+    The QTensor dialect is an adjusted variant of the standard tensor dialect of MLIR that supports linear typing for one-dimensional tensors of qubits.
+    In order to support linear typing, the extract operations of this dialect are modified so that they also return the updated tensor.
+    In addition, alloc/dealloc operations are added to the dialect to support the bulk allocation and deallocation of qubit tensors with linear types.
   }];
 
   let dependentDialects = ["::mlir::qco::QCODialect"];


mlir/lib/Dialect/QTensor/IR/Operations/InsertOp.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+    if (insertOp.getDest() != extractOp.getOutTensor()) {
+      return failure();
+    }
+    if (insertOp.getIndices() != extractOp.getIndices()) {
+      return failure();
+    }
+
+    rewriter.replaceOp(insertOp, extractOp.getTensor());
+    rewriter.eraseOp(extractOp);
```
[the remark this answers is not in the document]
li-mingbao: The extract operation supports linear typing, so each result can only be used once. Since this function also checks if the scalar value and the tensor stems from the extract operation this is the only user.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractOp.cpp
```diff
@@ -0,0 +1,100 @@
...
+    auto tensorCast = extract.getTensor().getDefiningOp<tensor::CastOp>();
+    if (!tensorCast) {
+      return failure();
+    }
+    if (!llvm::isa<RankedTensorType>(tensorCast.getSource().getType())) {
+      return failure();
+    }
+    rewriter.replaceOpWithNewOp<ExtractOp>(extract, tensorCast.getSource(),
+                                           extract.getIndices());
+    return success();
```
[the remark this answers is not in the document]
li-mingbao: This should be ok as the types still match. This does not cause any issues for example if this operation is followed by an insertOp that uses this operation.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -161,6 +168,203 @@ void QCOProgramBuilder::updateQubitTracking(Value inputQubit,
...
+  auto indexValue = utils::variantToValue(*this, getLoc(), index);
+  auto extractOp = qtensor::ExtractOp::create(*this, tensor, indexValue);
+  auto qubit = extractOp.getResult();
+  auto outTensor = extractOp.getOutTensor();
+
+  validQubits.insert(qubit);
+  updateTensorTracking(tensor, outTensor);
+
+  return {qubit, outTensor};
+}
```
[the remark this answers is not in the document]
li-mingbao: The updateTensorTracking function is in line with the updateQubitTracking function that is used in the other qco operations. There the state is also just updated in the end.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -203,6 +218,135 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * ```
+   * ```mlir
+   * %extractedSlice, %outTensor = qtensor.extract_slice %tensor[%c0][%c2][%c1]
+   * : tensor<3x!qco.qubit> to tensor<2x!qco.qubit>
+   * ```
+   */
+  std::pair<Value, Value>
+  extractSlice(Value tensor, const std::variant<int64_t, Value>& offset,
+               const std::variant<int64_t, Value>& sizes,
+               const std::variant<int64_t, Value>& strides);
```
denialhaag: If we limited ourselves to one-dimensional tensors, we would not need `strides`, right? I'm assuming this could simplify quite a few implementation details. At the same time, we would also diverge further from the original `tensor` dialect, which might not be desirable. Maybe @burgholzer has an opinion here. 🤔 

As I like making this comparison, Jeff restricts its `Qureg` to being one-dimensional.

This decision will, of course, affect many other parts of the code.

mlir/include/mlir/Dialect/QTensor/IR/QTensorDialect.h
```diff
@@ -0,0 +1,54 @@
...
+//===----------------------------------------------------------------------===//
+
+#define GET_TYPEDEF_CLASSES
+#include "mlir/Dialect/QTensor/IR/QTensorOpsTypes.h.inc" // IWYU pragma: export
+
+//===----------------------------------------------------------------------===//
+// QTensor Dialect Helpers
+//===----------------------------------------------------------------------===//
+
+namespace mlir::qtensor {} // namespace mlir::qtensor
```
denialhaag: I guess this can go. 🤔 

```suggestion
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+
+  let summary = "The QTensor dialect for tensors with linear typing.";
+
+  let description = [{
+  The Qtensor dialect is an adjusted variant of the standard tensor dialect
+  of MLIR that supports linear typing for tensor types. The extract operations
+  of this dialect are modified so that they also return the updated input tensor
+  as return value. In addition, the alloc/dealloc operations are added to the dialect
+  to support the bulk allocation and deallocation of tensors with linear types.
+  }];
```
denialhaag:
```suggestion
  let description = [{
    The Qtensor dialect is an adjusted variant of the standard tensor dialect of MLIR that supports linear typing for tensor types.
    The extract operations of this dialect are modified so that they also return the updated tensor.
    In addition, alloc/dealloc operations are added to the dialect to support the bulk allocation and deallocation of tensors with linear types.
  }];
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+
+def QTensor_DeallocOp : QTensorOp<"dealloc", [MemoryEffects<[MemFree]>]> {
+    let summary = "Deallocate a tensor";
+    let description = [{
+    This operation deallocates the tensor and the values it holds, releasing its resources.
+
+    Example:
+        ```mlir
+        qtensor.dealloc %tensor : <3x!qco.qubit>
+        ```
```
denialhaag: I'm not sure if this will render correctly in the docs.

```suggestion
    Example:
    ```mlir
    qtensor.dealloc %tensor : <3x!qco.qubit>
    ```
```

mlir/include/mlir/Dialect/QTensor/IR/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
+# Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+add_mlir_dialect(QTensorOps qtensor)
+add_mlir_doc(QTensorOps QTensorDialect Dialects/ -gen-dialect-doc)
```
denialhaag: It probably makes sense to add QTensor to the docs. For now, it should be enough to create a `docs/mlir/QTensor.md` similar to `docs/mlir/QC.md` with a link to it in `docs/mlir/index.md`. [This command](https://github.com/munich-quantum-toolkit/core/blob/d8f551d5d87939126eb805841b53a62760e04083/CMakeLists.txt#L147-L156) should already ensure that everything is copied together correctly.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+def QTensor_FromElementsOp : QTensorOp<"from_elements", [
+    Pure,
+    TypesMatchWith<"operand types match result element type",
+                   "result", "elements", "SmallVector<Type, 2>("
+                   "::llvm::cast<RankedTensorType>($_self).getNumElements(), "
+                   "::llvm::cast<RankedTensorType>($_self).getElementType())">
+  ]> {
+    let summary = "tensor from elements operation";
+    let description = [{
+        The `qtensor.from_elements`is a wrapper operation for the `tensor.from_elements` operation
```
denialhaag:
```suggestion
        The `qtensor.from_elements` is a wrapper operation for the `tensor.from_elements` operation
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+                   "dest", "result",
+                   "$_self">,
+    TypesMatchWith<"scalar type matches element type of dest",
+                   "dest", "scalar",
+                   "::llvm::cast<TensorType>($_self).getElementType()">]> {
+    let summary = "Insert element into tensor";
+    let description = [{
+        The `qtensor.insert` operation is a wrapper operation for the `tensor.insert` operation of the tensor dialect.
+        This operation inserts a scalar into a ranked tensor as specified by the operation`s indices.
+        During the canonicalization this  operation is converted into a `tensor.insert` operation.
```
denialhaag:
```suggestion
        During the canonicalization, this operation is converted into a `tensor.insert` operation.
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+          partially-static specification via attributes specified through the
+          `static_offsets`, `static_sizes` and `static_strides` arguments. A special
+          sentinel value ShapedType::kDynamic encodes that the corresponding entry has
+          a dynamic value.
+
+          Example:
+        ```mlir
+        %outTensor = qtensor.insert_slice %slicedTensor into %tensor[%c0][%c2][%c1]
+        : tensor<2x!qco.qubit> into tensor<3x!qco.qubit>
+        ```
```
denialhaag: The tabbing might need to be adjusted here for this to render correctly in the docs.

mlir/lib/Dialect/QTensor/IR/Operations/FromTensorOp.cpp
```diff
@@ -0,0 +1,74 @@
...
+  using OpRewritePattern<qtensor::FromElementsOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(qtensor::FromElementsOp fromElementsOp,
+                                PatternRewriter& rewriter) const final {
+    rewriter.replaceOpWithNewOp<tensor::FromElementsOp>(
+        fromElementsOp, fromElementsOp.getType(), fromElementsOp.getElements());
+
+    return success();
+  }
+};
```
denialhaag: I'm wondering whether we want canonicalizations like this. I'm assuming this allows us to use canonicalizations of the underlying `tensor` operation without reimplementing them, right? Staying in the `QTensor` world could make conversions easier because we would never need to match `tensor` operations, but I'm not sure if that's worth it. Maybe @burgholzer has an opinion here. 🤔

li-mingbao: This currently affects the `insert` and the `fromElements` op.
Looking at the implementations for [insertOp](https://github.com/llvm/llvm-project/blob/4434dabb69916856b824f68a64b029c67175e532/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L1613) and [fromElementsOp](https://github.com/llvm/llvm-project/blob/4434dabb69916856b824f68a64b029c67175e532/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L1441), there are not really any additional beneficial folds or canonicalizations implemented.
Maybe this affects some transformation passes in the `tensor` dialect but I havent looked deeply into them yet as I wanted to focus on the basic structure first.

So I guess removing the conversions would not change anything and simplify our conversions.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
burgholzer: I think it makes sense to restrict qubit registers to be one-dimensional. I am not so sure whether restricting general tensors of qubits to one dimension is fine. It might turn out to be helpful to aggregate qubits in a way that can be indexed with 2D coordinates.
On the other hand, one can always emulate multi-dimensional access in a one-dimensional tensor with the right index computations.
Feels like this could be a fair trade-off and it feels like it would simplify a lot of the implementation here.
Maybe pulling in one more opinion here, because we previously talked about related topics: @DRovara any opinions here?

mlir/lib/Dialect/QTensor/IR/Operations/FromTensorOp.cpp
burgholzer: > So I guess removing the conversions would not change anything and simplify our conversions.

That sounds like a convincing argument to me.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
DRovara: I have no doubt that there are use cases for multi-dimensional tensors.

However, as you say, we can also emulate them as 1D tensors, so the big question is: is there anything we lose by not supporting n-Dimensional qubit tensors?

- On the one hand, even something like a Python `lisŧ[list[qubit]]` where each individual list has dynamic size can be emulated by also keeping a secondary list of sublist-sizes
- The only thing one might lose is the round-trip expressibility for higher languages that support n-D Qubit Arrays. 

I'd say that: (1) higher-level languages that require round-trips are already niche (2) higher-level languages that support n-D Arrays are even more niche, and (3) actual use cases for such arrays even more.

So in my opinion, I see no big problem with explicitly only supporting 1D tensors.

li-mingbao: I am not really sure if this really simplifies the implementation here, as the current implementation of `extractSlice`and `insertSlice` is dependent on the `OffsetSizeAndStrideOpInterface`.
This interface requires a stride parameter and it provides quite a few utility functions that would need to be reimplemented here if we want to remove it.

burgholzer: I like your reasoning @DRovara 👍🏼 

@li-mingbao if the strides are necessary for that interface, then let's definitely keep it in order to reuse as much as possible from the existing dialects.
However, if we can agree that we limit ourselves to one-dimensional tensors for now, this might simplify other parts of the implementation.
An if it does not, then we do not need to limit ourselves.

mlir/lib/Dialect/QTensor/IR/Operations/FromTensorOp.cpp
li-mingbao: I removed the conversions now, every defined operation stays in the `qtensor` dialect.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
li-mingbao: I will have a look at it.
For now I adapted the builders so that the strides can be omitted and are set to a default value of 1.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -203,6 +218,135 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
   [[nodiscard]] ClassicalRegister
   allocClassicalBitRegister(int64_t size, std::string name = "c") const;
 
+  //===--------------------------------------------------------------------===//
+  // QTensor operations
+  //===--------------------------------------------------------------------===//
```
denialhaag, line 208: While I totally agree that it makes sense to have these in here, it might make sense to prepend the functions with `qtensor` to better reflect that these are not QCO operations. What do you think?

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -203,6 +218,135 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   *
+   * @par Example:
+   * ```c++
+   * auto tensor = builder.allocTensor(3);
+   * ```
+   * ```mlir
+   * %tensor = qtensor.alloc(3) : tensor<3x!qco.qubit>
+   * ```
+   */
+  Value allocTensor(int64_t size);
```
denialhaag: Regarding the comment above, maybe something like this: 🤔 

```suggestion
  Value qtensorAlloc(int64_t size);
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+//===----------------------------------------------------------------------===//
+
+def QTensor_AllocOp : QTensorOp<"alloc", [
+    MemoryEffects<[MemAlloc]>]> {
+    let summary = "Tensor alloc operation";
+    let description = [{
+        Allocates a qubit tensor with the given size and returns the allocated tensor.
+        The qubits are initialized to the |0> state.
+
+        Example :
```
denialhaag:
```suggestion
        Example:
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+}
+
+def QTensor_FromElementsOp : QTensorOp<"from_elements", [
+    Pure,
+    TypesMatchWith<"operand types match result element type",
+                   "result", "elements", "SmallVector<Type, 2>("
+                   "::llvm::cast<RankedTensorType>($_self).getNumElements(), "
+                   "::llvm::cast<RankedTensorType>($_self).getElementType())">
+  ]> {
+    let summary = "tensor from elements operation";
```
denialhaag: The summary could be improved a bit.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+    let description = [{
+        The `qtensor.from_elements` is a wrapper operation for the `tensor.from_elements` operation
+        of the tensor dialect. This operation creates a tensor using the operands as its values.
+
+        Example:
+
+        ```mlir
+        %tensor = qtensor.from_elements %q0, %q1, %q2 :  tensor<3x!qco.qubit>
+        ```
```
denialhaag, line 102:
```suggestion
        Example:
        ```mlir
        %tensor = qtensor.from_elements %q0, %q1, %q2 : tensor<3x!qco.qubit>
        ```
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+    Pure,
+    TypesMatchWith<"operand types match result element type",
+                   "result", "elements", "SmallVector<Type, 2>("
+                   "::llvm::cast<RankedTensorType>($_self).getNumElements(), "
+                   "::llvm::cast<RankedTensorType>($_self).getElementType())">
+  ]> {
+    let summary = "tensor from elements operation";
+    let description = [{
+        The `qtensor.from_elements` is a wrapper operation for the `tensor.from_elements` operation
+        of the tensor dialect. This operation creates a tensor using the operands as its values.
```
denialhaag:
```suggestion
        The `qtensor.from_elements` operation is a wrapper operation of the `tensor.from_elements` operation.
        This operation creates a tensor using the operands as its values.
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+        at the given indices. In addition, it also returns the updated input tensor as a result.
+        The indices must be of `index` type.
+
+        Example:
+        ```mlir
+        %q0, %outTensor = qtensor.extract %tensor[%c0] : tensor<4x!qco.qubit>
+        ```
+    }];
+
+    let arguments = (ins AnyRankedTensor:$tensor, Variadic<Index>:$indices);
```
denialhaag: Assuming this doesn't lead to complications, we can do this if we restrict ourselves to one-dimensional tensors, right?

```suggestion
    let arguments = (ins AnyRankedTensor:$tensor, Index:$index);
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+    }];
+
+    let arguments = (ins
+        AnyRankedTensor:$source,
+        Variadic<Index>:$offsets,
+        Variadic<Index>:$sizes,
+        Variadic<Index>:$strides,
+        DenseI64ArrayAttr:$static_offsets,
+        DenseI64ArrayAttr:$static_sizes,
+        DenseI64ArrayAttr:$static_strides
```
denialhaag: Same question here, but this would interfere with `OffsetSizeAndStrideOpInterface`. 

We _could_ be very strict here and reduce this like so, rendering the trait unnecessary:

```suggestion
        Index:$offset,
        Index:$size,
```

This should simplify the implementation quite a bit. One could argue that this aligns with our philosophy of preferring operands over attributes (see, e.g., parametrized gates).

Maybe @burgholzer and @DRovara have an opinion here.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+        Example:
+
+        ```mlir
+        %newTensor = qtensor.insert %q0 into %tensor[%c0] : tensor<3x!qco.qubit>
+        ```
+    }];
+
+    let arguments = (ins AnyType:$scalar,
+                        AnyRankedTensor:$dest,
+                        Variadic<Index>:$indices);
```
denialhaag: See https://github.com/munich-quantum-toolkit/core/pull/1542#discussion_r2919425654.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,491 @@
...
+
+    let arguments = (ins
+        AnyRankedTensor:$source,
+        AnyRankedTensor:$dest,
+        Variadic<Index>:$offsets,
+        Variadic<Index>:$sizes,
+        Variadic<Index>:$strides,
+        DenseI64ArrayAttr:$static_offsets,
+        DenseI64ArrayAttr:$static_sizes,
+        DenseI64ArrayAttr:$static_strides
```
denialhaag: See https://github.com/munich-quantum-toolkit/core/pull/1542#discussion_r2919460071.

mlir/lib/Dialect/QTensor/IR/Operations/AllocOp.cpp
denialhaag, line 1: This file should probably be renamed to `AllocOp.cpp` to align with the operation name.

mlir/lib/Dialect/QTensor/IR/Operations/DeallocOp.cpp
denialhaag, line 1: This file should probably be renamed to `DeallocOp.cpp` to align with the operation name.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
DRovara: I'm not sure I fully get the context. You're saying we should only use dynamic size and offset values for slices? Or for tensors in general?

I guess this really depends on what we are actually planning to do with the slices. Of course static arguments make optimizations a lot easier. Then again, simply checking if the operand is constant should do the trick as well. I have no strong opinion on it, especially if we are only talking about slices here.

denialhaag: Yes, I'm proposing to change `qtensor.extract_slice` and `qtensor.insert_slice` to only accept operands.

Interestingly, `tensor.extract` and `tensor.insert` (i.e., the simpler, non-slice versions) already only accept operands: https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorextract-tensorextractop

denialhaag: Zooming out a bit, the question is whether we want to stick with the initial strategy of adjusting the `tensor` dialect only where absolutely necessary, or do we just define a new, as-simple-as-can-be dialect.

li-mingbao: If we limit the operations to only dynamic size and offset and no strides, the `OffsetSizeAndStrideOpInterface` is not required anymore. 
This was mainly used to handle mixed static/dynamic values for size/offset/strides which would be redundant.

I can try to simplify it as much as possible and see how it will look like with these changes in mind.

burgholzer: > I can try to simplify it as much as possible and see how it will look like with these changes in mind.

I feel that this would be reasonable to try. While I still think sticking close to the original is worthwhile to keep in mind generally, it is also not particularly helpful for us to carry around complicated code that we do not even use just for the sake of it.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractSliceOp.cpp
```diff
@@ -0,0 +1,130 @@
...
+                           ArrayRef<NamedAttribute> attrs) {
+  auto sourceType = cast<RankedTensorType>(source.getType());
+  auto sizeValue = getConstantIntValue(size);
+  auto resultType = RankedTensorType::get(
+      {sizeValue ? *sizeValue : ShapedType::kDynamic},
+      sourceType.getElementType(), sourceType.getEncoding());
+
+  result.addAttributes(attrs);
+  build(b, result, {source.getType(), resultType}, source, offset, size);
+}
```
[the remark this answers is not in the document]
li-mingbao: This is correct as it returns the outSourceTensor first and then the extractedSlice.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractSliceOp.cpp
```diff
@@ -0,0 +1,129 @@
...
```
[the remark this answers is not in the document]
li-mingbao: The basic methods like build(), verify() and fold() do not have to have a doxygen-style comment. This is consistent with the other operation definitions in the other dialects.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: Hey,

I refactored the implementation so that it only accepts dynamic values now and limited it to just 1-D.

These limitations simplified the implementations by a lot and most of the operations now only boil down to the verification of the operation itself and the `extract`/`insert` folds and its variations. The normal `tensor` operations work a lot with `cast` operations and its canonicalizations to support static and dynamic values which I have removed for now.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,268 @@
...
+    TypesMatchWith<"result type matches type of dest",
+                   "dest", "result",
+                   "$_self">,
+    TypesMatchWith<"scalar type matches element type of dest",
+                   "dest", "scalar",
+                   "::llvm::cast<TensorType>($_self).getElementType()">]> {
+    let summary = "Insert element into tensor";
+    let description = [{
+        The `qtensor.insert` operation is a wrapper operation for the `tensor.insert` operation of the tensor dialect.
+        This operation inserts a scalar into a ranked tensor as specified by the operation`s index.
```
denialhaag:
```suggestion
        This operation inserts a scalar into a ranked tensor as specified by the operation's index.
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,268 @@
...
+def QTensor_InsertSliceOp : QTensorOp<"insert_slice", [
+    Pure,
+    TypesMatchWith<"expected result type to match dest type",
+                   "dest", "result", "$_self">
+  ]> {
+    let summary = "Insert slice into tensor";
+    let description = [{
+        The `qtensor.insert_slice` operation is a modified version of the `tensor.insert_slice` operation
+        of the tensor dialect. The operation inserts a tensor `source` into another tensor `dest` as specified by the
+        operation`s offset and size argument.
```
denialhaag:
```suggestion
        operation's offset and size arguments.
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,268 @@
...
+        %outSourceTensor, %extractedSlice = qtensor.extract_slice %sourceTensor[%c0][%c2] : tensor<3x!qco.qubit> to tensor<2x!qco.qubit>
+        ```
+    }];
+
+    let arguments = (ins
+        AnyRankedTensor:$source,
+        Index:$offset,
+        Index:$size
+    );
+    let results = (outs AnyRankedTensor:$out_source, AnyRankedTensor:$result);
```
denialhaag: I would maybe try to keep the argument names a bit more consistent, particularly with respect to `ExtractOp`. 🤔 

```suggestion
    let arguments = (ins
        AnyRankedTensor:$tensor,
        Index:$offset,
        Index:$size
    );
    let results = (outs AnyRankedTensor:$out_tensor, AnyRankedTensor:$result);
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,268 @@
...
+        The qubits are initialized to the |0> state.
+
+        Example:
+        ```mlir
+        %qtensor = qtensor.alloc(3) : tensor<3x!qco.qubit>
+        ```
+    }];
+
+    let arguments = (ins ConfinedAttr<I64Attr, [IntPositive]>:$size);
+    let results = (outs AnyStaticShapeTensor:$result);
```
denialhaag: We should be able to use this type now, right? 🤔 If yes, this would also apply to the rest of this file.

```suggestion
    let results = (outs 1DTensorOf<[AnyType]>:$result);
```

For reference: [CommonTypeConstraints.td#L831-L841](https://github.com/llvm/llvm-project/blob/e5927fecf8a6ce89e1a4eac5b828e7d42676452a/mlir/include/mlir/IR/CommonTypeConstraints.td#L831-L841)

mlir/lib/Dialect/QTensor/IR/Operations/AllocOp.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return emitOpError("Result must be a 1-D tensor");
+  }
+
+  if (resultType.getShape()[0] != size) {
+    return emitOpError("Tensor length must match size attribute (")
+           << size << "), but got " << resultType.getShape()[0];
+  }
+
+  return success();
+}
```
denialhaag, line 61: This verifier might be unnecessary if we type the tensor as suggested in https://github.com/munich-quantum-toolkit/core/pull/1542#discussion_r2932707807. The other verifiers could likely be simplified as well.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
denialhaag: There's even an argument for using QCO's `QubitType` instead of `AnyType`. I guess this depends on how "reusable" we want the QTensor dialect to be.

Maybe @burgholzer has an opinion here.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -161,6 +164,199 @@ void QCOProgramBuilder::updateQubitTracking(Value inputQubit,
...
+                                             Value outputTensor) {
+  // Validate the input tensor
+  validateTensorValue(inputTensor);
+
+  // Remove the input (consumed) value from tracking
+  validTensors.erase(inputTensor);
+
+  // Add the output (new) value to tracking
+  validTensors.insert(outputTensor);
+}
```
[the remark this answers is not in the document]
li-mingbao: It can be assumed that local tensor are handled before they leak into the global finalization. So only tracked tensors in the main entry block are deallocated.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: Thanks for the tip with the `1DTensorOf` constraint. I replaced the tensors with a `1DTensorOf<[QubitType]>` constraint as this was checked anyway in the verifiers. This simplified the verifiers by a lot.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -161,6 +164,200 @@ void QCOProgramBuilder::updateQubitTracking(Value inputQubit,
...
+    llvm::reportFatalUsageError("Elements must be of QubitType!");
+  }
+
+  validateTensorValue(tensor);
+  validTensors.erase(tensor);
+
+  qtensor::DeallocOp::create(*this, tensor);
+
+  return *this;
+}
```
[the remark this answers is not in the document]
li-mingbao: The comments are in the header file.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -161,6 +164,200 @@ void QCOProgramBuilder::updateQubitTracking(Value inputQubit,
...
+  if (!tensorType || tensorType.getRank() != 1) {
+    llvm::reportFatalUsageError("Tensor must be of 1-D RankedTensorType!");
+  }
+  if (!llvm::isa<QubitType>(tensorType.getElementType())) {
+    llvm::reportFatalUsageError("Elements must be of QubitType!");
+  }
+
+  auto indexValue = utils::variantToValue(*this, getLoc(), index);
+  auto extractOp = qtensor::ExtractOp::create(*this, tensor, indexValue);
+  auto qubit = extractOp.getResult();
```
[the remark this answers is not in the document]
li-mingbao: This should not be necessary as the variantToValue always the IndexType when it gets a int64_t variant which always happens for this function calls. So the returnType is guaranteed to be of IndexType.

docs/mlir/index.md
```diff
@@ -8,7 +8,9 @@ We define multiple dialects, each with its dedicated purpose:
 
 - The {doc}`QCO dialect <QCO>` uses value semantics and is mainly designed for running optimizations.
 
-Both dialects define various canonicalization and transformations that enable the compilation of quantum programs to native quantum hardware.
+- The {doc}`QTensor dialect <QTensor>` adds support for tensors with linear typing and is used in the QCO dialect to represent registers.
```
burgholzer:
```suggestion
- The {doc}`QTensor dialect <QTensor>` adds support for one-dimensional tensors of qubits with linear typing and is used in the QCO dialect to represent collections of qubits such as registers.
```
Made this a little more explicit.

mlir/include/mlir/Dialect/QTensor/IR/QTensorDialect.h
```diff
@@ -0,0 +1,33 @@
...
+//===----------------------------------------------------------------------===//
+// Types
+//===----------------------------------------------------------------------===//
+
+#define GET_TYPEDEF_CLASSES
+#include "mlir/Dialect/QTensor/IR/QTensorOpsTypes.h.inc" // IWYU pragma: export
+
+//===----------------------------------------------------------------------===//
+// QTensor Dialect Helpers
+//===----------------------------------------------------------------------===//
```
burgholzer:
```suggestion
```
Doesn't seem like there are any helpers here

mlir/lib/Dialect/QTensor/IR/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+# Licensed under the MIT License
+
+file(GLOB_RECURSE OPERATIONS "${CMAKE_CURRENT_SOURCE_DIR}/Operations/*.cpp")
+
+add_mlir_dialect_library(
+  MLIRQTensorDialect
+  QTensorOps.cpp
+  ${OPERATIONS}
+  ADDITIONAL_HEADER_DIRS
+  ${PROJECT_SOURCE_DIR}/mlir/include/mlir/Dialect/QTensor
```
burgholzer: This could also use `MQT_MLIR_SOURCE_INCLUDE_DIR`

```suggestion
  ${MQT_MLIR_SOURCE_INCLUDE_DIR}/mlir/Dialect/QTensor
```

mlir/lib/Dialect/QTensor/IR/QTensorOps.cpp
```diff
@@ -0,0 +1,78 @@
...
+  if (index1 == index2) {
+    return true;
+  }
+
+  auto val1 = getConstantIntValue(index1);
+  auto val2 = getConstantIntValue(index2);
+
+  return val1 && val2 && *val1 == *val2;
+}
+} // namespace mlir::qtensor
```
burgholzer: This feels a bit out of place here, but I suppose this is the only place it can reasonably be put in, right?

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+def QubitType :
+  Type<CPred<"llvm::isa<::mlir::qco::QubitType>($_self)">,
+       "Elements must be of qubit type">;
+
+def Static1DQubitTensor : Type<
+  And<[
+    1DTensorOf<[QubitType]>.predicate,
+    HasStaticShapePred
+  ]>,
+  "Tensor must be statically shaped and elements must be of qubit type">;
```
burgholzer: Isn't the `QubitType` superfluous here and the QCO qubit type can be directly used?

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+def QTensorDialect : Dialect {
+  let name = "qtensor";
+
+  let summary = "The QTensor dialect for tensors with linear typing.";
+
+  let description = [{
+      The Qtensor dialect is an adjusted variant of the standard tensor dialect of MLIR that supports linear typing for tensor types.
+      The extract operations of this dialect are modified so that they also return the updated tensor.
+      In addition, alloc/dealloc operations are added to the dialect to support the bulk allocation and deallocation of tensors with linear types.
+  }];
```
burgholzer: Both the brief and the description should be more explicit (qubits, 1D, ..).
Maybe take my rewording in the other comment as inspiration.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+def QTensor_AllocOp : QTensorOp<"alloc", [
+    MemoryEffects<[MemAlloc]>]> {
+    let summary = "Tensor alloc operation";
+    let description = [{
+        Allocates a qubit tensor with the given size and returns the allocated tensor.
+        The qubits are initialized to the |0> state.
+
+        Example:
+        ```mlir
+        %qtensor = qtensor.alloc(3) : tensor<3x!qco.qubit>
```
burgholzer: I am probably overlooking something, but shouldn't the `tensor` dialect be a dependent dialect if the return type here is a `tensor`?
Or is the docstring simply wrong and this should be `qtensor` or similar?

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+//===----------------------------------------------------------------------===//
+
+def QubitType :
+  Type<CPred<"llvm::isa<::mlir::qco::QubitType>($_self)">,
+       "Elements must be of qubit type">;
+
+def Static1DQubitTensor : Type<
+  And<[
+    1DTensorOf<[QubitType]>.predicate,
+    HasStaticShapePred
```
burgholzer, line 36: Isn't the limit on the tensor being statically shaped kind of arbitrary and limiting?
Is there a reason for not supporting tensors with dynamic sizes? Or am I just overlooking something and interpreting too much into the word "static" here?

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+        OpBuilder<(ins "int64_t":$size)>
+    ];
+
+    let hasVerifier = 1;
+}
+
+def QTensor_DeallocOp : QTensorOp<"dealloc", [MemoryEffects<[MemFree]>]> {
+    let summary = "Deallocate a tensor";
+    let description = [{
+        This operation deallocates the tensor and the values it holds, releasing its resources.
```
burgholzer, line 71: Should this be more explicit? In particular, should it state that individual values that have been extracted but not inserted again need to be deallocated separately?

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+    Pure,
+    TypesMatchWith<"operand types match result element type",
+                   "result", "elements", "SmallVector<Type, 2>("
+                   "::llvm::cast<RankedTensorType>($_self).getNumElements(), "
+                   "::llvm::cast<RankedTensorType>($_self).getElementType())">
+  ]> {
+    let summary = "Create a tensor from a range of values";
+    let description = [{
+        The `qtensor.from_elements` operation is a wrapper operation of the `tensor.from_elements` operation.
+        This operation creates a tensor using the operands as its values.
```
burgholzer: Should this note that it also consumes the values? This is implied by linear typing but it may not hurt to make this as explicit as possible throughout the file.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+                   "tensor", "result",
+                   "::llvm::cast<TensorType>($_self).getElementType()">,
+                   TypesMatchWith<
+      "returned tensor type matches input tensor", "tensor", "out_tensor", "$_self">]> {
+    let summary = "Extract element from tensor";
+    let description = [{
+        The `qtensor.extract` operation is the modified version of the standard `tensor.extract`
+        operation of the tensor dialect. It reads a ranked tensor and returns one element from the input tensor
+        at the given index. In addition, it also returns the updated input tensor as a result.
+        The index must be of `index` type.
```
burgholzer:
```suggestion
```
 Feels superfluous as it is conveyed by the type system.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+
+        Example:
+        ```mlir
+        %outTensor, %q0 = qtensor.extract %tensor[%c0] : tensor<4x!qco.qubit>
```
burgholzer, line 130: The type annotation cannot be correct here, can it? Why does it only list the tensor while there is another output type and another input type? Is this on purpose?
 Please double check also in other places.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+        The `qtensor.extract_slice` operation is the modified version of the standard `tensor.extract_slice`
+        operation of the tensor dialect. It reads a ranked tensor and returns the extracted tensor specified by the
+        offset and size argument. In addition, it also returns the updated input tensor as result.
+
+        The extract_slice operation supports the following arguments:
+
+        * tensor: the "base" tensor from which to extract a slice.
+        * offset: the starting position in the base tensor from which the slice is
+                  extracted.
+        * size: the length of the slice to extract from the base tensor.
```
burgholzer: Markdown lists are canonically written with `-` for the itemizer.

```suggestion
        - tensor: the "base" tensor from which to extract a slice.
        - offset: the starting position in the base tensor from which the slice is
                  extracted.
        - size: the length of the slice to extract from the base tensor.
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+
+        Example:
+        ```mlir
+        %outTensor, %extractedSlice = qtensor.extract_slice %tensor[%c0][%c2] : tensor<3x!qco.qubit> to tensor<2x!qco.qubit>
```
burgholzer, line 162: also here, the type annotation feels slightly off considering the linear typing.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+    TypesMatchWith<"result type matches type of dest",
+                   "dest", "result",
+                   "$_self">,
+    TypesMatchWith<"scalar type matches element type of dest",
+                   "dest", "scalar",
+                   "::llvm::cast<TensorType>($_self).getElementType()">]> {
+    let summary = "Insert element into tensor";
+    let description = [{
+        The `qtensor.insert` operation is a wrapper operation for the `tensor.insert` operation of the tensor dialect.
+        This operation inserts a scalar into a ranked tensor as specified by the operation's index.
```
burgholzer: Probably worth explicitly spelling out that this consumes the inserted value.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+        of the tensor dialect. The operation inserts a tensor `source` into another tensor `dest` as specified by the
+        operation's offset and size arguments.
+
+        The insert_slice operation supports the following arguments:
+
+          * source: the tensor that is inserted.
+          * dest: the tensor into which the source tensor is inserted.
+          * offset: the starting index in the `dest` tensor where the slice is inserted.
+          * size: the number of elements in the slice, which must match the size of the
+                  source tensor type.
```
burgholzer, line 237:
```suggestion
        The insert_slice operation supports the following arguments:

          - source: the tensor that is inserted.
          - dest: the tensor into which the source tensor is inserted.
          - offset: the starting index in the `dest` tensor where the slice is inserted.
          - size: the number of elements in the slice, which must match the size of the
                  source tensor type.
```

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
```diff
@@ -0,0 +1,282 @@
...
+def QTensor_InsertSliceOp : QTensorOp<"insert_slice", [
+    Pure,
+    TypesMatchWith<"expected result type to match dest type",
+                   "dest", "result", "$_self">
+  ]> {
+    let summary = "Insert slice into tensor";
+    let description = [{
+        The `qtensor.insert_slice` operation is a modified version of the `tensor.insert_slice` operation
+        of the tensor dialect. The operation inserts a tensor `source` into another tensor `dest` as specified by the
+        operation's offset and size arguments.
```
burgholzer: Probably also worth noting that the inserted tensor is consumed by this operation.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -203,6 +203,132 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * ```mlir
+   * %tensor = qtensor.alloc(3) : tensor<3x!qco.qubit>
+   * ```
+   */
+  Value qtensorAlloc(int64_t size);
+
+  /**
+   * @brief Allocate a qubit tensor from a list of qubit values
+   * @param elements Inserted Qubits
+   * @return The allocated tensor
```
burgholzer, line 242: Probably worth noting that the qubit values are consumed by this operation. Similar to the tablegen docs, everything written here is user facing documentation that should be as expressive as possible.

mlir/lib/Dialect/QTensor/IR/Operations/DeallocOp.cpp
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @brief Remove matching allocation and deallocation pairs without operations
+ * between them.
+ */
+struct RemoveAllocDeallocPair final : OpRewritePattern<DeallocOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(DeallocOp op,
+                                PatternRewriter& rewriter) const override {
+    // Check if the predecessor is an qtensor::AllocOp
```
burgholzer:
```suggestion
    // Check if the predecessor is a qtensor::AllocOp
```
I think "... a qtensor::.." is correct.

mlir/lib/Dialect/QTensor/IR/Operations/DeallocOp.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto tensor = op.getTensor();
+    auto allocOp = tensor.getDefiningOp<AllocOp>();
+    if (!allocOp || !tensor.hasOneUse()) {
```
burgholzer:
```suggestion
    if (!allocOp) {
```
superfluous due to linear typing.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractOp.cpp
```diff
@@ -0,0 +1,69 @@
...
+ */
+static InsertOp foldExtractAfterInsert(ExtractOp extractOp) {
+  auto insertOp = extractOp.getTensor().getDefiningOp<InsertOp>();
+  if (!insertOp) {
+    return nullptr;
+  }
+
+  if (insertOp.getScalar().getType() != extractOp.getResult().getType()) {
+    return nullptr;
+  }
```
burgholzer: this will never be true by design, right?

```suggestion
```

mlir/lib/Dialect/QTensor/IR/Operations/ExtractOp.cpp
```diff
@@ -0,0 +1,69 @@
...
+  return insertOp;
+}
+
+LogicalResult ExtractOp::fold(FoldAdaptor /*adaptor*/,
+                              SmallVectorImpl<OpFoldResult>& results) {
+  if (auto insertOp = foldExtractAfterInsert(*this)) {
+    results.push_back(insertOp.getDest());
+    results.push_back(insertOp.getScalar());
```
burgholzer:
```suggestion
    results.emplace_back(insertOp.getDest());
    results.emplace_back(insertOp.getScalar());
```
might not make much of a difference

mlir/lib/Dialect/QTensor/IR/Operations/ExtractSliceOp.cpp
```diff
@@ -0,0 +1,108 @@
...
+  auto insertSliceOp =
+      extractSliceOp.getTensor().getDefiningOp<InsertSliceOp>();
+  if (!insertSliceOp) {
+    return nullptr;
+  }
+
+  if (insertSliceOp.getSource().getType() !=
+      extractSliceOp.getResult().getType()) {
+    return nullptr;
+  }
```
burgholzer:
```suggestion
```
this will never evaluate to true by design, right?

mlir/lib/Dialect/QTensor/IR/Operations/ExtractSliceOp.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+  return insertSliceOp;
+}
+
+LogicalResult ExtractSliceOp::fold(FoldAdaptor /*adaptor*/,
+                                   SmallVectorImpl<OpFoldResult>& results) {
+  if (auto insertOp = foldExtractAfterInsertSlice(*this)) {
+    results.push_back(insertOp.getDest());
+    results.push_back(insertOp.getSource());
```
burgholzer:
```suggestion
    results.emplace_back(insertOp.getDest());
    results.emplace_back(insertOp.getSource());
```

mlir/lib/Dialect/QTensor/IR/Operations/InsertOp.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  if (!extractOp) {
+    return nullptr;
+  }
+  if (insertOp.getDest() != extractOp.getOutTensor()) {
+    return nullptr;
+  }
+  if (extractOp.getTensor().getType() != insertOp.getDest().getType()) {
+    return nullptr;
+  }
```
burgholzer:
```suggestion
```
Same as for the other occasions

mlir/lib/Dialect/QTensor/IR/Operations/InsertOp.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    qco::DeallocOp::create(rewriter, prevInsertOp.getLoc(),
+                           prevInsertOp.getScalar());
+    rewriter.replaceOpWithNewOp<InsertOp>(insertOp, insertOp.getScalar(),
+                                          prevInsertOp.getDest(), insertIndex);
+    rewriter.eraseOp(prevInsertOp);
+    return success();
+  }
+};
```
burgholzer: I am actually on the fence here. A program that would trigger this canonicalization is borderline invalid IR. I cannot imagine a situation where this would happen with qubit tensors. But maybe I am missing something.

mlir/lib/Dialect/QTensor/IR/Operations/InsertSliceOp.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  return {};
+}
+
+namespace {
+
+/**
+ * @brief Combine subsequent insertSlice operations with the same offset and
+ * size.
+ */
```
burgholzer: Same question here as in the InsertOp file. 
I do not really think the original IR that would trigger this should be considered valid IR.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -47,7 +49,8 @@ class QCOTest : public testing::TestWithParam<QCOTestCase> {
   void SetUp() override {
     // Register all necessary dialects
     DialectRegistry registry;
-    registry.insert<QCODialect, arith::ArithDialect, func::FuncDialect>();
+    registry.insert<QCODialect, arith::ArithDialect, func::FuncDialect,
+                    qtensor::QTensorDialect, tensor::TensorDialect>();
```
burgholzer: Just a note, feel free to resolve afterwards: Here the tensor dialect is listed explicitly.

mlir/unittests/programs/qco_programs.cpp
```diff
@@ -2133,4 +2133,113 @@ void nestedFalseIf(QCOProgramBuilder& b) {
         return llvm::to_vector(innerResult);
       });
 }
+
+void qtensorAlloc(QCOProgramBuilder& b) { b.qtensorAlloc(3); }
+
+void qtensorDealloc(QCOProgramBuilder& b) {
+  auto qtensor = b.qtensorAlloc(3);
+  b.qtensorDealloc(qtensor);
+}
```
burgholzer, line 2142: Definitely nothing for this PR, but maybe something for @denialhaag in a follow-up PR: I believe it would be quite convenient for the builder to be a little bit more ergonomic in the sense that the `qtensor` coming from `b.qtensorAlloc` would be usable just like a C++ class representing a vector of qubits.
So that one could do things like `qtensor[0]` for example and that would be translated to a `qtensorExtract` instruction internally in the builder.

Eventually, this could bring us very close to the `QuantumComputation` API that we currently have; and, hence, also to the respective Python bindings for building circuits.

mlir/unittests/programs/qco_programs.cpp
burgholzer, line 1: Can we try to get the coverage report to a satisfactory level by adding a couple of tests which trigger the "negative" paths in the folds and canonicalizations? e.g., subsequent inserts but at different indices, etc.

burgholzer (changes requested):
Hey @li-mingbao and @denialhaag 👋🏼 
Sorry that it took me a bit longer to get to this.
I have now managed to get through the PR in quite a bit of detail.
The good thing: I think none of the comments that I have are roadblocks or fundamental limitations. So, great work. Love to see it 🎉 
The bad thing: It's still quite a couple of comments and suggestions for slight rewordings and/or tweaks. I hope these make sense and it does not take too long to incorporate them.

This feels like the last round of reviews needed here and likely to be approved after the comments are addressed 🎉

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: This would require to include the `QCOOps.td` which somehow causes a lot of trouble. So this is more or less a workaround so that you can check for the type without directly including it in tablegen.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: The type `tensor` is a builtin type and not part of the `tensor` dialect (I was also surprised).

denialhaag: I was also surprised to find that out while working on Jeff!

For reference: https://mlir.llvm.org/docs/Dialects/TensorOps/

> On the `tensor` type itself, note that it is actually a builtin type (it lives in the builtin dialect), and does not live in this dialect.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: Yes, in this case here the output is always known. The outTensor has always the same type as the input tensor and the extracted element always has the same type as the element type of the input tensor.

This is how the `tensor` dialects define the assemblyformat and afaik the `qco.measure` also does not state the `i1` return type as this is always fix.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: This is more or less just a specialization because both `alloc` methods for `qtensors` have a `static` size.
During the construction it is always known what size the created tensor has (either through the int attribute or the number of input values).

Also it prevents that something like this is written:
`%tensor = qtensor.from_elements %q0, %q1, %q2 : tensor<?x!qco.qubit>`

While this is in theory correct, it makes sense to make it more precise so that we can check for boundaries better for example.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
li-mingbao: Thanks for the hint, I forgot to remove that. This was here since the original implementation still used a few standard `tensor` operations.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: Ah, interesting. Thanks for the info. Definitely did not know that. At least this helped to point out the one case where the tensor dialect was still being linked 😌

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: And the size for the allocs must be static? For the "fromElements" one, this is clear. For the regular alloc, I would have imagined that a runtime value can be used as well.
Imagine a quantum algorithm written as a function that takes the number of qubits as an input. It would allocate a dynamic array of qubits based on the function parameter. Is this supported at the moment?
If yes, all good. If no, then no need to fix this as part of this PR (if it's not trivial), but it would be good to denote the limitation somewhere.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: Fair enough. It does include all the necessary information I suppose. Sorry for the noise. Feel free to resolve.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: Currently the `alloc` operation only takes an int attribute but I can modify it so it takes a Value as its parameter and just check if the Value is constant in the builder and otherwise build a dynamic tensor.

mlir/lib/Dialect/QTensor/IR/QTensorOps.cpp
li-mingbao: The only other place that would fit is the `Utils.h` file probably. Should I move it there?

mlir/lib/Dialect/QTensor/IR/Operations/InsertOp.cpp
li-mingbao: Yeah I was unsure about that too, While in theory it is possible to write a program that triggers this, it is I think highly unlikely. I added this for completeness sake just in case this somehow happens. 

I am fine with letting it here or removing it.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractOp.cpp
li-mingbao: You are right, since I limited the inputs to `qco::QubitType` this should always be true.

mlir/lib/Dialect/QTensor/IR/QTensorOps.cpp
burgholzer: I just checked the MLIR source code a little more and I do not think this helper is necessary at all.
Looking at https://mlir.llvm.org/doxygen/TensorOps_8cpp_source.html#l02936 for example shows how this is done in the original source code. If you search for `isSame`, you'll find many more such patterns.
https://mlir.llvm.org/doxygen/TensorOps_8cpp_source.html#l01365 als has a nice example with `llvm::equal` and `Value` that seems to be applicable here.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: Could you quantify "a lot of trouble" a little bit? The current solution here very much feels like a hack to me.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: That seems reasonable to me. It's a use case that should definitely be supported eventually.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractOp.cpp
burgholzer: Then feel free to remove any such checks from the folds/canonicalizations.

mlir/lib/Dialect/QTensor/IR/Operations/InsertOp.cpp
burgholzer: Okay. I think I'd rather remove this then. It's yet another pattern that needs to be matched against but will likely never actually match in practice. As such it feels like dead code that I'd rather not like to maintain.

mlir/lib/Dialect/QTensor/IR/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+  ${MQT_MLIR_SOURCE_INCLUDE_DIR}/mlir/Dialect/QTensor
+  DEPENDS
+  MLIRQTensorOpsIncGen
+  LINK_LIBS
+  PUBLIC
+  MLIRIR
+  MLIRQCODialect
+  MLIRArithDialect
+  MLIRInferTypeOpInterface
+  MLIRSideEffectInterfaces
```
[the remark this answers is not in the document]
burgholzer: @li-mingbao the suggestion by @coderabbitai is very reasonable here

mlir/lib/Dialect/QTensor/IR/Operations/DeallocOp.cpp
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @brief Remove matching allocation and deallocation pairs without operations
+ * between them.
+ */
+struct RemoveAllocDeallocPair final : OpRewritePattern<DeallocOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(DeallocOp op,
+                                PatternRewriter& rewriter) const override {
+    // Check if the predecessor is a qtensor::AllocOp
```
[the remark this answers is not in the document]
burgholzer: @li-mingbao Although @coderabbitai is a bit pedantic here, I believe the suggestion is actually an improvement of the documentation

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: I changed it so it takes a value now instead of an attribute. If it is constant it creates a static sized tensor otherwise it is dynamically sized.

mlir/lib/Dialect/QTensor/IR/QTensorOps.cpp
li-mingbao: Thanks for the hint. I removed the helper function and used the `getAsOpFoldResult` for comparison.

mlir/unittests/programs/qco_programs.cpp
denialhaag: Sounds good! I can take care of this when we integrate the registers more into our infrastructure!

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: Apparently if I want to include the `QCOOps.td` in the `QTensorOps.td` file, it tries to generate the full implementations of the qco operations into the generated `QTensorOps.cpp.inc` file. This causes that the implementations are now generated twice. 

According to a LLM either this works or you need to split up the `QCOOps.td` file into an operation and a type tablegen file and just include the types.

mlir/lib/Dialect/QTensor/IR/Operations/ExtractSliceOp.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+  return insertSliceOp;
+}
+
+LogicalResult ExtractSliceOp::fold(FoldAdaptor /*adaptor*/,
+                                   SmallVectorImpl<OpFoldResult>& results) {
+  if (auto insertOp = foldExtractAfterInsertSlice(*this)) {
+    results.emplace_back(insertOp.getDest());
+    results.emplace_back(insertOp.getSource());
+    return success();
```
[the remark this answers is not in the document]
li-mingbao: This is not necessary as you check the index and the size before. This along with the linear type constraint and 1D -QubitTensor constraint guarantees that the types must match.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: Ah. I have seen this distinction being made a couple of times in the past.
I'll have a quick look once all the other feedback is addressed. Don't let this block you for now.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -161,6 +164,198 @@ void QCOProgramBuilder::updateQubitTracking(Value inputQubit,
...
+  checkFinalized();
+
+  auto tensorType = llvm::dyn_cast<RankedTensorType>(tensor.getType());
+
+  if (!tensorType || tensorType.getRank() != 1) {
+    llvm::reportFatalUsageError("Tensor must be of 1-D RankedTensorType!");
+  }
+  if (!llvm::isa<QubitType>(tensorType.getElementType())) {
+    llvm::reportFatalUsageError("Elements must be of QubitType!");
+  }
```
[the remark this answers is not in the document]
li-mingbao: This check is now done in the validateTensorValue function, that also checks if the tensor value is unconsumed.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
li-mingbao: I adjusted the descriptions description in most instances in this tablegen file and in the builder. I explicitly stated everywhere that the values are consumed when they are inserted in any way.
Feel free to remark if I need to change anything.

mlir/unittests/programs/qco_programs.cpp
li-mingbao: I added additional tests that improved the Coverage by a bit, the remaining lines are mostly outputting errors.

mlir/include/mlir/Dialect/QTensor/IR/QTensorOps.td
burgholzer: @denialhaag could you, once #1570 merges, make the corresponding adaptations here? Both structurally as well as the particular change proposed by me above?

denialhaag: Sure! Both done. 😌

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -1060,3 +1062,56 @@ INSTANTIATE_TEST_SUITE_P(
...
+INSTANTIATE_TEST_SUITE_P(
+    QTensorTest, QCOTest,
+    testing::Values(
+        QCOTestCase{"QTensorAlloc", MQT_NAMED_BUILDER(qtensorAlloc),
+                    MQT_NAMED_BUILDER(qtensorAlloc)},
+        QCOTestCase{"QTensorAllocDealloc", MQT_NAMED_BUILDER(qtensorDealloc),
+                    MQT_NAMED_BUILDER(qtensorAlloc)},
+        QCOTestCase{"QTensorFromElements",
+                    MQT_NAMED_BUILDER(qtensorFromElements),
+                    MQT_NAMED_BUILDER(qtensorFromElements)},
```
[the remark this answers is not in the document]
denialhaag: This is fine for now. The tests still ensure that the builder methods exit successfully.

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Add conversions between Jeff and QCO ([#1479], [#1548], [#1565]) ([**@denialhaag**])
 - ✨ Add a `place-and-route` pass for mapping circuits to architectures with restricted topologies ([#1537], [#1547], [#1568]) ([**@MatthiasReumann**])
 - ✨ Add initial infrastructure for new QC and QCO MLIR dialects
-  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1548], [#1550], [#1554], [#1570])
+  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1542], [#1548], [#1550], [#1554], [#1570])
```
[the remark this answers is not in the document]
denialhaag: I'd say QTensor is a helper dialect for QCO. I wouldn't put it on the same level in the changelog.

PR #1565, opened 2026-03-16
⬆️ Update `jeff-mlir`

denialhaag:
## Description

This PR updates `jeff-mlir` to the latest version, allowing us to use the conversion passes added in https://github.com/PennyLaneAI/jeff-mlir/pull/10.

## Checklist


cmake/ExternalDependencies.cmake
```diff
@@ -41,7 +41,7 @@ if(BUILD_MQT_CORE_MLIR)
   FetchContent_Declare(
     jeff-mlir
     GIT_REPOSITORY https://github.com/PennyLaneAI/jeff-mlir.git
-    GIT_TAG c22047e0ebe361126f042baf3c439e0042827a03)
+    GIT_TAG e76be7dd40e6f1e5b1d74ce34b547816d1b67390)
```
denialhaag: I'll update this to the merge commit once we have it.

mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
```diff
@@ -12,24 +12,26 @@ def JeffToQCO : Pass<"jeff-to-qco"> {
...
+    - Support for multiple functions is currently limited
+  }];
 
   let dependentDialects = [
-     "mlir::arith::ArithDialect",
-     "mlir::jeff::JeffDialect",
-     "mlir::qco::QCODialect",
+    "mlir::arith::ArithDialect",
+    "mlir::math::MathDialect",
+    "mlir::tensor::TensorDialect",
```
burgholzer, line 33: Do we really need these dependents here? They are now dependents of the Jeff dialect, so this should work transitively as well, right?

```suggestion
```

mlir/include/mlir/Conversion/QCOToJeff/QCOToJeff.td
```diff
@@ -12,14 +12,17 @@ def QCOToJeff : Pass<"qco-to-jeff"> {
...
-   }];
+    Note that this pass is still in development as QCO and Jeff do not have full feature parity yet.
+  }];
 
   let dependentDialects = [
-     "mlir::jeff::JeffDialect",
-     "mlir::qco::QCODialect",
+    "mlir::arith::ArithDialect",
+    "mlir::math::MathDialect",
+    "mlir::tensor::TensorDialect",
```
burgholzer, line 24:
```suggestion
```
Same question here.
But in the end, it also isn't too important and we can also keep these around.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -984,6 +914,15 @@ struct JeffToQCO final : impl::JeffToQCOBase<JeffToQCO> {
...
     auto* module = getOperation();
 
+    {
+      PassManager pm(context);
+      pm.addPass(createJeffToNative());
+      if (pm.run(module).failed()) {
+        signalPassFailure();
+        return;
+      }
+    }
```
burgholzer: This looks fine in principle. I was wondering for a bit, whether we should rather simply add the respective patterns defined in jeff-mlir to the conversion pipeline here, but probably the solution here is fine. Just wanted to bring up the thought. Maybe some inspiration can be taken from the various MLIR to LLVM conversion passes.

mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
denialhaag: I'm also a bit surprised that this is necessary, but the test execution fails with the following error if we remove them:

```
Assertion failed: (impl->multiThreadedExecutionContext == 0 && "appending to the MLIRContext dialect registry while in a " "multi-threaded execution context"), function appendDialectRegistry, file MLIRContext.cpp, line 422.
```

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: Yes, I guess we could define a function similar to [`arith::populateArithToLLVMConversionPatterns()`](https://github.com/llvm/llvm-project/blob/25981794438bf8d006e13a939c27d5f879fd4c0d/mlir/include/mlir/Conversion/ArithToLLVM/ArithToLLVM.h#L25-L26). 🤔

mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
burgholzer: Interesting. But then again, it does not really hurt to keep these. They are necessary after all (if only transitively). Feel free to resolve this.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: This would, however, also mean that we would have to redo the target definition. In particular, the conversion to Jeff is a bit more complicated because it requires marking some `arith` operations as legal before they are canonicalized away.

burgholzer: Ah, yeah. this is along the lines of what I had in mind.

On the second point: I believe we still have some work to be done there. As far as I know, canonicalization should not be relied on for correctness.
As such, it feels kind of wrong to have a half-baked conversion that then needs canonicalization in order for the code to be serializable.
Is there a way that we could add the respectively needed patterns to the conversion rewriter as well so that it is really just one go?

denialhaag: Hmm, maybe.

We start with this:

```
func.func @get_length(%arg0: tensor<?xi32>) -> i32 {
  %0 = jeff.int_array_length %arg0 : tensor<?xi32> -> i32
  return %0 : i32
}
func.func @main() {
  %0 = jeff.int_array_const32([1, 2, 3]) : tensor<?xi32>
  %1 = call @get_length(%0) : (tensor<?xi32>) -> i32
  return
}
```

After the conversion to built-in operations, we have this:

```
func.func @get_length(%arg0: tensor<?xi32>) -> i32 {
  %dim = tensor.dim %arg0, %c0 : tensor<?xi32>
  %0 = arith.index_cast %dim : index to i32
  return %0 : i32
}
func.func @main() {
  %cst = arith.constant dense<[1, 2, 3]> : tensor<3xi32>
  %cast = tensor.cast %cst : tensor<3xi32> to tensor<?xi32>
  %0 = call @get_length(%cast) : (tensor<?xi32>) -> i32
  return
}
```

Here, both cast operations are required for a valid IR; without them, the values would not comply with the function types.

Now the question is how we want to go back to Jeff's operations.

1. Currently, I'm adding another set of cast operations that then cancel out the ones from the first conversion. For this, the cast operations have to be marked as legal. The more I think about this, the more issues I see.
2. We could probably also mark the cast operations as illegal and simply remove them. The conversion only creates Jeff operations of the correct type, and we should never need to cast anything.

I'll try to implement option 2 and get back to this!

burgholzer: Option 2 sounds reasonable here. And I think it is worth to think about this for a bit to get it right instead of rushing out a solution. So the time trying this out is well spent in my opinion.

mlir/include/mlir/Conversion/JeffToQCO/JeffToQCO.td
denialhaag: This looks much cleaner now because we explicitly mark the dialects as illegal during conversion.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag: Done!

- We now use `jeff::populateNativeToJeffConversionPatterns()` and `jeff::populateJeffToNativeConversionPatterns()`.
- I'll post a comment with some more details on the implementation of option 2 over at https://github.com/PennyLaneAI/jeff-mlir/pull/10.

cmake/ExternalDependencies.cmake
```diff
@@ -41,7 +41,7 @@ if(BUILD_MQT_CORE_MLIR)
   FetchContent_Declare(
     jeff-mlir
     GIT_REPOSITORY https://github.com/PennyLaneAI/jeff-mlir.git
-    GIT_TAG c22047e0ebe361126f042baf3c439e0042827a03)
+    GIT_TAG 41d02b95ec509d586f4c9bd690b6f453a53918e6)
```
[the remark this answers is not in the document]
burgholzer:
```suggestion
    GIT_TAG c7e0d0340e470c9097a79a430b633b97dcd864e9)
```
new stable version just dropped

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
burgholzer: this feels like a really clean solution!

PR #1568, opened 2026-03-17
✨ Implement Trials For Mapping Pass

MatthiasReumann:
## Description

This pull request adds LightSABRE-like trials (configurable via pass options) to improve the mapping result. To avoid that the A* search gets stuck, this pull request also adds a configurable `maxIterations` parameter. 

## Checklist

### DIFF
--- a/mlir/lib/Dialect/QCO/Transforms/Mapping/Architecture.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Mapping/Architecture.cpp
@@ -13,6 +13,7 @@
 #include <llvm/Support/ErrorHandling.h>
 #include <mlir/Support/LLVM.h>
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdint>
 #include <string_view>


mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -406,12 +503,14 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
    * two-qubit operations inside the first layer (the front) executable.
+   * @details The parameter @p maxIterations determines how many nodes will
+   * be explored until the current search is considered a failure.
    * @returns a vector of hardware-index pairs (each denoting a SWAP) or
    * failure() if A* fails.
    */
   [[nodiscard]] static FailureOr<SmallVector<IndexGate>>
   search(ArrayRef<Layer> layers, const Layout& layout, const Architecture& arch,
-         const Parameters& params) {
+         const Parameters& params, const std::size_t maxIterations = 50'000) {
```
burgholzer: This feels like a fairly low number. Especially for larger architectures. Should this somewhat depend on the number of qubits of the architecture?

MatthiasReumann: Good idea. I thought about this a bit. This is what I came up with: 

$$
b := 4 \cdot \lceil \frac{N}{2} \rceil
$$

where $N$ is the number of qubits, the $\frac{N}{2}$ term is the maximum number of two qubit gates in any layer, and $4$ is the maximum connectivity of the architecture. The latter can be hard-coded for now but may depend on the architecture later. The $b$ term is essentially the effective branching factor of A*. Intuitively, this is the maximum amount of child nodes (SWAPs) we create when expanding. 

$$
\text{max}_{iterations} := b^{3}
$$

The maximum number of iterations is then simply the branching factor cubed. Intuitively, we consider the search a failure whenever it visited all paths of length $3$. In reality, the solution paths are likely to be much longer. From a "longest path perspective", if the algorithm visits $b^{3}$ nodes at most, the longest path $l$ should be 

$$
b^{3} = b\cdot l \rightarrow l = b^{2}
$$

because for every expanded node on the path the algorithm visits $b$ nodes. For $9$ qubits this longest path would be $400$. 

#### In Numbers 

| N | Maximum Iterations |
|--------|--------|
| 8 | 4096 |
| 16 | 32768 |
| 32 | 262144 |
| 64 | 2097152 |
| 128 | 16777216 | 
| 256 | 134217728 | 
| 512 | 1073741824 | 
| 1028 | 68719476736 | 

@coderabbitai What's your take on this?

MatthiasReumann: I think the `std::min` is good suggestion. What do you say @burgholzer?

burgholzer: @MatthiasReumann I believe both, your approach, and the suggestions from @coderabbitai seem very reasonable.
It would be good to also include this description as part of the documentation so that people can also read up on the reasoning without having to go to this PR comment.

I think it would make sense that the Architecture class exposes a `maxDegree` method as suggested by @coderabbitai. As this makes the formula truly generic.

The upper cap seems to make sense to me. I think I would personally put the cap a little bit higher though. Maybe somewhere around the range of the 128-150 qubit mark. I would hope that we can still reliably route 150 qubit circuits without running into too many failures. Probably requires some real-world statistics to tweak the actual value.

PR #1396, opened 2025-12-19
✨ Add non-linear control flow conversions between QC and QCO dialects

li-mingbao:
## Description

This PR adds support for the conversion of the `scf` operations `scf.while`, `scf.for`, `scf.if` and the conversion of additional functions between the `qc` and the `qco` dialect. This allows the conversion of programs with nonlinear controlflow.

Reopened since #1391 got automatically closed after the branch was deleted.

## Checklist:

- [ ] The pull request only contains commits that are focused and relevant to this change.
- [ ] I have added appropriate tests that cover the new/changed functionality.
- [ ] I have updated the documentation to reflect these changes.
- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [ ] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.
- [ ] I have reviewed my own code changes.

PR #1569, opened 2026-03-17
✨ Improve Static Qubit Handling

simon1hofmann:
## Description

Based on #1585.

## Checklist

### DIFF
--- a/mlir/include/mlir/Dialect/QC/IR/QCOps.td
+++ b/mlir/include/mlir/Dialect/QC/IR/QCOps.td
@@ -1014,13 +1014,13 @@ def InvOp : QCOp<"inv",
   let extraClassDeclaration = [{
         [[nodiscard]] UnitaryOpInterface getBodyUnitary();
         size_t getNumQubits() { return getBodyUnitary().getNumQubits(); }
-        size_t getNumTargets() { return getNumQubits(); }
-        static size_t getNumControls() { return 0; }
-        static ValueRange getControls() { return {}; }
+        size_t getNumTargets() { return getBodyUnitary().getNumTargets(); }
+        size_t getNumControls() { return getBodyUnitary().getNumControls(); }
         Value getQubit(size_t i) { return getBodyUnitary().getQubit(i); }
-        Value getTarget(size_t i) { return getQubit(i); }
+        Value getTarget(size_t i) { return getBodyUnitary().getTarget(i); }
         ValueRange getTargets() { return getBodyUnitary().getTargets(); }
-        static Value getControl(size_t i) { llvm::reportFatalUsageError("Operation does not have controls"); }
+        Value getControl(size_t i) { return getBodyUnitary().getControl(i); }
+        ValueRange getControls() { return getBodyUnitary().getControls(); }
         size_t getNumParams() { return getBodyUnitary().getNumParams(); }
         Value getParameter(size_t i) { return getBodyUnitary().getParameter(i); }
         ValueRange getParameters() { return getBodyUnitary().getParameters(); }


mlir/include/mlir/Dialect/QC/IR/QCOps.td
```diff
@@ -65,31 +65,42 @@ def AllocOp : QCOp<"alloc", [MemoryEffects<[MemAlloc]>]> {
...
 
+def DynamicQubit : Type<
+    And<[CPred<"::mlir::isa<::mlir::qc::QubitType>($_self)">,
+         CPred<"!::mlir::cast<::mlir::qc::QubitType>($_self).getIsStatic()">]>,
+    "dynamic qubit type (!qc.qubit)">;
+
+def StaticQubit : Type<
+    And<[CPred<"::mlir::isa<::mlir::qc::QubitType>($_self)">,
+         CPred<"::mlir::cast<::mlir::qc::QubitType>($_self).getIsStatic()">]>,
+    "static qubit type (!qc.qubit<static>)">;
```
denialhaag: I would maybe move these defintion in a separate section further above.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -81,7 +81,12 @@ def AllocOp : QCOOp<"alloc", [MemoryEffects<[MemAlloc]>]> {
 def DeallocOp : QCOOp<"dealloc", [MemoryEffects<[MemFree]>]> {
     let summary = "Deallocate a qubit";
     let description = [{
-        Deallocates a qubit, releasing its resources.
+        Deallocates a qubit.
+
+        In QCO's value/linear semantics, this operation also serves as the sink
+        that consumes the qubit SSA value (ensuring every qubit value is used
+        exactly once). When lowering back to QC (reference semantics), deallocs
+        corresponding to static qubits may be erased.
```
denialhaag: I would maybe not use the term "lowering" here. To me, only the QC-to-QIR conversion is a lowering. Maybe that's just my opinion, though.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -96,27 +96,24 @@ struct ConvertQCOAllocOp final : OpConversionPattern<qco::AllocOp> {
...
- *
- * Example transformation:
- * ```mlir
- * qco.dealloc %q_qco : !qco.qubit
- * // becomes:
- * qc.dealloc %q_qc : !qc.qubit
- * ```
+ * For dynamic qubits (`!qco.qubit`), lowers to `qc.dealloc`.
+ * For static qubits (`!qco.qubit<static>`), erases the op since QC does not
+ * require explicit deallocation of static qubits.
```
denialhaag: I would maybe provide two examples here, similar to how we write examples for the other conversions.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +111,77 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+ * @brief Converts func.return and sinks remaining live qubits.
+ *
+ * @details
+ * QC uses reference semantics and does not enforce linear typing for qubits.
+ * After conversion, QCO requires that every qubit SSA value is consumed
+ * exactly once. For allocations (including static qubits), the sink is
+ * `qco.dealloc`. This pattern inserts `qco.dealloc` operations for all
+ * still-live qubits tracked in the lowering state right before the return.
+ */
+struct ConvertFuncReturnOp final : StatefulOpConversionPattern<func::ReturnOp> {
```
denialhaag, line 267: Here, we probably have to be a bit more careful. Once functions can return qubits, this won't work anymore.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -240,7 +311,8 @@ struct ConvertQCStaticOp final : StatefulOpConversionPattern<qc::StaticOp> {
     auto qcQubit = op.getQubit();
 
     // Create new qco.static operation with the same index
-    auto qcoOp = qco::StaticOp::create(rewriter, op.getLoc(), op.getIndex());
+    auto qcoOp = qco::StaticOp::create(rewriter, op.getLoc(),
+                                       static_cast<int64_t>(op.getIndex()));
```
denialhaag: Is this cast necessary? The type should be the same for both operations, right?

simon1hofmann: Because `getIndex` returns `uint64_t` and `StaticOp::create` expects `int64_t`.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: Does the implementation not treat returned qubits correctly? `escapedQubits` holds the return operands, and `liveQubits` is built by skipping those, so returned qubits are never deallocated.

denialhaag: Ahh, my bad... After looking at the implementation more carefully now, I have three comments:

1. Is the check for nested regions necessary? I don't think our current implementation would allow a `func.return` in any of our modifiers.
2. Is the `seen` map necessary? This would be relevant if several QC qubits are mapped to the same QCO qubit, right? If that happened, something went horribly wrong, right?
3. The current implementation doesn't allow for multiple functions anyway, but at the moment, the qubit map is created once for the entire conversion, so we probably shouldn't clear it at the end of the first function. 🤔

simon1hofmann: Addressing your comments:
1. makes sense, can be removed.
2. I tried to simplify it.
3. Not entirely sure, but I think the qubit map has to be cleared for `func.return` to be considered legal and for multiple functions there would be individual qubit maps anyway right?

mlir/include/mlir/Dialect/QC/IR/QCOps.td
simon1hofmann: Moved them further up.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
simon1hofmann: Changed lowering to conversion.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
simon1hofmann: Added two examples for dynamic and static qubits.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -94,7 +113,7 @@ def DeallocOp : QCOOp<"dealloc", [MemoryEffects<[MemFree]>]> {
     let hasCanonicalizer = 1;
 }
 
-def StaticOp : QCOOp<"static", [Pure]> {
+def StaticOp : QCOOp<"static", [MemoryEffects<[MemAlloc]>]> {
```
burgholzer: Is there a particular reason for why we could not keep `Pure` here?
Or, asked differently, do we lose anything by not declaring this operation as pure?

mlir/include/mlir/Dialect/QC/IR/QCTypes.td
```diff
@@ -25,7 +25,18 @@ def QubitType : QCType<"Qubit", "qubit"> {
         QC dialect. Operations using this type modify qubits in place using
         reference semantics, similar to how classical imperative languages handle
         mutable references.
+
+        `!qc.qubit` (default) denotes a dynamically allocated qubit.
+        `!qc.qubit<static>` denotes a qubit that has been mapped to a fixed
+        hardware position.
```
burgholzer:
```suggestion
        `!qc.qubit<static>` denotes a qubit with a statically known identifier.
```
I believe it is helpful to keep this a little more generic.

mlir/include/mlir/Dialect/QCO/IR/QCOTypes.td
```diff
@@ -26,13 +26,24 @@ def QubitType : QCOType<"Qubit", "qubit"> {
         and produce new output qubits following value semantics and the SSA
         paradigm, enabling powerful dataflow analysis and optimization.
 
+        `!qco.qubit` (default) denotes a dynamically allocated qubit.
+        `!qco.qubit<static>` denotes a qubit that has been mapped to a fixed
+        hardware position.
```
burgholzer:
```suggestion
        `!qco.qubit<static>` denotes a qubit with a statically known identifier.
```
I believe it is helpful to keep this a little more generic.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -902,7 +902,7 @@ class JeffToQCOTypeConverter final : public TypeConverter {
     addConversion([](Type type) { return type; });
 
     addConversion([ctx](jeff::QubitType /*type*/) -> Type {
-      return qco::QubitType::get(ctx);
+      return qco::QubitType::get(ctx, /*isStatic=*/false);
```
burgholzer:
```suggestion
      return qco::QubitType::get(ctx);
```
Shouldn't this be the default?
I guess it doesn't really hurt to be explicit, but it also does not really feel necessary.

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/BarrierOp.cpp
```diff
@@ -77,6 +77,21 @@ struct MergeSubsequentBarrier final : OpRewritePattern<BarrierOp> {
...
+  // !qco.qubit)
+  const auto numTargets = getNumTargets();
+  for (size_t i = 0; i < numTargets; ++i) {
+    if (getQubitsIn()[i].getType() != getQubitsOut()[i].getType()) {
+      return emitOpError("qco.barrier qubit ")
+             << i << " input type must match output type (got "
+             << getQubitsIn()[i].getType() << " vs "
+             << getQubitsOut()[i].getType() << ")";
+    }
+  }
```
burgholzer: Haven't checked in the IDE, but there might be a more concise way to write this, either via `llvm::zip` or via something like `getQubitsIn().getTypes() == getQubitsOut().getTypes()` 
(i vaguely remember that something like that exists)

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -1060,6 +1115,24 @@ INSTANTIATE_TEST_SUITE_P(
...
+                    MQT_NAMED_BUILDER(staticQubitsWithTwoTargetOps)},
+        QCOTestCase{"StaticQubitsWithCtrl",
+                    MQT_NAMED_BUILDER(staticQubitsWithCtrl),
+                    MQT_NAMED_BUILDER(staticQubitsWithCtrl)},
+        QCOTestCase{"StaticQubitsWithInv",
+                    MQT_NAMED_BUILDER(staticQubitsWithInv),
+                    MQT_NAMED_BUILDER(staticQubitsWithInv)},
+        QCOTestCase{"MixedStaticDynamicQubits",
+                    MQT_NAMED_BUILDER(mixedStaticDynamicQubits),
+                    MQT_NAMED_BUILDER(mixedStaticDynamicQubits)},
```
burgholzer: I think it would make sense to also add the QC equivalent of these programs and add them to the pipeline and all other related tests.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -127,6 +127,61 @@ TEST_F(QCOTest, DirectIfBuilder) {
...
+  EXPECT_EQ(xxpyOp.getQubit0Out().getType(), staticType);
+  EXPECT_EQ(xxpyOp.getQubit1Out().getType(), dynamicType);
+
+  DeallocOp::create(builder, xxpyOp.getQubit0Out());
+  DeallocOp::create(builder, xxpyOp.getQubit1Out());
+  auto module = builder.finalize();
+  ASSERT_TRUE(module);
+  EXPECT_TRUE(verify(*module).succeeded());
+}
+
```
burgholzer: This test feels kind of redundant given the explicit type verifiers that ensure this. I'd have a slight preference to remove it.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -254,6 +256,26 @@ void CtrlOp::build(
...
+    }
+  }
+  for (size_t i = 0; i < getNumTargets(); ++i) {
+    if (getTargetsIn()[i].getType() != getTargetsOut()[i].getType()) {
+      return emitOpError("qco.ctrl target ")
+             << i << " input type must match output type (got "
+             << getTargetsIn()[i].getType() << " vs "
+             << getTargetsOut()[i].getType() << ")";
+    }
+  }
```
burgholzer: As per the other comment, I think there is a more concise way to write these.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -144,7 +156,8 @@ struct ConvertQCOStaticOp final : OpConversionPattern<qco::StaticOp> {
   matchAndRewrite(qco::StaticOp op, OpAdaptor /*adaptor*/,
                   ConversionPatternRewriter& rewriter) const override {
     // Create qc.static with the same index
-    rewriter.replaceOpWithNewOp<qc::StaticOp>(op, op.getIndex());
+    rewriter.replaceOpWithNewOp<qc::StaticOp>(
+        op, static_cast<int64_t>(op.getIndex()));
```
burgholzer: The casts added throughout this PR are a little distracting.
Did some underlying type change here or were we simply ignoring a warning that was there before?
(This also holds for other places throughout the PR)

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
burgholzer: Regarding 3): I think this is what @li-mingbao already had set up to some degree in #1396

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +111,71 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+    // Build return values from qubitMap (adaptor.getOperands() may carry stale
+    // root values because gate patterns use eraseOp instead of replaceOp).
+    llvm::SmallVector<Value> returnValues;
+    llvm::DenseSet<Value> escapedQubits;
+    returnValues.reserve(op.getNumOperands());
+    for (auto [qcOperand, adaptorOperand] :
+         llvm::zip(op.getOperands(), adaptor.getOperands())) {
+      if (state.qubitMap.contains(qcOperand)) {
+        auto latest = state.qubitMap[qcOperand];
+        returnValues.push_back(latest);
```
burgholzer:
```suggestion
        returnValues.emplace_back(latest);
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +111,71 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+        escapedQubits.insert(latest);
+      } else {
+        returnValues.push_back(adaptorOperand);
```
burgholzer:
```suggestion
        returnValues.emplace_back(adaptorOperand);
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +111,71 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+      }
+    }
+
+    // Collect non-escaped live qubits for deallocation.
+    llvm::DenseSet<Value> liveQubitsSet;
```
burgholzer:
```suggestion
    llvm::DenseSet<Value> liveQubits;
```
This is me just being pedantic, but it is clear from the type that this is a set.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +111,71 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+    for (const auto& [qcQubit, qcoQubit] : state.qubitMap) {
+      if (!escapedQubits.contains(qcoQubit)) {
+        liveQubitsSet.insert(qcoQubit);
+      }
+    }
```
burgholzer: I don't know if it changes much, but LLVM has the concept of intersections of sets, which could be a shorthand for some of the code here. Obviously not a must-have, but might create cleaner code overall.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +111,71 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+                                        liveQubitsSet.end());
+
+    // Sort deterministically (mirrors QCOProgramBuilder::finalize()).
+    llvm::sort(liveQubits, [](Value a, Value b) {
+      auto* opA = a.getDefiningOp();
+      auto* opB = b.getDefiningOp();
+      if (!opA || !opB || opA->getBlock() != opB->getBlock()) {
+        return a.getAsOpaquePointer() < b.getAsOpaquePointer();
+      }
+      return opA->isBeforeInBlock(opB);
```
burgholzer: Since this is also used in the QCO program builder, it might be nice to extract this into the utils file so that it can be reused and doesn't go out of sync over time.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
burgholzer: See also my other comment, I was also confused by this.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
simon1hofmann: With all of the recent changes, I think we can keep it as `Pure`.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
simon1hofmann: Yes makes sense, I added them to the pipeline and all other related tests.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
simon1hofmann: With all of the other new test cases, makes sense to remove these.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
simon1hofmann: Implemented similar to BarrierOp.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: We need map values minus escaped, not intersection of two sorted sets; `llvm::set_*` would require sorting copied ranges. Switched to `llvm::make_second_range(qubitMap)` so we only walk the QCO values.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: Makes sense, created a new utility function.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
simon1hofmann: Without, I get the following `clang-tidy` warnings: 
> narrowing conversion from 'uint64_t' (aka 'unsigned long') to signed type 'int64_t' (aka 'long') is implementation-define

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
simon1hofmann: Comment by CodeRabbit:

⚠️ Potential issue | 🟠 Major

qco.static still should not be Pure.

The type split does not change the generic MLIR optimization behavior here: identical qco.static N ops remain CSE/hoistable, which can merge distinct linear ownership chains for the same physical qubit. This op should stay allocation-like instead of pure.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: I tried to solve it using minimal adjustments to the qubit map similar to #1396.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -111,6 +112,64 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+/**
+ * @brief Converts func.return and sinks remaining live qubits.
+ *
+ * @details
+ * QC uses reference semantics and does not enforce linear typing for qubits.
+ * After conversion, QCO requires that every qubit SSA value is consumed
+ * exactly once. For allocations (including static qubits), the sink is
+ * `qco.dealloc`. This pattern inserts `qco.dealloc` operations for all
+ * still-live qubits tracked in the lowering state right before the return.
+ */
```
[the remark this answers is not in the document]
burgholzer: This feels like a false positive. The QC to QCO direction must explicitly insert deallocs for static qubits to act as sinks for the linear qubit type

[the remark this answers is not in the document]
simon1hofmann: > This feels like a false positive. The QC to QCO direction must explicitly insert deallocs for static qubits to act as sinks for the linear qubit type

That's what I thought as well. Still have to get used to critically assessing all coderabbit comments.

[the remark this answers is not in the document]
burgholzer: > > This feels like a false positive. The QC to QCO direction must explicitly insert deallocs for static qubits to act as sinks for the linear qubit type
> 
> That's what I thought as well. Still have to get used to critically assessing all coderabbit comments.

Yeah. It's good in many aspects; but one has to, very critically, assess all its comments and be proactive about disagreeing. It also really tends to be overly cautious with defensive checks that simply bloat the code.
In case you ever feel unsure, feel free to tag me (or Daniel) in any of the comments to clarify.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
burgholzer: I reworked the underlying logic a little bit to get rid of all the unnecessary casts in https://github.com/munich-quantum-toolkit/core/pull/1569/commits/d6ad7c4841901abe0d23daf85906a209ad89c2f8

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -72,8 +74,13 @@ namespace {
...
 struct LoweringState {
-  /// Map from original QC qubit references to their latest QCO SSA values
-  llvm::DenseMap<Value, Value> qubitMap;
+  /// Per-region map from QC qubit references to latest QCO SSA values.
+  ///
+  /// @details Keys are `Operation::getParentRegion()` for ops being converted
+  /// (typically a `func.func` body, or a `qc.ctrl` / `qc.inv` region). This
+  /// avoids clearing state at the first `func.return` while later functions
+  /// still convert.
+  llvm::DenseMap<Region*, llvm::DenseMap<Value, Value>> qubitMap;
```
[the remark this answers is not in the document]
burgholzer: Tests are passing just fine in CI and locally. So this does not seem to be a problem

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -79,16 +79,10 @@ Value QCProgramBuilder::allocQubit() {
...
 
-  if (index < 0) {
-    llvm::reportFatalUsageError("Index must be non-negative");
-  }
-
-  // Create the StaticOp with the given index
-  auto indexAttr = getI64IntegerAttr(index);
-  auto staticOp = StaticOp::create(*this, indexAttr);
+  auto staticOp = StaticOp::create(*this, index);
   return staticOp.getQubit();
```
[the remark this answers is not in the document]
burgholzer: This is fine as it will never happen in practice and other things are way more likely to break down before that

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: @coderabbitai given how the `qc.static` operation is pure and will be reduced by CSE and hoisted upwards in the program, there should never be two distinct linear ownership chains for the same physical qubit in QCO (by construction). Does that make sense?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -487,17 +545,17 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
       const auto hw = layout.getHardwareIndex(p);
       rewriter.setInsertionPoint(q.getDefiningOp());
       auto op = rewriter.replaceOpWithNewOp<StaticOp>(q.getDefiningOp(), hw);
-      statics[hw] = op.getQubit();
+      statics[hw] = llvm::cast<TypedValue<QubitType>>(op.getQubit());
```
MatthiasReumann: IntelliSense tells me that the return type here is already `::mlir::TypedValue< ::mlir::qco::QubitType>`. Why the cast?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -55,14 +57,67 @@ namespace mlir::qco {
...
+ * `!qco.qubit` results from before placement. Ops with `TypesMatchWith` /
+ * same-type traits then fail verification until results are refreshed.
+ */
+static void synchronizeMappedQubitTypes(Region& region) {
+  region.walk<WalkOrder::PreOrder>([](Operation* op) {
+    const auto synchronizeQubitTypes = [](Value input, Value output) {
+      const auto qIn = dyn_cast<QubitType>(input.getType());
+      const auto qOut = dyn_cast<QubitType>(output.getType());
+      if (qIn && qOut && qIn != qOut) {
+        output.setType(qIn);
```
MatthiasReumann: I think the best practice would be to replace each unitary op (which consumes and produces `!qco.qubit` types) with the same one but different typing. Generally speaking, any IR rewriting should be done via the `IRRewriter` (or `PatternRewriter` if you work on the conversions). 

From the documentation: 
> Note that this [.setType()] is an extremely dangerous operation which can create completely invalid IR very easily. It is strongly recommended that you recreate IR objects with the right types instead of mutating them in place.

That being said, I don't think there is a cleaner option than traversing the IR another time. Alternatively, one could do it in the `commitTrial` function - but things will get messy then. Gemini's suggestion was to use the Conversion framework and the `TypeConverter` but that's a lot of boilerplate code for little functionality.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Have you regenerated your build folder? I was wondering the same, but apparently something changed by adding the additional attribute so that the getter is no longer returning the actual type, but rather a generic type.
Let's look into this a little bit more.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Thanks for the hint with the IR rewriter 👍🏼 
The conversion framework seems a little overkill here.
I was hoping that we could do this "on the fly" as part of the commitTrial function.
Let's entertain that idea for a bit. I am not sure whether I will have the capacity to look at this today; maybe @simon1hofmann or @MatthiasReumann you got some time to take a look?

MatthiasReumann: Yeah. I think it is possible to do this in the `commitTrial` function, the question is how ugly does that get. Maybe it's not even _that_ bad. @simon1hofmann Added you on Discord. Maybe let's discuss this in person.

Maybe something else entirely that could help this process: Have you ever thought about implementing the `InferTypeOpInterface` interface? @burgholzer @denialhaag

burgholzer: fun fact: before this PR, most operations in this PR automatically got that interface, simply by including the respective header, which we found out a couple months back. Now, with this PR, we apparently lost some of that. It's a nice exercise to compare the generated dialect files in the build folder before and after this PR.

while working on this PR a couple days back, I had already implemented the interface for the modifier and SCF operations, where the benefit was most apparent. However, I reverted those again as there was little to no perceived benefit for quite some boilerplate.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: We just had a look at this. Apparently, this is because `qco.static` returns the `StaticQubit` type which exists in Tablegen but doesn't exist as class in C++ (In other words: Tablegen doesn't generate a class for it). Consequently, the `getQubit` uses the generic `::mlir::Type` class. If we update the `qco.static` operation to return the `QubitType` (which exists as generated C++ class) it correctly types the return value.

MatthiasReumann: Thus, there are two options:
1. Use the `QubitType` as return type for the `qco.static` operation (Instead of `StaticQubit`)
2. Cast to `QubitType` in the passes

burgholzer: Hm. The first one loses a bit of type information, but leads to cleaner pass code. I think I would prefer that.
Maybe as a third option: Find a solution so that different C++ types are emitted for the different qubit types. I remember playing with this for a bit while working on the PR, but did not find a solution that I was really happy with.

MatthiasReumann: I think the only option to generate two different types is to define two different types. Then the Unitary operations had to do something like 

```tablegen
def HOp : Op<"h", [ ... ]> {
  let arguments = (ins AnyTypeOf<[StaticQubitType, DynamicQubitType]>:$qubit);
}
```

burgholzer: Which is not particularly pretty as well..
Hm.. I am starting to dislike the complexity this distinction introduces..

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -55,14 +59,295 @@ namespace mlir::qco {
...
+                                                inv.getQubitsIn()) ||
+         qubitOperandAndResultTypesDiffer(inv);
+}
+
+/** @brief Copies op results into `SmallVector<Value>` (OpResult → Value). */
+[[nodiscard]] static llvm::SmallVector<Value> opResultsAsValues(Operation* op) {
+  llvm::SmallVector<Value> vals;
+  llvm::append_range(vals, op->getResults());
+  return vals;
+}
```
burgholzer: This is basically `llvm::to_vector`, right?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -55,14 +59,295 @@ namespace mlir::qco {
...
+ *             (`getOutputQubits()`).
+ * @return True when either (1) a body block argument type differs from the
+ *         matching `getTargetsIn()` operand, or (2) any qubit result type
+ *         differs from the paired input qubit type (via
+ *         @ref modifierBodyArgsMismatchTargetOperands and
+ *         @ref qubitOperandAndResultTypesDiffer).
+ *
+ * @details Pure predicate: no IR changes. Callers use this to skip redundant
+ *          `replaceOpWithNewOp` work.
+ */
```
burgholzer: Can't really help it. These docstrings feel very AI generated and somehow more complicated than they need to be. Also holds for other docstrings generated through this file as part of the PR.

simon1hofmann: I'm not really a fan of the changes in Mapping.cpp anyways and first wanted to get feedback from @MatthiasReumann. I added the docstrings for context, but I guess they just make it even more confusing.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: I was thinking the same. I discussed this a little bit with @simon1hofmann, but there is no _big_ benefit of adding another type besides less "management" logic in the conversion passes, right?

burgholzer:
> @burgholzer @simon1hofmann
> 
> 
> 
> > Am I correct in assuming that our problems would be entirely gone if we would simply deallocate all the qubits, no matter if they are dynamic or static? The only downside we identified from that was that the semantics in QC for static qubits are not "perfect" as the deallocation of a static qubit really does not make too much sense.
> 
> However, in the grand scheme of things, this might be the easiest solution.
> 
> 
> 
> I think you could even keep the "The QC dialect doesn't dealloc static qubits" with just one qubit type. It just requires one more datastructure in the conversion passes that remembers which static qubits it still has to deallocate (QC -> QCO) and which deallocs are to be removed (QCO -> QC). (Even though I still have the strong opinion that `qco.dealloc` should then be renamed 🙈) 
> 
> 
> 
> I am not even sure if that much more logic in the conversions would be required: 
> 
> - In the QC dialect we could simply iterate over all `qc.static` ops (which in the QC dialect don't have an associated `qc.dealloc`) and utilize the references semantics to add the missing `qc.dealloc`s. 
> 
> - In the QCO dialect it requires a `LoweringState` (analogously to the QC-to-QCO conversion) which tracks the current SSA values of static qubits and removes the `qco.dealloc` operations for these static qubits. 
> 
> 
> 
> Overall this is probably less code than the one currently added to `Mapping.cpp`. 
> 
> 
> 
> 

This feels like a very reasonable proposal. The only concern that I have is the one that @DRovara brought up elsewhere: can we really always track which qubits are static; especially in the presence of control flow that depends on runtime values?

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -1158,9 +1144,19 @@ struct QCToQCO final : impl::QCToQCOBase<QCToQCO> {
...
+    // to insert sink operations (`qco.sink`) for remaining live
+    // qubits. Therefore, we make it dynamically illegal unless the lowering
+    // state has no remaining qubits.
+    patterns.add<ConvertFuncReturnOp>(typeConverter, context, &state);
+    target.addDynamicallyLegalOp<func::ReturnOp>([&](const func::ReturnOp op) {
+      if (!typeConverter.isLegal(op)) {
+        return false;
+      }
+      const auto it = state.qubitMap.find(op->getParentRegion());
+      return it == state.qubitMap.end() || it->second.empty();
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai The distinction between different qubit types using `!qc.qubit<static>` does not exist anymore.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -26,13 +26,37 @@
...
+    return QubitMode::Mixed;
+  }
+  if (sawStatic) {
+    return QubitMode::StaticOnly;
+  }
+  if (sawAlloc) {
+    return QubitMode::DynamicOnly;
+  }
+  return QubitMode::Unknown;
+}
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai The distinction between different qubit types using !qco.qubit<static> does not exist anymore.

mlir/unittests/Conversion/QCToQIR/test_qc_to_qir.cpp
```diff
@@ -624,6 +624,21 @@ INSTANTIATE_TEST_SUITE_P(
...
+                        MQT_NAMED_BUILDER(qir::staticQubitsWithParametricOps)},
+        QCToQIRTestCase{"StaticQubitsWithTwoTargetOps",
+                        MQT_NAMED_BUILDER(qc::staticQubitsWithTwoTargetOps),
+                        MQT_NAMED_BUILDER(qir::staticQubitsWithTwoTargetOps)},
+        QCToQIRTestCase{"StaticQubitsWithCtrl",
+                        MQT_NAMED_BUILDER(qc::staticQubitsWithCtrl),
+                        MQT_NAMED_BUILDER(qir::staticQubitsWithCtrl)},
+        QCToQIRTestCase{"StaticQubitsWithInv",
+                        MQT_NAMED_BUILDER(qc::staticQubitsWithInv),
+                        MQT_NAMED_BUILDER(qir::staticQubitsWithInv)},
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai adding mixed static/dynamic success case: Not added because mixed static/dynamic is intentionally rejected earlier; adding as a normal equivalence case would conflict with intended behavior.

mlir/unittests/programs/qco_programs.cpp
```diff
@@ -38,9 +38,50 @@ void staticQubits(QCOProgramBuilder& b) {
...
+  auto q1 = b.allocQubit();
+  std::tie(q0, q1) = b.rzz(0.123, q0, q1);
+  q0 = b.h(q0);
+  q1 = b.h(q1);
+}
+
+void allocSinkPair(QCOProgramBuilder& b) {
   auto q = b.allocQubit();
-  b.dealloc(q);
+  b.sink(q);
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai explicit static-sink helper: Existing static program coverage already exercises static sink behavior (including finalize-time sinks), so no additional helper was necessary for this fix set.

MatthiasReumann:
> I currently do not have the capacity to review all the frameworks.. @munich-quantum-toolkit/mqt-cc anyone got some time for that?

I can do such a review and post it here (soon)!

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -26,13 +26,60 @@
...
+ * @details
+ * Used to track whether a function uses static or dynamic qubit allocation.
+ * This is used to determine whether to convert `qco.sink` to `qc.dealloc` (for
+ * dynamic qubits) or simply erase it (for static qubits). This is also used to
+ * catch cases of mixed addressing being used, which is not supported.
+ */
+struct LoweringState {
+  /// The qubit addressing mode used in the module
+  QubitAddressingMode mode = QubitAddressingMode::Unknown;
+};
```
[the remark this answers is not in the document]
burgholzer: There is no need to scan all functions in a module to infer the addressing mode.
An entire module will only ever use one particular mode.
Alloc and static operations always precede sink operations in the programs.
My assumption, which seems to be backed by all tests passing in CI, was that the rewriter visits the program in sequence, which will always guarantee a mode to be set when encountering a `sink`.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -1158,9 +1105,18 @@ struct QCToQCO final : impl::QCToQCOBase<QCToQCO> {
...
+    // to insert sink operations (`qco.sink`) for dead qubit values. Therefore,
+    // we mark it illegal as long as the qubit map of the region is not empty.
+    patterns.add<ConvertFuncReturnOp>(typeConverter, context, &state);
+    target.addDynamicallyLegalOp<func::ReturnOp>([&](const func::ReturnOp op) {
+      if (!typeConverter.isLegal(op)) {
+        return false;
+      }
+      const auto it = state.qubitMap.find(op->getParentRegion());
+      return it == state.qubitMap.end() || it->second.empty();
+    });
```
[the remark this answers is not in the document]
burgholzer: This seems to be working just fine for now as all tests are passing.
I'd prefer to leave this as-is and change it if problematic situations should come up.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -924,35 +920,15 @@ OwningOpRef<ModuleOp> QCOProgramBuilder::finalize() {
...
-  // Automatically deallocate all still-allocated tensors
-  // Sort tensors for deterministic output
-  llvm::SmallVector<Value> sortedTensors(validTensors.begin(),
-                                         validTensors.end());
-  llvm::sort(sortedTensors, blockOrderComparator);
-
-  for (auto tensor : sortedTensors) {
+  for (auto tensor : validTensors) {
     qtensor::DeallocOp::create(*this, tensor);
   }
```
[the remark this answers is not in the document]
burgholzer: None of that is necessary. The order of these operations is irrelevant as they are not connected. They may be reordered arbitrarily and the program would still have the same functionality.

mlir/include/mlir/Dialect/QC/IR/QCOps.td
```diff
@@ -1007,8 +1014,10 @@ def InvOp : QCOp<"inv",
         size_t getNumQubits() { return getBodyUnitary().getNumQubits(); }
         size_t getNumTargets() { return getNumQubits(); }
         static size_t getNumControls() { return 0; }
+        static ValueRange getControls() { return {}; }
         Value getQubit(size_t i) { return getBodyUnitary().getQubit(i); }
         Value getTarget(size_t i) { return getQubit(i); }
+        ValueRange getTargets() { return getBodyUnitary().getTargets(); }
```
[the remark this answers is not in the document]
burgholzer: This is a real problem. Thanks for catching this.
I am unsure about the solution here. For a canonicalized IR, this is a non-issue, CtrlOp and InvOp will always be canonicalized so that InvOp is nested inside of CtrlOp; never the other way around.
However, in general, the problem is that I have not yet found a way to return a ValueRange from a qc::CtrlOp that contains all the qubits. `ctrlOp.getControls()` returns a range of controls and a fictive `ctrlOp.getBodyUnitary().getQubits` would return the range of (target) qubits of the wrapped operation. However, I have not found a way to actually expose such a `getQubits` method, which combines the two ranges. Materializing a temporary SmallVector does not work as it would be destroyed when going out of scope. LLVM's `concat_range` does not work as it does not return a ValueRange.
@coderabbitai do you see any kind of solution here? Or is the solution to change the semantics of `getNumTargets` and `getTagret` for `InvOp` to actually only consider the targets of the underlying op and equally pipe-through the controls instead of erroring out?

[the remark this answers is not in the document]
burgholzer: Implemented in https://github.com/munich-quantum-toolkit/core/pull/1569/commits/ac68ee47d29cdae3f6f6d760beed394bae83b75e
My only concern is that now the semantics between the QC InvOp and the QCO InvOp are no longer strictly speaking identical. On the QCO side, where all qubits (no matter weither they are internally used as control or target) are passed as operands to the InvOp due to linear typing.
That means that in QCO there is no notion of a "control" for an InvOp that would just pipe through to the inside operation.
I am fine with this personally as it does not make a difference in the canonicalized situation and fixes the overall dilemma. However, I wanted to mention it here.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -848,25 +876,14 @@ struct ConvertQCBarrierOp final : StatefulOpConversionPattern<qc::BarrierOp> {
...
-    // Get QCO qubits from state map
-    auto qcQubits = op.getQubits();
-    SmallVector<Value> qcoQubits;
-    qcoQubits.reserve(qcQubits.size());
-    for (const auto& qcQubit : qcQubits) {
-      assert(qubitMap.contains(qcQubit) && "QC qubit not found");
-      qcoQubits.push_back(qubitMap[qcQubit]);
-    }
+    auto* operation = op.getOperation();
+    const auto qcQubits = llvm::to_vector(op.getQubits());
```
burgholzer, line 880:
```suggestion
    auto qcQubits = op.getQubits();
```
@denialhaag I forgot this here, but do not want to trigger another CI run in this PR to not hold everything up. This saves materializing one vector.
Would you mind just including this in your rebase of #1580?
You can simply resolve this comment and merge the PR here once you read this 😌

denialhaag: Sure, I'll do that!

PR #1574, opened 2026-03-20
✨Arena-Based A* Search For Mapping Pass

MatthiasReumann:
## Description

This pull request updates the A* search implementation of the mapping pass in two ways:
1. Use an Arena (of Nodes). Instead of storing the sequence of SWAPs for each node, a node now stores a pointer of its parent and reconstructs the sequence whenever a goal node is found. This reduces the memory footprint of a node significantly - although now all nodes need to be stored and kept in memory. 
2. If a layout is visited multiple times (via different sequences of SWAPs) child-nodes are only created whenever the previous depth is higher than the new one (and thus we've found a shorter sequence of SWAPs leading to the same layout). 

## Checklist

### DIFF
--- a/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
@@ -598,7 +598,12 @@ private:
     llvm::SpecificBumpPtrAllocator<Node> arena;
     std::priority_queue<Node*, std::vector<Node*>, Node::ComparePointer>
         frontier;
-    frontier.emplace(std::construct_at(arena.Allocate(), layout));
+
+    Node* root = std::construct_at(arena.Allocate(), layout);
+    if (root->isGoal(layers.front(), arch)) {
+      return SmallVector<IndexGate>{};
+    }
+    frontier.emplace(root);
 
     DenseMap<Layout, std::size_t, LayoutInfo> bestDepth;
     DenseSet<IndexGate> expansionSet;
@@ -630,10 +635,10 @@ private:
 
       if (curr->isGoal(layers.front(), arch)) {
         SmallVector<IndexGate> seq(curr->depth);
-        std::size_t i = seq.size() - 1;
+        std::size_t j = seq.size() - 1;
         for (Node* n = curr; n->parent != nullptr; n = n->parent) {
-          seq[i] = n->swap;
-          --i;
+          seq[j] = n->swap;
+          --j;
         }
         return seq;
       }


mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -15,6 +15,7 @@
 #include "mlir/Dialect/QCO/Transforms/Passes.h"
 #include "mlir/Dialect/QCO/Utils/WireIterator.h"
 
+#include <llvm/ADT/DenseMapInfo.h>
```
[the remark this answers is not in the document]
MatthiasReumann: My beloved bunny, first of all check the pipeline - it compiles! Secondly, the `SmallVector.h` has a specialization for `DenseMapInfo<SmallVector<T,N>>`. Could you please re-verify this statement?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -534,6 +560,7 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
 
     MinQueue frontier{};
     frontier.emplace(root);
+    DenseSet<Layout, LayoutInfo> discovered{root.layout};
```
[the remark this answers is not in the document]
MatthiasReumann: Can you provide me an example where, given a program-to-hardware mapping, a sequence of fewer SWAPs leads to the same program-to-hardware mapping as one with more SWAPs?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -575,43 +594,63 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+      // If the currently visited node is a goal node, reconstruct the sequence
+      // of SWAPs from this node to the root.
 
-      expansionSet.clear();
-      if (!curr.sequence.empty()) {
-        expansionSet.insert(curr.sequence.back());
+      if (curr->isGoal(layers.front(), arch)) {
+        SmallVector<IndexGate> seq;
+        for (Node* n = curr; n->parent != nullptr; n = n->parent) {
+          seq.push_back(n->swap);
```
```suggestion
          seq.emplace_back(n->swap);
```
🤷🏼

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -575,43 +594,63 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
-      // Given a layout, create child-nodes for each possible SWAP between
-      // two neighbouring hardware qubits.
+      // If the currently visited node is a goal node, reconstruct the sequence
+      // of SWAPs from this node to the root.
 
-      expansionSet.clear();
-      if (!curr.sequence.empty()) {
-        expansionSet.insert(curr.sequence.back());
+      if (curr->isGoal(layers.front(), arch)) {
+        SmallVector<IndexGate> seq;
```
burgholzer: Can probably be pre-allocated in size, right? since the depth is known beforehand.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -575,43 +594,63 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+        for (Node* n = curr; n->parent != nullptr; n = n->parent) {
+          seq.push_back(n->swap);
+        }
+        return to_vector(reverse(seq));
```
burgholzer: Instead of constructing and reversing, which materializes another vector, I am wondering whether it would be better to simply pre-allocate a vector and fill it in the right places from the beginning.

burgholzer (changes requested):
Just some true nitpicks on some of the C++ details.
Otherwise, this looks great already!

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Sure can! Good idea.

burgholzer (approved):
LGTM! Let's get this in now.
However, I also believe there must be some way to avoid these dominance issues without the "Keule" that is topological sorting.

PR #1581, opened 2026-03-24
🐛 Fix SSA Dominance Issues for Mapping Pass for Circuits With Barriers

MatthiasReumann:
## Description

This pull request fixes an issue which occurs when running the mapping pass on circuits using barriers. 

## Checklist

- [X] I have added appropriate tests that cover the new/changed functionality.
- [X] I have updated the documentation to reflect these changes.
- [X] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [X] I have added migration instructions to the upgrade guide (if needed).
- [X] The changes follow the project's style guidelines and introduce no new warnings.
- [X] The changes are fully tested and pass the CI checks.

### DIFF
--- a/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
@@ -350,8 +350,8 @@ private:
     /**
      * @brief Decrement ref count of op and potentially release its iterators.
      */
-    std::optional<SmallVector<WireIterator*, 0>> visit(Operation* op,
-                                                       WireIterator* it) {
+    std::optional<ArrayRef<WireIterator*>> visit(Operation* op,
+                                                 WireIterator* it) {
       assert(refCount.contains(op) && "expected sync map to contain op");
 
       // Add iterator for later release.
@@ -375,7 +375,7 @@ private:
 
   private:
     /// @brief Maps operations to to-be-released iterators.
-    DenseMap<Operation*, SmallVector<WireIterator*, 0>> onHold;
+    DenseMap<Operation*, SmallVector<WireIterator*, 2>> onHold;
     /// @brief Maps operations to ref counts.
     DenseMap<Operation*, std::size_t> refCount;
   };


mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -321,6 +321,65 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+     * @brief Clear the contents of the map.
+     */
+    void clear() {
+      onHold.clear();
+      refCount.clear();
+    }
+
+  private:
+    /// @brief Maps operations to to-be-released iterators.
+    DenseMap<Operation*, SmallVector<WireIterator*, 0>> onHold;
```
```suggestion
    DenseMap<Operation*, SmallVector<WireIterator*, 1>> onHold;
```
Wouldn't 0 imply that every entry would be guaranteed to trigger a reallocation?

MatthiasReumann: `SmallVector<Type, 0>` is the llvm equivalent of `std::vector<Type>`. See [here](https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h) for more information: 

> SmallVector has grown a few other minor advantages over std::vector, causing SmallVector<Type, 0> to be preferred over std::vector<Type>.

MatthiasReumann: Maybe it does make sense to set an inline-capacity here. In most cases we shouldn't probably just guesstimate this number but actually figure it out explicitly via a series of benchmarks.

MatthiasReumann: Mhm. Now that I think about it. In the mapping pass this most likely is almost always 2. The only outlier would be the barrier ops.

MatthiasReumann: Updated the inline-capacity to 2.

burgholzer: Sounds reasonable 👍🏻

PR #1605, opened 2026-04-02
:sparkles: Add Hadamard Lifting

### LINKED ISSUE #1386: ✨ MLIR - Add Hadamard lifting to support measurement lifting
### Problem Statement

In PR [1108](https://github.com/munich-quantum-toolkit/core/pull/1108), `lift-measurements` has been introduced. It lifts measurements before certain gates to save on quantum resources.
At the moment, the `lift-measurements` stops at Hadamard gates.

### Proposed Solution

I'd like to propose `hadamard-lifting`: Pulling Hadamard gates in front of certain gates using commutation rules (e.g. H * X = Z * H) in order to support `lift-measurements` and improve `lift-measurements`' impact on the circuit.

Conveniently, the gates that can be commuted with Hadamard (Pauli gates, CNOT) are also the ones that a measurement can be lifted over.

LiRem101:
## Description

This PR introduces the optimization pass Hadamard lifting.

The optimization pass lifts Hadamard gates in front of Pauli gates in order to improve the results bu measurement-lifting. The implemented liftings are:
`z` followed by `h` to `h` followed by `x`.
`x` followed by `h` to `h` followed by `z`.
`y` followed by `h` to `h` followed by `y`.

The transformation are also applied on controlled gates, provided both participating gates have the same target qubit and the same control qubits.
An exception is the `z` gate: The transformation exchanges the target and control qubits of a controlled z gate, if it is followed by a controlled Hadamard gate it can be swapped with afterwards.

Additionally, a transformation routine is implemented which swaps the target and control qubits of a CNOT operation (by adding four Hadamard gates before/after the target/control qubits). This is applied if a Hadamard gate is applied to the CNOT target, and afterwards the qubit is measured. The whole transformation routine is:
image

After the transformation routine, measurement-lifting can be used to lift the measurement in front of the control and to remove a costly two-qubit gate.

The paper in which the routine is published is currently under review.

Microsoft Copilot has been occasionally used to draft first versions of code, but was not integrated into an IDE.

Fixes #1386

## Checklist


burgholzer:
Just an fyi: #1580 is in 🚀 
This probably requires some moderate adjustments here. Should be pretty straight forward though.

burgholzer:
@LiRem101 before I review the code here, please try to work through the coderabbit comments.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,541 @@
...
+      rewriter.replaceOpWithNewOp<ZOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
+    } else if (gateName == "z") {
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<XOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
+    } else {
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<YOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
```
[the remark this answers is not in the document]
LiRem101: Fixed with [6af2a96](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/6af2a96553f904c6282c6526cdb7948930d333c2)

mlir/unittests/Dialect/QCO/Transforms/Optimization/test_qco_hadamard_lifting.cpp
```diff
@@ -0,0 +1,504 @@
...
+  qRef[1] = referenceBuilder.h(qRef[1]);
+  qRef[1] = referenceBuilder.x(qRef[1]);
+  qRef[2] = referenceBuilder.h(qRef[2]);
+  qRef[2] = referenceBuilder.y(qRef[2]);
+  reference = referenceBuilder.finalize();
+
+  ASSERT_TRUE(runHadamardLiftingPass(module.get()).succeeded());
+  PassManager pm(reference->getContext());
+  pm.addPass(createHadamardLifting());
+  EXPECT_TRUE(pm.run(*reference).succeeded());
```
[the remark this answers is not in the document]
LiRem101: Fixed with [a15d2c1](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/a15d2c135f1feefa32e2f6e978f70d6d7b753801)

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,541 @@
...
+ * @param range2 The second range.
+ */
+static bool containRangesOfSameElements(const std::vector<Value>& range1,
+                                        const std::vector<Value>& range2) {
+  bool result = true;
+  result &= range1.size() == range2.size();
+  for (auto element : range1) {
+    result &= std::ranges::find(range2, element) != range2.end();
+  }
+  return result;
```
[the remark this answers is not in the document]
LiRem101: Fixed with [233da3f](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/233da3f4c0277ac2b919ad9984492493f11ef391)

mlir/unittests/Dialect/QCO/Transforms/Optimization/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+          MLIRQCOProgramBuilder
+          MLIRQCOTransforms
+          MLIRSupportMQT
+          MLIRQTensorDialect
+          MLIRQCOUtils
+          MLIRParser
+          MLIRIR
+          MLIRPass
+          MLIRSupport
+          LLVMSupport)
```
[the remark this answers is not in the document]
LiRem101: Fixed with [26ca2ed](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/26ca2edf938d41603460be2a9f01dccf334e51a4)

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,57 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+    the following transformation is applied to circuits pattern, where a Hadamard gate follows a the target gate of a
+    CNOT, which is followed by a measurement:
+                              ┌───┐┌───┐┌───┐
+        ──■─────────────      ┤ H ├┤ X ├┤ H ├──
+        ┌─┴─┐┌───┐┌──────┐ => │───│└─┬─┘│───┘──┐
+        ┤ X ├┤ H ├┤ Meas │    ┤ H ├──■──┤ Meas │
+        └───┘└───┘└──────┘    └───┘     └──────┘
+
+     Afterwards, the measurement lifting routine could transform the CNOT into a classically controlled Pauli X.
+
```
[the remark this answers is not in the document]
LiRem101: Fixed with [cc97cfd](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/cc97cfd4e10a8b3b73436e2350138a4942511200)

mlir/unittests/Dialect/QCO/Transforms/Optimization/test_qco_hadamard_lifting.cpp
```diff
@@ -0,0 +1,533 @@
...
+  reference = referenceBuilder.finalize();
+
+  ASSERT_TRUE(runHadamardLiftingPass(module.get()).succeeded());
+  PassManager pm(reference->getContext());
+  pm.addPass(createCanonicalizerPass());
+  EXPECT_TRUE(pm.run(*reference).succeeded());
+
+  EXPECT_TRUE(
+      areModulesEquivalentWithPermutations(module.get(), reference.get()));
+}
```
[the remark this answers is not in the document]
LiRem101: Fixed with [cf245b1](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/cf245b1bc59480f60b4c8f0e6df85c11130d6968)

mlir/unittests/Dialect/QCO/Transforms/Optimization/test_qco_hadamard_lifting.cpp
```diff
@@ -0,0 +1,516 @@
...
+#include <mlir/IR/BuiltinOps.h>
+#include <mlir/IR/DialectRegistry.h>
+#include <mlir/IR/OwningOpRef.h>
+#include <mlir/IR/Value.h>
+#include <mlir/Pass/PassManager.h>
+#include <mlir/Support/LogicalResult.h>
+#include <mlir/Transforms/Passes.h>
+
+#include <cassert>
+#include <cmath>
```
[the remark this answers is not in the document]
LiRem101: Fixed with [57159ae](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/57159aecc9a92121aec268743e79dcb5147c0f85)

mlir/unittests/Dialect/QCO/Transforms/Optimization/test_qco_hadamard_lifting.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+  EXPECT_TRUE(
+      areModulesEquivalentWithPermutations(module.get(), reference.get()));
+}
+
+/**
+ * @brief Test: Checks that Hadamard gates are not lifted if they are controlled
+ * and the Pauli gate is a Pauli Y gate.
+ */
+TEST_F(QCOHadamardLiftingTest, doNotliftHadamardOverPauliYGateIfControlled) {
```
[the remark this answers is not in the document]
LiRem101: Fixed with [57159ae](https://github.com/munich-quantum-toolkit/core/pull/1605/commits/57159aecc9a92121aec268743e79dcb5147c0f85)

LiRem101:
@burgholzer I worked through all of coderabbit's comments :)

mlir/include/mlir/Compiler/CompilerPipeline.h
```diff
@@ -40,6 +40,9 @@ struct QuantumCompilerConfig {
 
   /// Print IR after each stage
   bool printIRAfterAllStages = false;
+
+  /// Enable Hadamard lifting
+  bool hadamardLifting = false;
```
burgholzer:
```suggestion
  bool enableHadamardLifting = false;
```
better fits the rest of the config options.

mlir/lib/Compiler/CompilerPipeline.cpp
```diff
@@ -136,9 +137,12 @@ QuantumCompilerPipeline::runPipeline(ModuleOp module,
...
   }
   // Stage 5: Optimization passes
-  // TODO: Add optimization passes
-  if (failed(
-          runStage([&](PassManager& pm) { populateQCOCleanupPipeline(pm); }))) {
+  if (failed(runStage([&](PassManager& pm) {
+        if (config_.hadamardLifting) {
+          pm.addPass(qco::createHadamardLifting());
+        }
+        populateQCOCleanupPipeline(pm);
```
burgholzer:
```suggestion
```
This is no longer needed as part of this stage as it is run in the follow-up stage of the compiler pipeline. This was just a placeholder.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
 }
 
+//===----------------------------------------------------------------------===//
+// Optimization Passes
+//===----------------------------------------------------------------------===//
+
+def HadamardLifting : Pass<"hadamard-lifting", "mlir::ModuleOp"> {
+  let dependentDialects = ["mlir::qco::QCODialect"];
+  let summary = "This pass attempts to move Hadamard gates as far away from "
+                "measurements as possible by flipping them "
```
burgholzer:
```suggestion
                "measurements as possible by commuting them "
```

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+                "with Pauli gates. Additionally, it can change target and "
+                "control qubits from Pauli-Z gates to make "
+                "Hadamard lifting applicable. It also lifts Hadamard gates "
+                "over CNOT gates if that moves a measurement "
+                "directly after a control.";
```
burgholzer: This summary could be slightly improved to make the purpose a little clearer; why someone would be using this pass.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+  let description = [{
+    This pass lifts Hadamard gates away from the measurements in order to apply measurement lifting more effectively. It
```
burgholzer: At this point, it is unclear to the reader what "measurement lifting" is supposed to be. This should be rephrased so that the description is mostly self contained.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+         ┌───┐ ┌───┐      ┌───┐ ┌───┐
+        ─┤ X ├─┤ H ├─ => ─┤ H ├─┤ Z ├─
+         └───┘ └───┘      └───┘ └───┘
+         ┌───┐ ┌───┐      ┌───┐ ┌───┐
+        ─┤ Z ├─┤ H ├─ => ─┤ H ├─┤ X ├─
+         └───┘ └───┘      └───┘ └───┘
+         ┌───┐ ┌───┐      ┌───┐ ┌───┐
+        ─┤ Y ├─┤ H ├─ => ─┤ H ├─┤ Y ├─
+         └───┘ └───┘      └───┘ └───┘
+    Hadamard lifting is only applied to lift Hadamard gates further in front in the circuit, not the other way around.
```
burgholzer: I am tripping over the wording here. "lifting in front" does not really sound like proper English. It is rather "lift up", in general.
Generally, I am under the impression that "Hadamard commutation" would be a more fitting description overall. Would you agree?
If so, I'd propose a bit of a more thorough rewrite of the wording throughout this PR.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+
+    If the Hadamard and Pauli gates are controlled, they are only lifted if both gates are controlled by exactly the
+    same qubits and the Pauli gate is not a Pauli Y gate. In that case, the swap of the Pauli Y gate and the Hadamard
+    gate introduces a phase (YH = -HY) which leads to a non-global phase for controlled gates.
```
burgholzer: Our optimizations should generally all be global-phase preserving, if possible. This widens the applicability of the passes.
As such, the Y-H commutation rule shall also introduce a global phase gate with the appropriate phase.

(I am fairly sure it is not worth to explicitly handle the CY-CH case as this will rarely pop up in practice)

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+
+    In the case of the controlled Pauli Z gate, controls and targets are interchangeable. Therefore, transformation
+    routines as follows are possible and applied:
+        ┌───┐
+        ┤ Z ├──■──    ──■────■──    ──■────■──
+        └─┬─┘┌─┴─┐ => ┌─┴─┐┌─┴─┐ => ┌─┴─┐┌─┴─┐
+        ──■──┤ H ├    ┤ Z ├┤ H ├    ┤ H ├┤ X ├
+             └───┘    └───┘└───┘    └───┘└───┘
+    The commutation of controlled gates is only done if there are no gates applied between the two controlled gates to
+    any of the qubits used on the controlled gates.
```
burgholzer: This seems stricter than it technically needs to be. as long as the gate in-between both only acts on the control qubit(s) of both considered operations, the same rules still apply; since controls commute.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -72,4 +72,64 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+        └───┘    └───┘     └───┘
+    the following transformation is applied to circuits where a Hadamard gate follows a target gate of a controlled
+    Pauli X gate, which is followed by a measurement:
+                              ┌───┐┌───┐┌───┐
+        ──■─────────────      ┤ H ├┤ X ├┤ H ├──
+        ┌─┴─┐┌───┐┌──────┐ => │───│└─┬─┘│───┘──┐
+        ┤ X ├┤ H ├┤ Meas │    ┤ H ├──■──┤ Meas │
+        └───┘└───┘└──────┘    └───┘     └──────┘
+
+     Afterward the measurement lifting routine could transform the CNOT into a classically controlled Pauli X.
```
burgholzer: Again, measurement lifting has not been explained and the reader lacks context here.

burgholzer: Typically people write `Pauli-X`, `Pauli-Y`, etc. instead of `Pauli X`. Might be worth adjusting everywhere.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+                                        const std::vector<Value>& range2) {
+  return range1.size() == range2.size() &&
+         std::ranges::is_permutation(range1, range2);
+}
+
+namespace {
+
+/**
+ * @brief This pattern changes the target of a controlled Pauli Z gate if a
+ * controlled hadamard gate is it successor.
```
burgholzer:
```suggestion
 * controlled Hadamard gate is it successor.
```
Please also check elsewhere

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ * If all out qubits of Pauli Z are equal to all in qubits of Hadamard, we can
+ * commute the gates and change Pauli Z to X. This is only possible if Hadamard
+ * and Pauli act on the same qubit as target. If the target of the Pauli gate is
+ * a ctrl at the hadamard and vice versa, we can change the target of Pauli Z to
+ * the Hadamard's. This is done in this pattern.
```
burgholzer: Just devil's advocate, but are there any practical examples of this pattern actually occurring out in the wild?
I think I personally have never really seen a controlled-H being used in a textbook algorithm or VQA; let alone one followed by a CNOT on the same qubits.
That might just be my limited overview of the field though.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+   * @return True if target qubit of gate1 is ctrl in gate2 and vice versa.
+   * False otherwise.
+   */
+  static bool areTargetsControlsAtTheOtherGates(CtrlOp gate1, CtrlOp gate2) {
+    const Value targetQubitGate2 = gate2.getInputTarget(0);
+    const Value targetQubitGate1 = gate1.getOutputTarget(0);
+    const auto inCtrlGate2 = gate2.getControlsIn();
+    const auto outCtrlGate1 = gate1.getControlsOut();
+    std::vector inCtrlGate2Vec(inCtrlGate2.begin(), inCtrlGate2.end());
+    std::vector outCtrlGate1Vec(outCtrlGate1.begin(), outCtrlGate1.end());
```
burgholzer: This is a general comment for the entire PR: please avoid using `std::vector` and prefer LLVM's own optimized data types for situations like these (here, `llvm::SmallVector`)

In a similar vein: try to avoid materializing temporary vectors when this is not absolutely necessary. This is what ranges are for and LLVM has great tooling support for handling the ranges being returned from the respective methods.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+namespace {
+
+/**
+ * @brief This pattern changes the target of a controlled Pauli Z gate if a
+ * controlled hadamard gate is it successor.
+ * If all out qubits of Pauli Z are equal to all in qubits of Hadamard, we can
+ * commute the gates and change Pauli Z to X. This is only possible if Hadamard
+ * and Pauli act on the same qubit as target. If the target of the Pauli gate is
+ * a ctrl at the hadamard and vice versa, we can change the target of Pauli Z to
```
burgholzer:
```suggestion
 * a control at the hadamard and vice versa, we can change the target of Pauli Z to
```
No obvious reason to not spell things out. Please also check elsewhere.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+    rewriter.replaceUsesWithIf(
+        qubit1, temporary,
+        [&](const OpOperand& operand) { return operand.getOwner() == gate; });
+    rewriter.replaceUsesWithIf(qubit2, qubit1, [&](const OpOperand& operand) {
+      return operand.getOwner() == gate;
+    });
+    rewriter.replaceUsesWithIf(
+        temporary, qubit2,
+        [&](const OpOperand& operand) { return operand.getOwner() == gate; });
+  }
```
burgholzer: This feels wrong as it creates a temporary operation and then calls replacement three times. At that point, it is way easier to simply replace the operation with a newly created one in one go.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ * @brief This method checks if two ranges contain of exactly the same
+ * elements.
+ *
+ * This method checks if two ranges contain of exactly the same elements.
+ *
+ * @param range1 The first range.
+ * @param range2 The second range.
+ */
+static bool containRangesOfSameElements(const std::vector<Value>& range1,
+                                        const std::vector<Value>& range2) {
```
burgholzer: These are not ranges; these are vectors.
And from what I can tell, you'd rather want to use a `Set` datatype here to avoid the complexity of the `is_permutation` call.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+   * controlled hadamard gate is it successor.
+   *
+   * @param op The operation to match (only Pauli gates trigger the rewrite)
+   * @param rewriter Pattern rewriter for applying transformations
+   * @return success() if circuit was changed, failure() otherwise
+   */
+  LogicalResult matchAndRewrite(CtrlOp op,
+                                PatternRewriter& rewriter) const override {
+    // op needs to be a Pauli Z gate and controlled
+    std::string opName = op.getBodyUnitary()->getName().stripDialect().str();
```
burgholzer: There's a dedicated method in the UnitaryOpInterface for getting the name of an operation.
Please take that into account everywhere.
I suppose the cleaner way overall is to use `llvm::isa` on the `op.getOperation` and compare to `ZOp` instead of relying on string matching.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+    if (op.getNumTargets() != 1 || opName != "z") {
```
burgholzer:
```suggestion
    if (opName != "z") {
```
Redundant condition.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+      return failure();
+    }
+
+    // op needs to be in front of a controlled hadamard gate
+    const auto& users = op->getUsers();
+    if (users.empty()) {
+      return failure();
+    }
+    auto* const user = *users.begin();
```
burgholzer:
```suggestion
    auto* const user = *op->getUsers().begin();
```
This can never happen due to linear typing.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+    if (user->getName().stripDialect().str() != "ctrl") {
+      return failure();
+    }
+    auto hadamardGate = llvm::dyn_cast<CtrlOp>(user);
```
burgholzer:
```suggestion
    auto hadamardGate = llvm::cast<CtrlOp>(user);
```
If you already know that it is a CtrlOp, then there is no need for the `dyn_` cast.
But, based on the suggestion above, you'd probably want to replace the name check with a 
```
auto hadamardGate = llvm::dyn_cast<CtrlOp>(user);
if (!hadamardGate) {
  return failure();
}
```
anyway.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+    if (hadamardGate.getNumTargets() != 1 ||
+        hadamardGate.getBodyUnitary()->getName().stripDialect().str() != "h") {
```
burgholzer: Redundant check. It is sufficient to type check for HOp.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+      return failure();
+    }
+
+    const std::vector<Value> outputsOp(op.getOutputQubits().begin(),
+                                       op.getOutputQubits().end());
+    const std::vector inputsH(hadamardGate.getInputQubits().begin(),
+                              hadamardGate.getInputQubits().end());
+    if (!containRangesOfSameElements(outputsOp, inputsH)) {
+      return failure();
+    }
```
burgholzer: This is a perfect example where materializing vectors (and even a SmallVector) is not at all necessary.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+        op.getInputForOutput(targetInputQubitHadamard);
+
+    exchangeTwoQubitsAtGate(op, originalInputTargetQubitZ, newTargetInputQubitZ,
+                            rewriter);
+
+    const Value newTargetInputQubitH =
+        op.getOutputForInput(newTargetInputQubitZ);
+
+    exchangeTwoQubitsAtGate(hadamardGate, targetInputQubitHadamard,
+                            newTargetInputQubitH, rewriter);
```
burgholzer: I basically already annotated this at the exchange function; but this feels like a very hacky and generally inefficient way of performing the exchange.
This should essentially boil down to two `rewriter.replaceOp...` calls

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ * @brief This pattern is responsible for lifting Hadamard gates above Pauli
+ * gates.
+ *
+ * This pattern swaps a Pauli gate with a Hadamard gate. This is done using the
+ * commutation rules of Pauli and Hadamard gates, which are:
+ * - X - H - = - H - Z -
+ * - Y - H - = - H - Y -
+ * - Z - H - = - H - X -
+ * This is applied to uncontrolled gates and controlled ones, if the controls
+ * are applied to the same qubits for both gates and the Pauli gates are X or Y.
```
burgholzer:
```suggestion
 * are applied to the same qubits for both gates and the Pauli gates are X or Z.
```

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ * In case of Pauli Y, the routine is only applied to uncontrolled gates, as
+ * HY = -YH, which leads to a relative phase if the Hadamard and Pauli gate are
+ * controlled.
```
burgholzer: I would be very tempted to simplify this to just consider uncontrolled gates.
As pointed out elsewhere, I am not aware of (m)any algorithms that show such a pattern. Happy to be mistaken though.
It would definitely simplify the implementation though.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ */
+struct LiftHadamardsAbovePauliGatesPattern final
+    : OpInterfaceRewritePattern<UnitaryOpInterface> {
+  explicit LiftHadamardsAbovePauliGatesPattern(MLIRContext* context)
+      : OpInterfaceRewritePattern(context) {}
+
+  /**
+   * This method checks whether the first and second controls are controlled by
```
burgholzer:
```suggestion
   * This method checks whether the first and second operations are controlled by
```

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+   */
+  static bool areControlsControlledBySameQubits(CtrlOp firstCtrl,
+                                                CtrlOp secondCtrl) {
+    const std::vector<Value> controlOutputsFirstGate(
+        firstCtrl.getControlsOut().begin(), firstCtrl.getControlsOut().end());
+    const std::vector controlInputsSecondGate(
+        secondCtrl.getControlsIn().begin(), secondCtrl.getControlsIn().end());
+    return containRangesOfSameElements(controlOutputsFirstGate,
+                                       controlInputsSecondGate);
+  }
```
burgholzer: Since this is one of the first larger reviews of your code, I'm rather annotating a little bit too much and too often. This is another example where materializing is not necessary.
This check essentially just boils down to asking each incoming control qubit of the second operation whether the defining operation of the qubit value is the first operation. So this is a simple `all_of` call on the `secondCtrol.getControlsIn()` range.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+   * uncontrolled.
+   *
+   * @param gate The Pauli gate.
+   * @param hadamardGate The Hadamard gate.
+   * @param controlled Whether the gates are controlled or not.
+   * @param rewriter The used rewriter.
+   * @return success() if circuit was changed, failure() otherwise
+   */
+  static LogicalResult swapPauliWithHadamard(UnitaryOpInterface gate,
+                                             UnitaryOpInterface hadamardGate,
```
burgholzer:
```suggestion
  static LogicalResult swapPauliWithHadamard(UnitaryOpInterface gate,
                                             HOp hadamardGate,
```
If you already know that this is supposed to be a Hadamard, than the function signature may as well expect it.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+      rewriter.replaceOpWithNewOp<ZOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
+    } else if (gateName == "z") {
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<XOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
+    } else {
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<YOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
```
burgholzer: I believe there is no need to set the insertion point here as you are directly replacing a particular operation. This already defines the place of insertion.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<ZOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
+    } else if (gateName == "z") {
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<XOp>(hadamardGate,
+                                       hadamardGate.getInputQubit(0));
+    } else {
+      rewriter.setInsertionPoint(hadamardGate);
+      rewriter.replaceOpWithNewOp<YOp>(hadamardGate,
```
burgholzer: Based on other recommendations of not doing string matching, this should probably be a LLVM `TypeSwitch` on `gate.getOperation`

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+        op.getOutputTarget(0) != hadamardGate.getInputTarget(0)) {
+      return failure();
+    }
+
+    return swapPauliWithHadamard(op, hadamardGate, false, rewriter);
+  }
+};
+
+/**
+ * @brief This pattern remove an H gate between a CNOT and a measurement.
```
burgholzer:
```suggestion
 * @brief This pattern removes an H gate between a CNOT and a measurement.
```

burgholzer: Technically, this brief description is not accurate, because it omits that it adds other Hadamards and flips the CNOT direction.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ *
+ * If there is a Hadamard gate between the target qubit of a CNOT and a
+ * measurement, we flip the CNOT and apply a hadamard gate to the incoming and
+ * outcoming qubits. As H * H = id, the measurement is then the direct successor
+ * of a CNOT ctrl, which is beneficial for the qubit reuse routine.
```
burgholzer: > beneficial for the qubit reuse routine.

Lacks context

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ * The procedure also works if there are additional ctrls. Only the target
+ * and ctrl involved in the transformation get hadamard gates assigned.
+ * For now, the involved ctrl to be flipped with the target is chosen randomly.
```
burgholzer:
```suggestion
 * The involved ctrl to be flipped with the target is chosen randomly.
```
"For now" is a stale phrase. It is highly likely to remain in the code for a pretty long time.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+    rewriter.replaceUsesWithIf(
+        inputQubit1, temporary,
+        [&](const OpOperand& operand) { return operand.getOwner() == gate; });
+    rewriter.replaceUsesWithIf(
+        inputQubit2, inputQubit1,
+        [&](const OpOperand& operand) { return operand.getOwner() == gate; });
+    rewriter.replaceUsesWithIf(
+        temporary, inputQubit2,
+        [&](const OpOperand& operand) { return operand.getOwner() == gate; });
+  }
```
burgholzer: There should be absolutely no need to create a temporary operation here and also no need to perform so many (conditional) replacements.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+    const HOp newHOPAfterCtrl = addHadamardGatesAfterGate(
+        cnotGate, relevantOutputQubitsForHadamard, rewriter);
+
+    // Flip CNOT targets and ctrl
+    swapQubits(cnotGate, cnotGate.getInputControl(0),
+               cnotGate.getInputTarget(0), op, newHOPAfterCtrl, rewriter);
+
+    return success();
+  }
+};
```
burgholzer, line 322: A lot of the same comments as above apply here. I stopped reviewing this in detail and will do so after an initial round of improvements.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
```diff
@@ -0,0 +1,545 @@
...
+ * @brief This pattern remove an H gate between a CNOT and a measurement.
+ *
+ * If there is a Hadamard gate between the target qubit of a CNOT and a
+ * measurement, we flip the CNOT and apply a hadamard gate to the incoming and
+ * outcoming qubits. As H * H = id, the measurement is then the direct successor
+ * of a CNOT ctrl, which is beneficial for the qubit reuse routine.
+ * The procedure also works if there are additional ctrls. Only the target
+ * and ctrl involved in the transformation get hadamard gates assigned.
+ * For now, the involved ctrl to be flipped with the target is chosen randomly.
+ */
```
burgholzer, line 137: Let me ask the devil's advocate question here as well: How many practical algorithms does this pattern apply to?

burgholzer (changes requested):
Thanks @LiRem101 for putting this PR together!
Since this is one of the first bigger contributions from you, I reviewed this in greater detail and rigor to provide you with ample feedback. Please do not get discouraged by the number of comments. It's all part of the process.

As you will probably see from the comments, the current code will require changes before it can be considered mergeable. A lot of it simply does not yet feel like idiomatic MLIR.

Besides some comments on the documentation itself, I also asked a few devil's advocate questions. The main purpose there is to find out if we really want to keep all of the functionality currently being proposed in this PR or if we'd rather focus on a selected subset that has true practical relevance. These questions are purposefully phrased in a way that you can challenge them.

Just as a recommendation, it is probably better if you switch the PR itself back to draft while you are working on it, to avoid CodeRabbit from spamming the PR with comments while you are still iterating.

Also, if you are unsure about something, make sure to check out the contributing guidelines, particularly https://mqt.readthedocs.io/projects/core/en/latest/contributing.html#pull-request-workflow, or simply ask 😌

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: I will add a brief explanation that this is done to improve measurement lifting (which I explain further in the description).

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: I added an explanation that it is a subroutine of Damian's qubit-reuse routine. I am just unsure how much I am supposed to explain the routine? As far as I got it, it's planned to implement the qubit-reuse routine for QCO (it exists for MqtOpt), so this would be clear as soon as the routine is implemented?

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: I will adapt the wording. The naming I'd like to discuss again after I worked through the review, as we discussed.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: I will add a gphase for the Pauli-Y case.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: Oh, you are right. Though I will remove the controlled cases anyway, so that is not too relevant anymore.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: I'll adapt that

LiRem101: I will add an explanation (see my comment on line R89).

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: No, you are right. I will remove the parts of the code targeting controlled gates.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: Thanks for the info, I'll change that

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: The function is removed alongside the routine for controlled Pauli-Z gates, but thanks for the info.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: Thanls for the info. While coding I wondered if there is a way to compare to `ZOp` directly, but I could not figure it out :D

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: Removed due to removed AdaptCtrldPauliZToLiftingPattern routine.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: Removed due to removed AdaptCtrldPauliZToLiftingPattern routine, but thanks for the example

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: Removed due to removed AdaptCtrldPauliZToLiftingPattern routine.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: I will remove the handling of controlled gates in the routine.

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: Method removed due to not handling controlled gates anymore. Nevertheless, thanks for the info

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: You are right, I'll remove that

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: I will adapat that

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: I wasn't sure how "brief" the brief description should be ^^'. I'll add more info

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: I'll add more context

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: I will change that to modifyOpInPlace calls

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: The method is removed, but thanks for the info!

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: I will try to apply the comments on the other parts of the code as good as possible

mlir/lib/Dialect/QCO/Transforms/Optimization/HadamardLifting.cpp
LiRem101: It is not very common, but there have been some occasions in the mqt.bench set if one casts the circuits to ones that only contain `["h", "x", "y", "z", "s", "sdg", "t", "tdg", "p", "sx", "rx", "ry", "rz", "swap", "cx", "cy", "cz", "cp", "crx", "cry", "crz", "ch", "cu"]` gates. The circuits containing this pattern are Cardinality Circuit, Copula Circuit, and instances of Quantum Walk, Grover (both if containing 4 qubits) and multiplier (if containing 4 or 8 qubits). Sometimes, the patterns only appear after measurement lifting took place. I do agree that it is not a very common pattern though. However, my argument here would be that it can still be useful to check for it, as it can potentially lead to reduce costly multi-qubit gates

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -101,4 +101,61 @@ def MappingPass : Pass<"place-and-route", "mlir::ModuleOp"> {
...
+    Hadamard lifting uses the following commutation rules:
+         ┌───┐ ┌───┐      ┌───┐ ┌───┐
+        ─┤ X ├─┤ H ├─ => ─┤ H ├─┤ Z ├─
+         └───┘ └───┘      └───┘ └───┘
+         ┌───┐ ┌───┐      ┌───┐ ┌───┐
+        ─┤ Z ├─┤ H ├─ => ─┤ H ├─┤ X ├─
+         └───┘ └───┘      └───┘ └───┘
+         ┌───┐ ┌───┐      ┌───┐ ┌───┐ ┌───────┐
+        ─┤ Y ├─┤ H ├─ => ─┤ H ├─┤ Y ├─┤ G(pi) ├─
+         └───┘ └───┘      └───┘ └───┘ └───────┘
```
burgholzer: The notation for the global phase here is slightly misleading as it looks like the phase would apply to a qubit; but it is a global phase.

mlir/include/mlir/Compiler/CompilerPipeline.h
```diff
@@ -43,6 +43,9 @@ struct QuantumCompilerConfig {
 
   /// Disable quaternion-based single-qubit rotation gate merging
   bool disableMergeSingleQubitRotationGates = false;
+
+  /// Enable Hadamard lifting
+  bool enableHadamardLifting = false;
```
burgholzer, line 48: Is there any reason why one would not want to run this as part of the default pipeline?
I'd rather follow a similar trend here as in the quaternion merge pass, and add a disable option that defaults to false.
Same for the `mqt-cc` CLI option and other places where it currently defaults to off.
Edit: After offline discussion, we may want to postpone this change to when the pass is actually beneficial to run. Let's leave it like it is for now.

burgholzer (approved):
As discussed already offline, this LGTM now. 
I pushed a couple of changes to simplify the pass implementation.
There's also one more open comment about an illustration in the documentation. That should be pretty quick to address though.
Potentially, CodeRabbit will also have a handful of comments since I switched the PR back to ready-for-review.
Afterwards, this should be good to go.

mlir/lib/Dialect/QCO/Transforms/Optimizations/HadamardLifting.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    // op needs to be in front of a Hadamard gate
+    auto hadamardGate = dyn_cast<HOp>(*op->getUsers().begin());
+
+    if (!hadamardGate) {
+      return failure();
+    }
+
+    return swapPauliWithHadamard(op, hadamardGate, rewriter);
```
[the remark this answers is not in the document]
LiRem101: @coderabbitai Due to linear typing, `op` will always have at least one User, and we do not need a guard

mlir/lib/Dialect/QCO/Transforms/Optimizations/HadamardLifting.cpp
```diff
@@ -0,0 +1,242 @@
...
+    for (unsigned int i = 0; i < cnotGate.getNumControls(); i++) {
+      if (isa<MeasureOp>(*cnotGate.getOutputControl(i).getUsers().begin())) {
+        if (i == cnotGate.getNumControls() - 1) {
+          return failure();
+        }
+      } else {
+        controlIndex = i;
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
LiRem101: @coderabbitai Due to linear typing, `op` will always have at least one User, and we do not need a guard. Additionally, a `CtrlOp` always has at least one control qubit. We know that cnotGate is a CtrlOp, as this has been checked in line 162.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
LiRem101: You are right, I changed the notation

burgholzer (approved):
LGTM 👍🏼 
I'll get this in once #1694 is in.

PR #1620, opened 2026-04-07
✨ Add GateTable registry and round-trip corpus tests for conversions

### LINKED ISSUE #1604: 🎨 Compact conversions using `std::index_sequence`
### Problem Statement

Inspired by the `JaspToQC` conversion in https://github.com/FullStaQD/compiler (https://github.com/FullStaQD/compiler/pull/18/changes#diff-90c1b2cd0a0990b9837f7d36e6c7ae839a07946ff693eb6f53a0183905ff7880R252), check if the same pattern can be applied to the conversions in MQT Core.

### Proposed Solution

Check if any of the conversions (`JeffToQCO`, `QCOToJeff`,  `QCOToQC`, QCToQCO`, `QCToQIR`) can also use this pattern.

simon1hofmann:
## Summary
- Introduce `mlir/include/mlir/Conversion/GateTable.h` as a single source of truth for supported gates across conversion passes.
- Refactor conversions to consume the gate table instead of duplicating per-gate boilerplate:
  - `QCToQCO`, `QCOToQC`, `QCToQIR`
- Add a new GateTable round-trip corpus test to validate conversion consistency across all table gates and common modifier variants (base/inv/ctrl).

## Motivation
Gate support was previously duplicated across multiple conversions, making it easy for backends to drift. Centralizing gate metadata improves maintainability and helps ensure conversions stay aligned.

## Details
- **GateTable**
  - Encodes targets/parameter arity per gate.
  - Captures QC→QIR unitary runtime function-name selector for controlled variants.
- **QC→QIR**
  - Uses gate-table metadata to register unitary gate patterns.
  - Structural/non-unitary ops (alloc/static/measure/reset/ctrl/yield/barrier) remain handled by dedicated patterns.

Fixes #1604.

## Checklist


burgholzer:
#1580 is in 🚀 
So registers are now first class citizens in QCO. This will probably require a bit of an advanced merge here.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
denialhaag, line 1: Unfortunately, we had to temporarily disable the Jeff tests in #1580. After rebasing, you will lose feedback on the implementation here. If you still want to see whether the code executes, you can uncomment this line:

https://github.com/munich-quantum-toolkit/core/blob/6f076d297ba9df74ce69915eeb44440f5c30e8a2/mlir/unittests/Conversion/JeffRoundTrip/CMakeLists.txt#L30

The tests will fail during equivalence checking because the tensor types don't match.

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -91,218 +93,77 @@ static void createModified(
...
-  }
+    rewriter.replaceOpWithNewOp<QCOOpType>(op, targets[TargetIndices]...,
+                                           parameters[ParamIndices]...);
+    return;
+  }
+
+  auto lambda = [&](ValueRange innerTargets) -> llvm::SmallVector<Value> {
+    auto qcoOp =
+        QCOOpType::create(rewriter, op.getLoc(), innerTargets[TargetIndices]...,
+                          parameters[ParamIndices]...);
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai Added table-driven `jeff.custom` arity validation (covers 726–754): we now check
`op.getInTargetQubits().size() == TARGETS` and `op.getParams().size() == PARAMS`
before any indexing/forwarding, and fail via `rewriter.notifyMatchFailure` on
mismatch. `createGateFromJeffArity` also enforces `targets.size()`/`parameters.size()`
before calling `createGateFromJeff`. Additionally, the `sx`/`barrier` fast paths
no longer silently drop extra params (`sx` routes through the arity check; `barrier`
rejects non-empty params).

mlir/lib/Conversion/QCToQIR/QCToQIR.cpp
```diff
@@ -1036,6 +1138,44 @@ struct ConvertQCYieldOp final : StatefulOpConversionPattern<YieldOp> {
...
+    KEY, TARGETS, PARAMS, QCO_OP, QC_OP, JEFF_KIND, JEFF_OP,                   \
+    JEFF_BASE_ADJOINT, JEFF_CUSTOM_NAME, JEFF_PPR, QIR_KIND, QIR_FN)           \
+  patterns.add<ConvertQCUnitaryOpQIR<QC_OP, (TARGETS), (PARAMS), &(QIR_FN)>>(  \
+      typeConverter, ctx, &state);
+  MQT_GATE_TABLE(MQT_ADD_QC_TO_QIR_UNITARY)
+#undef MQT_ADD_QC_TO_QIR_UNITARY
+
+  patterns.add<ConvertQCBarrierOp, ConvertQCCtrlOp, ConvertQCYieldOp>(
+      typeConverter, ctx, &state);
+}
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai This is already addressed in the current `QCToQIR.cpp`.

- **GateTable filtering**: `populateQCToQIRPatterns` now filters `MQT_GATE_TABLE` rows with
  `if constexpr ((QIR_KIND) == ::mlir::mqt::gates::QIRKind::Unitary)` before registering
  `ConvertQCUnitaryOpQIR<...>`, so non-unitary rows are ignored.

- **Legacy DEFINE_* registrations**: I checked the file and there are no remaining
  `DEFINE_*` macro-based QC→QIR pattern registrations (no `DEFINE_ONE_TARGET`,
  `DEFINE_TWO_TARGET`, etc.) to remove, so there’s no duplicate implementation left to drift.

simon1hofmann:
@coderabbitai review

simon1hofmann:
@denialhaag would be great to get your feedback on this 🙏

denialhaag:
> @denialhaag would be great to get your feedback on this 🙏

Sorry, this somehow got lost in my inbox. I'll have a look first thing tomorrow morning!

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,207 @@
...
+ */
+
+#include <cstddef>
+#include <cstdint>
+
+namespace mlir::mqt::gates {
+
+/** @brief Jeff lowering kind for a gate entry. */
+enum class JeffKind : std::uint8_t {
+  Native,       //!< Dedicated Jeff op exists.
```
denialhaag: `jeff` calls these `wellKnown`, so maybe we want to adopt this as well: https://github.com/unitaryfoundation/jeff/blob/15a32f4ad8611f457f7dac7420ea32154f7e202d/docs/spec.md?plain=1#L97

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,207 @@
...
+ */
+struct PPRPaulis {
+  std::int32_t p0 = 0;
+  std::int32_t p1 = 0;
+};
+
+// Helper macros to keep `MQT_GATE_TABLE` macro-friendly.
+#define MQT_PPR(P0, P1)                                                        \
+  ::mlir::mqt::gates::PPRPaulis { (P0), (P1) }
+#define MQT_PPR_NONE MQT_PPR(0, 0)
```
denialhaag: I think this has to be treated a bit differently. `jeff`'s PPR gate does accept `0` for the identity gate. Maybe we shouldn't use `0` as a default. 🤔

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -223,6 +225,72 @@ struct ConvertQTensorDeallocOp final : OpConversionPattern<qtensor::DeallocOp> {
...
+                         std::index_sequence<TargetIndices...> /*tgt*/,
+                         std::index_sequence<ParamIndices...> /*par*/) {
+    QCOpType::create(rewriter, loc, qcOperands[TargetIndices]...,
+                     op.getParameter(ParamIndices)...);
+  }
+
+  LogicalResult
+  matchAndRewrite(QCOOpType op, QCOOpType::Adaptor adaptor,
+                  ConversionPatternRewriter& rewriter) const override {
+    const auto qcOperands = adaptor.getOperands();
```
denialhaag: It looks like `getOperands()` returns an IR type. 

```suggestion
   auto qcOperands = adaptor.getOperands();
```

Some context:

> The design team since decided to change to a different model, which eschews `const` entirely for the core IR types: you should never see a `const` method on `Operation`, should never see the type `const Value`, and you shouldn’t feel bad about this. That said, you should use `const` for non-IR types, like `SmallVector`’s and many other things.

From here: https://mlir.llvm.org/docs/Rationale/UsageOfConst/

This also applies in other files.

mlir/lib/Conversion/QCToQIR/QCToQIR.cpp
denialhaag, line 1: I guess it should be possible to remove some of the "old" conversion structures here, right? 🤔

mlir/include/mlir/Conversion/GateTable.h
denialhaag, line 1: This is more of a general comment and is very influenced by my personal opinion, so take this with a grain of salt.

I'm not really sure if I like that the `jeff` conversion is part of this. It feels like `jeff`'s particularities make the table quite a bit more complex (five entries are only due to `jeff`), while making the conversions between QCO and `jeff` seemingly more implicit (at least at first glance).

At the same time, the conversions between QC and QCO and the QC-to-QIR conversion seem to really benefit from this table. They have all gotten easier to read.

I'm not really sure if I have a point. I think I just have an aversion to fitting things together that are maybe not meant to fit. If we ever decide to convert to yet another (more particular ✨) dialect that requires another five (or more) entries to this table, this might just develop into a headache. Maybe @burgholzer can chime in here.

mlir/unittests/Conversion/GateTableRoundTripCorpus/test_gate_table_round_trip_corpus.cpp
denialhaag, line 1: I don't know if this is really necessary. I feel like the coverage wouldn't decrease by a line if we removed this because it more or less reimplements the existing conversion tests. 🤔

denialhaag (dismissed):
Thanks a lot for working on this! 🙂 I really like how this improves the conversions between QC and QCO and the QC-to-QIR conversion. I'm not really sure if we want to fit the `jeff` conversions into this, though. You can find some more details below.

mlir/include/mlir/Conversion/GateTable.h
burgholzer: Yeah, I am with you on this one.
Also briefly discussed with @simon1hofmann already while on a phone call earlier.
Splitting off Jeff makes sense here. The other conversions are things that we have under full control and where we can ensure feature parity so that special cases can be avoided as much as possible.
Jeff support is too experimental for now and likely subject to change over time, which makes me think it deserved remaining separate.

mlir/include/mlir/Conversion/GateTable.h
simon1hofmann: Changed `Native` to `WellKnown`.

mlir/include/mlir/Conversion/GateTable.h
simon1hofmann: Makes sense, removed the `0` initialization.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
simon1hofmann: Thanks for the hint, removed `const` for returned IR types across files.

mlir/lib/Conversion/QCToQIR/QCToQIR.cpp
simon1hofmann: Removed dead code.

mlir/unittests/Conversion/GateTableRoundTripCorpus/test_gate_table_round_trip_corpus.cpp
simon1hofmann: Removed the gate table round trip corpus tests.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -1468,6 +1126,82 @@ class QCOToJeffTypeConverter final : public TypeConverter {
...
+  } else if constexpr (Kind == ::mlir::mqt::gates::JeffKind::SpecialU2ToU) {
+    static_assert(std::is_same_v<QCOOpType, qco::U2Op> && Targets == 1 &&
+                      Params == 2,
+                  "QCOToJeff SpecialU2ToU is only implemented for qco.u2");
+    patterns.add<ConvertQCOU2OpToJeff>(typeConverter, context, &state);
+  } else {
+    static_assert(AlwaysFalse<Kind, Targets, Params>::value,
+                  "MQT_ADD_QCO_TO_JEFF_GATE: unhandled JeffKind");
+  }
+}
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai GateTable should only be used for QC/QCO/QIR conversions.

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,116 @@
...
+
+#pragma once
+
+/**
+ * @file GateTable.h
+ * @brief Central gate registry for QC / QCO / QIR conversions.
+ *
+ * @details
+ * This header defines a single source of truth for the set of supported gates
+ * shared by QCO↔QC lowering and QC→QIR lowering in `mlir/lib/Conversion/`.
```
burgholzer:
```suggestion
 * shared by QCO↔QC conversion and QC→QIR lowering in `mlir/lib/Conversion/`.
```

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Jeff lowering (`QCOToJeff`, `JeffToQCO`) enumerates its supported gates
+ * directly in those conversion implementations.
```
burgholzer:
```suggestion
```
probably not really necessary to list all exceptions here.

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Each entry specifies:
+ * - a canonical gate key (identifier),
+ * - number of targets,
+ * - number of parameters,
+ * - the corresponding QCO and QC op types,
+ * - QIR lowering kind and function-name selector (for QC→QIR only).
+ *
+ * Conversions consume this table as an X-macro list without introducing ODR
+ * issues across translation units.
```
burgholzer:
```suggestion
 * - QIR lowering kind and function-name selector (for QC→QIR only).
```
Feels like LLM-generated noise.

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,116 @@
...
+#include <cstddef>
+#include <cstdint>
+
+namespace mlir::mqt::gates {
+
+/** @brief QIR lowering kind for a gate entry. */
+enum class QIRKind : std::uint8_t {
+  Unitary,     //!< Lower to a QIR runtime call (name depends on num ctrls).
+  Unsupported, //!< Not supported in QC-to-QIR lowering.
+};
```
burgholzer: This enum feels kind of pointless. Every single table entry uses `Unitary`.
Might as well remove that entirely, right?

mlir/include/mlir/Conversion/GateTable.h
```diff
@@ -0,0 +1,116 @@
...
+
+} // namespace mlir::mqt::gates
+
+/**
+ * @brief Central gate table (QC / QCO / QIR only).
```
burgholzer:
```suggestion
 * @brief Central gate table.
```
LLM Noise

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -125,6 +127,119 @@ static void handleResult(Operation* op, ConversionPatternRewriter& rewriter,
...
+ * @return llvm::SmallVector<Value> Collected parameters in index order
+ */
+template <typename OpType, std::size_t... Indices>
+static llvm::SmallVector<Value>
+collectParams(OpType op, std::index_sequence<Indices...> /*indices*/) {
+  llvm::SmallVector<Value> params;
+  params.reserve(sizeof...(Indices));
+  (params.push_back(op.getParameter(Indices)), ...);
+  return params;
+}
```
burgholzer: This kind of triggered me. 
I recently added a `getParameters()` method to the QC UnitaryOpInterface.
This might be enough justification to add that as well for QCO.
Should be pretty similar to add and makes this function redundant.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -1222,10 +911,11 @@ struct ConvertQCOBarrierOpToJeff final
...
 
     llvm::SmallVector<Value> targets;
-    if (!state.inModifier()) {
-      targets = adaptor.getQubitsIn();
-    } else {
+    if (state.inModifier()) {
       targets = state.targetsIn;
+    } else {
+      targets.append(adaptor.getOperands().begin(),
+                     adaptor.getOperands().end());
```
burgholzer: IIRC there is something like `llvm::append_range` that could be used here.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -223,6 +225,74 @@ struct ConvertQTensorDeallocOp final : OpConversionPattern<qtensor::DeallocOp> {
   }
 };
 
+template <typename QCOOpType, typename QCOpType, std::size_t NumTargets,
+          std::size_t NumParams>
+struct ConvertQCOGateToQC final : OpConversionPattern<QCOOpType> {
+  using OpConversionPattern<QCOOpType>::OpConversionPattern;
+
+  /**
+   * @brief Generic QCO gate lowering helper (value semantics -> reference).
```
burgholzer:
```suggestion
   * @brief Generic QCO gate conversion helper (value semantics -> reference).
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -742,321 +743,65 @@ struct ConvertQCResetOp final : StatefulOpConversionPattern<qc::ResetOp> {
...
-
-    assignMappedQubit(state, operation, qcQubit0, qcoOp.getQubit0Out());
-    assignMappedQubit(state, operation, qcQubit1, qcoOp.getQubit1Out());
-
-    rewriter.eraseOp(op);
-
-    return success();
+  template <std::size_t... ParamIndices>
+  static auto createGate(ConversionPatternRewriter& rewriter, Location loc,
+                         const llvm::SmallVector<Value>& qcoTargets,
```
burgholzer: Could this be a ValueRange as well here instead of a SmallVector?
I am pretty sure we could add `getInputTargets()`, `getOutputTargets()`, `getParameters()`, `getInputControls()`, and `getOutputControls()` methods to the UnitaryOpInterface of QCO. I am pretty certain that would simplify some of the code here.

burgholzer (changes requested):
I like this very much!
Makes the code much more compact 🙌🏼 
You'll find some feedback inline for some further improvements that I believe could make sense here.
Some of them might be debatable.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
simon1hofmann: Yes that makes sense, added a `getParameters()` method to the QC UnitaryOpInterface and removed the `collectParams` function.

mlir/include/mlir/Conversion/GateTable.h
simon1hofmann: Removed `QIRKind` from `GateTable`.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: I added the necessary methods, but I'm not entirely sure if everything is correct this way, would be great to get your feedback on that.

burgholzer:
@simon1hofmann would you mind resolving the conflicts here real quick?

simon1hofmann:
> @simon1hofmann would you mind resolving the conflicts here real quick?

Done.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -78,14 +78,22 @@ template <size_t T, size_t P> class TargetAndParameterArityTrait {
...
+    OperandRange getInputTargets() { return getInputQubits(); }
     Value getOutputTarget(const size_t i) { return getOutputQubit(i); }
+    ResultRange getOutputTargets() { return getOutputQubits(); }
 
     static Value getInputControl([[maybe_unused]] size_t i) {
       llvm::reportFatalUsageError("Operation does not have controls");
     }
+    OperandRange getInputControls() {
+      return this->getOperation()->getOperands().slice(T, 0);
+    }
```
burgholzer:
```suggestion
    static OperandRange getInputControls() {
      return OperandRange{};
    }
```
Should be cleaner here.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
```diff
@@ -78,14 +78,22 @@ template <size_t T, size_t P> class TargetAndParameterArityTrait {
...
     static Value getOutputControl([[maybe_unused]] size_t i) {
       llvm::reportFatalUsageError("Operation does not have controls");
     }
+    ResultRange getOutputControls() {
+      return this->getOperation()->getResults().slice(T, 0);
+    }
```
burgholzer:
```suggestion
    static ResultRange getOutputControls() {
      return ResultRange{};
    }
```

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1014,16 +1014,21 @@ def BarrierOp : QCOOp<"barrier", traits = [UnitaryOpInterface]> {
...
         OperandRange getInputQubits() { return getQubitsIn(); }
+        OperandRange getInputTargets() { return getInputQubits(); }
         Value getOutputQubit(size_t i) { return getOutputTarget(i); }
         ResultRange getOutputQubits() { return getQubitsOut(); }
         Value getInputTarget(size_t i);
         Value getOutputTarget(size_t i);
         static Value getInputControl(size_t i) { llvm::reportFatalUsageError("BarrierOp cannot be controlled"); }
+        OperandRange getInputControls() { return getInputQubits().slice(getNumTargets(), 0); }
         static Value getOutputControl(size_t i) { llvm::reportFatalUsageError("BarrierOp cannot be controlled"); }
+        ResultRange getOutputControls() { return getOutputQubits().slice(getNumTargets(), 0); }
```
```suggestion
        static OperandRange getInputControls() { return OperandRange{}; }
static Value getOutputControl(size_t i) { llvm::reportFatalUsageError("BarrierOp cannot be controlled"); }
        static ResultRange getOutputControls() { return ResultRange{}; }
```

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1168,17 +1178,22 @@ def InvOp
...
+        OperandRange getInputTargets() { return getInputQubits(); }
         Value getOutputQubit(size_t i);
-        ResultRange getOutputQubits() { return getResults(); }
+        ResultRange getOutputQubits() { return getQubitsOut(); }
         Value getInputTarget(size_t i) { return getInputQubit(i); }
         Value getOutputTarget(size_t i) { return getOutputQubit(i); }
         static Value getInputControl(size_t i) { llvm::reportFatalUsageError("Operation does not have controls"); }
+        OperandRange getInputControls() { return getInputQubits().slice(getNumTargets(), 0); }
         static Value getOutputControl(size_t i) { llvm::reportFatalUsageError("Operation does not have controls"); }
+        ResultRange getOutputControls() { return getOutputQubits().slice(getNumTargets(), 0); }
```
```suggestion
        static OperandRange getInputControls() { return OperandRange{}; }
static Value getOutputControl(size_t i) { llvm::reportFatalUsageError("Operation does not have controls"); }
        static ResultRange getOutputControls() { return ResultRange{}; }
```

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -149,6 +151,102 @@ static void handleResult(Operation* op, ConversionPatternRewriter& rewriter,
...
+    const bool useState, std::index_sequence<Indices...> /*indices*/) {
+  llvm::SmallVector<Value> selected;
+  selected.reserve(sizeof...(Indices));
+
+  auto src = useState ? ValueRange(stateOperands) : adaptorOperands;
+  assert(src.size() >= sizeof...(Indices) &&
+         "Not enough operands available for conversion");
+  (selected.push_back(src[Indices]), ...);
+  return selected;
+}
```
burgholzer, line 171: Can't this function effectively be eliminated and be replaced by something similar to the parameter handling below, i.e.
```
ValueRange params = op.getParameters();
params[ParamIndices]...
```
Related to that: can't the two functions below not be combined into one that then uses something like `targets[TargetIndices]...` for the target operands?

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -599,92 +697,99 @@ struct ConvertQCOOneTargetZeroParameterToJeff final
...
     if (!state.inModifier()) {
-      target = adaptor.getQubitIn();
-    } else {
-      target = state.targetsIn[0];
+      const auto expected = NumTargets + NumParams;
+      if (adaptor.getOperands().size() != expected) {
+        return op.emitOpError()
+               << "expected " << expected
+               << " operands (targets + parameters) for QCO→Jeff custom gate "
+                  "lowering, got "
```
burgholzer:
```suggestion
                  "conversion, got "
```
Given how there is a back translation as well, this is much more a conversion than a lowering. Might be worth checking that elsewhere as well.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -599,92 +697,99 @@ struct ConvertQCOOneTargetZeroParameterToJeff final
...
-    createCustomOp(op, rewriter, state, {target}, {}, false, "sx");
+    auto targets = selectOperands(adaptor.getOperands(), state.targetsIn,
+                                  state.inModifier(),
+                                  std::make_index_sequence<NumTargets>{});
+
+    llvm::SmallVector<Value> params;
+    params.reserve(NumParams);
+    if constexpr (NumParams != 0) {
+      params.append(op.getParameters().begin(), op.getParameters().end());
+    }
```
burgholzer: Can't you just pass `op.getParameters()` to the createCustomOp function here? saves materializing a vector.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -1535,6 +1207,109 @@ class QCOToJeffTypeConverter final : public TypeConverter {
...
+          .add<ConvertQCOOneTargetOneParameterToJeff<QCOOpType, JeffOpType>>(
+              typeConverter, context, &state);
+    } else if constexpr (Targets == 1 && Params == 3) {
+      patterns
+          .add<ConvertQCOOneTargetThreeParameterToJeff<QCOOpType, JeffOpType>>(
+              typeConverter, context, &state);
+    } else if constexpr (Targets == 2 && Params == 0) {
+      patterns
+          .add<ConvertQCOTwoTargetZeroParameterToJeff<QCOOpType, JeffOpType>>(
+              typeConverter, context, &state);
```
burgholzer: This is probably just me overoptimizing, but do we really need this kind of distinction? Couldn't this just be one function similar to how this is handled in the FullStaQD compiler repo?

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -1535,6 +1207,109 @@ class QCOToJeffTypeConverter final : public TypeConverter {
...
+ * pattern registration. Ill-formed combinations trigger `static_assert` with a
+ * message referencing this function.
+ */
+template <JeffKind Kind, std::size_t Targets, std::size_t Params,
+          typename QCOOpType, typename JeffOpType, bool JeffBaseAdjoint>
+static void addQCOToJeffGatePattern(RewritePatternSet& patterns,
+                                    TypeConverter& typeConverter,
+                                    MLIRContext* context, LoweringState& state,
+                                    StringRef customName,
+                                    const PPRPaulis& ppr) {
```
burgholzer: Couldn't you simply add default arguments for these so the signatures below aren't so repetitive?

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -779,321 +780,44 @@ struct ConvertQCResetOp final : StatefulOpConversionPattern<qc::ResetOp> {
...
+    auto params = op.getParameters();
+    if constexpr (NumTargets == 0) {
+      return QCOOpType::create(rewriter, loc, params[ParamIndices]...);
+    } else if constexpr (NumTargets == 1) {
+      return QCOOpType::create(rewriter, loc, qcoTargets[0],
+                               params[ParamIndices]...);
+    } else {
+      return QCOOpType::create(rewriter, loc, qcoTargets[0], qcoTargets[1],
+                               params[ParamIndices]...);
+    }
```
burgholzer: Can't this also be handled with an index sequence like in the QCO to QC conversion?
Then there would also not be the arbitrary limitation to two targets max.

burgholzer (changes requested):
This feels very clean already! But it also feels like it's not yet fully consistent between the different conversions.
I believe this should be the last iteration we need here 😌 
I hope the suggestions make sense and work.

mlir/include/mlir/Dialect/QCO/IR/QCODialect.h
simon1hofmann: This raises 

```
Call to implicitly-deleted default constructor of 'OperandRange'clang(ovl_deleted_special_init)
ValueRange.h(42, 28): Default constructor of 'OperandRange' is implicitly deleted because base class 'llvm::detail::indexed_accessor_range_base<OperandRange, OpOperand *, Value, Value, Value>' has no default constructor
```

Similar for the other suggested changes.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
simon1hofmann: Yes that works as well 👍

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
simon1hofmann: Also works, saves some more lines.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
simon1hofmann: Added defaults.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: Good idea, refactored the conversion accordingly.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
simon1hofmann: Yes, can be eliminated and the two functions below can be combined.

simon1hofmann:
> This feels very clean already! But it also feels like it's not yet fully consistent between the different conversions. I believe this should be the last iteration we need here 😌 I hope the suggestions make sense and work.

Thanks a lot for the feedback, helped me to further improve the conversions and remove more code. Probably makes sense that you have another look at the last commit and the issue with constructing empty `OperandRange`s and `ResultRange`s.

burgholzer:
> 
> > [!NOTE]
> > Unit test generation is a public access feature. Expect some limitations and changes as we gather feedback and continue to improve it.
> 
> ---
> 
> Generating unit tests... This may take up to 20 minutes.
> 
> 

Second time this happened in the past weeks. I just clicked "resume reviews" for the Rabbit and it started something else 😅

burgholzer:
@coderabbitai full review

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -243,6 +245,74 @@ struct ConvertQTensorDeallocOp final : OpConversionPattern<qtensor::DeallocOp> {
...
+                  ConversionPatternRewriter& rewriter) const override {
+    auto qcOperands = adaptor.getOperands();
+    assert(qcOperands.size() == (NumTargets + NumParams) &&
+           "Unexpected number of operands for QCO->QC gate conversion");
+    auto qcTargets = qcOperands.take_front(NumTargets);
+
+    createGate(rewriter, op.getLoc(), qcOperands,
+               std::make_index_sequence<NumTargets>{},
+               std::make_index_sequence<NumParams>{});
+    rewriter.replaceOp(op, qcTargets);
```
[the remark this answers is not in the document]
burgholzer: As we regularly test debug builds in CI, this is fine as an assertion.

burgholzer (approved):
Thanks @simon1hofmann for your effort on this!
I just pushed a couple of cleanup commits to get this over the finish line 🏁 
I really like how this essentially safes 1k+ lines of code.

PR #1626, opened 2026-04-09
✨ Prevent mixing qubit addressing modes

### LINKED ISSUE #1601: 🚸 Add QC and QCO verifier that catch programs using mixed qubit addressing modes
While working on #1569, I struggled with the fact that it is up to the conversions between QC and QCO to catch an ill-formed program that mixes static qubits with dynamically allocated ones in a single program.
This should not be the responsibility of a conversion, but rather of a verifier in the respective dialects. The conversions may assume that they are passed valid IR.

I have removed the respective code from the conversions in #1569. This is meant as a tracking issue for adding back respective IR verifiers in a follow-up.

simon1hofmann:
## Description

~This update introduces a new helper function, `verifyNoMixedQubitAddressingModes`, which checks for conflicts between dynamically allocated (`qc.alloc` and `qco.alloc`) and statically addressed (`qc.static` and `qco.static`) qubits within the same isolated region. The function is integrated into the `AllocOp` and `StaticOp` operations for both QC and QCO dialects, ensuring that any attempt to mix these addressing modes will result in a verification error. Additionally, unit tests have been added to validate this behavior.~

Enforces a single qubit-addressing mode per program (either static indices or dynamic allocation, including registers / `qtensor.alloc`): program builders abort with a clear usage error on mix, and lowering passes fail with `emitOpError` if mixed IR is converted. 

Fixes #1601 

## Checklist


simon1hofmann:
While implementing #1626, there are (at least) two viable ways to ensure that **static qubit addressing**
(`qc.static` / `qco.static`) is not mixed with **dynamic allocation** (`qc.alloc` / `qco.alloc`) in the
same program scope.

#### Approach A (implemented): op-local verifiers on `alloc`/`static`

**What it does**
- Add `verify()` hooks to `qc.alloc` and `qc.static` (and likewise for QCO).
- Each verifier finds the nearest parent op with `OpTrait::IsIsolatedFromAbove` (typically `func.func`)
  and checks whether the opposite addressing op exists anywhere in that isolated region.
- If both exist, emit an error.

**Pros**
- Minimal IR changes; no new container op required.
- Runs automatically with standard `mlir::verify()`; conversions/passes can assume valid IR.
- Works regardless of nesting (e.g., inside modifier regions).

**Cons**
- Even with early-exit walks, it can still do redundant work (potentially many verifiers scanning the
  same function).
- Diagnostics are attached to whichever op is being verified, not necessarily the “first conflicting site”.

#### Approach B (more efficient): single verifier on a dialect-owned container op

**What it would look like**
- Introduce a QC/QCO “program scope” operation, e.g. `qc.program` / `qco.program`, with
  `OpTrait::IsIsolatedFromAbove`.
- Ensure all QC/QCO code lives under that op (builders and conversions would target it).
- Put the mixed-addressing check in **one** `verify()` on the container op:
  - Walk the program body once, track `sawAlloc` / `sawStatic`.
  - If both become true, error immediately (optionally point at the second mode encountered).

**Pros**
- Most efficient: one linear scan per program scope (no repeated per-op checks).

**Cons**
- Larger design change: introduces new ops and requires updating builders/tests/conversions/IR patterns.
- Potentially disruptive if downstream tooling assumes “QC/QCO directly inside func/module” structure.

I'm not sure if compilation time of approach A is even a bottleneck, but for large circuits with many qubits it would mean O(n^2) complexity when each allocated qubit is compared with all other operations.

Any thoughts on this?

burgholzer:
> I'm not sure if compilation time of approach A is even a bottleneck, but for large circuits with many qubits it would mean O(n^2) complexity when each allocated qubit is compared with all other operations.
> 
> Any thoughts on this?

Hm. It is an undesirable complexity for something that is just verifying an invariant that we assume to be true.
At the moment, I am not fully sure when MLIR runs the `verify` checks as part of our pipeline. Does it even do that? Or does it only do so when explicitly prompted or when printing?

We are already catching mixed usage in the conversions themselves, right?
Maybe that is sufficient?
I am not yet sure though.

simon1hofmann:
> > I'm not sure if compilation time of approach A is even a bottleneck, but for large circuits with many qubits it would mean O(n^2) complexity when each allocated qubit is compared with all other operations.
> > Any thoughts on this?
> 
> Hm. It is an undesirable complexity for something that is just verifying an invariant that we assume to be true. At the moment, I am not fully sure when MLIR runs the `verify` checks as part of our pipeline. Does it even do that? Or does it only do so when explicitly prompted or when printing?
> 
> We are already catching mixed usage in the conversions themselves, right? Maybe that is sufficient? I am not yet sure though.

Currently, `verify` checks are only run in the unit tests, but not as part of the pipeline (`mqt-cc`). 
Only the `QCOToQC` conversion currently catches mixed usage, as it has to decide between replacing `qco.sinks` with `qc.deallocs` or deleting them.

mlir/include/mlir/Dialect/Utils/QubitAddressingVerifier.h
```diff
@@ -0,0 +1,73 @@
...
+    const char* thisSpelling, const char* oppositeSpelling) {
+  ::mlir::Operation* scope =
+      op->template getParentWithTrait<::mlir::OpTrait::IsIsolatedFromAbove>();
+  if (scope == nullptr) {
+    return op.emitOpError(
+        "must be nested within an operation with IsIsolatedFromAbove");
+  }
+
+  bool foundOpposite = false;
+  (void)scope->walk([&](::mlir::Operation* nestedOp) -> ::mlir::WalkResult {
```
burgholzer: We typically prefer the C++-idiomatic `std::ignore =` over C-style void casts.

```suggestion
  std::ignore = scope->walk([&](::mlir::Operation* nestedOp) -> ::mlir::WalkResult {
```

mlir/unittests/Dialect/QC/IR/test_qc_ir.cpp
```diff
@@ -95,6 +98,33 @@ TEST_P(QCTest, ProgramEquivalence) {
...
 }
 
+TEST_F(QCTest, VerifierRejectsMixedQubitAddressingModes) {
+  QCProgramBuilder builder(context.get());
+  builder.initialize();
+
+  // Mixing `qc.alloc` (dynamic) and `qc.static` (static) in the same function
+  // must be rejected by the dialect verifier.
+  (void)builder.allocQubit();
+  (void)builder.staticQubit(0);
```
burgholzer:
```suggestion
  std::ignore = builder.allocQubit();
  std::ignore = builder.staticQubit(0);
```

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -95,6 +98,33 @@ TEST_P(QCOTest, ProgramEquivalence) {
...
 }
 
+TEST_F(QCOTest, VerifierRejectsMixedQubitAddressingModes) {
+  QCOProgramBuilder builder(context.get());
+  builder.initialize();
+
+  // Mixing `qco.alloc` (dynamic) and `qco.static` (static) in the same function
+  // must be rejected by the dialect verifier.
+  (void)builder.allocQubit();
+  (void)builder.staticQubit(0);
```
burgholzer:
```suggestion
  std::ignore = builder.allocQubit();
  std::ignore = builder.staticQubit(0);
```

burgholzer (changes requested):
This looks pretty clean already.
I have two overall concerns: 

This is not yet a complete verifier for the situation.
Since registers landed in #1580, qubits can also be dynamically allocated with `memref.alloc` and `qtensor.alloc`, which also count as dynamic allocations that must not be mixed with static qubits. This makes the whole situation a little bit more complex.

I am also still a bit afraid about the complexity here.
From https://www.jeremykun.com/2023/09/13/mlir-verifiers/
> Verifiers ensure the types and operations in a concrete MLIR program are well-formed. Verifiers are run before and after every pass, and help to ensure that individual passes, folders, rewrite patterns, etc., emit proper IR. This allows you to enforce invariants of each op, and it makes passes simpler because they can rely on the invariants to avoid edge case checking.

So these are, in fact, run after every pass automatically.
I don't quite feel comfortable with the corresponding overhead.
In the case where everything is fine (the common case), this will scan the entire program (of size M) for every qubit allocation (assume N).
Then this has a complexity of `N*M`.

Based on the above, I am leaning towards rather going a slightly different route.
- We should be able to easily verify this in the Builders (by keeping track of which allocation type has already been used). This should be easy to add.
- Furthermore, the conversions should catch further mismatches and error out. I don't think much has to be done in that regard.
- Lastly, we should update the documentation of the dialects themselves and the individual operations to make note of this assumption.

Overall, this feels like the cleaner solution to the problem, doesn't it?

simon1hofmann:
> Based on the above, I am leaning towards rather going a slightly different route.
> 
> * We should be able to easily verify this in the Builders (by keeping track of which allocation type has already been used). This should be easy to add.
> * Furthermore, the conversions should catch further mismatches and error out. I don't think much has to be done in that regard.
> * Lastly, we should update the documentation of the dialects themselves and the individual operations to make note of this assumption.
> 
> Overall, this feels like the cleaner solution to the problem, doesn't it?

Yes that makes more sense and is also what I had in mind.
I will add verifiers to the builders (`QC`, `QCO`, and `QIR`) and also to all conversions.

I just noticed that `QCOToJeff` does not convert `qco.static`, is that concept not supported in Jeff?

simon1hofmann:
> > Based on the above, I am leaning towards rather going a slightly different route.
> > 
> > * We should be able to easily verify this in the Builders (by keeping track of which allocation type has already been used). This should be easy to add.
> > * Furthermore, the conversions should catch further mismatches and error out. I don't think much has to be done in that regard.
> > * Lastly, we should update the documentation of the dialects themselves and the individual operations to make note of this assumption.
> > 
> > Overall, this feels like the cleaner solution to the problem, doesn't it?
> 
> Yes that makes more sense and is also what I had in mind. I will add verifiers to the builders (`QC`, `QCO`, and `QIR`) and also to all conversions.
> 
> I just noticed that `QCOToJeff` does not convert `qco.static`, is that concept not supported in Jeff?



> > Based on the above, I am leaning towards rather going a slightly different route.
> > 
> > * We should be able to easily verify this in the Builders (by keeping track of which allocation type has already been used). This should be easy to add.
> > * Furthermore, the conversions should catch further mismatches and error out. I don't think much has to be done in that regard.
> > * Lastly, we should update the documentation of the dialects themselves and the individual operations to make note of this assumption.
> > 
> > Overall, this feels like the cleaner solution to the problem, doesn't it?
> 
> Yes that makes more sense and is also what I had in mind. I will add verifiers to the builders (`QC`, `QCO`, and `QIR`) and also to all conversions.
> 
> I just noticed that `QCOToJeff` does not convert `qco.static`, is that concept not supported in Jeff?

@denialhaag Also wanted to tag you here concerning static qubits in Jeff, as I think they're currently not handled (besides raising the IllegalDialect exception). Would it make sense to convert static qubits to dynamic qubits in the conversion? Maybe also something to directly address in Jeff?

denialhaag:
> > I just noticed that `QCOToJeff` does not convert `qco.static`, is that concept not supported in Jeff?
> 
> @denialhaag Also wanted to tag you here concerning static qubits in Jeff, as I think they're currently not handled (besides raising the IllegalDialect exception). Would it make sense to convert static qubits to dynamic qubits in the conversion? Maybe also something to directly address in Jeff?

Yes, the concept behind `qco.static` does not exist in `jeff`. We can consider bringing this up in an upcoming meeting. 🤔 For now, I'd almost say it's fine to simply not support programs that include static qubits, since the current `jeff` support is a bit of a patchwork anyway. If the concept of static qubits is not added to `jeff`, we can still convert `qco.static` to `jeff.qubit_alloc` then. If that happens, we would lose the possibility to round-trip.

burgholzer:
> > > I just noticed that `QCOToJeff` does not convert `qco.static`, is that concept not supported in Jeff?
> 
> > 
> 
> > @denialhaag Also wanted to tag you here concerning static qubits in Jeff, as I think they're currently not handled (besides raising the IllegalDialect exception). Would it make sense to convert static qubits to dynamic qubits in the conversion? Maybe also something to directly address in Jeff?
> 
> 
> 
> Yes, the concept behind `qco.static` does not exist in `jeff`. We can consider bringing this up in an upcoming meeting. 🤔 For now, I'd almost say it's fine to simply not support programs that include static qubits, since the current `jeff` support is a bit of a patchwork anyway. If the concept of static qubits is not added to `jeff`, we can still convert `qco.static` to `jeff.qubit_alloc` then. If that happens, we would lose the possibility to round-trip. 

I'd say that
- at least create an issue tracking this deficiency in the Jeff conversion
- bring this up in a meeting or separate issue in the Jeff repository itself
- add a temporary fix for our conversions that simply converts to dynamic allocations; excluding roundtrip tests for programs with static qubits, but potentially adding tests that still assert on the expected output.
- potentially find a long-term solution afterwards.

Does that sound reasonable?

denialhaag:
> > > > I just noticed that `QCOToJeff` does not convert `qco.static`, is that concept not supported in Jeff?
> > 
> > 
> > > 
> > 
> > 
> > > @denialhaag Also wanted to tag you here concerning static qubits in Jeff, as I think they're currently not handled (besides raising the IllegalDialect exception). Would it make sense to convert static qubits to dynamic qubits in the conversion? Maybe also something to directly address in Jeff?
> > 
> > 
> > Yes, the concept behind `qco.static` does not exist in `jeff`. We can consider bringing this up in an upcoming meeting. 🤔 For now, I'd almost say it's fine to simply not support programs that include static qubits, since the current `jeff` support is a bit of a patchwork anyway. If the concept of static qubits is not added to `jeff`, we can still convert `qco.static` to `jeff.qubit_alloc` then. If that happens, we would lose the possibility to round-trip.
> 
> I'd say that
> 
> * at least create an issue tracking this deficiency in the Jeff conversion
> * bring this up in a meeting or separate issue in the Jeff repository itself
> * add a temporary fix for our conversions that simply converts to dynamic allocations; excluding roundtrip tests for programs with static qubits, but potentially adding tests that still assert on the expected output.
> * potentially find a long-term solution afterwards.
> 
> Does that sound reasonable?

Fine with me! I have created https://github.com/munich-quantum-toolkit/core/issues/1629 and https://github.com/unitaryfoundation/jeff/issues/53. Feel free to add comments if I missed something. 🙂

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -36,6 +36,11 @@ namespace mlir::qc {
  * place without producing new SSA values, providing a natural mapping to
  * hardware execution models.
  *
+ * @par Qubit addressing:
+ * A program must use either static qubits (`staticQubit`) or dynamic allocation
+ * (`allocQubit` / qubit registers), never both. The builder terminates with a
```
burgholzer:
```suggestion
 * (`allocQubit` / `allocQubitRegister`), never both. The builder terminates with a
```
For consistency with the other docstrings?

mlir/include/mlir/Dialect/QIR/Builder/QIRProgramBuilder.h
```diff
@@ -59,6 +59,12 @@ namespace mlir::qir {
  * - Measurements block: Measurements, resets, deallocations
  * - Output block: Output recording calls (array-based, grouped by register)
  *
+ * @par Qubit addressing:
+ * A program must use either static qubits (`staticQubit`) or dynamic allocation
+ * (`allocQubit`, `allocQubitRegister`), never both. The builder terminates
+ * with a usage error if the modes are mixed. Lowering to QIR likewise rejects
+ * mixed-mode QC input.
```
burgholzer:
```suggestion
 * with a usage error if the modes are mixed.
```
For consistency with the other docstrings?

mlir/lib/Conversion/JeffToQCO/JeffToQCO.cpp
```diff
@@ -491,6 +491,11 @@ struct ConvertJeffQuregFreeZeroOpToQCO final
 /**
  * @brief Converts jeff.qubit_alloc to qco.alloc
  *
+ * @details
+ * Jeff does not distinguish static qubits from heap allocation. Qubits that
+ * were `qco.static` before `qco-to-jeff` therefore reappear as `qco.alloc`
+ * after `jeff-to-qco`.
+ *
```
burgholzer:
```suggestion
```
Tbh, this does not really feel relevant here. It just matters for the reverse direction.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -74,6 +81,23 @@ struct LoweringState {
...
+  ensureAddressingMode(QubitAddressingMode requestedMode, Operation* op) {
+    if (mode == QubitAddressingMode::Unknown) {
+      mode = requestedMode;
+      return success();
+    }
+    if (mode == requestedMode) {
+      return success();
+    }
+    return op->emitOpError(
+        "cannot mix static and dynamic qubit allocation modes in conversion");
```
burgholzer:
```suggestion
        "cannot mix static and dynamic qubit allocation modes in QCO program");
```

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -266,6 +290,15 @@ struct ConvertQTensorAllocOp final
...
                   ConversionPatternRewriter& rewriter) const override {
+    const auto tensorType = llvm::dyn_cast<RankedTensorType>(op.getType());
+    if (!tensorType ||
+        !llvm::isa<qco::QubitType>(tensorType.getElementType())) {
+      return failure();
+    }
+    if (failed(getState().ensureAddressingMode(QubitAddressingMode::Dynamic,
+                                               op.getOperation()))) {
+      return failure();
+    }
```
burgholzer: This can be simplified. QTensors are always tensors of qubits.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -61,6 +61,20 @@ enum class QubitAddressingMode : std::uint8_t {
...
+  ensureAddressingMode(QubitAddressingMode requestedMode, Operation* op) {
+    if (mode == QubitAddressingMode::Unknown) {
+      mode = requestedMode;
+      return success();
+    }
+    if (mode == requestedMode) {
+      return success();
+    }
+    return op->emitOpError(
+        "cannot mix static and dynamic qubit allocation modes in conversion");
```
burgholzer:
```suggestion
        "cannot mix static and dynamic qubit allocation modes in QCO program");
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -91,6 +92,12 @@ struct QubitInfo {
  * - %q2 after the X gate
  */
 struct LoweringState {
+  enum class AllocationMode : std::uint8_t {
+    Unset,
+    Static,
+    Dynamic,
+  };
```
burgholzer: This lacks docstrings to be consistent with the other conversions

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -116,6 +123,22 @@ struct LoweringState {
 
   /// Stack of active modifier regions
   SmallVector<ModifierFrame> modifierFrames;
+
+  AllocationMode allocationMode = AllocationMode::Unset;
```
burgholzer, line 129: docstring?

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -116,6 +123,22 @@ struct LoweringState {
...
+                                                   Operation* op) {
+    if (allocationMode == AllocationMode::Unset) {
+      allocationMode = mode;
+      return success();
+    }
+    if (allocationMode == mode) {
+      return success();
+    }
+    return op->emitOpError(
+        "cannot mix static and dynamic qubit allocation modes in conversion");
```
burgholzer:
```suggestion
        "cannot mix static and dynamic qubit allocation modes in QC program");
```

mlir/lib/Conversion/QCToQIR/QCToQIR.cpp
```diff
@@ -68,6 +68,12 @@ namespace {
  * @brief State object for tracking lowering information during QIR conversion
  */
 struct LoweringState : QIRMetadata {
+  enum class AllocationMode : std::uint8_t {
+    Unset,
+    Static,
+    Dynamic,
+  };
```
burgholzer: Docstrings?

mlir/lib/Conversion/QCToQIR/QCToQIR.cpp
```diff
@@ -95,6 +101,22 @@ struct LoweringState : QIRMetadata {
   Block* entryBlock{};
   Block* measurementsBlock{};
   Block* outputBlock{};
+
+  AllocationMode allocationMode = AllocationMode::Unset;
```
burgholzer, line 107: Docstring?

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -472,6 +476,26 @@ void QCProgramBuilder::checkFinalized() const {
   }
 }
 
+void QCProgramBuilder::ensureAllocationMode(const AllocationMode requestedMode,
+                                            const char* requestedName) {
```
burgholzer: having to supply the name here feels redundant and inconsistent with below where the string is also directly derived from the enum value.
This also holds for all other places where this kind of function is used.

mlir/unittests/Conversion/QCOToQC/test_qco_to_qc.cpp
```diff
@@ -27,6 +27,7 @@
 #include <mlir/IR/DialectRegistry.h>
 #include <mlir/IR/MLIRContext.h>
 #include <mlir/IR/Verifier.h>
+#include <mlir/Parser/Parser.h>
```
burgholzer: Why is that include required now and was not required before?
also holds for some of the other test files here.

mlir/unittests/Dialect/QC/IR/test_qc_ir.cpp
```diff
@@ -12,6 +12,7 @@
 #include "mlir/Dialect/QC/Builder/QCProgramBuilder.h"
 #include "mlir/Dialect/QC/IR/QCDialect.h"
 #include "mlir/Support/IRVerification.h"
+#include "mlir/Support/LogicalResult.h"
```
burgholzer:
```suggestion
```
Don't see where this is used in the added code.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -15,6 +15,7 @@
 #include "mlir/Dialect/QTensor/IR/QTensorDialect.h"
 #include "mlir/Dialect/QTensor/IR/QTensorOps.h"
 #include "mlir/Support/IRVerification.h"
+#include "mlir/Support/LogicalResult.h"
```
burgholzer:
```suggestion
```

burgholzer (changes requested):
This looks pretty good to me and like a clean solution.
Practically all of my comments are nitpicks and small details.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -116,6 +123,22 @@ struct LoweringState {
...
+    if (allocationMode == AllocationMode::Unset) {
+      allocationMode = mode;
+      return success();
+    }
+    if (allocationMode == mode) {
+      return success();
+    }
+    return op->emitOpError(
+        "cannot mix static and dynamic qubit allocation modes in QC program");
+  }
```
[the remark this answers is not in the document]
simon1hofmann: @burgholzer Is this something we should already keep in mind?

mlir/unittests/Conversion/QCOToQC/test_qco_to_qc.cpp
simon1hofmann: I first added and then removed some conversions tests again and missed those added includes.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
simon1hofmann: Yes makes sense, can be inferred from the `requestedMode`.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
simon1hofmann: Added missing docstrings and made the naming more consistent.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
simon1hofmann: Removed the redundant first check.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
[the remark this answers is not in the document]
burgholzer: @simon1hofmann @coderabbitai tracking this globally is entirely fine for now. We will not be mixing addressing modes in any of the programs that we generate ourselves and qubit routing will convert any program using dynamic addressing to static addressing.
Let's keep it simple here and make it more fine granular when/if this should actually be needed.

burgholzer (approved):
This looks great now. Many thanks! 🙏🏼

PR #1638, opened 2026-04-14
✨ Add SCF conversions between QC and QCO dialects

li-mingbao:
## Description

This PR adds support for the conversion of the `scf` operations `scf.while`, `scf.for`, `scf.if` between the `QC` and the `QCO` dialect. This allows the conversion of programs with nonlinear controlflow.

This PR is the revamped version of the #1396 PR to integrate the newest changes.
Changes compared to the PR and new features:

- Conversions support registers and qubits inside the `scf` operations.
- The func conversion is not added since this also requires a `qtensor.cast` operation first.
- `scf.if` is converted into `qco.if` 
- `qco.if` also supports tensors of qubits now as input types.
- `qco.yield` also supports tensors of qubits now as input types.
 
## Checklist:


li-mingbao:
Hey @burgholzer @denialhaag , 
While I probably still have to go through a bunch of coderabbit comments before a full review, I want to ask regarding the `QCOProgramBuilder`, mainly on how to define programs with them.

At the moment the builder of the `scf` operations inserts all the extracted qubits from the current region and updates the tensor tracking accordingly before creating the scf operation itself. This saves the user the effort to manually insert every qubit before using a tensor, especially since they are extracted at the beginning. The following extraction sadly does not work and the user has to manually extract the qubits they want. 

You can look at the tests to see how the `QCOProgramBuilder` works at the moment and if this "style" is fine with you.

burgholzer:
> Hey @burgholzer @denialhaag , While I probably still have to go through a bunch of coderabbit comments before a full review, I want to ask regarding the `QCOProgramBuilder`, mainly on how to define programs with them.
> 
> At the moment the builder of the `scf` operations inserts all the extracted qubits from the current region and updates the tensor tracking accordingly before creating the scf operation itself. This saves the user the effort to manually insert every qubit before using a tensor, especially since they are extracted at the beginning. The following extraction sadly does not work and the user has to manually extract the qubits they want.
> 
> You can look at the tests to see how the `QCOProgramBuilder` works at the moment and if this "style" is fine with you.

I just briefly had a look at the Program Builder and the related tests for QCO.
Overall, this looks really good in my opinion 🙌🏼 
The only thing I am concerned about a little bit is that the current re-insertion means that I cannot pass a register _and_ one of its qubits to an SCF construct at the same time.
This is entirely fine in my opinion. I am just wondering whether we need some kind of check for that somewhere. But maybe such a situation would never come up in the first place.

I will give this a proper review once the Rabbit has unleashed its wrath here 🙃
However, from a first look, this already looks really promising!
Thanks for pushing this @li-mingbao 🙏🏼

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -91,90 +92,132 @@ parseTargetAliasing(OpAsmParser& parser, Region& region,
...
-
-static void printIfOpAliasing(OpAsmPrinter& printer, Operation* op,
-                              Region& thenRegion, Region& elseRegion,
-                              OperandRange qubits) {
-  printer << "qubits";
-  printTargetAliasing(printer, op, thenRegion, qubits);
-  printer << " else ";
-  printer << "qubits";
-  printTargetAliasing(printer, op, elseRegion, qubits);
+  printer.printRegion(region, false);
```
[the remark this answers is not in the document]
li-mingbao: The InvOp and the CtrlOp still requires it.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -973,30 +1208,362 @@ struct ConvertQCYieldOp final : StatefulOpConversionPattern<qc::YieldOp> {
...
+
+    // Create the new ForOp
+    auto newForOp =
+        scf::ForOp::create(rewriter, op.getLoc(), op.getLowerBound(),
+                           op.getUpperBound(), op.getStep(), qcoTargets);
+
+    assignMappedTensors(state, op.getOperation(), registerMap,
+                        newForOp.getResults().take_front(numRegisters));
+    assignMappedQubits(state, op.getOperation(), qubitMap,
+                       newForOp->getResults().take_back(numQubits));
```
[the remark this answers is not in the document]
li-mingbao: It is assumed that operations do not support mixed qubit/classical values in scf operations so operations only have either qubit values or classical values

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -330,16 +412,167 @@ template <typename OpType>
...
+  } else {
+    llvm::append_range(memrefs, state.regionRegisterMap[anchor]);
+    llvm::append_range(qcQubits, state.regionQubitMap[anchor]);
+  }
+
+  SmallVector<Value> qcoTargets;
+  qcoTargets.reserve(memrefs.size() + qcQubits.size());
+  llvm::append_range(qcoTargets, resolveMappedTensors(state, anchor, memrefs));
+  llvm::append_range(qcoTargets, resolveMappedQubits(state, anchor, qcQubits));
+  return qcoTargets;
```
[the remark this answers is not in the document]
li-mingbao: This should be ok since a loaded qubit and the register itself should not be used in the same scf region. It is assumed that required qubits from a memref are loaded inside the region and never come from the outside.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -444,6 +458,96 @@ QCProgramBuilder::inv(const llvm::function_ref<void()>& body) {
...
+  }
+  return *this;
+}
+
+QCProgramBuilder& QCProgramBuilder::scfCondition(Value condition) {
+  checkFinalized();
+
+  scf::ConditionOp::create(*this, condition, ValueRange{});
+  return *this;
+}
```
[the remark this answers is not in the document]
li-mingbao: It can be assumed that this does not happen and allocation always happens at the beginning.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -801,25 +873,160 @@ QCOProgramBuilder& QCOProgramBuilder::sink(Value qubit) {
...
+    llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> beforeBody,
+    llvm::function_ref<llvm::SmallVector<Value>(ValueRange)> afterBody) {
+  checkFinalized();
+  checkQubitType(initArgs);
+
+  // Get the updated arguments after inserting the extracted qubits
+  auto updatedArgs = insertExtractedQubits(initArgs);
+  // Create the empty while operation
+  auto whileOp = scf::WhileOp::create(*this, initArgs.getTypes(), updatedArgs);
+  const llvm::SmallVector<Location> locs(initArgs.size(), getLoc());
```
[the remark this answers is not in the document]
li-mingbao: Ignore the old implementation, the current implementation also works for scf.while and qco.if<-> scf.if conversion. See the tests.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -91,90 +92,132 @@ parseTargetAliasing(OpAsmParser& parser, Region& region,
...
+  // Parse result type list
+  if (parser.parseArrowTypeList(result.types)) {
+    return failure();
+  }
+  // Resolve the operands
+  if (failed(parser.resolveOperands(thenOperands, result.types,
+                                    parser.getCurrentLocation(),
+                                    result.operands))) {
+    return failure();
+  }
```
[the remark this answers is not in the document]
li-mingbao: They should be used now and are used to check if the then and else branch use the same operands.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -194,6 +195,58 @@ void QCOProgramBuilder::updateTensorTracking(Value inputTensor,
...
+      // from this tensor
+      for (auto [qubit, qubitInfo] : validQubits) {
+        if (qubitInfo.regId == regId) {
+          // Create an InsertOp for the qubit
+          auto newTensor = qtensor::InsertOp::create(
+                               *this, qubit, currentTensor, qubitInfo.regIndex)
+                               .getResult();
+          // Update the tensor tracking
+          validQubits.erase(qubit);
+          updateTensorTracking(currentTensor, newTensor);
```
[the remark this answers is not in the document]
li-mingbao: This should be fixed now by incrementing a pointer.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -763,10 +972,28 @@ struct QCOToQC final : impl::QCOToQCBase<QCOToQC> {
...
+        auto tensorType = llvm::dyn_cast<RankedTensorType>(t);
+        auto qMemref =
+            memrefType && llvm::isa<qc::QubitType>(memrefType.getElementType());
+        auto qTensor = tensorType &&
+                       llvm::isa<qco::QubitType>(tensorType.getElementType());
+
+        return (memrefType && qMemref) || (tensorType && qTensor);
+      };
+      return !llvm::any_of(op->getOperandTypes(), isQubitType);
+    });
```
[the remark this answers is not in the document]
li-mingbao: scf.while and qco.if can also be converted now. The conversion expects that each of these scf operations only have either quantum or classical types as iterArgs

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -174,6 +176,24 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * @par Example:
+   * ```c++
+   * auto q0 = builder.memrefLoad(memref, index);
+   * ```
+   * ```mlir
+   * %q0 = memref.load %memref[%index] : memref<3x!qc.qubit>
+   * ```
+   */
+  Value memrefLoad(QubitRegister& memref,
+                   const std::variant<int64_t, Value>& index);
```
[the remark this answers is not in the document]
li-mingbao: The load operation is ok and can be handled as seen by the tests. This is necessary to use the induction variable in a for loop. A warning is added to the builder so that only unique values can be used as indices and the value must be defined in the same region as the loadOp.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -194,6 +195,74 @@ void QCOProgramBuilder::updateTensorTracking(Value inputTensor,
...
+  updatedArgs.reserve(initArgs.size());
+
+  // Iterate through the initial values and add the latest value to the updated
+  // arguments
+  for (auto initArg : initArgs) {
+    if (llvm::isa<QubitType>(initArg.getType())) {
+      // Directly insert qubits
+      updatedArgs.emplace_back(initArg);
+    } else {
+      validateTensorValue(initArg);
```
denialhaag: This could probably use [`llvm::TypeSwitch`](https://llvm.org/doxygen/classllvm_1_1TypeSwitch.html). I think we should be a bit more careful and not assume that something is a tensor just because it's not a qubit. 🤔

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -194,6 +195,74 @@ void QCOProgramBuilder::updateTensorTracking(Value inputTensor,
...
+  for (Type type : values.getTypes()) {
+    if (llvm::isa<QubitType>(type)) {
+      continue;
+    }
+    if (auto tensor = llvm::dyn_cast<RankedTensorType>(type);
+        tensor && llvm::isa<QubitType>(tensor.getElementType())) {
+      continue;
+    }
+
+    llvm::reportFatalUsageError("Elements must be qubit values");
```
denialhaag: This could probably also use [`llvm::TypeSwitch`](https://llvm.org/doxygen/classllvm_1_1TypeSwitch.html).

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1253,19 +1255,109 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
 
+  /**
+   * @brief Construct a scf.for operation with iter args
+   *
+   * @param lowerbound Lowerbound of the loop
+   * @param upperbound Upperbound of the loop
+   * @param step Stepsize of the loop
+   * @param initArgs Initial arguments for the iter args
+   * @param body Function that builds the body of the for operation
+   * @return ValueRange of the results
```
denialhaag, line 1288: I would maybe add a `@details` section here that explains that qubits are automatically inserted via `insertExtractedQubits()`. Same for the methods below.

denialhaag:
> Hey @burgholzer @denialhaag , While I probably still have to go through a bunch of coderabbit comments before a full review, I want to ask regarding the `QCOProgramBuilder`, mainly on how to define programs with them.
> 
> At the moment the builder of the `scf` operations inserts all the extracted qubits from the current region and updates the tensor tracking accordingly before creating the scf operation itself. This saves the user the effort to manually insert every qubit before using a tensor, especially since they are extracted at the beginning. The following extraction sadly does not work and the user has to manually extract the qubits they want.
> 
> You can look at the tests to see how the `QCOProgramBuilder` works at the moment and if this "style" is fine with you.

Thanks a lot for working on this from my side as well! 🙂 

I also had a look at the test programs, and they look good to me. The `insertExtractedQubits()` is definitely a good idea!

denialhaag (changes requested):
Below are three things I stumbled across while going through some of the files. I'm happy to do a fuller review once you have prevailed over the Rabbit! 🙂

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: I think in this case it should be fine since the types are checked before to be valid qubit types.

denialhaag: Ahh, I had overseen this. Then it should be fine!

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: I added a `llvm::TypeSwitch` there.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
li-mingbao: I added a `@details` section to the scf operations.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -330,16 +413,167 @@ template <typename OpType>
...
+  for (auto& region : regions) {
+    auto& qubitInfoMap = state->qubitInfoMap[&region];
+    // Skip empty regions e.g. empty else region of an If operation
+    if (region.empty()) {
+      continue;
+    }
+    // Check that the region has only one block
+    assert(region.hasOneBlock() && "Expected single-block region");
+    // Iterate through all operations of the current region
+    for (auto& operation : region.front().getOperations()) {
```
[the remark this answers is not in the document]
li-mingbao: This should be okay as we only expect regions with one block anyway.

li-mingbao:
@denialhaag 
This should be ready for a review now.
Some changes are regarding the `qco.if` and the `qco.yield` operation so that they can also handle tensors.
These changes could also be applied to the `ctrl` and `inv` modifiers but I ignored them for now in most cases to not bloat this PR too much.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -174,6 +176,24 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   *
+   * @par Example:
+   * ```c++
+   * auto q0 = builder.memrefLoad(memref, index);
+   * ```
+   * ```mlir
+   * %q0 = memref.load %memref[%index] : memref<3x!qc.qubit>
+   * ```
+   */
+  Value memrefLoad(Value memref, Value index);
```
denialhaag, line 200: This should also work by passing an integer.

```suggestion
  Value memrefLoad(Value memref, const std::variant<int64_t, Value>& index);
```

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -937,6 +957,110 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+  //===--------------------------------------------------------------------===//
+  // SCF operations
+  //===--------------------------------------------------------------------===//
+
+  /**
+   * @brief Construct a scf.for operation
+   *
+   * @param lowerbound Lowerbound of the loop
+   * @param upperbound Upperbound of the loop
+   * @param step Stepsize of the loop
```
denialhaag: I'd at least adapt the English versions here. I'm not sure if I'd prefer `lowerBound` or `lb` over `lowerbound`.

```suggestion
   * @param lowerbound Lower bound of the loop
   * @param upperbound Upper bound of the loop
   * @param step Step size of the loop
```

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -937,6 +957,110 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * ```
+   * ```mlir
+   * scf.while : () -> () {
+   *   %res = qc.measure %q0 : !qc.qubit -> i1
+   *   scf.condition(%res)
+   * } do {
+   *   qc.h %q0 : !qc.qubit
+   *   scf.yield
+   * }
+   * ```
```
denialhaag: Not _super_ important, but this example is not particularly realistic. The loop either enters the do block immediately or is stuck forever.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -937,6 +957,110 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   */
+  QCProgramBuilder& scfWhile(const llvm::function_ref<void()>& beforeBody,
+                             const llvm::function_ref<void()>& afterBody);
+
+  /**
+   * @brief Construct a scf.if operation
+   *
+   * @param condition Condition for the if operation
+   * @param thenBody Function that builds the then body of the if
+   * operation
```
denialhaag:
```suggestion
   * @param thenBody Function that builds the then body of the if operation
```

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -937,6 +957,110 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * ```c++
+   * builder.scfIf(condition, [&] {
+   *   builder.h(q0);
+   * });
+   * ```
+   * ```mlir
+   * scf.if %condition {
+   *   qc.h %q0 : !qc.qubit
+   * }
+   * ```
```
denialhaag, line 1052: Maybe it's nice to show the more complex case with a non-trivial else block here. 🤔

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
denialhaag, line 1: Most of the points I'm making in `QCProgramBuilder.h` also apply here.

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1253,19 +1257,121 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   *   return {insert};
+   * });
+   * ```
+   * ```mlir
+   * %q2 = scf.for %iv = %lb to %ub step %step iter_args(%arg0 = %q0)
+   * -> (tensor<3x!qco.qubit>) {
+   *   %outTensor, %q0 = qtensor.extract %arg0[%iv] : tensor<3x!qco.qubit>
+   *   %q1 = qco.h %q0 : !qco.qubit -> !qco.qubit
+   *   %insert = qtensor.insert %q1 into %outTensor[%iv] : tensor<3x!qco.qubit>
+   *   scf.yield %insert : tensor<3x!qco.qubit>
```
denialhaag: This may be easier to read.

```suggestion
   *   %t0, %q0 = qtensor.extract %arg0[%iv] : tensor<3x!qco.qubit>
   *   %q1 = qco.h %q0 : !qco.qubit -> !qco.qubit
   *   %t1 = qtensor.insert %q1 into %t0[%iv] : tensor<3x!qco.qubit>
   *   scf.yield %t1 : tensor<3x!qco.qubit>
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1253,19 +1257,121 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
    * ```
    */
   ValueRange
-  qcoIf(const std::variant<bool, Value>& condition, ValueRange qubits,
+  qcoIf(const std::variant<bool, Value>& condition, ValueRange initArgs,
         llvm::function_ref<SmallVector<Value>(ValueRange)> thenBody,
         llvm::function_ref<SmallVector<Value>(ValueRange)> elseBody = nullptr);
 
+  /**
+   * @brief Construct a scf.for operation with iter args
```
denialhaag: I think this is a bit more in line with the other docstrings in this file.

```suggestion
   * @brief Construct a scf.for operation
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1253,19 +1257,121 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * ```
+   */
+  ValueRange
+  scfFor(const std::variant<int64_t, Value>& lowerbound,
+         const std::variant<int64_t, Value>& upperbound,
+         const std::variant<int64_t, Value>& step, ValueRange initArgs,
+         llvm::function_ref<llvm::SmallVector<Value>(Value, ValueRange)> body);
+
+  /**
+   * @brief Construct a scf.while operation with arguments
```
denialhaag:
```suggestion
   * @brief Construct a scf.while operation
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1253,19 +1257,121 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+  /**
+   * @brief Construct a scf.condition operation with yielded values
+   *
+   * @param condition Condition for the condition operation
+   * @param yieldedValues ValueRange of the yieldedValues
+   * @return Reference to this builder for method chaining
+   *
+   * @par Example:
+   * ```c++
+   * builder.scfCondition(condition, yieldedValues);
```
denialhaag:
```suggestion
   * builder.scfCondition(condition, q0);
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1358,6 +1464,22 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
    */
   void updateTensorTracking(Value inputTensor, Value outputTensor);
 
+  /**
+   * @brief Updates the latest QCO values of the initial values after inserting
+   * all extracted qubits back to the tensor.
+   * @param initArgs ValueRange of the initial QCO values
+   * @return SmallVector of the updated values after inserting the qubits back
+   * to the tensors.
+   */
```
denialhaag, line 1482: I think this could convey the method's behavior a bit better. `initArgs` can be qubits and QTensor values. Qubits are returned without modification, but tensors are modified.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -104,6 +114,38 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
 };
 } // namespace
 
+/**
+ * @brief Moves the operations from one region into another.
+ *
+ * @details Moves the operations from the source region into the target region.
+ * The target region replaces the uses of the old blockarguments with the
+ * replacementValues and erases the unused blockarguments.
```
denialhaag:
```suggestion
 * The target region replaces the uses of the old block arguments with the
 * @p replacementValues and erases the unused block arguments.
```

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -104,6 +114,38 @@ class StatefulOpConversionPattern : public OpConversionPattern<OpType> {
...
+ *
+ * @param sourceRegion Source region where the operations are moved from
+ * @param targetRegion Target region where the operations are moved to
+ * @param offset Offset to the arguments that are dropped
+ * @param numArgs Number of arguments that are dropped
+ * @param replacementValues Values to replace the uses of the arguments
+ * @param rewriter PatternRewriter of the current conversion pass
+ *
+ */
```
denialhaag, line 130:
```suggestion
 * @param rewriter PatternRewriter of the current conversion pass
 */
```

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -712,7 +724,210 @@ struct ConvertQCOYieldOp final : OpConversionPattern<qco::YieldOp> {
...
+ * values.
+ *
+ * @par Example:
+ * ```mlir
+ * %targets_out = scf.for %iv = %lb to %ub step %step iter_args(%arg0 =
+ * %qtensor) -> (tensor<3x!qco.qubit) {
+ *   %outTensor, %q0 = qtensor.extract %arg0[%iv] : tensor<3x!qco.qubit>
+ *   %q1 = qco.h %q0 : !qco.qubit -> !qco.qubit
+ *   %insert = qtensor.insert %q1 into %outTensor[%iv] : tensor<3x!qco.qubit>
+ *   scf.yield %insert : tensor<3x!qco.qubit>
```
denialhaag: See my other comment.

```suggestion
 *   %t0, %q0 = qtensor.extract %arg0[%iv] : tensor<3x!qco.qubit>
 *   %q1 = qco.h %q0 : !qco.qubit -> !qco.qubit
 *   %insert = qtensor.insert %q1 into %t1[%iv] : tensor<3x!qco.qubit>
 *   scf.yield %t1 : tensor<3x!qco.qubit>
```

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -763,10 +978,27 @@ struct QCOToQC final : impl::QCOToQCBase<QCOToQC> {
...
+              return llvm::isa<qc::QubitType>(t.getElementType());
+            })
+            .Case<RankedTensorType>([](RankedTensorType t) {
+              return llvm::isa<qco::QubitType>(t.getElementType());
+            })
+            .Default([](auto) { return false; });
+      };
+
+      return !llvm::any_of(op->getOperandTypes(), isQubitType);
+    });
```
denialhaag, line 995: This confuses me a bit. Shouldn't SCF operations be illegal if and only if operands of `!qco.qubit` type are involved?

Or is the illegality of `!qc.qubit` operands necessary because the type converter converts them first? 🤔

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -123,6 +134,16 @@ struct LoweringState {
   /// information
   llvm::DenseMap<Region*, llvm::DenseMap<Value, QubitInfo>> qubitInfoMap;
 
+  /// Per-region map from original its QC register to its extracted QC qubits
```
denialhaag:
```suggestion
  /// Per-region map from original QC register to its extracted QC qubits
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -288,6 +331,16 @@ resolveMappedQubits(LoweringState& state, Operation* anchor,
   }));
 }
 
+/** @brief Resolves a range of QC memrefs to their latest QCO tensor values. */
```
denialhaag:
```suggestion
/** @brief Resolves a range of QC memrefs to their latest QTensor values. */
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -298,14 +351,44 @@ static void assignMappedQubits(LoweringState& state, Operation* anchor,
   }
 }
 
+/** @brief Updates mappings for matching QC memref and QCO tensor ranges. */
```
denialhaag:
```suggestion
/** @brief Updates mappings for matching QC memref and QTensor ranges. */
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -330,16 +413,167 @@ template <typename OpType>
...
+      auto insertOp =
+          qtensor::ExtractOp::create(rewriter, target->getLoc(), tensor, index);
+      assignMappedTensor(state, anchor, memref, insertOp.getOutTensor());
+      assignMappedQubit(state, anchor, qubit, insertOp.getResult());
+    }
+  }
+}
+
+/** @brief Resolves all QC qubit and memref values to their
+ * latest QCO values.
```
denialhaag:
```suggestion
 * latest QCO and QTensor values.
```

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -330,16 +413,167 @@ template <typename OpType>
...
  *
  * @details
- * QC uses reference semantics and does not enforce linear typing for qubits.
- * After conversion, QCO requires that every qubit SSA value is consumed
- * exactly once. For allocations (including static qubits), the sink is
- * `qco.sink`. This pattern inserts `qco.sink` operations for all
+ * QC uses reference semantics and does not enforce linear typing for
+ * qubits. After conversion, QCO requires that every qubit SSA value is
+ * consumed exactly once. For allocations (including static qubits), the
+ * sink is `qco.sink`. This pattern inserts `qco.sink` operations for all
```
denialhaag: These and some other docstring changes below seem accidental. Maybe it's better to revert them to keep the PR's footprint a bit smaller.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -973,30 +1209,362 @@ struct ConvertQCYieldOp final : StatefulOpConversionPattern<qc::YieldOp> {
...
+ * }
+ * ```
+ * is converted to
+ * ```mlir
+ * %targets_out = scf.for %iv = %lb to %ub step %step iter_args(%arg0 =
+ * %qtensor) -> (tensor<3x!qco.qubit) {
+ *   %outTensor, %q0 = qtensor.extract %arg0[%iv] : tensor<3x!qco.qubit>
+ *   %q1 = qco.h %q0 : !qco.qubit -> !qco.qubit
+ *   %insert = qtensor.insert %q1 into %outTensor[%iv] : tensor<3x!qco.qubit>
+ *   scf.yield %insert : tensor<3x!qco.qubit>
```
denialhaag:
```suggestion
 *   %t0, %q0 = qtensor.extract %arg0[%iv] : tensor<3x!qco.qubit>
 *   %q1 = qco.h %q0 : !qco.qubit -> !qco.qubit
 *   %t1 = qtensor.insert %q1 into %t0[%iv] : tensor<3x!qco.qubit>
 *   scf.yield %t1 : tensor<3x!qco.qubit>
```

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -444,6 +457,96 @@ QCProgramBuilder::inv(const llvm::function_ref<void()>& body) {
...
+//===----------------------------------------------------------------------===//
+
+QCProgramBuilder&
+QCProgramBuilder::scfFor(const std::variant<int64_t, Value>& lowerbound,
+                         const std::variant<int64_t, Value>& upperbound,
+                         const std::variant<int64_t, Value>& step,
+                         const llvm::function_ref<void(Value)>& body) {
+  checkFinalized();
+
+  const auto loc = getLoc();
```
denialhaag: I believe the location is an IR type and the `const` can be dropped: https://mlir.llvm.org/docs/Rationale/UsageOfConst/

```suggestion
  auto loc = getLoc();
```

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -444,6 +457,96 @@ QCProgramBuilder::inv(const llvm::function_ref<void()>& body) {
...
+  const auto lb = utils::variantToValue(*this, loc, lowerbound);
+  const auto ub = utils::variantToValue(*this, loc, upperbound);
+  const auto stepSize = utils::variantToValue(*this, loc, step);
```
denialhaag:
```suggestion
  auto lb = utils::variantToValue(*this, loc, lowerbound);
  auto ub = utils::variantToValue(*this, loc, upperbound);
  auto stepSize = utils::variantToValue(*this, loc, step);
```

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -444,6 +457,96 @@ QCProgramBuilder::inv(const llvm::function_ref<void()>& body) {
...
+  return *this;
+}
+
+QCProgramBuilder&
+QCProgramBuilder::scfIf(const std::variant<bool, Value>& cond,
+                        const llvm::function_ref<void()>& thenBody,
+                        const llvm::function_ref<void()>& elseBody) {
+  checkFinalized();
+
+  const auto condition = utils::variantToValue(*this, getLoc(), cond);
```
denialhaag:
```suggestion
  auto condition = utils::variantToValue(*this, getLoc(), cond);
```

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -194,6 +196,93 @@ void QCOProgramBuilder::updateTensorTracking(Value inputTensor,
...
+                           .Case<RankedTensorType>([](RankedTensorType t) {
+                             return llvm::isa<QubitType>(t.getElementType());
+                           })
+                           .Default([](Type) { return false; });
+
+    if (!isQubitType) {
+      llvm::reportFatalUsageError("Elements must be qubit values");
+    }
+  }
+}
```
denialhaag: Does it make sense to already declare this in `QCOProgramBuilder.h`? 🤔

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -801,67 +892,180 @@ QCOProgramBuilder& QCOProgramBuilder::sink(Value qubit) {
...
 
+ValueRange QCOProgramBuilder::scfFor(
+    const std::variant<int64_t, Value>& lowerbound,
+    const std::variant<int64_t, Value>& upperbound,
+    const std::variant<int64_t, Value>& step, ValueRange initArgs,
+    llvm::function_ref<llvm::SmallVector<Value>(Value, ValueRange)> body) {
+  checkFinalized();
+  checkQubitType(initArgs);
+
+  const auto loc = getLoc();
```
denialhaag:
```suggestion
  auto loc = getLoc();
```

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -801,67 +892,180 @@ QCOProgramBuilder& QCOProgramBuilder::sink(Value qubit) {
 // SCF Operations
 //===----------------------------------------------------------------------===//
 
+ValueRange QCOProgramBuilder::scfFor(
+    const std::variant<int64_t, Value>& lowerbound,
+    const std::variant<int64_t, Value>& upperbound,
+    const std::variant<int64_t, Value>& step, ValueRange initArgs,
+    llvm::function_ref<llvm::SmallVector<Value>(Value, ValueRange)> body) {
+  checkFinalized();
+  checkQubitType(initArgs);
```
denialhaag: Would it make sense to call this within `insertExtractedQubits()` to ensure that the two are always called together? 🤔

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -801,67 +892,180 @@ QCOProgramBuilder& QCOProgramBuilder::sink(Value qubit) {
...
+    }
   }
 
-  return ifResults;
-}
+  scf::ConditionOp::create(*this, condition, yieldedValues);
 
+  return *this;
+}
 //===----------------------------------------------------------------------===//
```
denialhaag, line 1068:
```suggestion
}

//===----------------------------------------------------------------------===//
```

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1253,19 +1257,121 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   *   scf.condition(%cond) %q1 : !qco.qubit
+   * } do {
+   * ^bb0(%arg0 : !qco.qubit):
+   *   %q1 = qco.h %arg0 : !qco.qubit -> !qco.qubit
+   *   scf.yield %q1 : !qco.qubit
+   * }
+   * ```
+   */
+  ValueRange
+  scfWhile(ValueRange args,
```
denialhaag: For consistency? ✨ 


```suggestion
  scfWhile(ValueRange initArgs,
```

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -91,90 +92,152 @@ parseTargetAliasing(OpAsmParser& parser, Region& region,
...
+    return failure();
   }
-  printer << ") ";
 
-  printer.printRegion(region, false);
-}
+  SmallVector<OpAsmParser::Argument, 4> thenArgs;
+  SmallVector<OpAsmParser::Argument, 4> elseRegionArgs;
+  SmallVector<OpAsmParser::UnresolvedOperand, 4> thenOperands;
+  SmallVector<OpAsmParser::UnresolvedOperand, 4> elseOperands;
```
denialhaag:
```suggestion
  SmallVector<OpAsmParser::Argument> thenArgs;
  SmallVector<OpAsmParser::Argument> elseRegionArgs;
  SmallVector<OpAsmParser::UnresolvedOperand> thenOperands;
  SmallVector<OpAsmParser::UnresolvedOperand> elseOperands;
```

> In the absence of a well-motivated choice for the number of inlined elements N, it is recommended to use SmallVector<T> (that is, omitting the N). This will choose a default number of inlined elements reasonable for allocation on the stack (for example, trying to keep sizeof(SmallVector<T>) around 64 bytes).

From here: https://llvm.org/doxygen/classllvm_1_1SmallVector.html

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -330,16 +413,167 @@ template <typename OpType>
...
+          }
+        }
+      }
+    }
+  }
+  // Mark scf operations that need to be changed afterwards
+  if ((!regionQubitMap.empty() || !regionRegisterMap.empty()) &&
+      op->getDialect()->getNamespace() == "scf") {
+    op->setAttr("needChange", StringAttr::get(op->getContext(), "yes"));
+  }
```
denialhaag: I guess this is necessary because there's no other way to identify that an SCF operation uses reference semantics when checking legality? 🤔

denialhaag (changes requested):
Hi, @li-mingbao! Thanks for all your work on this! I really like this a lot already! 🙂

Don't be surprised by the number of comments below. They are almost all cosmetic and straightforward to resolve.

This PR adds another level of complexity to the conversions, but it seems necessary. I didn't look at every line of code, but the tests cover different (nested) cases and give me confidence.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -91,90 +92,152 @@ parseTargetAliasing(OpAsmParser& parser, Region& region,
...
-  const auto numTargets = targetsIn.size();
-  for (unsigned i = 0; i < numTargets; ++i) {
-    if (i > 0) {
-      printer << ", ";
-    }
-    printer.printOperand(entryBlock.getArgument(i));
-    printer << " = ";
-    printer.printOperand(targetsIn[i]);
+ParseResult IfOp::parse(::mlir::OpAsmParser& parser,
+                        ::mlir::OperationState& result) {
```
denialhaag, line 96: Something else I just noticed: The parser is currently not tested. Maybe it makes sense to add a test in `test_qco_ir.cpp`. 🤔

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
li-mingbao: This is done on purpose since every `int64_t` this would take is invalid. It is either already extracted from the beginning or it would be out of bounds. 
So the only valid inputs are dynamic index values.

burgholzer: Repeated extracts from a memref are valid though. They would just be eliminated by canonicalization later on.

li-mingbao: From what I have this does not always get canonicalized. When I build and canonicalize the following program:
``` 
auto reg = b.allocQubitRegister(1);
  b.h(reg[0]);
  auto load = b.memrefLoad(reg.value,0);
  b.h(load);
```
I get this IR:
```
 module {                                                                                                             
     %alloc = memref.alloc() : memref<1x!qc.qubit>                                                                    
     %0 = memref.load %alloc[%c0] : memref<1x!qc.qubit>                                                               
     qc.h %0 : !qc.qubit                                                                                              
     %1 = memref.load %alloc[%c0] : memref<1x!qc.qubit>                                                               
     qc.h %1 : !qc.qubit                                                                                              
     memref.dealloc %alloc : memref<1x!qc.qubit>                                                                      
     return %c0_i64 : i64                                                                                             
   }                                                                                                                  
 }  
```
I also initially thought that the 2nd load operation is removed and the gates would be cancelled but this did not happen.

The IR should still be valid but there are two different SSA values for the same qubit now and I dont know if you can do any shenanigans in the QC dialect with that. 
If this is fine I can remove the restrictions of the  `memref.load` operation and allow any load operations.

Edit: This actually causes some issue when converting the operation to QCO as a value would be extracted from the tensor twice at the same index.

denialhaag: Hmm, maybe the canonicalization doesn't work if there's an operation between two loads? 🤔 

Something else: After looking at it again, I don't think the current check is correct. Because it is based on the SSA value and not the underlying integer, it is currently possible to load the same qubit twice. Otherwise, the current implementation wouldn't even work because qubits are loaded by `allocQubitRegister()` and then again by `memrefLoad()` inside the for loop.

li-mingbao: Yeah I somehow want to allow dynamic indices to be used for a load operation, otherwise it would not be possible to define a for loop.

Would it make sense to check if the index value is a constant value? I guess constant values would all be illegal since they were extracted already or out of bounds. The remaining dynamic indices are then checked with the current map.

burgholzer: Had a brief conversation with Claude about this; specifically why the two loads cannnot be folded.

> The root cause is the **`[MemRead, MemWrite]` memory effect annotation on the `!qc.qubit` argument** of `qc.h` (and all other gate ops).
> 
> Here is the precise chain of reasoning:
> 
> ### Why the load is not eliminated
> 
> MLIR's `memref` canonicalization (specifically the **load-after-load** redundancy elimination / CSE) can only fold two `memref.load` operations at the same index into one if there is **no intervening operation that may write to that memory location**.
> 
> Between the two loads in your program:
> ```mlir
> %0 = memref.load %alloc[%c0] : memref<1x!qc.qubit>
> qc.h %0 : !qc.qubit                                // <-- HERE
> %1 = memref.load %alloc[%c0] : memref<1x!qc.qubit>
> ```
> 
> `qc.h` is declared with:
> ```tablegen
> let arguments = (ins Arg<QubitType, "the target qubit", [MemRead, MemWrite]>:$qubit_in);
> ```
> 
> The `MemWrite` effect on `$qubit_in` tells MLIR's alias analysis that `qc.h` **writes to memory through the qubit reference**. Since `%0` was loaded from `%alloc[%c0]`, and `qc.h %0` is declared as writing memory via `%0`, the compiler must conservatively assume the underlying allocation `%alloc` may have been mutated. Therefore, the second `memref.load` cannot be considered redundant — it must re-read the (potentially changed) value.
> 
> ### Why `MemWrite` is there
> 
> This is by design for the reference-semantics model of the QC dialect. A `!qc.qubit` is a **mutable reference** to a quantum state. The `MemWrite` annotation reflects the fact that the gate modifies the qubit in place (the quantum state changes). Without it, alias analysis would be incorrect for ordering / dependency tracking of quantum operations.
> 
> ### What would need to change
> 
> If you want the two loads to be folded, you would need to teach the compiler that even though `qc.h` writes to the *quantum state* via the qubit reference, it does **not** change the *identity* of the qubit stored in the `memref` slot `%alloc[%c0]`. This is a semantic distinction that the generic `MemWrite` effect cannot express.
> 
> Concretely, you have two options:
> 
> 1. **Use a custom alias analysis / memory SSA pass** that understands that `qc.*` operations write through the qubit *value* (quantum state) but never change what is stored in the `memref` slot itself, so loads of the slot can still be forwarded.
> 
> 2. **Change the program representation** so qubits are not loaded from a `memref` between gates — e.g., load once, use `%0` for both gates, and avoid the redundant loads in the first place. This would be the idiomatic QC dialect usage:
>    ```mlir
>    %0 = memref.load %alloc[%c0] : memref<1x!qc.qubit>
>    qc.h %0 : !qc.qubit
>    qc.h %0 : !qc.qubit   // reuse %0, no second load needed
>    ```

So I would argue it's best to avoid the loads in the first place, as recommened above. At least if there are any operations happening inbetween

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
li-mingbao: In theory all operand types should already be converted to `!qc.qubit` or `memref` types but some types were not converted yet, so I also have to explicitly match them. 
In the end any `scf` operation that still uses a qubit operand of any kind is illegal in the `QC` dialect.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
li-mingbao: I added a test for the parser.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
li-mingbao: Actually it is not, since you can just check if the regionmaps of the operations have any qubits or memref.
I deleted this and adjusted the legality condition.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: I moved it there and removed it from the `scf` operations.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: I added the declaration to the header file.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
li-mingbao: Sorry for that! I should have reverted all accidental changes.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
li-mingbao: I added an else branch to this example here and added the missing `scf.yield`

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
li-mingbao: I adjusted the the description.
Let me know if this still requires changes.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
denialhaag: > Yeah I somehow want to allow dynamic indices to be used for a load operation, otherwise it would not be possible to define a for loop.
> 
> Would it make sense to check if the index value is a constant value? I guess constant values would all be illegal since they were extracted already or out of bounds. The remaining dynamic indices are then checked with the current map.

An alternative to this would be a region-based map, right? Might be a bit cleaner than checking whether an index is constant-like. This will still allow a qubit to be loaded twice if two different operands end up referring to the same index, but I don't see a way of catching that. 🤔

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
li-mingbao: I also applied the changes in `QCOProgramBuilder.h`.
Let me know if I missed something.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
denialhaag: Perfect!

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
denialhaag: Looks good now!

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
denialhaag: I like it!

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
denialhaag: No worries! Happens to me all the time when I forget a bracket somewhere and `prek` reformats all docstrings. 🤠

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
denialhaag: Looks good!

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
li-mingbao: I modified the tracking with a region-based map and check if the index is already used to load a qubit in the current region or in any outer regions.

I manually have to track the current nested regions via the `regionStacks` now since during the creation of the region it is still detached from the operation (as the operation is still being built).

denialhaag: Looks good to me!

One last comment on this: Maybe it makes sense to add a `@details` section stating that this method should (more or less) only be used when defining the body of `scfFor` (and maybe `scfWhile`?). That is correct, right? I cannot think of another scenario where this method would be useful. 🤔

denialhaag: One could even consider disallowing the method completely if the region stack is in the root region. 🤔

This would prevent a user from defining a constant and loading a qubit in the root module.

(I might be getting too hung up on this.)

li-mingbao: Sounds good to me. I added a `@details` section to explain that the same index cannot be used to load a qubit multiple times in the same region. Also, I limited it to any nested regions inside the main function.
I guess you can have a `scf.for` operation and another nested operation inside it (e.g. another scf operation or one of the modifiers) and load the qubit there. This technically loads the qubit inside the body of the inner operation which would still be valid.

There are still some ways to circumvent this but I dont know how to restrict every single possibility in an efficient way, especially since you have to do some canonicalizations before the operation is constructed.

denialhaag: Yeah, let's go with this for now! Thanks for your patience with me here! 🙂

burgholzer: Good discussion here! Keep in mind that in almost all cases, programs in practice will come into the compiler collection through other means than the builder (OpenQASM, `jeff`, Qiskit, QIR). These languages have their own rules for creating sane programs. And the rest then just has to be handled properly in the conversions and/or translations into our framework.

denialhaag: Yeah, that's a good point! I think we can be happy with this now and adapt later if necessary. 😌

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -174,6 +177,28 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   *
+   * @par Example:
+   * ```c++
+   * auto q0 = builder.memrefLoad(memref, index);
+   * ```
+   * ```mlir
+   * %q0 = memref.load %memref[%index] : memref<3x!qc.qubit>
+   * ```
+   */
+  Value memrefLoad(Value memref, Value index);
```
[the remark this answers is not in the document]
li-mingbao: This is not required anymore since we verify the used index via the regionStack map.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -712,7 +723,210 @@ struct ConvertQCOYieldOp final : OpConversionPattern<qco::YieldOp> {
   LogicalResult
   matchAndRewrite(qco::YieldOp op, OpAdaptor /*adaptor*/,
                   ConversionPatternRewriter& rewriter) const override {
-    rewriter.replaceOpWithNewOp<qc::YieldOp>(op);
+    if (llvm::isa<scf::IfOp>(op->getParentOp())) {
+      rewriter.replaceOpWithNewOp<scf::YieldOp>(op);
+    } else {
+      rewriter.replaceOpWithNewOp<qc::YieldOp>(op);
+    }
```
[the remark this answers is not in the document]
li-mingbao: this should never happen as the parent is always converted first.

li-mingbao:
Hey @burgholzer,
Just wanted to quickly ask when you have time for a review since it's been a while since I (or in this case @denialhaag) requested it.

burgholzer:
> Hey @burgholzer,
> 
> Just wanted to quickly ask when you have time for a review since it's been a while since I (or in this case @denialhaag) requested it.

Hey 👋🏻 
Sorry, this somehow fell through the crack a little bit. I'll try to get to it asap. Although this may mean early next week.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
```diff
@@ -710,7 +720,210 @@ struct ConvertQCOYieldOp final : OpConversionPattern<qco::YieldOp> {
...
+    // Erase default block
+    rewriter.eraseBlock(&newFor.getRegion().front());
+
+    // Inline the region and replace the blockarguments
+    inlineRegion(op.getRegion(), newFor.getRegion(), 1,
+                 adaptor.getInitArgs().size(), adaptor.getInitArgs(), rewriter);
+
+    rewriter.replaceOp(op, adaptor.getInitArgs());
+
+    return success();
```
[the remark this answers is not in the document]
li-mingbao: It can be assumed that the operations return the results in the same order as the input. Therefore the replacement should be correct.

denialhaag:
@li-mingbao, I wanted to bring this PR up to speed after the merge of #1673. Unfortunately, my activity triggered CodeRabbit, which then found issues unrelated to the rebase. 😕

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -971,13 +1198,343 @@ struct ConvertQCYieldOp final : StatefulOpConversionPattern<qc::YieldOp> {
     auto& state = getState();
     auto* operation = op.getOperation();
     auto& frame = currentModifierFrame(state);
-    auto targets = resolveMappedQubits(state, operation, frame.yieldOrder);
+    auto targets = resolveMappedQubits(state, operation, frame.qcQubits);
     rewriter.replaceOpWithNewOp<qco::YieldOp>(op, targets);
     popModifierFrame(state);
```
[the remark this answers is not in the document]
li-mingbao: This should be fine, since replaceOpWithNewOp erases the old operation.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -971,13 +1198,343 @@ struct ConvertQCYieldOp final : StatefulOpConversionPattern<qc::YieldOp> {
...
+    assignMappedTensors(state, op.getOperation(), registerMap,
+                        newForOp.getResults().take_front(numRegisters));
+    assignMappedQubits(state, op.getOperation(), qubitMap,
+                       newForOp->getResults().take_back(numQubits));
+
+    // Move the the contents from the old block into the new block
+    auto& srcBlock = op.getRegion().front();
+    auto& dstBlock = newForOp.getRegion().front();
+    dstBlock.getOperations().splice(dstBlock.end(), srcBlock.getOperations());
+    rewriter.replaceAllUsesWith(op.getInductionVar(),
```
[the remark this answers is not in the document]
li-mingbao: This does not create a scf.yield since this operation has iter args. If this is the case, the trivial scf.yield is not added.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
```diff
@@ -971,13 +1198,343 @@ struct ConvertQCYieldOp final : StatefulOpConversionPattern<qc::YieldOp> {
...
+                                newForOp.getInductionVar());
```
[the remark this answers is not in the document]
li-mingbao: In the QC dialect scf operations do not have qubit values as iter_args when they are used in the body. So scf operations that are converted never have additional iter args when converted from QC to QCO.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -924,6 +946,115 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
+   *   auto q0 = builder.memrefLoad(memref, iv);
+   *   builder.h(q0);
+   * });
+   * ```
+   * ```mlir
+   * scf.for %iv = %lb to %ub step %step {
+   *   %q0 = memref.load %memref[%iv] : memref<3x!qc.qubit>
+   *   qc.h %q0 : !qc.qubit
+   *   scf.yield
+   * }
```
burgholzer, line 973: scf.for has an implicit terminator trait, so I would have guessed that the `scf.yield` is never printed in valid IR. Am I wrong here?
This also holds for the other builder examples down below.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -924,6 +946,115 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
    */
   QCProgramBuilder& dealloc(Value qubit);
 
+  //===--------------------------------------------------------------------===//
+  // SCF operations
+  //===--------------------------------------------------------------------===//
+
+  /**
+   * @brief Construct a scf.for operation
```
burgholzer: I am pretty sure that the correct spelling is "... _an_ scf.<...>" because the pronounciation of `scf` starts with an "es" sound. Applies everywhere across the entire PR.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -91,90 +90,151 @@ parseTargetAliasing(OpAsmParser& parser, Region& region,
...
+  SmallVector<OpAsmParser::Argument> thenArgs;
+  SmallVector<OpAsmParser::Argument> elseRegionArgs;
+  SmallVector<OpAsmParser::UnresolvedOperand> thenOperands;
+  SmallVector<OpAsmParser::UnresolvedOperand> elseOperands;
 
-static ParseResult
-parseIfOpAliasing(OpAsmParser& parser, Region& thenRegion, Region& elseRegion,
-                  SmallVectorImpl<OpAsmParser::UnresolvedOperand>& operands) {
-  // Parse the qubits keyword
   if (parser.parseKeyword("qubits")) {
```
burgholzer: Is "qubits" still the right keyword here? Especially since this may now also contain qtensor values?

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -942,15 +1149,21 @@ OwningOpRef<ModuleOp> QCOProgramBuilder::finalize() {
...
-  for (auto& [tensor, tensorInfo] : validTensors) {
+  for (auto& [tensor, tensorInfo] : sortedTensors) {
     auto currentTensor = tensor;
     // Filter out qubits belonging to this tensor
     for (auto& [qubit, qubitInfo] : qubitsByRegister[tensorInfo.regId]) {
-      auto indexValue = constantFromScalar(*this, getLoc(), qubitInfo.regIndex);
-      currentTensor =
-          qtensor::InsertOp::create(*this, qubit, currentTensor, indexValue)
-              .getResult();
+      currentTensor = qtensor::InsertOp::create(*this, qubit, currentTensor,
```
burgholzer, line 1153: There was some sorting of the tensor values here before. I typically avoided such sorting in the conversions as it is not strictly necessary for equivalence and costs time. However, I noticed that one of the tests would sporadically (very rarely fail) specifically the `QCOTest.IfOpParser` test. Could it be that the module verifier is not yet flexible enough to catch these differences? If so, can we extend it so that this is supported?

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
burgholzer: I have to say that this also confuses me a little bit. Are we missing something elsewhere that makes this necessary?

burgholzer (changes requested):
Thanks @li-mingbao for all the work on this PR! 🙏🏼 
I finally found some time to get to this.
I pushed a couple of commits that cleaned up a couple of smaller details and improved small bits and pieces.
I also have a handful of smaller comments left, which you will find inline. These should be quick to address and I believe this should be ready to go in afterwards.

I have the feeling that the conversions between QC and QCO have accumulated quite some tracking code and data structures that feel a bit much. And I am wondering whether some of the respective code could be simplified. However, these are all concerns for a potential follow-up PR.
Let's get this one in sooner rather than later 😄

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
li-mingbao: I think I added the sorting since the IfOpParser test would fail sometimes because it creates 2 different registers and the deallocation order of the registers was not deterministic. So the `areModulesEquivalentWithPermutations` function would fail.

mlir/lib/Conversion/QCOToQC/QCOToQC.cpp
li-mingbao: The conversion pass only marks operations changes and does not apply the changes directly. The legality checker only checks the original operand types of the operations and does not use the Adaptor for it. The exception is when the block arguments are replaced, as their types are directly replaced and change the operand types of the operation that uses them.
So in the example below, the scf.while operation would have to check for `qco.qubit` types and the scf.yield operation has to check for `qc.qubit` types.
```
 %1 = scf.while (%arg0 = %0) : (!qco.qubit) -> !qco.qubit {
      %2 = qco.h %arg0 : !qco.qubit -> !qco.qubit
      %qubit_out, %result = qco.measure %2 : !qco.qubit
      scf.condition(%result) %qubit_out : !qco.qubit
    } do {
    ^bb0(%arg0: !qco.qubit):
      scf.yield %arg0 : !qco.qubit
    }
```

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
li-mingbao: You are right for scf.for and scf.if. I removed these instances from the docstrings. 
In the scf.while operation the yield operation is still printed.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
li-mingbao: I should have corrected all instances of this.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
li-mingbao: I am not sure what a good word it would be so I decided to replace them with a generic `args` keyword to represent both qubits and qtensors here. This is inline with other scf operations like `iter_args` for scf.for.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -133,6 +137,27 @@ QCProgramBuilder::ClassicalRegister::operator[](const int64_t index) const {
...
+        "Qubit cannot be loaded in the main function region");
+  }
+  for (Region* curr : regionStack) {
+    if (loadedQubits[curr][memref].contains(index)) {
+      llvm::reportFatalUsageError("Qubit already loaded in enclosing region");
+    }
+  }
+
+  auto loadOp = memref::LoadOp::create(*this, memref, index);
+  loadedQubits[region][memref].insert(index);
```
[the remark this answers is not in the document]
li-mingbao: This is ok for now, as this should not happen anyway. The load operation is typically only used for the induction variable of the scf for operation. Other cases can be ignored for now.

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -251,19 +345,21 @@ Value QCOProgramBuilder::qtensorFromElements(ValueRange elements) {
...
   auto qubit = extractOp.getResult();
   auto outTensor = extractOp.getOutTensor();
 
   validateTensorValue(tensor);
   const auto regId = validTensors[tensor].regId;
 
-  validQubits.try_emplace(qubit, QubitInfo{.regId = regId, .regIndex = index});
+  validQubits.try_emplace(qubit,
+                          QubitInfo{.regId = regId, .regIndex = indexValue});
   updateTensorTracking(tensor, outTensor);
```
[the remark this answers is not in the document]
li-mingbao: This is okay, as it is expected that values are inserted before they are yielded by the scf operations. So they should never be used in the finalize() function.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
burgholzer: `args` seems like a good fit 👍🏼

burgholzer (approved):
Thanks for all the hard work here @li-mingbao 🙏🏼 
This LGTM now 🎉 

I'll get this in once #1694 is in. 🚀 
Edit: Screw that. CodeRabbit requested further changes there. Getting this in first.

PR #1673, opened 2026-04-28
🎨 Clean up `llvm` includes

denialhaag:
## Description

This PR cleans up `llvm` includes by relying on [`mlir/Support/LLVM.h`](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Support/LLVM.h) as much as possible in header files.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Add conversions between `jeff` and QCO ([#1479], [#1548], [#1565], [#1637]) ([**@denialhaag**])
 - ✨ Add a `place-and-route` pass for mapping circuits to architectures with restricted topologies ([#1537], [#1547], [#1568], [#1581], [#1583], [#1588], [#1664]) ([**@MatthiasReumann**], [**@burgholzer**])
 - ✨ Add initial infrastructure for new QC and QCO MLIR dialects
-  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1542], [#1548], [#1550], [#1554], [#1567], [#1569], [#1570], [#1572], [#1573], [#1580], [#1602], [#1620], [#1623], [#1624], [#1626], [#1627], [#1635])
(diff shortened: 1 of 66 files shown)


denialhaag:
This is a bit of a full-circle moment, but I believe it's a good idea to never include `mlir/Support/LLVM.h` in `.cpp` files, since it's (almost exclusively) a bunch of forward declarations. Removing `mlir/Support/LLVM.h` has allowed the include cleaner to reliably detect missing headers. I also believe the change improves IDE support, but I haven't experimented with this enough so far.

For header files, the forward declarations in `mlir/Support/LLVM.h` are very convenient.

This change forces us to explicitly specify the `llvm` namespace for all LLVM classes (e.g., `llvm::SmallVector`). I have honestly come to like this because it clearly conveys whether something is an MLIR or a LLVM concept. In addition, this helps with consistency because some useful classes (e.g., `llvm::StringMap`) are not forward-declared in `mlir/Support/LLVM.h` and cannot be used without the explicit `llvm` namespace.

That said, these are just my thoughts, and I'm open to differing opinions.

burgholzer:
I just pushed two commits that take this a step further.
The whole goal was to get rid of the annoying `mlir::` and `llvm::` prefixes wherever possible. So this is what I tried in those two commits.
I am pretty sure clang-tidy will scream at me for that. However, it makes the code so much more comfortable to read in my opinion and has led to an overall reduction of more than 100 LoC.
Let's see how hard it is to convince the include-cleaner that all of this is intended.

denialhaag:
> The whole goal was to get rid of the annoying `mlir::` and `llvm::` prefixes wherever possible.

As described in https://github.com/munich-quantum-toolkit/core/pull/1673#issuecomment-4340357848, I had consciously changed my mind about the goal, but I'm perfectly willing to go with this.

burgholzer:
> > The whole goal was to get rid of the annoying `mlir::` and `llvm::` prefixes wherever possible.
> 
> As described in [#1673 (comment)](https://github.com/munich-quantum-toolkit/core/pull/1673#issuecomment-4340357848), I had consciously changed my mind about the goal, but I'm perfectly willing to go with this.

Yeah, saw that. 
But I believe the original post from the LLVM discourse that triggered this also had the conclusion that `llvm::` can be avoided practically everywhere in MLIR code and that this is the intended way.
Additionally, it 
- saves quite some lines of code
- disambiguates how to write these constructs (the definitive answer is "without namespace qualifiers wherever possible")
- does not really seem to upset the include-cleaner.

So I would consider the current state of this PR a win in all aspects.

burgholzer (approved):
I am all happy with this in its current state.
@denialhaag feel free to take another look and proceed with the merge if you are happy as well.
Maybe you could go through some of the open PRs that will likely have conflicts after this is merged.

Issue #1363, opened 2025-12-05
✨ `BASEURL` support through QDMI authentication


marcelwa:
### Problem Statement

#1355 introduces support for QDMI authentication via FoMaC. However, it doesn't currently offer a way to provide the `BASEURL` parameter because this is a `DEVICE_SESSION_PARAMETER`, not a `SESSION_PARAMETER`.

### Proposed Solution

I don't know a good solution for this problem yet, but it will be closely related to solutions for #1361 and #1362.

burgholzer:
This relates to `https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/161`, although that issue is a bit broader in scope.

For the matter here, there currently is no way to provide a `BASEURL` for a device from the client side.
We could simply add such a parameter to the `QDMI_SESSION_PARAMETER` enum as a workaround until we figure out a proper way for clients to forward device session parameters through a driver.

However, this does more broadly relate to https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/215 as well and the question of whether the client interface is actually necessary. This has come up multiple times so far, and the more I think about it, the more often I think it would be beneficial to simply drop and build up on the QDMI Device interface.
This would still mean that a driver exists that loads the name-shifted shared or static libraries, but that it wouldn't have to go through a C API, again, before being wrapped in another C++ abstraction.
Dropping the interface would also allow us to directly set parameters for the individual devices, because we would be the ones in control of how the sessions are created with the devices.
At the same time, we keep thinking about this "stack-as-a-device" concept. And all the discussions so far have led more and more towards the insight, that a separate interface will probably be required for exposing stacks. Such an interface would be highly inspired by QDMI, particularly the client interface supposedly, but would be separate.

Concrete proposal:
- Drop QDMI client interface
- Merge Driver and FoMaC as C++ abstraction over QDMI
- Expose the resulting C++ API to Python (almost identical to now)
- Work on a separate Stack-Interface (QSMI) inspired by the client interface (once worked out, it would be up to the individual drivers to implement logic for dynamically loading libraries implementing the new interface)

@ystade your input would also be valuable here.

Issue #1358, opened 2025-12-04
✨ Refactor FoMaC into a `qdmi` namespace


marcelwa:
### Problem Statement

The term FoMaC doesn't immediately convey the notion of QDMI. Furthermore, using QDMI through MQT Core is currently only possible with an (incomplete) C++ abstraction and respective Python bindings. All of this creates friction points for developers and users. 

### Proposed Solution

Let us refactor the use of QDMI in MQT Core to provide the following modules in a dedicated `qdmi` namespace:
- The QDMI C API
- A QDMI C++ Wrapper
- Python bindings of the C++ Wrapper

Issue #1226, opened 2025-09-26
✨ Dynamically add NA QDMI Devices from JSON


ystade:
### Problem Statement

The current solution offers an elegant way to transform a JSON representation of an NA device into a corresponding QDMI device as a shared library. As mentioned in [this thread](https://github.com/munich-quantum-toolkit/core/pull/1223#discussion_r2381960949), it would also be convinient, if one would be able to add a NA QDMI device based on a JSON specification during runtime without having to compile the JSON into a shared library that can then be loaded dynammically.

### Proposed Solution

Implement a generic QDMI device that expects a JSON file to parse during runtime. It the reports the corresponding figures stored in the JSON specification. The biggest challenge is to pass the location of the JSON file to the QDMI device.

PR #1687, opened 2026-05-06
:sparkles: Integrate QDMI Devices

### LINKED ISSUE #1082: ✨ MLIR - Feeding Architecture Information into MLIR Mapping Conversion
### Problem Statement

The conversion from `MQTOpt` to `MQTMem` requires information about the hardware. This must somehow be fed to the conversion pass.

### Proposed Solution

The end-user should be able to define the architecture as a parameter to the compilation process, in particular, the architecture should not be part of the quantum program.

MatthiasReumann:
🤖 *AI text below* 🤖

## Summary

This is the final integration slice of the compiler-target series. It:

- snapshots a configured circuit-model QDMI device as an immutable
  `mlir::CompilerTarget`, including names, topology, coherence data, operation
  capabilities, and available calibration;
- exposes the same target compilation workflow through C++, Python, and
  source-build `mqt-cc`;
- adds `mqt-cc` device listing, explicit registry configuration, and stable-ID
  device selection without introducing another driver or target abstraction;
- keeps generic optimization before target-native synthesis and final
  conformance verification.

The bridge validates that QDMI operation support fits the compiler's homogeneous
bidirectional target model. Directional operations must report both ordered
orientations on every supported pair, while operand-symmetric operations such as
CZ may report each undirected edge once. Devices without topology and operation
site restrictions retain the all-to-all convention used by DDSIM. Neutral-atom
zone models are rejected with a focused diagnostic.

## Design

`CompilerTarget` remains owned by the MLIR compiler and independent of FoMaC,
QDMI, and the legacy CoreIR target. The optional adapter is the only library
linking FoMaC to the compiler target. It snapshots all data eagerly, so
compilation does not retain a device or session handle.

Python exposes `CompilerTarget.from_device`, `QCOProgram.compile_for_target`,
and `compile_program(..., target=...)`. The C++ adapter and `mqt-cc` workflow are
source-build interfaces; this PR does not create a partial installed MLIR SDK
boundary or change wheel shipment for `mqt-cc`.

## Validation

- 223 compiler unit tests
- 10 focused adapter and `mqt-cc` CTests
- 29 focused Python MLIR tests
- provider-disabled compiler build
- authoritative stub regeneration
- changed-source clang-tidy
- strict warning-free documentation build
- complete repository lint and `git diff --check`
- independent exact-head design and correctness review

Closes #1082


MatthiasReumann:
@burgholzer (cc: @ystade)

This pull request now implements a very basic QDMI V1 integration into our compiler collection. Before continuing, I want to discuss the following issues.

####  Optionally enable QDMI
As of now, the current implementation tightly couples the `mqt-cc` with QDMI. Wouldn't it be nice to have QDMI as a compile-option in CMake? Something like `ENABLE_QDMI_INTEGRATION` (analogous to LLVM's `ENABLE_ASSERTIONS`) which adds the CLI Options, and generally the QDMI logic via preprocessor definitions? If so, we could still compile the `mqt-cc` without QDMI. 

#### Extract Nested Classes 

In order to augment the FoMaC device classes, I think it would be nice to extract them outside the `Session` class. Any opinions on that? It is kind of awkward to always write out `fomac::Session::Device`. Especially if we ever want to have specializations such as `SuperconductingDevice`, etc. 

####  Trailing Return Types (Nitpick, sorry!)

I am really not that sure if the FoMaC classes benefit from the benefits of trailing return types. The `auto` in the following examples just feels somewhat wrong. Modern & fancy but wrong. 
```cpp
// fomac/FoMaC.cpp
auto getParametersNum(const std::vector<Site>& sites = {},
                       const std::vector<double>& params = {}) const -> size_t;
// qdmi/sc/Device.hpp
auto sessionFree(MQT_SC_QDMI_Device_Session session) -> void;
```

I've taken the liberty to implement some of the above FoMaC changes this morning (+ some improvements using modern C++ concepts) in [`FoMaC.hpp`](https://github.com/munich-quantum-toolkit/core/blob/enh/refactor-fomac/include/mqt-core/fomac/FoMaC.hpp) and [`FoMaC.cpp`](https://github.com/munich-quantum-toolkit/core/blob/enh/refactor-fomac/src/fomac/FoMaC.cpp) for you to better understand what I meant. 

---

Any comments appreciated! Many thanks 🙏

burgholzer:
> @burgholzer (cc: @ystade)
> 
> This pull request now implements a very basic QDMI V1 integration into our compiler collection. Before continuing, I want to discuss the following issues.
> 
> #### Optionally enable QDMI
> As of now, the current implementation tightly couples the `mqt-cc` with QDMI. Wouldn't it be nice to have QDMI as a compile-option in CMake? Something like `ENABLE_QDMI_INTEGRATION` (analogous to LLVM's `ENABLE_ASSERTIONS`) which adds the CLI Options, and generally the QDMI logic via preprocessor definitions? If so, we could still compile the `mqt-cc` without QDMI.

Hm. I'd argue that if we view QDMI as the primary way to add architecture information to `mqt-cc`, it should be(come) an essential part of it. Would we ever compile `mqt-cc` without QDMI support? In what kind of circumstances would that yield a benefit?

> #### Extract Nested Classes
> In order to augment the FoMaC device classes, I think it would be nice to extract them outside the `Session` class. Any opinions on that? It is kind of awkward to always write out `fomac::Session::Device`. Especially if we ever want to have specializations such as `SuperconductingDevice`, etc.

We have an open tracking issue going in that direction https://github.com/munich-quantum-toolkit/core/issues/1358. See also https://github.com/munich-quantum-toolkit/core/issues/1363#issuecomment-3617031698
Your proposed refactoring sounds fine as a start.

> #### Trailing Return Types (Nitpick, sorry!)
> I am really not that sure if the FoMaC classes benefit from the benefits of trailing return types. The `auto` in the following examples just feels somewhat wrong. Modern & fancy but wrong.
> 
> ```c++
> // fomac/FoMaC.cpp
> auto getParametersNum(const std::vector<Site>& sites = {},
>                        const std::vector<double>& params = {}) const -> size_t;
> // qdmi/sc/Device.hpp
> auto sessionFree(MQT_SC_QDMI_Device_Session session) -> void;
> ```

Feel free to get rid of them. I am personally not the biggest fan of these either way.

> I've taken the liberty to implement some of the above FoMaC changes this morning (+ some improvements using modern C++ concepts) in [`FoMaC.hpp`](https://github.com/munich-quantum-toolkit/core/blob/enh/refactor-fomac/include/mqt-core/fomac/FoMaC.hpp) and [`FoMaC.cpp`](https://github.com/munich-quantum-toolkit/core/blob/enh/refactor-fomac/src/fomac/FoMaC.cpp) for you to better understand what I meant.

Thanks, I'll take a look! 👍🏼 

> Any comments appreciated! Many thanks 🙏

MatthiasReumann:
> In what kind of circumstances would that yield a benefit?

I was kind of thinking about IBM Benchpress. Not sure if we could still use QDMI here that easily.  

>We have an open tracking issue going in that direction https://github.com/munich-quantum-toolkit/core/issues/1358. See also https://github.com/munich-quantum-toolkit/core/issues/1363#issuecomment-3617031698

Oh. Great! I'll have a look at these and try to incorporate them in the refactoring. 

> Thanks, I'll take a look! 

I've also opened #1849 to ease the comparison to the old implementation.

burgholzer:
> > In what kind of circumstances would that yield a benefit?
> 
> I was kind of thinking about IBM Benchpress. Not sure if we could still use QDMI here that easily.

I was hoping that we could take the input from benchpress and wrap it in QDMI at runtime.
Somewhat relates to https://github.com/munich-quantum-toolkit/core/issues/1226 but on a broader scale

> > We have an open tracking issue going in that direction #1358. See also [#1363 (comment)](https://github.com/munich-quantum-toolkit/core/issues/1363#issuecomment-3617031698)
> 
> Oh. Great! I'll have a look at these and try to incorporate them in the refactoring.

Take those with a grain of salt though. It's been a while that they have been written and some circumstances might have changed since then. Still good to have a look though.

ystade:
A little late, but since I was tagged, I also wanted to comment:

### Trailing return types

I guess I am responsible for them. I love them because when you read `auto` as `func` or so, the function signature becomes very similar to other languages. For me, it feels natural to have the return type at the end of the function. Anyway, I agree that this depends very much on taste, and I will go along with the majority.

### FoMaC Concept

I guess everything has already been said here. I can just repeat here that in the future we want to directly build on QDMI and want to get rid of FoMaC or rather replace that by a C++ abstraction of QDMI.

MatthiasReumann:
@coderabbitai full review

MatthiasReumann:
@coderabbitai full review

mlir/lib/Support/SuperconductingDevice.cpp
```diff
@@ -0,0 +1,78 @@
...
+size_t SuperconductingDevice::maxDegree() const {
+  return coupling_.getMaxDegree();
+}
+
+Graph SuperconductingDevice::getCouplingGraph(
+    const std::shared_ptr<fomac::Device>& device) {
+  const auto sites = device->getSites();
+  const auto siteCoupling = device->getCouplingMap();
+  assert(siteCoupling.has_value() &&
+         "expected QDMI device with a coupling map");
```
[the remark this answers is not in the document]
MatthiasReumann: To avoid the use of exceptions, we will assume (which is explicitly stated in the description) that the device is a superconducting device which must have a coupling map.

MatthiasReumann:
@burgholzer @ystade 

I think this one is ready to discuss. 

- I've added a `SuperconductingDevice`[^1] class that acts as an _adapter_ for QDMI, coupling maps, and whatever is to follow (V2 at some point, I guess). 
- The CLI is defined as follows (prefix: `qdmi`) 
```cpp
--qdmi-list-devices         // List all available QDMI devices. 
--qdmi-device=iqm-radiance  // Specify the device used for hardware-specific passes. 
--qdmi-config=./qdmi.json   // Specify session config, additional devices and their respective configs. 
```
- The specified device is loaded and stored in the `QuantumCompilerConfig`. The QDMI device is then used in the hardware-specific phase to construct the `SuperconductingDevice`. I assume superconducting devices only. Future work might implement a "dispatcher" to create the correct pass pipeline for the respective device's constraints. 

---

#### CLI Example 

```console
matthias:~$ cat ~/Downloads/qdmi.json
{
    "devices": [
        {
            "libName": "/Users/matthias/Documents/projects/core/build/src/qdmi/devices/sc/libmqt-core-qdmi-sc-device.dylib",
            "prefix": "MQT_SC",
            "deviceConfig": {
                "baseUrl": "...",
                "token": "...",
                "authFile": "...",
                "authUrl": "...",
                "username": "...",
                "password": "...",
                "custom1": "...",
                "custom2": "...",
                "custom3": "...",
                "custom4": "...",
                "custom5": "..."
            }
        }
    ]
}
matthias:~$ mqt-cc --qdmi-list-devices --qdmi-config ~/Downloads/qdmi.json
[2026-07-09 14:03:58.473] [info] [Driver.cpp:193] Device session parameter BASE URL not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter TOKEN not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter AUTH FILE not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter AUTH URL not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter USERNAME not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter PASSWORD not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter CUSTOM1 not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter CUSTOM2 not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter CUSTOM3 not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter CUSTOM4 not supported by device (skipped)
[2026-07-09 14:03:58.474] [info] [Driver.cpp:193] Device session parameter CUSTOM5 not supported by device (skipped)
Available QDMI devices:
	MQT NA Default QDMI Device // Statically loaded.
	MQT SC Default QDMI Device // Statically loaded.
	MQT Core DDSIM QDMI Device // Statically loaded.
	MQT SC Default QDMI Device // Dynamically loaded.
```


[^1]: I consciously avoided `SCDevice` because of "super-computing"; and generally to avoid acronyms as much as possible :)

burgholzer:
I have not yet had the chance to look through everything in detail here yet, but I have one general high-level observation already:

A lot of the QDMI-related functionality is built "outside" of QDMI here (building on top of FoMaC; relying on `addDevice..`).
I would argue that we need to take this one step further. These enhancements should directly go into the QDMI Driver so that the driver itself reads the config file and uses that to make its devices available.
The build should likely write a config file as part of the build, where one needs to be a bit careful about the library paths being embedded in the config because one needs to ensure that the config file also works in an install tree of MQT Core (such as in the Python bindings). We already perform some kind of shenanigans around that with the builtin devices and how they are hardcoded in the driver. The config file should replace the existing approach.
More generally, I believe that quite some of the functionality in the `QDMI.h` header should likely be functionality that is added to the FoMaC and/or Driver classes directly.

@MatthiasReumann does that make sense? Can you work with that?

MatthiasReumann:
> @burgholzer: I would argue that we need to take this one step further. 

I'll try to summarize my thoughts on this. Let me know, if the following is valid. 

> These enhancements should directly go into the QDMI Driver so that the driver itself reads the config file and uses that to make its devices available.

Generally understood.

> The build should likely write a config file as part of the build, where one needs to be a bit careful about the library paths being embedded in the config because one needs to ensure that the config file also works in an install tree of MQT Core (such as in the Python bindings). We already perform some kind of shenanigans around that with the builtin devices and how they are hardcoded in the driver. The config file should replace the existing approach.

I wonder: Is there any valid reason to link devices statically in the `mqt-core` driver? Wouldn't it be quite elegant if the `mqt-core` driver takes a config file at runtime (via function call or CLI) and loads the specified devices at boot-time?[^1] I guess this would avoid any of the build shenanigans described by you. However, we could still supply a `builtin-devices.json` (maybe even auto-generate it) to the QDMI devices generated by the generator:

```
<cmake world>
generateIQMSpark()
generateIQMRadiance()
generateIBMHeron()

generateJSONConfig()

<c++ world>
Driver::get().init("config.json");
```

Because isn't the purpose of the `mqt-core` driver to be "plug-and-play"?

> More generally, I believe that quite some of the functionality in the QDMI.h header should likely be functionality that is added to the FoMaC and/or Driver classes directly.

Understood. 

---

Currently, what really confuses me - and I just realized - is that the `Driver.cpp` implements the QDMI client functions. So the driver *is* in fact the client in `mqt-core`. 

[^1]: which is actually already happening (just via CMake magic).

burgholzer:
> > The build should likely write a config file as part of the build, where one needs to be a bit careful about the library paths being embedded in the config because one needs to ensure that the config file also works in an install tree of MQT Core (such as in the Python bindings). We already perform some kind of shenanigans around that with the builtin devices and how they are hardcoded in the driver. The config file should replace the existing approach.
> 
> I wonder: Is there any valid reason to link devices statically in the `mqt-core` driver? Wouldn't it be quite elegant if the `mqt-core` driver takes a config file at runtime (via function call or CLI) and loads the specified devices at boot-time?[1](#user-content-fn-1-bedd083af414b77c3daee32d716ccc6c) I guess this would avoid any of the build shenanigans described by you. However, we could still supply a `builtin-devices.json` (maybe even auto-generate it) to the QDMI devices generated by the generator:
> 
> ```
> <cmake world>
> generateIQMSpark()
> generateIQMRadiance()
> generateIBMHeron()
> 
> generateJSONConfig()
> 
> <c++ world>
> Driver::get().init("config.json");
> ```
> 
> Because isn't the purpose of the `mqt-core` driver to be "plug-and-play"?

IIRC we no longer statically link the QDMI devices that come with MQT Core.
As you point out in the footnote above, we do some CMake magic to populate the devices at driver bootup.
My recommendation from above was precisely to replace that magic with configuration file logic.
The example driver in the main QDMI repository already has some boilerplate code for this https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/cc94309662a8cdc265821f642707d44e34378274/examples/driver/qdmi_example_driver.cpp#L233-L311
This includes default search locations as well as the option to provide the config path via an environment variable.

I would argue that the following things should be possible:
- a default config file should be (auto-)generated at build/install time that contains all the devices built into MQT Core. This should be placed somewhere where the driver looks for it (in the build folder or the install tree; it should not be written to the source tree as it is a generated artifact)
- providing an additional configuration file to be considered. This can happen either through an environment variable, or potentially through a parameter of the driver constructor (does that make sense?). Both of these options allow configuration at boot time. Additionally, we likely want configurability at runtime (similar to `addDynamicDeviceLibrary`) either programmatically or through the CLI.
- We could potentially use a MQT Core QDMI CLI that provides prefixes and library paths for the builtin devices so that other libraries that, e.g., want to use the DDSIM device could programmatically get that information. Ideally that CLI would be in C++ and somewhat exposed to Python so that it can be used with and without Python.
 
> > More generally, I believe that quite some of the functionality in the QDMI.h header should likely be functionality that is added to the FoMaC and/or Driver classes directly.
> 
> Understood.
> 
> Currently, what really confuses me - and I just realized - is that the `Driver.cpp` implements the QDMI client functions. So the driver _is_ in fact the client in `mqt-core`.

I believe I get where your confusion is coming from. This is a consequence of the initial design of QDMI v1. QDMI v1 defines the device and the client interface. HW Providers implement the (prefixed) device interface; clients can use the (non-prefixed) client interface with all the devices they have access to. The driver mediates between both interfaces.
The FoMaC library builds on top of the _client_ interface that is exposed by the driver; except for the `addDynamicDeviceLibrary` method of the driver, which is not officially part of the QDMI client interface.

We have a rather lengthy exposition of that topic here: https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/215 that was written almost a year ago. So far, no one dared to make the move and eliminate the client interface in MQT Core to combine the FoMaC and Driver libraries (also getting rid of the FoMaC name 😅)
Maybe you can give that issue a read and make up your mind based on that and the discussion here.

MatthiasReumann:
@burgholzer  Okay, I think I got a plan. 

**Step 0: Device Generation**

1. Create a few more `sc` device JSON files. Maybe real-hardware architectures such as those from IQM.
2. Setup CMake to create `.dylib`s for each of those JSON files. 
3. Setup CMake to create a `builtin-devices.json` somewhere in the `build` folder, referencing those `.dylib`s. Store the path to this JSON as a CMake variable. 

**Step 1: Driver Boot-up**

1. Keep the current CMake magic; but parse the `builtin-devices.json` in the driver's constructor and load the devices described in it. 
2. Add a CMake Configuration Variable, explicitly stating a list of additional device JSON files. Parse and load these in the same step as 1.1.

**Step 2: Driver Runtime**

1. The interface that parses and loads the devices used in Step 1, should also be usable for the MLIR CLI. Essentially something like `Driver::get().load(<json_file>)`. 

----

Essentially, the JSON config files will be the canonical way to load devices into `mqt-core` QDMI driver.

```
┌──────────────────────────────────────────────────┐                                                       
│  CMake                                           │                                                       
│                                                  │                                                       
│      ┌───────────┐     ┌─────────┐               │                                                       
│      │ Generator │    ┌│         │               │                                                       
│      └───┬─┬─┬───┘  ┌─││.dylib   │               │                                                       
│          │ │ └──────│ ││         │               │  ┌──────────────────┐                                 
│          │ └────────│ │└───┬─────┘               │  │ Outside World    │                                 
│          └──────────│ └─┬──┼───┘                 │  │                  │                                 
│                     └─┬─┼──┼─┘                   │  │┌────────────────┐│                                 
│                       │ │  │                     │  ││lrz-devices.json││                                 
│                    ┌──▼─▼──▼──────────────┐      │  │└────────────────┘│                                 
│                    │ builtin-devices.json │      │  │┌────────────────┐│                                 
│                    └──────────────────────┘      │  ││aws-devices.json││                                 
└───────────────────────────────────┬──────────────┘  │└────────────────┘│                                 
                                    │                 └──────┬───────────┘                                 
┌───────────────────────────────────┼──────────────┐         │                                             
│ CPP                               │                        │                                             
│              ┌────────────────────▼──────┐       │         │                                             
│              │Driver::Driver()           │       │         │                                             
│              │  for each: load(...)      ◄───────┼─────────┘                                             
│              └───────────────────────────┘       │  -DQDMI_DEVICE_PATHS=lrz-devices.json;aws-devices.json
│                                                  │                                                       
│              ┌───────────────────────────┐       │                                                       
│              │Driver::get().load(...)    │       │                                                       
│              └────────────▲──────────────┘       │                                                       
│                           │                      │                                                       
└───────────────────────────┼──────────────────────┘                                                       
                            │     --qdmi-configs=iqm-devices.json                                                          
                         ┌──┴──┐                                                                           
                         │MLIR │                                                                           
                         └─────┘                                                                           
```

I'll leave it at that for now for this pull request. 

Any objections? @ystade Any opinions?

burgholzer:
> @burgholzer  Okay, I think I got a plan. 
> 
> 
> 
> **Step 0: Device Generation**
> 
> 
> 
> 1. Create a few more `sc` device JSON files. Maybe real-hardware architectures such as those from IQM.
> 
> 2. Setup CMake to create `.dylib`s for each of those JSON files. 
> 
> 3. Setup CMake to create a `builtin-devices.json` somewhere in the `build` folder, referencing those `.dylib`s. Store the path to this JSON as a CMake variable. 
> 
> 
> 
> **Step 1: Driver Boot-up**
> 
> 
> 
> 1. Keep the current CMake magic; but parse the `builtin-devices.json` in the driver's constructor and load the devices described in it. 
> 
> 2. Add a CMake Configuration Variable, explicitly stating a list of additional device JSON files. Parse and load these in the same step as 1.1.
> 
> 
> 
> **Step 2: Driver Runtime**
> 
> 
> 
> 1. The interface that parses and loads the devices used in Step 1, should also be usable for the MLIR CLI. Essentially something like `Driver::get().load(<json_file>)`. 
> 
> 
> 
> ----
> 
> 
> 
> Essentially, the JSON config files will be the canonical way to load devices into `mqt-core` QDMI driver.
> 
> 
> 
> ```
> 
> ┌──────────────────────────────────────────────────┐                                                       
> 
> │  CMake                                           │                                                       
> 
> │                                                  │                                                       
> 
> │      ┌───────────┐     ┌─────────┐               │                                                       
> 
> │      │ Generator │    ┌│         │               │                                                       
> 
> │      └───┬─┬─┬───┘  ┌─││.dylib   │               │                                                       
> 
> │          │ │ └──────│ ││         │               │  ┌──────────────────┐                                 
> 
> │          │ └────────│ │└───┬─────┘               │  │ Outside World    │                                 
> 
> │          └──────────│ └─┬──┼───┘                 │  │                  │                                 
> 
> │                     └─┬─┼──┼─┘                   │  │┌────────────────┐│                                 
> 
> │                       │ │  │                     │  ││lrz-devices.json││                                 
> 
> │                    ┌──▼─▼──▼──────────────┐      │  │└────────────────┘│                                 
> 
> │                    │ builtin-devices.json │      │  │┌────────────────┐│                                 
> 
> │                    └──────────────────────┘      │  ││aws-devices.json││                                 
> 
> └───────────────────────────────────┬──────────────┘  │└────────────────┘│                                 
> 
>                                     │                 └──────┬───────────┘                                 
> 
> ┌───────────────────────────────────┼──────────────┐         │                                             
> 
> │ CPP                               │                        │                                             
> 
> │              ┌────────────────────▼──────┐       │         │                                             
> 
> │              │Driver::Driver()           │       │         │                                             
> 
> │              │  for each: load(...)      ◄───────┼─────────┘                                             
> 
> │              └───────────────────────────┘       │  -DQDMI_DEVICE_PATHS=lrz-devices.json;aws-devices.json
> 
> │                                                  │                                                       
> 
> │              ┌───────────────────────────┐       │                                                       
> 
> │              │Driver::get().load(...)    │       │                                                       
> 
> │              └────────────▲──────────────┘       │                                                       
> 
> │                           │                      │                                                       
> 
> └───────────────────────────┼──────────────────────┘                                                       
> 
>                             │     --qdmi-configs=iqm-devices.json                                                          
> 
>                          ┌──┴──┐                                                                           
> 
>                          │MLIR │                                                                           
> 
>                          └─────┘                                                                           
> 
> ```
> 
> 
> 
> I'll leave it at that for now for this pull request. 
> 
> 
> 
> Any objections? @ystade Any opinions? 

In principle, this plan sounds solid.
Let me try to collect my thoughts a little bit:
- I'd like to avoid creating too many shared libraries that are bundled with MQT Core because each one increases our binary sizes. Especially in case of the SC device, these will just be repetitive and produce artificial bloat. Instead, I'd look into making the SC device configurable at session initialization time. We could use a Custom Session parameter to describe the path to a SC JSON config file describing the architecture. If this parameter is not set, the default one that comes with MQT Core is chosen. But this also allows to create generic backends on demand by providing different config files. I am kind of looking for the equivalent of Qiskit's GenericBackendV2, but through QDMI. In QDMI v2 this would be a provider that takes a single config file with multiple architecture and makes all of them available through a single shared library. In v1, we need to hack this together a little bit.

- I think it would be really beneficial to not only allow a configure-time CMake option but some kind of search mechanism for config files. Quite a bit of tooling out there has good precedents for this kind of logic (consider uv, ruff, or ty as state of the art here). Can't we add config file detection and precedence rules based on these tools? The respective logic is not really special for those tools; nor does it need to be special for our purposes. This also includes defining a proper schema for these config files.

- I believe the proposal from above does not yet address the problem that we also need to provide a config file in the install tree that needs to resolve to the right paths and not to the build dir. Not rocket science, but needs to be taken care of.

- I really like the idea of using the config file as the single source of truth for QDMI device registration/loading. Exposing this at various levels makes a lot of sense to me.

Overall, this is definitely the right direction. Happy to discuss further 🙂

MatthiasReumann:
@coderabbitai full review

mlir/lib/Support/CMakeLists.txt
```diff
@@ -6,14 +6,16 @@
...
 # Licensed under the MIT License
 
+file(GLOB_RECURSE SUPPORT_CPP "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
+
 add_mlir_library(
   MLIRSupportMQT
-  IRVerification.cpp
-  Passes.cpp
-  PrettyPrinting.cpp
+  ${SUPPORT_CPP}
```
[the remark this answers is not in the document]
MatthiasReumann: @burgholzer If this works as described, we might want to add this everywhere.

ystade:
@MatthiasReumann I skimmed through the conversation here, and to me, everything makes a lot of sense. I do not have additional comments that go beyond Lukas' ones right now. Let me know whether you would like to have feedback on any particular point.

cmake/ExternalDependencies.cmake
```diff
@@ -131,6 +131,9 @@ endif()
 
 # Make all declared dependencies available.
 FetchContent_MakeAvailable(${FETCH_PACKAGES})
+if(TARGET qdmi)
+  set_property(TARGET qdmi PROPERTY SYSTEM ON)
+endif()
```
burgholzer:
```suggestion
```
No need to keep this.

bindings/fomac/fomac.cpp
```diff
@@ -113,6 +116,153 @@ template <typename Query>
...
+
+  auto targetSite = nb::class_<fomac::TargetSite>(
+      m, "TargetSite", "A target site and its coherence data.");
+  targetSite
+      .def(nb::init<std::size_t, std::optional<uint64_t>,
+                    std::optional<uint64_t>>(),
+           "site_id"_a, nb::kw_only(), "t1"_a = std::nullopt,
+           "t2"_a = std::nullopt)
+      .def_prop_ro("id", &fomac::TargetSite::id,
+                   "The provider-defined site ID.")
```
burgholzer: QDMI distinguishes between a site index (or ID) and site name. The id is mandatory and needs to be unique for the session. The name is optional, but helps to identify the site. It likely makes sense to also optionally store the name here and use it, e.g., as part of the QIR emission to determine proper labels for the recorded qubit outcomes. Even if it is not really used anywhere, it is available from QDMI and should be included in the snapshot.

docs/mlir/index.md
```diff
@@ -25,6 +25,43 @@ These dialects define various canonicalization and transformation passes that
...
+operation as native. This makes topology-only and synthesis-only targets
+possible without separate compiler configuration. An operation set may also be
+partial: compilation succeeds when the input already conforms or the target
+contains the single- and two-qubit primitives required to lower the input.
+
+The combined target pipeline does not hide the constituent passes. C++ pass
+tooling can construct `createMappingPass(target, options)` directly, while
+two-qubit native synthesis remains available through
+`QCOProgram::fuseTwoQubitUnitaryRuns`. This keeps mapping and synthesis
+independently benchmarkable without adding another architecture API.
```
burgholzer: This feels like an iteration artifact describing the way we arrived at the solution. Reword the docs additions here so they reflect meaningful documentation for the newly added feature.

docs/mlir/python_compiler_collection.md
```diff
@@ -143,6 +143,70 @@ The {code}`qco_pipeline` argument replaces the default QCO optimization
...
+    output=OutputFormat.QCO_OPTIMIZED,
+    target=target,
+)
+```
+
+The QCO pipeline runs before target compilation. For lifetime-disjoint qubits,
+`qco_pipeline="reuse-qubits,mqt-qco-default"` can therefore reduce the logical
+qubit demand before placement, routing, and native synthesis. Use
+`mqt-qubit-reuse` instead of `reuse-qubits` when measurement lifting and
+classical-control replacement are desired as preparation steps.
```
burgholzer: Anyone that wants to use qubit reuse will want to run the combined pipeline with `mqt-qubit-reuse`. Reword accordingly.

docs/mlir/python_compiler_collection.md
```diff
@@ -143,6 +143,70 @@ The {code}`qco_pipeline` argument replaces the default QCO optimization
...
+
+Target mapping accepts scalar and tensor-backed dynamic allocations in the entry
+function body. Allocations nested in control flow are not supported because
+their placement requires lifetime-aware resource planning.
```
burgholzer: Feels like an iteration artifact --> reword the entire section.
Also holds for the paragraphs below.

docs/qdmi/sc_device.md
```diff
@@ -74,3 +74,20 @@ are rejected.
...
+
+MQT Core also ships `json/sc/iqm-garnet.json` and `json/sc/iqm-emerald.json`.
+They contain the 20-qubit Garnet and 54-qubit Emerald topologies, respectively,
+and expose the compiler-facing IQM operation set `r`, `cz`, and `measure`. Their
+calibration snapshots contain the available per-qubit T1 and T2 coherence times
+and per-locus operation fidelities from the default calibration sets retrieved
+on 2 August 2026. Coherence times are rounded to the nearest nanosecond and
+stored as raw values with a `us` unit and a `0.001` scale factor. IQM does not
+expose operation durations through this interface, so those values remain
+absent. The snapshots contain no service credentials or calibration identifiers.
```
burgholzer:
```suggestion
absent.
```
Iteration artifact

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -165,20 +165,20 @@ def QuantumLoopUnroll
...
   let dependentDialects = ["mlir::qco::QCODialect"];
-  let summary = "This pass ensures that a program meets the connectivity "
-                "constraints of a target architecture.";
+  let summary = "Place and route a program for a target architecture.";
   let description = [{
     This pass maps the dynamically allocated qubits in a quantum program to the static qubits of a target architecture.
-    The pass performs both placement and routing of the qubits to ensure that all two-qubit operations in the program
-    can be executed on the target architecture.
+    It enforces both the coupling graph and, when present, the target's ordered operation loci. The input must already
+    use target-native operations; target pipeline construction runs native synthesis before this pass.
```
burgholzer: This is not true in general. The only pre-condition for the placement and routing pass is that the circuit has been decomposed to 1- and 2-qubit gates. synthesis of the two-qubit gates to the native gate set is typically performed _after_ routing, because the routing will introduce SWAP gates that need to be decomposed as well.
Typically, some kind of gate optimization is performed before the placement and routing, to shrink the circuits as much as possible before the routing takes place. 
Reword and refactor the design accordingly.

mlir/lib/Compiler/Programs.cpp
```diff
@@ -163,7 +164,15 @@ template <class PopulatePasses>
...
-  std::forward<PopulatePasses>(populatePasses)(pm);
+  if constexpr (std::is_same_v<
+                    std::invoke_result_t<PopulatePasses, OpPassManager&>,
+                    LogicalResult>) {
+    if (failed(std::forward<PopulatePasses>(populatePasses)(pm))) {
+      return failure();
+    }
+  } else {
+    std::forward<PopulatePasses>(populatePasses)(pm);
+  }
```
burgholzer: Feels overly complex. Is this necessary?

mlir/lib/Compiler/Targeting.cpp
```diff
@@ -0,0 +1,52 @@
...
+  if (target.hasCouplings() || target.hasOperations()) {
+    populateDecomposeMultiControlledPipeline(pm, 2);
+  }
+
+  if (target.hasOperations()) {
+    llvm::SmallVector<llvm::StringRef> operationNames;
+    operationNames.reserve(target.operations().size());
+    for (const auto& operation : target.operations()) {
+      operationNames.emplace_back(operation.name());
+    }
```
burgholzer: This feels like something the `Target` could expose on its own.
Given how the Compiler Target is tailored for MLIR and the Compiler Collection, it feels right to have it use MLIR datatypes wherever feasible.
To keep the FoMaC path MLIR-free, the CMake targets should be cleanly separated so that only the Compiler Target links against MLIR.
Then it should be treated as part of the Compiler Collection and the same rules and guidelines should apply.
It likely makes sense to move the `Target`-related files into the `mlir/` subtree to make this even cleaner/clearer.

mlir/lib/Compiler/Targeting.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    nativeGates.emplace(
+        qco::decomposition::NativeGateset::fromOperationNames(operationNames));
+    pm.addPass(qco::createFuseTwoQubitUnitaryRuns(*nativeGates));
```
burgholzer: As pointed out in the other comment, two-qubit optimization makes sense even when no native gates are present; just to reduce the complexity of the circuit.
For this first, (almost) target-independent step, it likely makes sense to still try to preserve any gate that is already native on the target, while also generally preserving individual single- and two-qubit gates that cannot really be optimized. This may be something to tackle in a follow-up.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -189,6 +185,148 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+    }
+    for (const auto& locus : operation.loci()) {
+      capability->loci.insert(
+          TargetLocus(locus.sites().begin(), locus.sites().end()));
+    }
+  }
+
+  [[nodiscard]] bool
+  supportsTargetOperation(const decomposition::OperationSemantics& semantics,
+                          const std::span<const size_t> locus) const {
```
burgholzer: These functions kind of feel duplicated with the functionality that is already in the Target class. Try to simplify as much as possible. These kind of queries feel like they would not be needed for the Mapping pass.

Holds all the way down to `supportsOperationAt`.
Make sure there is no unnecessary bloat here.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -189,6 +185,148 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+    return supportsTargetOperation(semantics, locus);
+  }
+
+  /**
+   * @brief Return whether a routing SWAP should use reversed operand order.
+   *
+   * An empty result means that the target cannot synthesize a SWAP on the edge.
+   */
+  [[nodiscard]] std::optional<bool>
+  swapUsesReversedOperands(const size_t source, const size_t target) const {
```
burgholzer: This entire function feels a bit overkill. Given the current set of entanglers, it is always possible to synthesize a SWAP, even when the architecture is directional.
In general, I'd think this PR should be simpler and just require the architecture to be symmetric. This is what the current code on `main` assumes and it feels pretty reasonable. We may extend this in the future to more generic architectures.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -241,16 +377,20 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
-        const auto [hw0, hw1] = layout.getHardwareIndices(prog0, prog1);
-        const size_t nswaps = device.distanceBetween(hw0, hw1) - 1;
-        costs += decay * static_cast<float>(nswaps);
+      for (const auto& request : window) {
+        if (request.programs.size() == 2 &&
+            !mapping.isExecutable(request, layout)) {
+          const auto [hw0, hw1] = layout.getHardwareIndices(
+              request.programs[0], request.programs[1]);
+          const auto distance = mapping.device->distanceBetween(hw0, hw1);
+          assert(distance && "expected a connected target");
```
burgholzer: This assumption should also be verified once, earlier and then removed here.
Also holds below.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -362,6 +519,29 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+                  << "cannot route an operation acting on "
+                  << unitary.getNumQubits()
+                  << " qubits; decompose it to one- and two-qubit operations "
+                     "first";
+              return WalkResult::interrupt();
+            })
+            .wasInterrupted()) {
+      signalPassFailure();
+      return;
+    }
```
burgholzer: This will walk the entire function once. Feels a bit overkill for something that we could also detect on the fly during routing where we iterate over the ops anyway.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -559,13 +759,29 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+        return WalkResult::advance();
+      }
+      op->emitError()
+          << "target mapping requires dynamic qubit allocations in the entry "
+             "function body";
+      return WalkResult::interrupt();
+    });
+    if (nestedAllocation.wasInterrupted()) {
+      return failure();
     }
```
burgholzer: This also feels like an avoidable walk of the program that could simply be included in the main routing method.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer, line 1: Because of the changes here, this PR (number) should likely be added to the respective changelog for the mapping as well

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/FuseTwoQubitUnitaryRuns.cpp
burgholzer, line 1: Because of the changes here, this PR (number) should likely be added to the respective changelog for the mapping as well

burgholzer (changes requested):
Alright. Iterated on this quite a bit with the help of gpt-5.6-sol.
The comments in this review are not intended for @MatthiasReumann, but for iteration work on the PR that I am planning.
Let's see how quickly the iteration converges.
But I think this is starting to look really clean.

burgholzer (approved):
Alright. This feels like it is ready 🚀 
Feels great to have QDMI integrated into the Compiler Collection 🥳

PR #1696, opened 2026-05-11
:sparkles: Add support for the IQM QDMI device

marcelwa:
## Description

This PR adds support for the [IQM QDMI device](https://github.com/iqm-finland/QDMI-on-IQM/).

The code in this PR was generated in part by GPT-5.4 via GitHub Copilot.

## Checklist


marcelwa:
@coderabbitai full review

marcelwa:
@burgholzer I would love your feedback on this as I'm currently unsure how best to test the new device in the CI. In the most ideal case, we would add a Resonance key as a secret and test directly through the real Server API.

burgholzer:
> @burgholzer I would love your feedback on this as I'm currently unsure how best to test the new device in the CI. In the most ideal case, we would add a Resonance key as a secret and test directly through the real Server API.

I'd generally hold off on this PR until a new release for the IQM QDMI device is out that adds the Qiskit extra (more specifically, the respective code). We should not be duplicating the backend code here, but just consume it from the `iqm-qdmi` library.

I agree that testing would be good through the real server API.
I'd be happy to add such a secret to the repository, but that only solves half of the problem. We also need to get that key to the workflows somehow and, at the top of my head, I am not really aware of a clean solution for that.
We could hardcode that potential secret into the workflows, but that doesn't sound too scalable to me (thinking about potential future additions).
Any ideas from your side on this?

marcelwa:
> I'd generally hold off on this PR until a new release for the IQM QDMI device is out that adds the Qiskit extra (more specifically, the respective code). We should not be duplicating the backend code here, but just consume it from the `iqm-qdmi` library.

Totally agree on this!

> We also need to get that key to the workflows somehow and, at the top of my head, I am not really aware of a clean solution for that. We could hardcode that potential secret into the workflows, but that doesn't sound too scalable to me (thinking about potential future additions). Any ideas from your side on this?

Ah right, because we're dealing with the reusable workflows here. There is probably no easy way to pass arbitrary secrets to them so they can be exposed as environment variables, right?

burgholzer:
> Ah right, because we're dealing with the reusable workflows here. There is probably no easy way to pass arbitrary secrets to them so they can be exposed as environment variables, right?

Not at the moment. But it also sounds kind of unsafe to me to add this. We'd have to brainstorm a bit (probably with LLM assistance) whether there is a potential solution that doesn't risk us exposing secrets. I'll see what I can find in that regard.

marcelwa:
I have decoupled the `iqm-qdmi` dependency from its `qiskit` extra in MQT Core's `iqm` extra. It now looks as follows:
```toml
iqm = [
  "iqm-qdmi>=1.1.0",
  "qiskit[qasm3-import]>=1.1.0",
]
```

When https://github.com/iqm-finland/QDMI-on-IQM/pull/78 gets merged, we could change this to:
```toml
iqm = [
  "iqm-qdmi[qiskit]>=1.1.0"
]
```
with clean resolution, if we felt like it. However, that would then not provide the `qasm3-import` extra. Makes me wonder if that shouldn't be added to QDMI-on-IQM to begin with.

burgholzer:
> I have decoupled the `iqm-qdmi` dependency from its `qiskit` extra in MQT Core's `iqm` extra. It now looks as follows:
> 
> ```toml
> iqm = [
>   "iqm-qdmi>=1.1.0",
>   "qiskit[qasm3-import]>=1.1.0",
> ]
> ```
> 
> When [iqm-finland/QDMI-on-IQM#78](https://github.com/iqm-finland/QDMI-on-IQM/pull/78) gets merged, we could change this to:
> 
> ```toml
> iqm = [
>   "iqm-qdmi[qiskit]>=1.1.0"
> ]
> ```
> 
> with clean resolution, if we felt like it. However, that would then not provide the `qasm3-import` extra. Makes me wonder if that shouldn't be added to QDMI-on-IQM to begin with.

I made the respective changes in `iqm-qdmi` and released as `v1.1.1`. However, that still led to resolution errors with the simplified IQM extra shown above.
Having played around with various options, in its current form there is hardly any benefit of the additions in this PR.
The PR simply reexports symbols from the `iqm-qdmi` library and does _not_ tightly integrate them into the systems available from the QDMI Provider per default.
At the same time, it adds a cyclic dependency problem we have to work with/around.
Additionally, the PR here does not add proper testing for the QDMI device; and if it would, this would complicate the testing setup of MQT Core, which would have to become conditional on the precense of IQM credentials.

Considering all of the above, I believe the only logical conclusion is that we should simply close this PR unmerged.
Anyone wanting to work with the IQM QDMI Device from Qiskit will inevitably use MQT Core to do so; and we test that path excessively in the `iqm-qdmi` repo.

@marcelwa Feel free to re-open if you think otherwise.

PR #1709, opened 2026-05-15
✨ Inject Coupling Map Into Mapping Pass

MatthiasReumann:
## Description

This pull request addresses the open issue of injecting a coupling map into the mapping pass. Contrary to #1687 a `AugmentedDevice` class is constructed from the coupling map instead of a QDMI device handle. 

## Changes 
- ✨Add a new constructor to the mapping pass.
- ♻️ Removes the dependency of the QC dialect for the mapping pass unit-test. 
- 🔥 Remove `Architecture.h` and `Architecture.cpp`

## Checklist


mlir/include/mlir/Dialect/QCO/Utils/Algorithms.h
```diff
@@ -0,0 +1,38 @@
...
+#include <utility>
+
+namespace mlir::qco {
+
+using Matrix = llvm::SmallVector<llvm::SmallVector<size_t, 0>, 0>;
+using Edges = llvm::DenseSet<std::pair<size_t, size_t>>;
+
+/**
+ * @brief Find all shortest paths between two nodes in a graph.
+ * @details Has a time and memory complexity of O(n^3).
```
burgholzer: correct me if I am wrong, but isn't the memory complexity only O(n^2)?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -62,34 +59,14 @@ using namespace mlir::qtensor;
...
-
 namespace {
 
 struct MappingPass : impl::MappingPassBase<MappingPass> {
 private:
   using IndexType = size_t;
   using IndexPairType = std::pair<IndexType, IndexType>;
-  using Window = SmallVector<IndexPairType>;
+  using Window = llvm::SmallVector<IndexPairType>;
+  using Neighbours = llvm::SmallVector<llvm::SmallVector<size_t, 4>>;
```
burgholzer: We generally decided to drop the `llvm::` prefixes wherever possible. This is definitely such a case. Also holds elsewhere throughout the file. If this is just because clang-tidy is complaining, then simply add the LLVM support header to the list of includes and it should be quiet.

burgholzer (changes requested):
LGTM besides two small details.
Before merging, I'd also prompt at least one full review from CodeRabbit. Probably worth waiting until you addressed the two points.

mlir/include/mlir/Dialect/QCO/Utils/Algorithms.h
MatthiasReumann: ah, yeah! good catch 👍🏻

The sentence previously was "has a time and memory complexity of O(n^3) and O(n^2), respectively". I stripped the memory complexity part because it returns the matrix instead of storing it in the `Architecture` class.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Thought it was the other way around 🫠 Should've followed that PR more closely...

MatthiasReumann:
@coderabbitai full review

mlir/unittests/Dialect/QCO/Transforms/Mapping/test_mapping.cpp
```diff
@@ -26,161 +28,182 @@
...
+      if (op.getNumQubits() > 1) {
+        const auto q0 = cast<TypedValue<QubitType>>(op.getInputQubit(0));
+        const auto q1 = cast<TypedValue<QubitType>>(op.getInputQubit(1));
+        const auto i0 = qubits.getIndex(q0);
+        const auto i1 = qubits.getIndex(q1);
+
+        if (!coupling.contains(std::make_pair(i0, i1))) {
+          return WalkResult::interrupt();
+        }
+      }
```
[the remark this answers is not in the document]
MatthiasReumann: I have made the assumption that only circuits with zero, one, and two-qubit gates are used explicitly by adding a `assert`.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: We should probably document that somewhere in our development guide in the long run. Or in the AGENTS.md file so that AI-based reviews and code generation will get this.

burgholzer (approved):
Nice. LGTM 👍🏻

PR #1716, opened 2026-05-18
♻️ Implement Explicit Assumption Checks for Mapping Pass

MatthiasReumann:
## Description

This pull request implements verification logic, which validates if an input IR satisfies the mapping pass' assumptions. 

The assumptions are as follows:
1. There is only one monolithic `qtensor`. 
2. There are no `qco.alloc` operations. 
3. There is an "extraction" as well as an "insertion" phase. 

These assumptions hold for the `.qasm` programs generated by MQT-Bench. 

## Checklist


MatthiasReumann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Utils/Utils.cpp
```diff
@@ -0,0 +1,39 @@
...
+    const auto strAttr = dyn_cast<StringAttr>(attr);
+    return strAttr && strAttr.getValue() == ENTRY_POINT_LABEL;
+  };
+
+  for (auto func : op.getOps<func::FuncOp>()) {
+    const auto passthrough = func->getAttrOfType<ArrayAttr>(PASSTHROUGH_LABEL);
+    if (passthrough && llvm::any_of(passthrough, isEntry)) {
+      return func;
+    }
+  }
```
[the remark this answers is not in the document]
MatthiasReumann: We consider an IR with multiple entry point functions as invalid.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -383,120 +383,223 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+              })
+              .Default([&](Operation* op) {
+                report_fatal_error("unknown op in def-use chain: " +
+                                   op->getName().getStringRef());
+                return WalkResult::interrupt();
+              });
+
+      if (res.wasInterrupted()) {
+        return failure();
+      }
```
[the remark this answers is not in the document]
MatthiasReumann: The default case is considered a "programmer" error - we have missed something in the implementation. Hence, the report_fatal_error. We use the same pattern for the drivers and the `WireIterator`. Otherwise, good input bunny 🐇

mlir/lib/Dialect/QCO/Utils/Utils.cpp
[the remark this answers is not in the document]
MatthiasReumann: I have added a respective comment in the description of the function.

mlir/lib/Dialect/QCO/Utils/Utils.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+  return nullptr;
+}
```
burgholzer: This is more so a general observation also partly inspired by discussions over at FullStaQD. It doesn't actually make much sense for us to denote the entrypoint behind a `passthrough` array attribute. This is a constraint that bubbled up from the QIR conversion. But there is absolutely no reason why we could not define our own entry point signature and convert that signature during the QIR conversion.

See https://github.com/FullStaQD/compiler/blob/e92a921b2fdf6174a86c1f75fc79a06ac9ad82cd/mlir/include/qcc/Constants.h#L16-L17

Nothing to be changed in this PR, but a rather direct follow-up could make sense. At the very least, an issue tracking this.

mlir/include/mlir/Dialect/QCO/Utils/Utils.h
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @details
+ * Searches for the function marked with the "entry_point" attribute in
+ * the passthrough attributes. If multiple functions are marked, returns the
+ * first one encountered.
+ *
+ * @param op The module operation to search in.
+ * @returns the entry point function, or nullptr if not found.
+ */
+func::FuncOp getEntryPoint(ModuleOp op);
```
burgholzer: As you probably know by now, I am not the biggest fan of `Utils` files and like to avoid them as much as possible. 
Is there any immediate other use of this function besides its current use in the mapping pass?
If not, I'd rather place it there as a private function and avoid all the boilerplate around the utils file.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -383,120 +383,222 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
     }
 
-    // 1.1 Handle tensors.
-    for (auto op : make_early_inc_range(func.getOps<qtensor::DeallocOp>())) {
-      rewriter.eraseOp(op);
+    const auto tensors = func.getOps<qtensor::AllocOp>();
+    if (range_size(tensors) != 1) {
+      func.emitError() << "must contain a single qtensor.alloc operation";
+      return failure();
     }
```
burgholzer, line 474: This limitation is a bit arbitrary, right? Is there any technical obstacle to allowing multiple registers?
Totally fine to really nail down the current limitations here. Some of these feel like they should be fairly easy to lift though and we should probably do that in follow-ups.

burgholzer (changes requested):
LGTM in principle. Just a few small comments on this one.

mlir/include/mlir/Dialect/QCO/Utils/Utils.h
MatthiasReumann: Yes. 

This utility comes in quite handy in the unit tests. Maybe we can put it to the `QCODialect.h` file? But where to put the definition then?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Not entirely arbitrary: When translating the QASM programs from mqt-bench to the QC dialect, exactly one tensor is created. 

But you are definitely right. Supporting more than one tensor should be fairly simple to implement.

mlir/include/mlir/Dialect/QCO/Utils/Utils.h
burgholzer: Technically, the definition could simply be inline. It's not too complex of a function and does not really pull in many headers, does it?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: That cannot be true for all the benchmarks. I am fairly sure that https://github.com/munich-quantum-toolkit/bench/blob/main/src/mqt/bench/benchmarks/qpeexact.py will create two tensors.

In a similar line of thought, supporting individual allocations of qubits should be moderately easy to support, as long as these allocations are not in entirely inconvenient places (SCF constructs for example), right?

mlir/include/mlir/Dialect/QCO/Utils/Utils.h
MatthiasReumann: Sounds good 👍🏻

burgholzer:
Okay, one of the changes in my commit prompted the clang-tidy warning. I don't quite know which one yet. It's definitely a false-positive though as actually implementing it leads to test failures.
@MatthiasReumann maybe you could take it from here.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Mhm. True. 

> In a similar line of thought, supporting individual allocations of qubits should be moderately easy to support, as long as these allocations are not in entirely inconvenient places (SCF constructs for example), right?

Yes. As long as the `alloc-phase -> computation-phase -> dealloc-phase` assumption is satisfied, this should be easy. Essentially, we have to ensure that the DAG has `n < N_{arch}` entry qubit nodes. (... the recent qubit-reuse discussion we had)

MatthiasReumann: > That cannot be true for all the benchmarks. I am fairly sure that https://github.com/munich-quantum-toolkit/bench/blob/main/src/mqt/bench/benchmarks/qpeexact.py will create two tensors.

Maybe let's start to lift this assumption then next to actually support all mqt-bench programs.

burgholzer: Sounds like a plan! 👍🏼

burgholzer (approved):
LGTM 👍🏼 (feel free to merge once everything is green)

PR #1718, opened 2026-05-19
✨ Add `QuantumLoopUnroll` pass

MatthiasReumann:
## Description

This pull request adds a `QuantumLoopUnroll` pass using the built-in unrolling functionality provided by MLIR. Thanks to these utilities we can unroll nested `scf.for` operations, without having to implement the logic ourselves. I would also argue that a separate pass is the cleaner option to using `conversion + affine` or the `transform` dialect. 

Moreover, this pull request is part of a larger series of PRs to follow with the goal of supporting structured control flow. 

Some QIR programs don't support backwards branching. By executing this pass before lowering to QIR, a user can make his/her quantum program (utilizing `scf.for` operations) QIR compliant. 

## Changes 
- ✨ Add `QuantumLoopUnroll` pass and associated unit tests

## Potential Follow-Ups

As a consequence of unrolling loops, a lot of (extract-insert) pairs are created which potentially could be merged. I don't think there is canonicalization logic achieving this - although this would be quite useful. 

For example, with such a canonicalization pattern, we could already map the GHZ-using-a-loop program to superconducting architectures.

## Checklist


MatthiasReumann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Transforms/Optimizations/QuantumLoopUnroll.cpp
```diff
@@ -0,0 +1,112 @@
...
+ * @param loop The loop to test.
+ * @returns true, if the loop is a quantum loop.
+ */
+static bool isQuantumLoop(scf::ForOp loop) {
+  return llvm::any_of(loop.getInitArgs(), [](Value arg) {
+    const auto type = arg.getType();
+    return isa<QubitType>(type) ||
+           (isa<RankedTensorType>(type) &&
+            isa<QubitType>(cast<RankedTensorType>(type).getElementType()));
+  });
```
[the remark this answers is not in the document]
MatthiasReumann: The QCO dialect enforces linear typing. Because for now we assume that no qubits are allocated inside a loop, a valid quantum loop in the QCO dialect has either qubits or a qubit tensor as init argument.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -40,6 +40,22 @@ def MergeSingleQubitRotationGates
...
 
+def QuantumLoopUnroll
+    : InterfacePass<"quantum-loop-unroll", "FunctionOpInterface"> {
+  let dependentDialects = ["mlir::qco::QCODialect", "mlir::scf::SCFDialect"];
+  let summary = "Unroll bounded quantum loops";
+  let description = [{
+    This pass unrolls `scf.for` operations which contain quantum operations. Classical loops are not unrolled.
+
+    Because the pass uses the `mlir::loopUnrollByFactor` utility internally, the same assumptions and restrictions hold for both.
+    See the MLIR documentation for more information.
```
burgholzer: Can we link to some stable reference here so that people directly know where to look?

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -40,6 +40,22 @@ def MergeSingleQubitRotationGates
   }];
 }
 
+def QuantumLoopUnroll
+    : InterfacePass<"quantum-loop-unroll", "FunctionOpInterface"> {
```
burgholzer, line 44: Just an idea that came to mind, but wouldn't this be better called "qco-loop-unroll"? It is a bit closer to the `affine-loop-unroll`, although technically it is not the `qco` operations being unrolled. So maybe it is not the best idea. Still wanted to bring it up because "quantum" is so generic.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -40,6 +40,22 @@ def MergeSingleQubitRotationGates
...
+  let dependentDialects = ["mlir::qco::QCODialect", "mlir::scf::SCFDialect"];
+  let summary = "Unroll bounded quantum loops";
```
burgholzer: Just playing devil's advocate here: Is it clear what a "quantum loop" is?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add `quantum-loop-unroll` pass that unrolls for-loop operations ([#1718]) ([**@MatthiasReumann**])
```
burgholzer: This could be a bit more precise (it does not unroll all for loops for example). It's also missing an "a" in the beginning of the sentence to be consistent with the remaining entries.

burgholzer (changes requested):
Nice stuff 👍🏼 
I pushed a couple of cleanups.
I think particularly the `walk` change is interesting because it simplified the implementation a bit and might apply to further parts in the code base.

I have a couple of comments left, mostly on wording and naming. Otherwise, this should be good to go 😌

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
MatthiasReumann: Yeah. The naming is still a bit off here. Particularly, also the "quantum-loop" 🧐

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
MatthiasReumann: Probably not. Maybe we should just spell it out: `Unroll bounded loops that contain quantum operations`

MatthiasReumann: Maybe we can then also keep the "quantum-loop-unroll" pass name...?

burgholzer: Hm. I like 
> Unroll bounded loops containing quantum operations

as a summary. That's clear enough.
let's use that and stick with the OG name.
`qco-loop-unroll` isnt fully true either because the pass also considers qtensor ops. The generic "quantum" is probably best afterall.

MatthiasReumann: Done ☑️

burgholzer (approved):
LGTM now 🚀

PR #1730, opened 2026-05-22
✨ Add `TensorIterator`

MatthiasReumann:
## Description

This pull request adds the bi-directional `TensorIterator` class (the analog of the `WireIterator` for tensors). Cherry-picked from #1728 to ease the reviewing effort. 

## Checklist


MatthiasReumann:
@coderabbitai full review

MatthiasReumann:
@burgholzer Any ideas what to do with the codecov coverage tests? As far as I know, there is no way to force the execution of the `llvm::report_internal_error` functions. The default cases exist to catch developer errors in the future.

burgholzer:
> @burgholzer Any ideas what to do with the codecov coverage tests? As far as I know, there is no way to force the execution of the `llvm::report_internal_error` functions. The default cases exist to catch developer errors in the future.

I am entirely fine with fatal error lines not being covered.
That's one of the reasons why the codecov check is not blocking in CI.
If it were easy in C++ to mock functions, this would be a different story; but it isn't.

From time to time, I am wondering whether it would be better to simply mark places like these as `llvm::unreachable()` because the compiler could probably do a better job at optimizing the program, but that would probably hurt debuggability in the long run.

mlir/lib/Dialect/QTensor/Utils/TensorIterator.cpp
```diff
@@ -0,0 +1,118 @@
...
+        .Case<scf::ForOp>([&](scf::ForOp op) {
+          tensor_ = cast<TypedValue<RankedTensorType>>(
+              op.getTiedLoopResult(&*(tensor_.use_begin())));
+        })
+        .Default([&](Operation* op) {
+          report_fatal_error("unknown op in def-use chain: " +
+                             op->getName().getStringRef());
+        });
+  }
+}
```
burgholzer, line 77: What about a `qco::IfOp` that uses a tensor value? Same below

burgholzer (approved):
Just pushed another cleanup for the namespace qualifiers (my inner monk just gets triggered by the greyed-out code in the IDE 👀)

Just one open question here.

denialhaag:
Note that the linter will fail until https://github.com/munich-quantum-toolkit/workflows/pull/377 is merged and released.

mlir/lib/Dialect/QTensor/Utils/TensorIterator.cpp
```diff
@@ -0,0 +1,132 @@
...
+  if (!(isa<AllocOp, FromElementsOp>(op_))) {
+    TypeSwitch<Operation*>(op_)
+        .Case<ExtractOp>([&](ExtractOp op) { tensor_ = op.getOutTensor(); })
+        .Case<InsertOp>([&](InsertOp op) { tensor_ = op.getResult(); })
+        .Case<scf::ForOp>([&](scf::ForOp op) {
+          tensor_ = cast<TypedValue<RankedTensorType>>(
+              op.getTiedLoopResult(&*(tensor_.use_begin())));
+        })
+        .Case<qco::IfOp>([&](qco::IfOp op) {
+          tensor_ = cast<TypedValue<RankedTensorType>>(op.getResults().front());
```
burgholzer: This feels a bit brittle. Is there a guarantee that the tensor currently being iterated is the first result?
What about if's with multiple results (multiple tensors or mixed qubits and tensors)?

MatthiasReumann: That is actually a really good point. I don't know why but I thought only one tensor was possible.

MatthiasReumann: Done ☑️

mlir/lib/Dialect/QTensor/Utils/TensorIterator.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <cassert>
+#include <iterator>
+
+namespace mlir::qtensor {
+TypedValue<RankedTensorType> TensorIterator::tensor() const {
+  if (op_ == nullptr) {
+    return tensor_;
+  }
+
+  // A tensor deallocation doesn't have an OpResult.
```
burgholzer: The comment here is not fully up to date

mlir/lib/Dialect/QTensor/Utils/TensorIterator.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  // Find the output from the input tensor SSA value.
+  if (!(isa<AllocOp, FromElementsOp>(op_))) {
+    TypeSwitch<Operation*>(op_)
+        .Case<ExtractOp>([&](ExtractOp op) { tensor_ = op.getOutTensor(); })
+        .Case<InsertOp>([&](InsertOp op) { tensor_ = op.getResult(); })
+        .Case<scf::ForOp>([&](scf::ForOp op) {
+          tensor_ = cast<TypedValue<RankedTensorType>>(
+              op.getTiedLoopResult(&*(tensor_.use_begin())));
```
burgholzer, line 64: Probably something for a follow-up, but it seems useful to define a similar method for our qcoIf. Resolves a bit of boilerplate and kind of makes sense to me.
Scf::if probably has such a method, right?
Same below for the reverse direction.

mlir/lib/Dialect/QTensor/Utils/TensorIterator.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  // If the op is a nullptr, the tensor value is a block argument and thus the
+  // beginning of the tensor's life-chain.
+  if (op_ == nullptr) {
+    return;
+  }
+
+  // For deallocations and scf::YieldOps, tensor_ is an OpOperand.
+  // Hence, only get the def-op.
```
burgholzer: The comment here is not fully up to date

burgholzer (changes requested):
This LGTM now! 🙏
There's two outdated comments and a suggestion for a useful helper method.
Then this can go in.

mlir/lib/Dialect/QTensor/Utils/TensorIterator.cpp
MatthiasReumann: yes, definitely 👍🏻 thought about the same!

MatthiasReumann: Created #1746 ☑️

burgholzer (approved):

burgholzer (approved):

MatthiasReumann:
@burgholzer I think the workflow got stuck here. I unfortunately can't stop it and rerun it.

burgholzer:
> @burgholzer I think the workflow got stuck here. I unfortunately can't stop it and rerun it. 

Yeah.. GitHub Actions had an outage yesterday..
Could you push another commit onto this PR? Or rewrite the history an force push? Something that will trigger another CI run.

PR #1661, opened 2026-04-23
✨ Add `-absorb-swaps` pass

jmoosburger:
## Description

 - add new qco pass as warmup task (in cooperation with @MatthiasReumann)
 - remove leading swap-gates in a circuit
 - reorder static qubits instead

## Checklist

- [ ] The pull request only contains commits that are focused and relevant to this change.
- [ ] I have added appropriate tests that cover the new/changed functionality.
- [ ] I have updated the documentation to reflect these changes.
- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [ ] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.
- [ ] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [ ] ~~I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).~~
- [ ] ~~AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.~~
- [ ] ~~I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.~~

PR #1750, opened 2026-05-28
✨ Add `absorb-swaps` pass

jmoosburger:
## Description

- add new qco pass as warmup task (in cooperation with @MatthiasReumann)
- remove leading swap-gates in a circuit
- reorder static qubits instead

## Checklist

- [x] ~I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).~
- [x] ~AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.~
- [x] ~I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.~


jmoosburger:
@MatthiasReumann

works fine now :muscle: 
nice setup with the CMakePresets by the way... :)

i have another conceptional question though: which of the following circuits is correct?
```
circuit 120

q0 ----X---------- q1
       |
q1 ----X----X----- q2
            |
q2 ---------X----- q0
```
or:
```
circuit 201

q0 ----X---------- q2
       |
q1 ----X----X----- q0
            |
q2 ---------X----- q1
```

in my opinion its the second circuit 201, but the implementation we discussed produces the first result.
i just wanted to make sure that i'm correct in my assumption before i fix it :)

jmoosburger:
the linter produces warnings on the one hand, but failes because of a http-error 403 (unauthorized) on the other hand. see here: https://github.com/munich-quantum-toolkit/core/actions/runs/26597942119/job/78373906843?pr=1750

is there anything i can do about it?

denialhaag:
> the linter produces warnings on the one hand, but failes because of a http-error 403 (unauthorized) on the other hand. see here: https://github.com/munich-quantum-toolkit/core/actions/runs/26597942119/job/78373906843?pr=1750
>
> is there anything i can do about it?

Unfortunately, the action cannot post comments on PRs coming from a fork. 😕 

That said, the report can also be found in the summary of the run: https://github.com/munich-quantum-toolkit/core/actions/runs/26597942119/attempts/1#summary-78373906843

MatthiasReumann:
@jmoosburger 

> i have another conceptional question though: which of the following circuits is correct?

The best way to visualize this is to apply the SWAPs sequentially:

```
circuit 120

q0 ----X-- q1 ------------ q1
       |
q1 ----X-- q0 --X-- q2 --- q2
                |
q2 -------------X-- q0 --- q0
```

jmoosburger:
> > the linter produces warnings on the one hand, but failes because of a http-error 403 (unauthorized) on the other hand. see here: https://github.com/munich-quantum-toolkit/core/actions/runs/26597942119/job/78373906843?pr=1750
> > is there anything i can do about it?
> 
> Unfortunately, the action cannot post comments on PRs coming from a fork. 😕
> 
> That said, the report can also be found in the summary of the run: https://github.com/munich-quantum-toolkit/core/actions/runs/26597942119/attempts/1#summary-78373906843

so if i resolve all the warnings the lint check will succeed, as it does not post a comment?

MatthiasReumann:
> so if i resolve all the warnings the lint check will succeed, as it does not post a comment?

If you resolve all warnings the `CI / Lint / 🚨 Lint (pull request)` workflow will succeed. The pull request will look like [this one](https://github.com/munich-quantum-toolkit/core/pull/1696).

jmoosburger:
> > so if i resolve all the warnings the lint check will succeed, as it does not post a comment?
> 
> If you resolve all warnings the `CI / Lint / 🚨 Lint (pull request)` workflow will succeed. The pull request will look like [this one](https://github.com/munich-quantum-toolkit/core/pull/1696).

Which way is recommended?
the _contribution.md_ suggests `clang-tidy` and the section _Development Setup_ in _installation.md_ suggests `nox -s lint`...

denialhaag:
> Which way is recommended? the _contribution.md_ suggests `clang-tidy` and the section _Development Setup_ in _installation.md_ suggests `nox -s lint`...

`nox -s lint` lints everything but C++ files, so you would need to run `clang-tidy` locally. That said, I usually just use the error messages from the CI because `clang-tidy` might report different errors locally. 😬

jmoosburger:
> `nox -s lint` lints everything but C++ files, so you would need to run `clang-tidy` locally. That said, I usually just use the error messages from the CI because `clang-tidy` might report different errors locally. 😬

ahh, that explains a thing :+1: 
does the PR get squashed on merge then, right?

denialhaag:
> ahh, that explains a thing 👍 does the PR get squashed on merge then, right?

Yeah, we often squash, but it's not a must. Either way, we generally don't stress too much about having the perfect commit history in a PR.

jmoosburger:
the linter suggests to pass a const reference in `SwapAbsoption.findSwapsReadyForAbsorption()`:

> Warning: the parameter 'wires' of type 'SmallVector<WireIterator>' is copied for each invocation but only used as a const reference; consider making it a const reference

but this results in `SegementationFaultExceptions` ...
shall I rewrite the code or ignore the warning °^^

the other lint warnings are fixed :+1:

denialhaag:
> the linter suggests to pass a const reference in `SwapAbsoption.findSwapsReadyForAbsorption()`:
> 
> > Warning: the parameter 'wires' of type 'SmallVector' is copied for each invocation but only used as a const reference; consider making it a const reference
> 
> but this results in `SegementationFaultExceptions` ... shall I rewrite the code or ignore the warning °^^

I tried around a bit locally and just pushed a potential fix. I think this was actually pointing to an issue. 🤔

Let me know if this looks good to you! 🙂

jmoosburger:
> I tried around a bit locally and just pushed a potential fix. I think this was actually pointing to an issue. 🤔
> 
> Let me know if this looks good to you! 🙂

fine with me :+1:

do you have any other feedback, as this is pretty much what we've discussed on this feature?

MatthiasReumann:
@jmoosburger If you think the pull request is ready to review, feel happy to request a review from someone (me) for on the top right of the pull request page.

jmoosburger:
> @jmoosburger If you think the pull request is ready to review, feel happy to request a review from someone (me) for on the top right of the pull request page.

Unfortunately i cannot edit the reviewers-section °^^
image

But i recognized, i need to add a summary and a description for the new pass first :point_up:

jmoosburger:
> But i recognized, i need to add a summary and a description for the new pass first ☝️

I added a rather short description as i think it's not that complicated. feel free to share your personal opinion on that!

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -174,4 +174,21 @@ def HadamardLifting : Pass<"hadamard-lifting", "mlir::ModuleOp"> {
   }];
 }
 
+def SwapAbsorptionPass : Pass<"absorb-swaps", "mlir::ModuleOp"> {
+  let dependentDialects = ["mlir::qco::QCODialect"];
+  let summary = "Eliminate SWAP operations by rewiring circuit";
+  let options = [];
```
MatthiasReumann:
```suggestion
```

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
```diff
@@ -0,0 +1,190 @@
...
+using namespace mlir;
+using namespace mlir::qco;
+
+namespace {
+
+class SwapAbsorbPassTest : public testing::Test {
+
+protected:
+  void SetUp() override {
+    // Register all necessary dialects
```
MatthiasReumann, line 51:
```suggestion
```

As this is the same everywhere else, I tend to omit this comment nowadays.

mlir/lib/Dialect/QCO/Transforms/Optimizations/SwapAbsorption.cpp
```diff
@@ -0,0 +1,84 @@
...
+    ModuleOp anchor = getOperation();
+    IRRewriter rewriter(&getContext());
+
+    for (auto func : anchor.getOps<func::FuncOp>()) {
+      SmallVector<SWAPOp> readyToAbsorb;
+      do {
+        SmallVector<WireIterator> wires;
+        for (auto op : func.getOps<StaticOp>()) {
+          wires.emplace_back(op.getQubit());
+        }
```
MatthiasReumann: If these wires are empty, we should probably early-exit the `runOnOperation` function. 

Something like 
```cpp
if (wires.empty()) {
    return;
}
```

MatthiasReumann: As according to the [LLVM best practices](https://llvm.org/docs/ProgrammersManual.html#vector), you could probably also move the `SmallVector<WireIterator> wires` above the `do` and clear it in each iteration.

MatthiasReumann (changes requested):
It took us a pretty long time and three pull requests but we are finally here! Congratulations on your potentially first contribution to the MLIR compiler collection 🎉

I've left some small remarks which should be fairly easy to integrate. Otherwise, great job! 🚀

jmoosburger:
> It took us a pretty long time and three pull requests but we are finally here! Congratulations on your potentially first contribution to the MLIR compiler collection 🎉
> 
> I've left some small remarks which should be fairly easy to integrate. Otherwise, great job! 🚀

Thanks for your help and patience!
Hopefully there will be more in the future :partying_face:

MatthiasReumann:
@jmoosburger I've pushed some minor changes directly. 

Let's see what @burgholzer has to say, but I think this one should be good to go 🚀

MatthiasReumann (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ with the exception that minor releases may include breaking changes.
 
 ### Added
 
+- ✨ Add a `absorb-swaps` pass for absorbing initial SWAPs ([#1750])
+  ([**@jmoosburger**], [**@MatthiasReumann**])
```
```suggestion
- ✨ Add an `absorb-swaps` pass for absorbing initial SWAPs ([#1750])
  ([**@jmoosburger**], [**@MatthiasReumann**], [**@burgholzer**])
```

mlir/lib/Dialect/QCO/Transforms/Optimizations/SwapAbsorption.cpp
```diff
@@ -0,0 +1,74 @@
...
+  void runOnOperation() override {
+    ModuleOp anchor = getOperation();
+    IRRewriter rewriter(&getContext());
+
+    for (auto func : anchor.getOps<func::FuncOp>()) {
+      SmallVector<SWAPOp> readyToAbsorb;
+      SmallVector<WireIterator> wires;
+      do {
+        wires.clear();
+        for (auto op : func.getOps<StaticOp>()) {
```
burgholzer: This will collect all static operations repeatedly in every iteration of the `do...while` loop, which seems fairly wasteful. Pretty sure there is a more efficient way to do this.

mlir/lib/Dialect/QCO/Transforms/Optimizations/SwapAbsorption.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace {
+struct SwapAbsorption : impl::SwapAbsorptionBase<SwapAbsorption> {
+  using SwapAbsorptionBase::SwapAbsorptionBase;
+
+protected:
+  void runOnOperation() override {
+    ModuleOp anchor = getOperation();
+    IRRewriter rewriter(&getContext());
+
+    for (auto func : anchor.getOps<func::FuncOp>()) {
```
burgholzer: I am really wondering about the efficiency of this pass. I already pushed a couple of improvements, but I can't help it and feel like this is not the way to go.

Is there any particular reason why this pass cannot simply be one that matches on `StaticOp` and checks for subsequent `SWAPOp` that also have a `StaticOp` as there second input?
Something as simple as
```c++
/**
 * @brief Absorb SWAP operations into static qubit allocations
 */
struct RemoveSWAPAfterStaticAllocation final : OpRewritePattern<SWAPOp> {
  using OpRewritePattern::OpRewritePattern;

  LogicalResult matchAndRewrite(SWAPOp op,
    auto qubit0 = op.getInputQubit(0);
    if (!isa<StaticOp>(qubit0.getDefiningOp())) {
      return failure();
    }
    auto qubit1 = op.getInputQubit(1);
    if (!isa<StaticOp>(qubit1.getDefiningOp())) {
      return failure();
    }
    rewriter.replaceOp(op, {qubit1, qubit0});
    return success();
  }
};
```
This feels way simpler _and_ more efficient than the pass implementation here.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -200,4 +200,18 @@ def HadamardLifting : Pass<"hadamard-lifting", "mlir::ModuleOp"> {
...
+  let summary = "This pass absorbs SWAP operations into the initial "
+                "program-to-hardware mapping.";
+  let description = [{
+    For a SWAP operation exchanging static qubits q0 and q1, the pass replaces the use of the
+    first (second) input qubit with the second (first) output qubit of the SWAP and subsequently
+    removes the operation. As a result, the initial program-to-hardware mapping is changed.
+    This process is repeated until no more SWAP operations can be absorbed.
+
+    The pass assumes that the quantum program is already mapped to static qubits.
+  }];
```
burgholzer, line 214: I think the wording here could be improved; maybe even the naming of the pass itself. As it stands, the name of the pass is not self explanatory.
"initial program-to-hardware mapping" is something that is oddly specific for a pass that just generally matches static operations and checks if any SWAPs are directly applied after the "allocation".

I am even wondering whether this should actually be limited to `StaticOp`. Why not extend the logic to `AllocOp` as well? In the pattern rewrite proposed in the other comment, that is likely to be fairly simple. Taking it one step further, `qtensor.load` could also qualify for such a simplification. And to go even further beyond, any mixture of `StaticOp`, `AllocOp` or `LoadOp` can equally be simplified.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
```diff
@@ -0,0 +1,189 @@
...
+  builder.sink(q11);
+
+  auto moduleThroughPass = builder.finalize();
+  auto originalModule = moduleThroughPass->clone();
+
+  applySwapAbsorb(moduleThroughPass);
+  ASSERT_TRUE(mlir::OperationEquivalence::isEquivalentTo(
+      moduleThroughPass.get(), originalModule,
+      mlir::OperationEquivalence::Flags::IgnoreLocations));
+}
```
burgholzer: Kind of a pointless test. Why would the pass change a program that does not contain the single operation the pass actually matches on?

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
```diff
@@ -0,0 +1,189 @@
...
+  qco::QCOProgramBuilder builder(context.get());
+  builder.initialize();
+
+  const auto q00 = builder.staticQubit(0);
+  const auto q10 = builder.staticQubit(1);
+
+  const auto [q01, q11] = builder.swap(q00, q10);
+
+  const auto q02 = builder.id(q01);
+  const auto q12 = builder.id(q11);
```
burgholzer: Kind of pointless to use identity gates here, as they would generally be optimized away. Prefer to use non-trivial gates, if at all.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  builder.sink(q02);
+  builder.sink(q12);
+
+  auto moduleThroughPass = builder.finalize();
+  applySwapAbsorb(moduleThroughPass);
+
+  ASSERT_EQ(q10, ((IdOp)q02.getDefiningOp()).getInputQubit(0));
+  ASSERT_EQ(q00, ((IdOp)q12.getDefiningOp()).getInputQubit(0));
```
burgholzer: Do not use C-style casts. I am surprised clang-tidy is not screaming at this.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
burgholzer, line 1: These tests do not really follow the structure of most of the other tests, which always build a reference program and use our IR verifier to check for equivalence.
I'd prefer to have this as uniform as possible.

burgholzer (changes requested):
Finally found some time to look at this. Thanks for your work on this @jmoosburger and @MatthiasReumann 🙏🏼 

As you will probably see from the comments, I am not quite happy with this yet. Most of the comments are inline and should hopefully be rather self-explanatory. Let me know if anything is not clear.

One other issue that I still have with this is that the new pass is not added to the default pass pipeline, which is an oversight in my opinion. Why would we implement a pass when we do not intend to use it?

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
jmoosburger: it doesn't matter, which single qubit gate is used at this place, to test this particular pass. a gate is required to check, whether the pass reorders the circuit correctly. i took identity as it is the simplest one and does not add functionality to the test circuit

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
jmoosburger: i'll have a look at that :+1:

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
jmoosburger: for me its a requirement that circuits without swaps remain untouched.
from the tests point of view it's not possible to determine where exactly the pass matches on.
you cannot be sure a future implementation of this pass accidentally changes a circuit without a swap unless you test it.

nevertheless, feel free to reach out, if you are uncomfortable with this. i'll remove this test then :+1:

mlir/lib/Dialect/QCO/Transforms/Optimizations/SwapAbsorption.cpp
jmoosburger: i'll have a look at that :+1:

jmoosburger:
@burgholzer 

thanks for your feedback :pray: 
i'll work it through as soon as possible

> One other issue that I still have with this is that the new pass is not added to the default pass pipeline, which is an oversight in my opinion. Why would we implement a pass when we do not intend to use it?

this is actually no oversight but results from this comment: https://github.com/munich-quantum-toolkit/core/pull/1661#issuecomment-4310806755

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
jmoosburger: done and replaced with `mlir:cast`

burgholzer:
> > One other issue that I still have with this is that the new pass is not added to the default pass pipeline, which is an oversight in my opinion. Why would we implement a pass when we do not intend to use it?
> 
> 
> 
> this is actually no oversight but results from this comment: https://github.com/munich-quantum-toolkit/core/pull/1661#issuecomment-4310806755

I see. I think it would be a waste not to run this pass. So I'd argue it should be registered as part of the CompilerPipeline as one of the last optimization passes.

mlir/lib/Dialect/QCO/Transforms/Optimizations/SwapAbsorption.cpp
MatthiasReumann: I agree with @burgholzer here. 

One could even probably make an argument to add this pattern to the canonicalization pass instead of adding a new pass altogether (and all the boilerplate which comes along with it). This would again improve performance by applying the patterns in the same driver run as all the other patterns.

jmoosburger: question in terms of clarification: this would run per operation in the circuit, right?

it definitely seems more efficient, but does it cover all our required cases?
example:

```
q0 --H--X-----      q1 --H--
        |
q1 --H--X--X--  ->  q2 --H--
           |
q2 --H-----X--      q0 --H--
```

MatthiasReumann: You can still use the `WireIterator` in the pattern implementation. The above solution is just a blueprint, I assume.

burgholzer: I would argue against a canonicalization pattern here. I had this suggestion in my review already. However, this would mean that the pattern would try to match every single swap operation in a circuit while it would only really match very little.
As such, this feels cleaner as a separate pass that is invoked at a specific point in time than a canonicalization run ever so often.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
burgholzer: If the pass gets added to the regular pass pipeline, this will directly be covered by almost every single pipeline test (most of them do not contain SWAPs).
So I would argue there is truly not that much value in this test.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
burgholzer: Based on the other comment about unifying the test structure with the rest of the repo, this will likely need some changes anyway. At that point it's probably better to use some kind of multi-qubit gate whose inputs are permuted by swaps.

jmoosburger:
> I see. I think it would be a waste not to run this pass. So I'd argue it should be registered as part of the CompilerPipeline as one of the last optimization passes.

currently the passes are registered dependent on an enable-/disable-flag from the config:

```
        if (!config_.disableMergeSingleQubitRotationGates) {
          pm.addPass(qco::createMergeSingleQubitRotationGates());
        }
        if (config_.enableHadamardLifting) {
          pm.addPass(qco::createHadamardLifting());
        }
```

shall there be one for swap absorption as well?

burgholzer:
> > I see. I think it would be a waste not to run this pass. So I'd argue it should be registered as part of the CompilerPipeline as one of the last optimization passes.
> 
> 
> 
> currently the passes are registered dependent on an enable-/disable-flag from the config:
> 
> 
> 
> ```
> 
>         if (!config_.disableMergeSingleQubitRotationGates) {
> 
>           pm.addPass(qco::createMergeSingleQubitRotationGates());
> 
>         }
> 
>         if (config_.enableHadamardLifting) {
> 
>           pm.addPass(qco::createHadamardLifting());
> 
>         }
> 
> ```
> 
> 
> 
> shall there be one for swap absorption as well?

I think such a flag should exist out of consistency. The default should be "on", so I would argue it should be a "disable..." one.

mlir/lib/Dialect/QCO/Transforms/Optimizations/SwapAbsorption.cpp
```diff
@@ -32,42 +32,13 @@ struct SwapAbsorption : impl::SwapAbsorptionBase<SwapAbsorption> {
...
-          }
-          return WalkResult::interrupt();
-        });
+    anchor.walk([&rewriter](mlir::Operation *op) {
+      if(auto swap = mlir::dyn_cast<SWAPOp>(op))
+      {
+          rewriter.replaceOp(swap,
+                             {swap.getQubit1In(), swap.getQubit0In()});
+      }
+    });
```
MatthiasReumann: Because I stumbled upon this recent change, a comment for educational purposes:

This is invalid for two reasons. 
1. **Semantics:** This would rewire each SWAP in the circuit. But this pass should only rewire those at the very beginning without any dependency to a two qubit gate. Essentially what you are trying to do here is reverse the mapping pass (which inserts SWAPs). Maybe revisit the previous `walkProgramGraph` implementation for which SWAPs should be absorbed. 
2. **Correctness:** rewriting inside a `walk` callback invalidates the iterators and can (or will) lead to undefined behavior or hence segmentation faults. If you want to do something like this in MLIR, you first have to collect the operation in a `SmallVector` and then in a second step apply the rewrite.

jmoosburger: to 1.: sure, i'm in the middle of reimplementing the whole pass, this is, why i marked the PR as draft again :)
to 2.: i thought about that, but i was not sure as this was not relevant for the current TDD step, thanks for this input

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_swapabsorption.cpp
```diff
@@ -0,0 +1,172 @@
...
+  qRef3 = referenceBuilder.h(qRef3);
+  reference = referenceBuilder.finalize();
+
+  ASSERT_TRUE(applySwapAbsorbPass(swapModule.get()).succeeded());
+
+  EXPECT_TRUE(
+      areModulesEquivalentWithPermutations(swapModule.get(), reference.get()));
+}
+
+TEST_F(SwapAbsorbPassTest, PassReordersCircuitWithDependentSwap) {
```
jmoosburger, line 149: i found out for these dependent swaps, that if i remove them with `rewriter.replaceOp()` (like suggested above), the pass does not reorder the single qubit gates (as in the previous test cases), but just the inputs. the resulting circuit looks like, so:
```
module {
    %0 = qco.alloc : !qco.qubit
    %1 = qco.alloc : !qco.qubit
    %2 = qco.alloc : !qco.qubit
    %3 = qco.x %1 : !qco.qubit -> !qco.qubit
    %4 = qco.y %2 : !qco.qubit -> !qco.qubit
    %5 = qco.z %0 : !qco.qubit -> !qco.qubit
    qco.sink %5 : !qco.qubit
    qco.sink %4 : !qco.qubit
    qco.sink %3 : !qco.qubit
    return %c0_i64 : i64
  }
}
```
note, that %3, %4 and %5 receive the qubits as reordered input, the gates are in the same sequence as in the original circuit (*x y z* and not *y z x*).
expressed graphically, the following two circuits are semanticly the same, but the method `areModulesEquivalentWithPermutations()` considers them as different:

```
q0 --y-- q0     q0 --x--\ /------- q1
                         X  
q1 --z-- q1     q1 --y--/ \--\ /-- q2
                              X  
q2 --x-- q2     q2 --z-------/ \-- q0
```
as already mentioned, this did not occur in the previous test cases, with non-dependent swaps.
this is why the `referenceModule` is built differently than the other ones.

is this of any concern for the future? it might be less readable (if this is a requirement in this state of optimisation) and i'm not sure if it might cause problems if the wires cross each other.

i hope i explained clearly, what my question is. please ask if that's not the case :)

burgholzer: This can supposedly be "fixed" by using a qubit register instead of a plain alloc of three qubits in the test case definition.
An alloc is just an alloc. There is no telling them apart.

jmoosburger: I've already tried that, and it results in the same phenomenon. same implementation than with plainly allocated qubits, just with 3-qubit-registers through `programBuilder.allocQubitRegister(3)`:

```
module {
    %c3 = arith.constant 3 : index
    %0 = qtensor.alloc(%c3) : tensor<3x!qco.qubit>
    %out_tensor, %result = qtensor.extract %0[%c0] : tensor<3x!qco.qubit>
    %out_tensor_0, %result_1 = qtensor.extract %out_tensor[%c1] : tensor<3x!qco.qubit>
    %out_tensor_2, %result_3 = qtensor.extract %out_tensor_0[%c2] : tensor<3x!qco.qubit>
    %1 = qco.x %result_1 : !qco.qubit -> !qco.qubit
    %2 = qco.y %result_3 : !qco.qubit -> !qco.qubit
    %3 = qco.z %result : !qco.qubit -> !qco.qubit
    %4 = qtensor.insert %1 into %out_tensor_2[%c0] : tensor<3x!qco.qubit>
    %5 = qtensor.insert %2 into %4[%c1] : tensor<3x!qco.qubit>
    %6 = qtensor.insert %3 into %5[%c2] : tensor<3x!qco.qubit>
    qtensor.dealloc %6 : tensor<3x!qco.qubit>
    return %c0_i64 : i64
  }
}
```
i intentionally tried to figure out the pass with no registers, as there are these additional *extract* and *insert*-steps that affect the equivalence of circuits. see the following reference circuit which executes the inserts in a different order and is therefore considered not equal to the upper one:

```
module {
    %c3 = arith.constant 3 : index
    %0 = qtensor.alloc(%c3) : tensor<3x!qco.qubit>
    %out_tensor, %result = qtensor.extract %0[%c0] : tensor<3x!qco.qubit>
    %out_tensor_0, %result_1 = qtensor.extract %out_tensor[%c1] : tensor<3x!qco.qubit>
    %out_tensor_2, %result_3 = qtensor.extract %out_tensor_0[%c2] : tensor<3x!qco.qubit>
    %1 = qco.x %result_1 : !qco.qubit -> !qco.qubit
    %2 = qco.y %result_3 : !qco.qubit -> !qco.qubit
    %3 = qco.z %result : !qco.qubit -> !qco.qubit
    %4 = qtensor.insert %3 into %out_tensor_2[%c0] : tensor<3x!qco.qubit>
    %5 = qtensor.insert %2 into %4[%c2] : tensor<3x!qco.qubit>
    %6 = qtensor.insert %1 into %5[%c1] : tensor<3x!qco.qubit>
    qtensor.dealloc %6 : tensor<3x!qco.qubit>
    return %c0_i64 : i64
  }
}
```

MatthiasReumann: 👋🏻 Quick side note:

> the following two circuits are semanticly the same, but the method areModulesEquivalentWithPermutations() considers them as different. 

That's because they are different. The description of the function states:
```cpp
/// Compare two (quantum) module operations for structural equivalence, allowing
/// some permutations. This function is especially tailored to compare quantum
/// computations.
```

There is no mentioning of _semantical_ equivalence. So, why are they _structurally_ different? Well, because there are two additional SWAPs in the right circuit! 

**Edit: I just saw that you drew wire crossings, not SWAPs. Sorry. The point on "semantical equivalence" still holds, I guess.**

MatthiasReumann: > see the following reference circuit which executes the inserts in a different order and is therefore considered not equal to the upper one:

This has been fixed #1781. Make sure to merge the newest main into your local branch.

jmoosburger: > This has been fixed #1781. Make sure to merge the newest main into your local branch.

apparently this already is in my branch, check: https://github.com/jmoosburger/mqt-core/commits/feat/swap-absorption-pass/

MatthiasReumann: Okay, got it. Well these two programs aren't structurally equivalent! 

The above snippet extracts a value from `c0` and inserts it with `c2`, which is invalid.
```mlir
%out_tensor, %result = qtensor.extract %0[%c0] : tensor<3x!qco.qubit>
%3 = qco.z %result : !qco.qubit -> !qco.qubit
%6 = qtensor.insert %3 into %5[%c2] : tensor<3x!qco.qubit>
```

The reference does it correctly:
```mlir
%out_tensor, %result = qtensor.extract %0[%c0] : tensor<3x!qco.qubit>
%3 = qco.z %result : !qco.qubit -> !qco.qubit
%4 = qtensor.insert %3 into %out_tensor_2[%c0] : tensor<3x!qco.qubit>
```

jmoosburger: true, i'll try to find out where this comes from.

additionally for clarification: as you can see, the circuit wires don't get reordered on replacing a swap with its switched inputs, no matter if with plain qubits or within a register. there must be a wire crossing in the `reference`-module then, like this one in the `swapModule`, right?

MatthiasReumann: What _exactly_ do you mean with wire-crossing? The graphical representation of the quantum program?

jmoosburger:
Thank you for your time and for the feedback on this PR.

After giving it some thought, I've decided to close this PR for now. As I can't currently dedicate the time needed to finish this contribution to the standard I'd expect from my own work, closing it seems like the most responsible course of action.

Nevertheless, I'd like to sincerely thank you for all your efforts, for the opportunity to contribute, and for the valuable insights I gained while working on this project. I really appreciate that!

PR #1710, opened 2026-05-15
✨ Add QC to QIR Adaptive Profile Conversion

li-mingbao:
## Description

This PR adds the conversion of QC to the Adaptive Profile of QIR including the conversion of scf operations. 
The original conversion from QC to QIR is split up into two different conversion passes, separating the Base and the Adaptive Profile.
The Base QIR conversion now only supports operation that are available in the Base Profile. Any dynamic alllocation operations are converted into equivalent static qubit and result pointers during the conversion.
The QIRProgramBuilder is also updated to handle the Base and the Adaptive Profile.

## Checklist

- [ ] The pull request only contains commits that are focused and relevant to this change.
- [ ] I have added appropriate tests that cover the new/changed functionality.
- [ ] I have updated the documentation to reflect these changes.
- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [ ] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.
- [ ] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [ ] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).
- [ ] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.
- [ ] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

PR #1755, opened 2026-06-01
:sparkles: Add Dead Gate Elimination Pattern

DRovara:
## Description

This PR implements Dead Gate Elimination for unused operations on qubits.

An operation is considered unused if all of its outputs' users are dealloc operations.
This means a measurement can also be dead if its classical outcome is not used.
Similarly, `reset` and `if` operations can also be considered unused.

This PR marks all quantum gates as `Pure` (and marks the `RecursiveMemoryEffects` operations as `Pure` as well, as otherwise, any recursive memory effects would be overwritten by the compiler not knowing whether these region operations are pure to begin with). This allows us to simply check for memory effects before removing region operations.

Currently, this only considers individual qubits, rather than `qtensor` cases to keep this PR contained.
An extension to qtensors should also be possible, though it may require some additional assumptions to maintain overview on what specific qubits end up going unused.

---

The following new features/changes also had to be implemented to make this PR work with the rest of the framework:

### Return values for builders and tests
Without return values, most quantum programs constructed by the builder would basically be "dead". Therefore, the `QC` and `QCO` `ProgramBuilder`s now require all build functions to also indicate what values and types they return. The `initialize(...)` method now takes a function signature for the `main` function as an argument, while the `finalize(...)` method takes the actual return values. Similarly, the new `retype(...)` changes the type of the `main` function during the build process.

### Translations between IRs:
As `QC` and `QCO` programs now have return values, other IRs also need some notion to represent that concept and ways to translate between them. We have adapted the conversion/translation methods so that:
- *QIR (Base and Adaptive)* now uses the `record` QIR methods to indicate the output of measurement outcomes. (During translation from `QC` to `QIR`, return values have to be stripped out if they are direct measurement outcomes)
- *QuantumComputation* now considers all measured qubits to be returned.
- *OpenQASM 2* also considers all measured qubits to be returned.
- *OpenQASM 3* only considers bit values marked with the `output` keyword to be returned.
- *Jeff* supports return values, so the previous code that stripped away all returns during conversion to *jeff* was simply removed.

### OpenQASM Parsing:
Consequently, handling for the `output` keyword was added to the OpenQASM parser code. Similarly, the translation from OpenQASM to QC was adapted. Note that the `output` keyword is ignored when parsing into `QuantumComputation` objects, rather than into the `QC` dialect.

### QuantumComputation
For convenience (to keep the tests minimal), the `measureAll` method in the `QuantumComputation` was adapted to also take the `addBarrier` argument. If it is not set to true, barriers will no longer be inserted between the main circuit body and the measurements.

### QIR Builder
To no longer make the `QIRProgramBuilder` consider all measurement outcomes for reporting, the `measure` method now also takes the `record` flag as an argument. Outcomes will only be reported if the flag is set to `true`.

### Other
Various further helper methods have been added to make working with more complex programs easier.

--- 

*AI Usage note*: Some design decisions were taken after discussions with Gemini 3.1 Pro. Parts of the code were written with help from Gemini 3.1 Pro and Claude Opus 4.6.

## Checklist


DRovara:
@coderabbitai full review

DRovara:
@coderabbitai review

DRovara:
@burgholzer the implementation itself should be done now, but the tests will still fail until we fixed the ordering issue in the tester that we discussed internally. I guess this PR can either be reviewed now already or once that fix has been merged in here.

burgholzer:
> @burgholzer the implementation itself should be done now, but the tests will still fail until we fixed the ordering issue in the tester that we discussed internally. I guess this PR can either be reviewed now already or once that fix has been merged in here.

Thanks! I'll try to review this asap. Let's see if that turns out to be before or after the fix.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -96,7 +96,7 @@ def StaticOp : QCOOp<"static", [Pure]> {
 // Measurement and Reset Operations
 //===----------------------------------------------------------------------===//
 
-def MeasureOp : QCOOp<"measure"> {
+def MeasureOp : QCOOp<"measure", [Pure]> {
   let summary = "Measure a qubit in the computational basis";
```
burgholzer, line 100: Instead of adding the `Pure` trait to all individual operations, wouldn't it make sense/be possible to simply add the trait to the `UnitaryOpInterface` in QCO? Then we wouldn't have to specify it everywhere and it would be even clearer that any unitary operation is considered Pure in QCO

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
```diff
@@ -30,6 +35,34 @@
...
+    if (!isMemoryEffectFree(op)) {
+      // This effectively ignores the GPhase operation and variants such as its
+      // inverse or `ctrl` ops containing it, which should never be considered
+      // dead.
+      return failure();
+    }
+    return checkAndRemoveDeadGate(op.getOperation(), rewriter);
+  }
+};
+} // namespace
```
burgholzer: I am wondering whether this cannot easily be replaced with a fold instead of a canonicalization pattern.
From what I can tell from a conversation with Claude, a trait that defines the `foldTrait` method could do the trick here to avoid having to define the same fold over and over again. A boilerplate draft for this could be
```td
include "mlir/IR/OpBase.td"

def DeadIfOnlySinksFoldTrait : NativeOpTrait<"DeadIfOnlySinksFold"> {
  let description = [{
    Folds unitary-like ops away when all users are qco.sink by forwarding
    output qubits to corresponding input qubits.
  }];
}
```
and (in QCOOps.cpp)
```c++
namespace mlir::qco {

template <typename ConcreteType>
class DeadIfOnlySinksFold
: public mlir::OpTrait::TraitBase<ConcreteType, DeadIfOnlySinksFold> {
public:
static mlir::LogicalResult foldTrait(
mlir::Operation* op, mlir::ArrayRef<mlir::Attribute> /*operands*/,
mlir::SmallVectorImpl<mlir::OpFoldResult>& results) {
// Only for side-effect-free ops.
if (!mlir::isMemoryEffectFree(op)) {
return mlir::failure();
}

    // Only fold when every user is qco::SinkOp.
    if (!llvm::all_of(op->getUsers(),
                      [](mlir::Operation* user) { return mlir::isa<SinkOp>(user); })) {
      return mlir::failure();
    }

    // Restrict to UnitaryOpInterface-bearing ops.
    auto u = mlir::dyn_cast<UnitaryOpInterface>(op);
    if (!u) {
      return mlir::failure();
    }

    // Fold by forwarding outputs to corresponding inputs.
    for (mlir::Value v : u.getInputQubits()) {
      results.push_back(v);
    }
    return mlir::success();
}
};

} // namespace mlir::qco
```

It seems to me that it should always be possible to implement the currently implemented canonicalizations as folds.

mlir/lib/Dialect/QCO/IR/Operations/MeasureOp.cpp
```diff
@@ -9,12 +9,31 @@
...
+ * @brief Remove dead measurements.
+ */
+struct DeadMeasurementRemoval final : OpRewritePattern<MeasureOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(MeasureOp op,
+                                PatternRewriter& rewriter) const override {
+    return checkAndRemoveDeadGate(op, rewriter);
+  }
+};
```
burgholzer: At this point, I am not quite sure there is much value in the shared helper function, when the implementation here would essentially be 5 lines of code.
Maybe this also goes well with the suggestion of replacing this with a fold, which necessarily has to be defined for the op itself.

Same comment applies to the ResetOp

mlir/lib/Dialect/QCO/IR/Operations/ResetOp.cpp
```diff
@@ -114,4 +127,5 @@ OpFoldResult ResetOp::fold(FoldAdaptor /*adaptor*/) {
 void ResetOp::getCanonicalizationPatterns(RewritePatternSet& results,
                                           MLIRContext* context) {
   results.add<RemoveResetAfterExtract>(context);
+  results.add<DeadResetRemoval>(context);
```
burgholzer:
```suggestion
results.add<RemoveResetAfterExtract,DeadResetRemoval>(context);
```
Just shorter

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -234,11 +236,28 @@ struct ConditionPropagation : public OpRewritePattern<IfOp> {
...
+ */
+struct DeadIfRemoval final : OpRewritePattern<IfOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(IfOp op,
+                                PatternRewriter& rewriter) const override {
+    if (!isMemoryEffectFree(op)) {
+      // This effectively ignores `IfOp`s with memory effects.
+      return failure();
+    }
```
burgholzer: Probably just me being stupid, but isn't this always true now that the IfOp is marked as `Pure`? Same comment actually also applies for the general case.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -114,6 +114,114 @@ TEST_F(QCOTest, BuilderRejectsMixedStaticAndDynamicQubitAllocationModes) {
...
+  builder.sink(q0S2);
+  builder.sink(q1S2);
+  auto module = builder.finalize();
+
+  QCOProgramBuilder reference(context.get());
+  reference.initialize();
+  auto r0 = reference.allocQubit();
+  auto r1 = reference.allocQubit();
+  reference.sink(r0);
+  reference.sink(r1);
```
burgholzer: This feels like a slighlty pointless test. The consecutive pairs of alloc and sink would be eliminated by canonicalization, which would leave this as an empty program. I find it dangerous to define a reference that will actually (knowingly) be further transformed before the equivalence check.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -114,6 +114,114 @@ TEST_F(QCOTest, BuilderRejectsMixedStaticAndDynamicQubitAllocationModes) {
...
+        return SmallVector<Value>{q1Then};
+      },
+      [&](ValueRange qubits) -> SmallVector<Value> {
+        auto q1Else = reference.h(qubits[0]);
+        return SmallVector<Value>{q1Else};
+      })[0];
+
+  reference.sink(r0S2);
+  reference.sink(r1S1);
+  auto refModule = reference.finalize();
```
burgholzer: I am probably too dense to see it, but where does this `if` have memory effects? wouldn't it also simply be optimized away and both the reference program and the initial program optimize to the empty program?

burgholzer (changes requested):
@DRovara Thanks for pushing this. I pushed very small cleanups directly. The bigger part of my review is in the comments attached with this review. This will need more work before we can merge.

One of the bigger points in the review comments is that this could probably all be folds instead of canonicalizations, which should generally be more efficient.

Additionally, and this is something not in the individual comments, this is currently breaking almost our entire testing pipeline. If you inspect the MLIR outputs of the Pipeline tests for example, then most test cases simply get optimized to empty programs. While most tests still pass, this is obviously not the point of these unit tests. We need to find a way to allow for this kind of optimization, but make sure that it only applies when it actually should apply. In particular, we need to make sure that the default programs generated by our builders are "complete" programs that do not all reduce to the empty program. My best guess would be, that we need some notion of the "output of a quantum program". In QIR terms, this would be the output recording. Possibilities here include adding a dedicated "record_output" operation (similar to what is currently being done in FullStaQD) or to (somewhat) have our main entry point actually _return_ something, which would also make the values "used".

In its present state, I would argue the PR here does not work as intended.

@munich-quantum-toolkit/mqt-cc for awareness. Feel free to jump in if you have ideas.

DRovara:
@burgholzer Thanks for the review! Some comments:
> One of the bigger points in the review comments is that this could probably all be folds instead of canonicalizations, which should generally be more efficient.

That was my original plan, but I had some discussions with LLMs and they all agreed this should be a canonicalisation for API reasons. For instance, Gemini says:
> The MLIR folding infrastructure (OperationFolder) and the GreedyPatternRewriteDriver rely on a specific worklist mechanics. If you write a fold that decides to simplify an operation based on who is using it (e.g., calling op->getUsers()), you break the folder's dependency tracking. If that user operation is later modified or deleted, the framework will not know to re-evaluate your original operation's fold method. Your optimization will silently fail to trigger when it should.

---

> If you inspect the MLIR outputs of the Pipeline tests for example, then most test cases simply get optimized to empty programs.

Ah right, I completely forgot about that. I actually already found a solution for that, I just didn't implement it for the existing tests. If you view [the QCOProgramBuilder](https://github.com/munich-quantum-toolkit/core/blob/ad1be84e1ab39bcea17b63416ee6e0c85b3cb375/mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h) in #1756, I have extended the `initialize` and `finalize` methods so that they can also  take arbitrary return types and values. That way, if we just measure and return all qubits, all values will remain used. Does that sound like a good solution?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
DRovara: Apparently, it is not possible to do that. The only thing one could do is implement a second operation base class like

```
class QCOUnitaryOp<string mnemonic, list<Trait> traits = []> 
    : QCOOp<mnemonic, !listconcat(traits, [UnitaryOpInterface, Pure])> {
```

And then have the unitary operations inherit that rather than `QCOOp`. But it feels a bit weird to have a abstract base class and an interface for unitaries...

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
DRovara: See my high-level comment. The docs never mention it not being allowed to access the users in a fold, but Gemini, Claude, and Perplexity all said it's best not to.

Apparently, the fold framework in MLIR does not add elements to the worklist when their user changes. This means (from my interpretation) that there could be situations where we have the following chain of instructions:

"A => B => X"

Now if we have a rule that says "any operation that is followed by an X can be folded", then the worker might first look at "A", determine it can'f be folded, then look at "B", fold it, but then not check "A" again which could now also be folded.

Unfortunately, none of the AIs could give me actual sources for that (Gemini said "There are precedents of PRs in the LLVM repo being rejected because they accessed the user list inside a fold" but couldn't provide a link), but they all seem to agree on that being the reason.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
DRovara: In addition to being marked as `Pure`, the `IfOp` is also marked as `RecursiveMemoryEffects`. That means that, if any operation inside it has memory effects, then the `IfOp` also has memory effects, independent of it being marked as `Pure`. 

The "override" actually goes in the other direction:
If the `IfOp` was not marked as `Pure`, then even if all its child instructions were `Pure` and `IfOp` has `RecursiveMemoryEffects`, MLIR would still not be certain that the `IfOp` is memory-effect free because the region operation itself might also have memory effects.

The test case `CheckIfOpDeadGateElimination` actually shows exactly that situation, where one of the `IfOp`s contains a `GPhaseOp` and is therefore no longer memory-effect-free.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
DRovara: That's a good point. I have now slightly changed up the test so that some operations remain in the program, that's probably the cleanest solution.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
DRovara: The `gphase` Operation is marked as having memory effects (which it needs to have, after all it has no return values).

Just to be sure I quickly dumped the outcome and it is as expected, with just a single `if` (while the other one gets eliminated):

```
module {
    %cst = arith.constant 5.000000e-01 : f64
    %0 = qco.alloc : !qco.qubit
    %1 = qco.alloc : !qco.qubit
    %2 = qco.h %0 : !qco.qubit -> !qco.qubit
    %qubit_out, %result = qco.measure %2 : !qco.qubit
    %3 = qco.if %result args(%arg0 = %1) -> (!qco.qubit) {
      %4 = qco.x %arg0 : !qco.qubit -> !qco.qubit
      qco.gphase(%cst)
      qco.yield %4 : !qco.qubit
    } else args(%arg0 = %1) {
      %4 = qco.h %arg0 : !qco.qubit -> !qco.qubit
      qco.yield %4 : !qco.qubit
    }
    qco.sink %qubit_out : !qco.qubit
    qco.sink %3 : !qco.qubit
    return %c0_i64 : i64
  }
}
```

mlir/lib/Dialect/QCO/IR/Operations/MeasureOp.cpp
DRovara: Yeah, true. When I originally implemented it I was able to use it everywhere but due to all the edge cases it ended up getting s specialized that I might as well have individual implementations for all uses. I did that now.

Anyways, I have done something else for now: The helper function is now called `checkDeadGate` and it only checks if a gate is dead. This can be implemented in a general way. At least that way, if we ever decide to update the notion of a dead gate, it only takes a single change.

In an ideal world where we really want to minimize code reuse, we might even want to implement some "remove this operation" helper methods for every `qco` operation. Because that's essentially what's left in the current implementation after the `checkIfDead` check is done. But for now, I kept the individual implementations in the different operation files because it's not really an issue yet.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
burgholzer: Yeah, this feels fairly weird indeed.
In the end, this is not really bad; being explicit for every operation might actually be good. It just struck me as fairly repetitive in the diff.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
burgholzer: I tried to look around a little because I simply was not quite trusting the LLMs.
After a bit of digging, I conclude that you/they are right. From what I could gather

> A fold is meant to be:
> 
> - Purely local to the operation
> - Cheap and side-effect-free
> - Context-independent
> - Deterministic based only on operands and attributes
> 
> Concretely, a fold implementation:
> 
> - Operates only on:
> 
>   - The operation itself
>   - Its operands (via FoldAdaptor)
>   - Its attributes
> 
> - Returns:
> 
>   - Existing values, or
>   - Constants (Attribute → constant materialization)
> 
> - Must not mutate IR directly (no erasing/rewriting arbitrary ops)
> - Must not depend on analysis or global structure

In essence, looking at operation results and their users is not allowed as it is non-local and context-dependent.

So, after all, a canonicalization it is.
This got me thinking though: Couldn't this be a single canonicalization that is implemented for the `qco.sink` operation, which would look at the defining op of the incoming qubit value?

mlir/lib/Dialect/QCO/IR/Operations/MeasureOp.cpp
burgholzer: Related to the comment above: Wouldn't it be simpler if this were simply a canonicalization of the `qco.sink` op? would be a fairly central place for all related code.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer: Ah, it's the GPhaseOp that makes it have a memory effect; gotcha. That makes a lit of sense.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
burgholzer: Yeah, I almost expected that this was me being stupid. I overlooked the `gphase` operation. Might be worth adding a comment that explicitly states that.

burgholzer (commented):
> > If you inspect the MLIR outputs of the Pipeline tests for example, then most test cases simply get optimized to empty programs.
> 
> Ah right, I completely forgot about that. I actually already found a solution for that, I just didn't implement it for the existing tests. If you view [the QCOProgramBuilder](https://github.com/munich-quantum-toolkit/core/blob/ad1be84e1ab39bcea17b63416ee6e0c85b3cb375/mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h) in #1756, I have extended the `initialize` and `finalize` methods so that they can also take arbitrary return types and values. That way, if we just measure and return all qubits, all values will remain used. Does that sound like a good solution?

Generally sounds good to me. You might have meant this, but I would have expected the classical measurement results to be returned instead of the qubits.
This may also be a nice way of improving the QIR conversion, because classical return values could be transformed to output recordings in QIR.  @denialhaag would that make sense?

denialhaag:
> > > If you inspect the MLIR outputs of the Pipeline tests for example, then most test cases simply get optimized to empty programs.
> > 
> > 
> > Ah right, I completely forgot about that. I actually already found a solution for that, I just didn't implement it for the existing tests. If you view [the QCOProgramBuilder](https://github.com/munich-quantum-toolkit/core/blob/ad1be84e1ab39bcea17b63416ee6e0c85b3cb375/mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h) in #1756, I have extended the `initialize` and `finalize` methods so that they can also take arbitrary return types and values. That way, if we just measure and return all qubits, all values will remain used. Does that sound like a good solution?
> 
> Generally sounds good to me. You might have meant this, but I would have expected the classical measurement results to be returned instead of the qubits. This may also be a nice way of improving the QIR conversion, because classical return values could be transformed to output recordings in QIR. @denialhaag would that make sense?

Yeah, returning classical values from the main function and using that information for the output recording sounds sensible to me!

Two thoughts:

1. This has an impact on [#1710](https://github.com/munich-quantum-toolkit/core/pull/1710) (CC @li-mingbao). The change will necessarily require tweaks to the QC-to-QIR conversion here. We should decide which PR we want to merge first to avoid unnecessary conflicts.
2. This will have quite an impact on the test programs because pretty much all qubits should be measured at the end of the current programs, but I think that's fine.

burgholzer:
> Yeah, returning classical values from the main function and using that information for the output recording sounds sensible to me!
> 
> Two thoughts:
> 
> 1. This has an impact on [#1710](https://github.com/munich-quantum-toolkit/core/pull/1710) (CC @li-mingbao). The change will necessarily require tweaks to the QC-to-QIR conversion here. We should decide which PR we want to merge first to avoid unnecessary conflicts.

I would argue that this PR here is currently blocking more PRs than the QIR one is. So my tendency would be to merge this here first and afterwards adjust the QIR one.
If we make this change, a lot of the guesswork for output recording is potentially removed. I'd call that a win.

> 2. This will have quite an impact on the test programs because pretty much all qubits should be measured at the end of the current programs, but I think that's fine.

Yeah; there is almost no way around this. Generally, I believe this very much improves the test programs though. It makes them more complete quantum programs.

denialhaag:
> I would argue that this PR here is currently blocking more PRs than the QIR one is. So my tendency would be to merge this here first and afterwards adjust the QIR one. If we make this change, a lot of the guesswork for output recording is potentially removed. I'd call that a win.

Makes sense to me! It might make sense to have a look at the implementation in #1710 and fix the output recording directly before we have another version of a not-exactly-correct implementation. 🤔 

> Yeah; there is almost no way around this. Generally, I believe this very much improves the test programs though. It makes them more complete quantum programs.

Agreed!

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
DRovara: I'm almost always against repetitiveness, but I guess in this specific case it's then probably best to just leave it as it is.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
DRovara: I think it's a valid consideration, but my gut feeling tells me it's better to be on the gates themselves.

Mainly because a transformation pattern should aim to target/modify mainly the operation it matches. So matching `qco.sink` and then only modifying the predecessors is not best practise.

But also conceptually, removing dead gates is not a property of the sinks, being dead is a property of the gates. If we ever decide that there's another type of operation that randers gates dead, then it is a straightforward addition to modify the current version, but it would be more of a change to add another canonicalisation for that.

mlir/lib/Dialect/QCO/IR/Operations/MeasureOp.cpp
DRovara: See the answer above. I honestly don't really feel like the way it's implemented right now is wrong. Having the special operation handling in the gate files tells you exectly: "Okay I'm looking at the handling for an `IfOp` right now" etc. and you don't have to chain a bunch of `if (isa<qco::IfOp>(op)) ...`

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
burgholzer: I see your point. I believe my main concern was that this feels so repetitive.
And matching the `sink` operation would centralize all the code, but also the pattern matching itself. Instead of matching every single unitary and having to check its successor, it would only match a single operation and check the predecessor.
In terms of efficiency, this sounds way better to me.
And yeah, if another type of operation comes up, this would mean some duplication; but I would argue that even in that case, it seems more efficient to match one of two operations than to match against every single one.

At the same time, I agree that it does not feel quite right for the canonicalization to not modify the operation it matches on. But is that really so bad?

mlir/lib/Dialect/QCO/IR/Operations/MeasureOp.cpp
burgholzer: I am not saying it is wrong by any means; it is obviously working.
But I am concerned that this might not be particularly efficient compared to the alternative.
The sink canonicalization would just contain a TypeSwitch, which is pretty efficient.

DRovara:
Alright, @burgholzer, @denialhaag, I have now updated all the IR tests to return the values of the measured qubits.
It was quite some effort to get all of this done. In some cases where the IR is kind of weird, I didn't return all qubits because access to them is sometimes difficult, but in any case none of the operations should be considered dead now.

That being said, this probably still needs a second look from you, as I was previously not too involved in the tests, so I'm not sure if I handled any situations incorrectly, and I am quite sure that there could also be other ways to implement all of this cleaner.

In any case, the tests now still cause errors because of the ordering issue during the equivalence check, so that still needs to be fixed somehow for all of this to work.

Also, I have not touched the transformation tests. But most transformation tests don't even touch the canonicalisations, so it should not be an issue to begin with. The only exception is `test_qco_hadamard_lifting`, I will check that in more detail later, but most likely we would require @LiRem101 to fix the tests, because it's quite hard for me to judge where I can add measurements without breaking the ideas behind the tests.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -10,8 +10,10 @@
 
 #pragma once
 
+#include <llvm/ADT/TypeSwitch.h>
```
denialhaag: Is this include still needed? 🤔

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
```diff
@@ -69,6 +71,16 @@ void QCOProgramBuilder::initialize() {
   setInsertionPointToStart(&entryBlock);
 }
 
+void QCOProgramBuilder::retype(TypeRange returnTypes) {
+  auto mainFunc = dyn_cast<ModuleOp>(module).lookupSymbol<func::FuncOp>("main");
```
denialhaag: The entry point is generally defined as the function that has the `entry_point` attribute (see, e.g., [`getEntryPoint()`](https://github.com/munich-quantum-toolkit/core/blob/946116d51ba583010dbb3a692c7c2c5558070d24/mlir/include/mlir/Dialect/QCO/IR/QCODialect.h#L133-L161)).

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -237,4 +239,21 @@ mergeTwoTargetOneParameterWithSwappedTargets(OpType op,
...
+ * @brief Check if given quantum operation is unused (i.e., only used by
+ * sinks and no memory effects).
+ *
+ * @param op The operation to check.
+ * @return bool True if the operation is unused, false otherwise.
+ */
+inline bool checkDeadGate(Operation* op) {
+  if (!isMemoryEffectFree(op)) {
+    // This ignores operations with and regions that have children with memory
+    // effects, which should never be considered dead.
```
denialhaag:
```suggestion
    // This ignores operations and regions that have children with memory
    // effects, which should never be considered dead.
```

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -114,10 +116,124 @@ TEST_F(QCOTest, BuilderRejectsMixedStaticAndDynamicQubitAllocationModes) {
       "Cannot mix dynamic and static qubit allocation modes");
 }
 
+TEST_F(QCOTest, CheckDeadGateElimination) {
```
denialhaag: I understand that it does not necessarily make sense to define such specialized programs in `qco_programs.h`, but would it be possible to define the programs via similar functions in this file and then use `QCOTestCase`? 🤔

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
```diff
@@ -1388,6 +1425,22 @@ class QCOProgramBuilder final : public ImplicitLocOpBuilder {
...
+   * and uses it to build the desired quantum program. The builder will be
+   * properly initialized before calling this function, and the resulting module
+   * will be finalized using the returned ValueRange after this function
+   * completes.
+   * @return The module containing the quantum program built by buildFunc.
+   */
+  static OwningOpRef<ModuleOp> buildWithReturn(
+      MLIRContext* context,
+      const function_ref<std::pair<SmallVector<Value>, SmallVector<Type>>(
+          QCOProgramBuilder&)>& buildFunc);
```
denialhaag: Without having thought about it too much right now, is there still a real need for the old `build()` method, or should _any_ real QCO program return _something_ (other than the exit code)? 🤔

mlir/unittests/programs/qco_programs.h
```diff
@@ -10,1072 +10,1354 @@
 
 #pragma once
 
+#include <mlir/IR/ValueRange.h>
```
denialhaag: In header files, we try to only include `mlir/Support/LLVM.h`, which mainly contains forward declarations.

```suggestion
#include <mlir/Support/LLVM.h>
```

mlir/unittests/programs/qco_programs.cpp
```diff
@@ -14,1931 +14,2975 @@
...
+  }
+  return {bits, bitTypes};
+}
+
 namespace mlir::qco {
 
-void emptyQCO([[maybe_unused]] QCOProgramBuilder& builder) {}
+std::pair<SmallVector<Value>, SmallVector<Type>>
+emptyQCO([[maybe_unused]] QCOProgramBuilder& builder) {
+  return measureAndReturn(builder, {});
```
denialhaag: `[[maybe_unused]]` should no longer be needed, right? 🤔 

```suggestion
emptyQCO(QCOProgramBuilder& b) {
  return measureAndReturn(b, {});
```

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -1118,16 +1235,9 @@ INSTANTIATE_TEST_SUITE_P(
...
-        QCOTestCase{"AllocQubitRegister", MQT_NAMED_BUILDER(allocQubitRegister),
-                    MQT_NAMED_BUILDER(emptyQCO)},
-        QCOTestCase{"AllocMultipleQubitRegisters",
-                    MQT_NAMED_BUILDER(allocMultipleQubitRegisters),
-                    MQT_NAMED_BUILDER(emptyQCO)},
-        QCOTestCase{"AllocLargeRegister", MQT_NAMED_BUILDER(allocLargeRegister),
+        QCOTestCase{"AllocQubit", MQT_NAMED_BUILDER(allocQubitNoMeasure),
                     MQT_NAMED_BUILDER(emptyQCO)},
-        QCOTestCase{"StaticQubits", MQT_NAMED_BUILDER(staticQubits),
+        QCOTestCase{"StaticQubits", MQT_NAMED_BUILDER(staticQubitsNoMeasure),
```
denialhaag:
```suggestion
        QCOTestCase{"StaticQubitsNoMeasure", MQT_NAMED_BUILDER(staticQubitsNoMeasure),
```

denialhaag (dismissed):
Thanks a lot for the changes, @DRovara! I like where this is going!

Besides the comments below, I still see one major issue: Don't we also need to return the measurement results in QC? At the very least, the QC-to-QCO conversion would need to adjust the main function signature. Otherwise, the converted QCO programs _should_ be empty right now (after canonicalization, that is).

(For some reason, the build is currently failing. Could it be that you only built the `mqt-core-mlir-unittest-qco-ir` target locally? 🤔)

mlir/unittests/programs/qco_programs.h
DRovara: Gotcha, thanks for clarifying!

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
DRovara: Good question. I didn't feel confident enough to remove the original `build` function.

I can imagine scenarios where it would still be useful, the question is just (1) are they valid/important enough? and (2) are they maybe that rare that users could just use `buildWithReturn` with a function pointer that just returns an empty type range or a single ì64` exit code instead?

Example:
A program that is supposed to be translated down to QIR and should then run the `record_outputs` (don't remember the exact name) operation. Or, if that's what the new returns will be translated into, then maybe programs with intermediate print operations.

In any case, I have no problem with just removing the original `build` method, but I feel like that should be your call.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
DRovara: I agree that this is not the cleanest place to put and handle the test. I feel like, ideally, there should be a specialised test file for general canonicalisations. I'll have a quick look and then either try that or implement it the way you suggested.

mlir/unittests/programs/qco_programs.h
DRovara: Turns out that's not enough, as it does not declare `mlir::Value` I need `<mlir/IR/Value-h>` for that as well

mlir/lib/Dialect/QCO/Builder/QCOProgramBuilder.cpp
DRovara: Done, thanks for the suggestion!

mlir/include/mlir/Dialect/QCO/QCOUtils.h
DRovara: You're right, I removed it now

DRovara:
> Don't we also need to return the measurement results in QC?

Yeah you're absolutely right. My initial though was "This canonicalization only happens in QCO so we don't have to worry about QC programs", but of course if we start from QC and translate down to QCO, then we also need return values there. Though, I guess this thought can also be extended in other directions. The translation from other input formats would also require to track some return values then.

> Could it be that you only built the `mqt-core-mlir-unittest-qco-ir` target locally

Ah you're right. Turns out the build error is related to the other point: Currently, the `named_builder` requires a program function that returns the return values, because I added that for `QCO`. But the same class is also used for the `QC` and `QIR` tests. So I guess the only reasonable solution is to add the return types for them too. But I guess that means that this PR is going to become quite large (and quite some effort) - unless we have any other ideas, @denialhaag, @burgholzer?

mlir/include/mlir/Dialect/QCO/Builder/QCOProgramBuilder.h
burgholzer: I'd be fine with dropping the original build method. One can always emulate it with the existing builder by returning an exit code. So the new builder is strictly more expressive.
Edit: the return values of entry point functions will be translated to output recording calls in QIR.

burgholzer:
> > Don't we also need to return the measurement results in QC?
> 
> 
> 
> Yeah you're absolutely right. My initial though was "This canonicalization only happens in QCO so we don't have to worry about QC programs", but of course if we start from QC and translate down to QCO, then we also need return values there. Though, I guess this thought can also be extended in other directions. The translation from other input formats would also require to track some return values then.

I would argue it is quite natural for the QC programs to also return results from the main entry points. OpenQASM has the concept of outputs and defaults to output all values of classical registers in the order of definition if no explicit output statements are present.
Qiskit handles it similarly if I am not mistaken.
@denialhaag i think this also still works with Jeff programs, right? If it works with that, it should also work with Hugr/Guppy and Catalyst/PennyLane.
Which leaves CUDA-Q, which defines quantum kernels that. Quite naturally return values.
And for reading in QIR, we would replace the output recordings with return values supposedly. So I believe the overall design makes sense.

> 
> > Could it be that you only built the `mqt-core-mlir-unittest-qco-ir` target locally
> 
> 
> 
> Ah you're right. Turns out the build error is related to the other point: Currently, the `named_builder` requires a program function that returns the return values, because I added that for `QCO`. But the same class is also used for the `QC` and `QIR` tests. So I guess the only reasonable solution is to add the return types for them too. But I guess that means that this PR is going to become quite large (and quite some effort) - unless we have any other ideas, @denialhaag, @burgholzer?

I have no real other ideas. A large PR isn't bad per-se, if most of the changes are mechanical (updating the test programs).
We have a couple of open or upcoming PRs that touch the fundamentals of how we represent programs. None of them are particularly small or little work.
But I would argue that this comes quite naturally as we use the compiler collection more and more.
No real way around it, I suppose.

denialhaag:
> @denialhaag i think this also still works with Jeff programs, right? If it works with that, it should also work with Hugr/Guppy and Catalyst/PennyLane.

Yeah, `jeff` should be quite flexible here in the sense that the function marked as the entry point can have `targets` as any other region. The conversions between `jeff` and QCO likely have to be adjusted slightly to preserve that information, but I don't think this should be too big an issue. 🤔 

Edit: This Catalyst example, for example, returns a measurement result from the entry-point function: https://github.com/unitaryfoundation/jeff/blob/main/examples/catalyst_simple/catalyst_simple.txt

burgholzer:
@DRovara The QIR PR is in. The Verifier PR looks practically ready. So I would argue work on this can continue 🙂

MatthiasReumann:
@DRovara FYI: The new verifier is now merged into `main`.

burgholzer:
@DRovara Just to make you aware in case you missed it: This PR is currently in conflict with main. It may be worth to merge `main` into this branch before continuing adding further changes 😌

DRovara:
> @DRovara Just to make you aware in case you missed it: This PR is currently in conflict with main. It may be worth to merge `main` into this branch before continuing adding further changes 😌 

Yeah I noticed. My goal was to first at least make the tests pass for the current version so that I know if errors are caused by the merge or by previous issues. Hopefully I should be able to complete everything tomorrow.

noxfile.py
```diff
@@ -218,6 +218,7 @@ def stubs(session: nox.Session) -> None:
     session.run(
         "uv",
         "sync",
+        "--no-editable",
```
denialhaag: This won't be necessary once https://github.com/munich-quantum-toolkit/core/pull/1864 is merged. I have also created https://github.com/wjakob/nanobind/issues/1391 earlier to ensure that `nanobind.stubgen` is compatible with `scikit-build-core` 1.0.0 out of the box.

That said, I don't dislike it as a solution if https://github.com/wjakob/nanobind/issues/1391 doesn't go through. 🤔 

CC @burgholzer

burgholzer: Yeah. This is an acceptable solution; especially fine. That this is typically run only once anyway and not continuously where one could benefit from editable installs.

mlir/lib/Dialect/QCO/IR/QCOOps.cpp
DRovara: I thought some more on it and I think it is actually better to centralize the code after all. It doesn't really save a lot of duplicated code, but since the number of `UnitaryOps` is likely much higher than the number of `SinkOp`s (and in most cases probably by more than a constant factor), it's really way more efficient to have the transformation target the `SinkOp` instead

mlir/lib/Dialect/QCO/IR/Operations/MeasureOp.cpp
DRovara: Yeah, that's how I implemented it now. I agree.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
DRovara: I ended up moving the program definitions to `qco_programs.cpp` after all (as well as some other programs in this file). Reason for that is mostly so that we can reuse the `measureAndReturn(...)` function from in there, and we never know when we might want to reuse these programs. 
Better to have them in one shared place.

DRovara:
@coderabbitai full review

mlir/unittests/Dialect/QTensor/IR/test_qtensor_ir.cpp
```diff
@@ -189,8 +194,12 @@ TEST_F(QTensorTest, AllocOpStaticTypeWithDynamicSizeOperandFailsVerification) {
...
 /// An alloc immediately followed by dealloc should be eliminated entirely.
 TEST_F(QTensorTest, DeallocOpAllocDeallocPairIsRemoved) {
-  auto canonicalized =
-      buildAndCanonicalize([](QCOProgramBuilder& b) { b.qtensorAlloc(3); });
+  auto canonicalized = buildAndCanonicalize(
+      [](QCOProgramBuilder& b)
+          -> std::pair<SmallVector<Value>, SmallVector<Type>> {
+        b.qtensorAlloc(3);
+        return {{b.intConstant(0)}, {b.getI64Type()}};
+      });
```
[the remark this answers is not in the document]
DRovara: @coderabbitai I'm not 100% sure, but I think that the deallod operation is added automatically by the builder.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -361,6 +400,11 @@ class MLIRQasmImporter final : public qasm3::InstVisitor {
     case qasm3::Int:
     case qasm3::Uint: {
       classicalRegisters[id] = builder.allocClassicalBitRegister(size, id);
+      if ((stmt->isOutput || openQASM2CompatMode) &&
+          sizedType->type == qasm3::Bit) {
+        // We return `output` bits in QASM3, or all named bits in QASM2.
+        outputRegisters.push_back(id);
+      }
```
burgholzer: This is not entirely spec compliant. If an OpenQASM 3 program does not contain any output statements.
The spec states

> Similarly, the output modifier can be used to indicate that one or more variables are to be provided as an explicit output of the quantum procedure. Note that OpenQASM 2 did not allow the programmer to specify that only a subset of its variables should be returned as output, and so it would return all classical variables (which were all creg variables) as output. For compatibility, OpenQASM 3 does not require an output declaration to be provided: in this case it assumes that all of the declared variables are to be returned as output. If the programmer provides one or more output declarations, then only those variables described as outputs will be returned as an output of the quantum process. A variable may not be marked as both input and output.

burgholzer (changes requested):
Thanks @DRovara for the Herculean effort here!
I went through almost everything here. In the process, I significantly simplified the signatures of the tests by automatically deducing the types and providing a couple of shortcuts.
This brought the diff down by 1.5k lines or so.

I still have to go through the tests once more to make sure they are fine as they are. However, there was one thing in the main code that will require changes (if I understood everything correctly). See the inline comment. This should also allow us to revert quite some of the changes in the `qasm_programs.cpp` file.
Could you please take a look at that? (and at the other changes as well; to check whether they make sense to you as well)

DRovara:
Thanks a lot for the updates @burgholzer!

Automatically inferring the return types is smart, no idea how I missed that.

Regarding the built-in method to get corresponding inputs/outputs: I saw that too, but I didn't know we can just cast `Value`s to `OpOperand`. In any case though, it feels a bit inconsistent that the methods have different names and signatures for `IfOp`s, compared to all the unitaries. Is there a reason for that?

Also, regarding the OPENQASM 3 output thing: I can fix that, shouldn't be too much effort. And yeah, that should enable us to rever most `qasm_programs`

burgholzer:
> Regarding the built-in method to get corresponding inputs/outputs: I saw that too, but I didn't know we can just cast `Value`s to `OpOperand`. In any case though, it feels a bit inconsistent that the methods have different names and signatures for `IfOp`s, compared to all the unitaries. Is there a reason for that?

This is mainly following the syntax and wording of the corresponding SCF operation. because the `IfOP` is much more an SCF operation than it is a unitary operation.
We could consider adjusting the naming of the QCO functions as well if we feel like that would help.
 
> Also, regarding the OPENQASM 3 output thing: I can fix that, shouldn't be too much effort. And yeah, that should enable us to rever most `qasm_programs`

Would be great 👍🏼 I can take it from there again.

DRovara:
> Would be great 👍🏼 I can take it from there again.

Update is ready.

burgholzer (approved):
LGTM 👍🏼 
Let's get this in before we cause more conflicts with other open PRs 😅 
Thanks again @DRovara for this nice addition and for all the work put in!

PR #1756, opened 2026-06-02
:sparkles: Add pass and patterns for Classical Control Replacement

DRovara:
## Description

*Part 2 out of 3 of the Qubit Reuse Implementation*

This PR ports the previous implementation of the Classical Control Replacement pass to the new dialect infrastructure.

This pass has the goal of replacing controlled gates with "classically controlled gates" whenever a classical value for the control qubit is available because it has just been measured. 
In those cases, we remove the control from the `CtrlOp` and instead introduce a `qco.ifOp`.

This is one of 3 PRs that will, in total, constitute qubit reuse:
1. Measurement Lifitng
2. **Replacing Classical Controls**
3. Reuse Qubits

AI Notice: Gemini 3.1 and Claude Sonnet 4.6 were used for the generation of some code.

## Checklist


DRovara:
@coderabbit full review

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Added
 
+- ✨ Add Dead Gate Elimination Pattern ([#1755]) ([**@DRovara**])
```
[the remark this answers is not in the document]
DRovara: @coderabbitai I'm just merging in another PR here that has not been merged yet. So this changelog entry is still for that other PR.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -237,4 +237,42 @@ mergeTwoTargetOneParameterWithSwappedTargets(OpType op,
...
+      rewriter.replaceOp(op, i.getQubits());
+      return success();
+    } else {
+      // This currently only includes the `Reset` operation.
+      rewriter.replaceOp(op, op->getOperands());
+      return success();
+    }
+  }
+  return failure();
+}
```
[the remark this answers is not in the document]
DRovara: @coderabbitai can you tell me what specific includes/instructions to add?

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,196 @@
...
+ * applies a phase to the target qubit(s) and does not change their state.
+ * @param op The operation to check
+ * @return true if the operation is a diagonal gate, false otherwise
+ */
+static bool isDiagonal(Operation* op) {
+  if (auto i = dyn_cast<InvOp>(op)) {
+    return isDiagonal(i.getBodyUnitary());
+  }
+  return isa<ZOp, SOp, TOp, POp, SdgOp, TdgOp, IdOp>(op);
+}
```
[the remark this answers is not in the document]
DRovara: @coderabbitai maybe we can change the name of this function, but actually, in this context, we don't look for diagonal gates, but specifically diagonal gates that have `1` as their value in all diagonal entries except for the final one. This does not include rz.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,196 @@
...
+          auto newControls = qubits.slice(0, toKeep.size());
+          auto newTargets =
+              qubits.slice(toKeep.size(), qubits.size() - toKeep.size());
+
+          auto newCtrl =
+              CtrlOp::create(rewriter, op->getLoc(), newControls, newTargets);
+          rewriter.inlineRegionBefore(op.getRegion(), newCtrl.getRegion(),
+                                      newCtrl.getRegion().begin());
+          return newCtrl.getOutputQubits();
+        });
```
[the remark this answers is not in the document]
DRovara: @coderabbitai yes, but aren't there folds for the CtrlOp that automatically cancel `CtrlOp`s without controls? At least that's what happens in the tests.

DRovara:
@coderabbitai full review

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
```diff
@@ -0,0 +1,486 @@
...
+                             programBuilder.getI1Type()});
+  auto q0S0 = programBuilder.allocQubit();
+  auto q1S0 = programBuilder.allocQubit();
+  auto q2S0 = programBuilder.allocQubit();
+
+  auto [q0S1, c0] = programBuilder.measure(q0S0);
+
+  auto [q01S1, q2S1] =
+      programBuilder.ctrl({q0S1, q1S0}, {q2S0},
+                          [&](const ValueRange targets) -> SmallVector<Value> {
```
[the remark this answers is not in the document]
DRovara: Done

DRovara:
@coderabbitai full review

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,197 @@
...
+ * with `if` constructs.
+ */
+struct ReplaceBasisStateControlsWithIfPattern final
+    : mlir::OpRewritePattern<CtrlOp> {
+
+  explicit ReplaceBasisStateControlsWithIfPattern(mlir::MLIRContext* context)
+      : OpRewritePattern(context) {}
+
+  mlir::LogicalResult
+  matchAndRewrite(CtrlOp op, mlir::PatternRewriter& rewriter) const override {
```
burgholzer: This does not quite feel like the most efficient way to implement this pattern.
Instead of matching on CtrlOp, which there will be _many_ of in quantum programs; out of which almost none will potentially trigger the optimization; this pattern should match on measurement operations and check the successor operation. This is much more likely to succeed and requires way fewer checks.
It also makes some of the checks below a little simpler.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,197 @@
...
+    }
+    rewriter.replaceAllUsesWith(control, target);
+    rewriter.modifyOpInPlace(
+        op, [&]() { op.getTargetsInMutable()[0].set(control); });
+    auto dummyTarget = AllocOp::create(rewriter, op->getLoc());
+    rewriter.replaceAllUsesWith(op.getOutputForInput(target), dummyTarget);
+    rewriter.replaceAllUsesWith(op.getOutputForInput(control),
+                                op.getOutputForInput(target));
+    rewriter.replaceAllUsesWith(dummyTarget, op.getOutputForInput(control));
+    rewriter.eraseOp(dummyTarget);
```
burgholzer: This kind of feels inefficient. There has to be a more elegant way than introducing a dummy here.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,197 @@
...
+ * @return true if the operation is a diagonal gate, false otherwise
+ */
+static bool isPhaseGate(Operation* op) {
+  if (op == nullptr) {
+    return false;
+  }
+  if (auto i = dyn_cast<InvOp>(op)) {
+    return isPhaseGate(utils::getSoleBodyUnitary<UnitaryOpInterface>(
+        *op->getRegion(0).getBlocks().begin()));
+  }
```
burgholzer: I am not sure this will ever be needed. Canonicalization eliminates all `inv{diag}` cases so that none of them should survive. Might simplify the code a little.

burgholzer (changes requested):
Thanks @DRovara 🙏🏼 
I haven't checked all of the PR's code yet because I have a fairly general request for changes already. I'll check in more detail once that is addressed.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
DRovara: Yeah, I think the new implementation is a lot cleaner now. Let's still wait what the tests say (maybe I'm using undefined behaviour, not 100% sure), but if everything passes I'm happy with the new version.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
DRovara: I did that, but unfortunately it doesn't really make the code much cleaner.

I could rewrite the pass to focus on the entire process from the perspective of the measurement. However, that becomes a bit awkward if we have multiple controls that can be replaced. Because then the first call of the pattern would replace one control with an `IfOp`, but then all of a sudden, for the remaining `MeasureOp`s, the `CtrlOp` would no longer be the direct successor and we would have to implement some value forwarding through the `IfOp` block parameters.

So what I instead did was: The pattern now targets `MeasureOp`s, but immediately looks for a successor `CtrlOp` and continues from there with the old code. This has the added benefit of only requiring a single pattern execution per `CtrlOp` and not relying on intermediate canonicalizer calls, but of course the code is a little bit less intuitive that way. What do you think?

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,205 @@
...
+  mlir::LogicalResult
+  matchAndRewrite(MeasureOp measure,
+                  mlir::PatternRewriter& rewriter) const override {
+    auto op = dyn_cast<CtrlOp>(*measure.getQubitOut().getUsers().begin());
+    if (!op) {
+      return mlir::failure();
+    }
+    rewriter.setInsertionPointAfter(op);
+
+    if (op.getNumBodyUnitaries() == 1 && isPhaseGate(op.getBodyUnitary(0))) {
```
burgholzer: This can be slightly improved by using the getSoleBodyUnitary method, which avoids one extra traversal of the CtrlOp body.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,205 @@
...
+      trySwapControlsOfDiagonalGate(op, rewriter);
+    }
+
+    SmallVector<std::pair<mlir::Value, mlir::Value>> toReplace;
+    SmallVector<mlir::Value> toKeep;
```
burgholzer: you can drop _a lot_ (if not all) of the `mlir::` namespace qualifiers throughout this file.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,205 @@
...
+          return newCtrl.getOutputQubits();
+        });
+
+    for (auto replace : toReplace) {
+      rewriter.replaceAllUsesWith(op.getOutputForInput(replace.first),
+                                  replace.first);
+    }
+    for (auto [oldInput, result] : llvm::zip(allQubits, ifOp.getResults())) {
+      rewriter.replaceAllUsesWith(op.getOutputForInput(oldInput), result);
+    }
```
burgholzer: the `getOutputForInput` calls here feel slightly inefficient as each of them involves a map lookup. We know exactly which original value corresponds to which new value, so I believe this can be written more explicitly.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+/**
+ * @brief Checks if the given operation is a phase gate, i.e., it only
+ * applies a phase to the target qubit(s) in the 1 state.
+ * @param op The operation to check
+ * @return true if the operation is a diagonal gate, false otherwise
+ */
+static bool isPhaseGate(Operation* op) {
+  return isa<ZOp, SOp, TOp, POp, SdgOp, TdgOp, IdOp>(op);
```
burgholzer, line 56: this is missing the RZZ gate, although the implementation currently only handles single-qubit gates. I do not think that controlled RZZ gates are too important practically; so this is not a high priority.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,205 @@
...
+  size_t controlIndex = 0;
+  for (auto control : op.getControlsIn()) {
+    auto controlOutcome = getPredecessorMeasurementOutcome(control);
+    if (controlOutcome) {
+      controlIndex++;
+      continue;
+    }
+
+    Value controlOut = op.getOutputForInput(control);
+    Value targetOut = op.getOutputForInput(target);
```
burgholzer: No need for a map lookup on this one. This is simply

```suggestion
    Value targetOut = op.getTargetsOut()[0];
```

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  size_t controlIndex = 0;
+  for (auto control : op.getControlsIn()) {
+    auto controlOutcome = getPredecessorMeasurementOutcome(control);
+    if (controlOutcome) {
+      controlIndex++;
+      continue;
+    }
+
+    Value controlOut = op.getOutputForInput(control);
```
burgholzer: Also no need for a map lookup here. You are tracking the control index already

```suggestion
    Value controlOut = op.getControlsOut()[controlIndex];
```

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
```diff
@@ -0,0 +1,561 @@
...
+using namespace mlir;
+using namespace mlir::qco;
+
+class QCOReplaceClassicalControlsTest : public testing::Test {
+
+protected:
+  MLIRContext context;
+  QCOProgramBuilder programBuilder;
+  QCOProgramBuilder referenceBuilder;
+  OwningOpRef<ModuleOp> module;
```
burgholzer: We try to avoid the variable name `module` as it is a keyword in C++20 (still do not understand why this is not a straight up compilation error) and the syntax highlighting is weird if that name is used. If you look in the surrounding files, you'll find alternative suggestions.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
```diff
@@ -0,0 +1,561 @@
...
+  auto r0 = referenceBuilder.allocQubit();
+  auto r1 = referenceBuilder.allocQubit();
+
+  Value cr0;
+  std::tie(r0, cr0) = referenceBuilder.measure(r0);
+
+  r1 = referenceBuilder.qcoIf(
+      cr0, {r1}, [&](ValueRange qubits) -> SmallVector<Value> {
+        return SmallVector<Value>{referenceBuilder.x(qubits[0])};
+      })[0];
```
burgholzer: We recently updated the builders of modifiers to also support a shortcut for single-qubit bodies that avoids the boilerplate of wrapping everything in SmallVectors. 
If it turns out that we have not added such methods to the IfOp, this may be the perfect opportunity to do so and simplify the respective code.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
```diff
@@ -0,0 +1,561 @@
...
+                               referenceBuilder.getI1Type(),
+                               referenceBuilder.getI1Type()});
+  auto r0 = referenceBuilder.allocQubit();
+  auto r1 = referenceBuilder.allocQubit();
+  auto r2 = referenceBuilder.allocQubit();
+
+  Value cr0;
+  std::tie(r0, cr0) = referenceBuilder.measure(r0);
+  Value cr1;
+  std::tie(r1, cr1) = referenceBuilder.measure(r1);
```
burgholzer, line 698: Just an observation: we may need/should make these programs a little more complex. I could foresee a canonicalization that removes measurements directly after alloc operations (and replaces the classical values by fixed Booleans), which would trivialize the program here further.

I think the fix is rather easy; adding Hadamard's to the qubits should solve it and make the programs non-trivial no matter the optimization.

burgholzer (changes requested):
Thanks for the iteration @DRovara 🙏🏼 
I think this direction is fine for now. We can always try to improve later.
Most of the comments are on little performance or usability aspects. Should not be too hard to address. I am fairly optimistic that we can merge this afterwards.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
DRovara: Ah good point, that was still a relic of copying the old code over.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
DRovara: That's true but I'm not even sure if this is compatible with the RZZ gate. For instance, it also doesn't work with RZ. It's not enough that the gate is diagonal, it has to have a entry of 1 for all diagonal entries except for the last one, which does not apply to RZZ. Maybe it could be extended to other gates, but I'd say we keep it like this for now.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
DRovara: Gotcha. I think I originally copied this from another test file, but maybe I'm also just imagining that

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
DRovara: Didn't exist yet, but I have created the wrapper for it now.

mlir/unittests/Dialect/QCO/Transforms/Optimizations/test_qco_replace_classical_controls.cpp
DRovara: That's a very good point. Better make these preparations early rather than a huge fixing PR again. I have added `H` gates to all qubits.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
DRovara: The problem is, it's actually kind of hard to track exactly where all these values are coming from. We are slicing the set of the qubits quite a bit, and afterwards, it's not easy to find the matching qubits in the outputs of the original operation.

We could fix that by tracking all that information in additional vectors, but honestly, that requires more variables/memory and also uglier code, which feels worse than the map lookup, which of course is slower asymptotically, but also, since these instances will be so small (how many gates with 1000 controls will we really have?) it shouldn't be that costly.

I tried to ask AI for a cleaner implementation, but Claude Sonnet 5 at least didn't manage to find anything that doesn't require even more lines of code or vectors.

DRovara: Update: With some prompt engineering I managed to get Claude to implement it in a clean manner that only uses slightly more vectors.

mlir/lib/Dialect/QCO/Transforms/Optimizations/ReplaceClassicalControls.cpp
burgholzer: Yeah. Let's leave this as is for now.
However, the name is slightly misleading in case this does not apply to RZ at the moment. Because RZ is by all means a diagonal gate.
It's fine for me to postpone this to a follow-up, but we should open a tracking issue for that.

burgholzer (approved):
Pushed a couple of adjustments to the PR. LGTM now 👍🏼 
Also opened #1928 for a follow-up investigation.

PR #1766, opened 2026-06-05
Add QIR Program Format Support to the DDSIM QDMI Device

### LINKED ISSUE #1695: ✨ Add QIR Program Format Support to the QDMI DDSim Device
> [!NOTE]
> 
> This issue is part of [unitaryHACK 2026](https://unitaryhack.dev/); note that you have to be [registered](https://unitaryhack.dev/register/) to complete this issue. Learn more about the PR review process [here](https://unitaryhack.dev/hacker-guide/). 
>
> A message from unitaryHACK about [AI Slop](https://en.wikipedia.org/wiki/AI_slop): "While we allow LLMs as collaborative tools, fully AI-generated PRs are prohibited. Maintainers have the right to reject any contribution that appears to be automated or lacks human oversight."
>
> Learn more about unitaryHACK rules and opportunities [here](https://unitaryhack.dev/rules/) and about the usage of AI, in particular, [here](https://unitaryhack.dev/ai-guide/)!

### Description

The MQT Core QDMI DD device (`include/mqt-core/qdmi/devices/dd/Device.hpp`, `src/qdmi/devices/dd/Device.cpp`) is a Decision-Diagram-based quantum simulator exposed via the [Quantum Device Management Interface (QDMI)](https://github.com/Munich-Quantum-Software-Stack/QDMI) standard.

Currently, the device supports only **OpenQASM 2** and **OpenQASM 3** as input program formats (`QDMI_PROGRAM_FORMAT_QASM2`, `QDMI_PROGRAM_FORMAT_QASM3`). The QDMI standard also defines four [Quantum Intermediate Representation (QIR)](https://www.qir-alliance.org/) formats:

| Format constant | Description |
|---|---|
| `QDMI_PROGRAM_FORMAT_QIRBASESTRING` | QIR Base Profile as LLVM IR text (`.ll`) |
| `QDMI_PROGRAM_FORMAT_QIRBASEMODULE` | QIR Base Profile as LLVM bitcode binary (`.bc`) |
| `QDMI_PROGRAM_FORMAT_QIRADAPTIVESTRING` | QIR Adaptive Profile as LLVM IR text (`.ll`) |
| `QDMI_PROGRAM_FORMAT_QIRADAPTIVEMODULE` | QIR Adaptive Profile as LLVM bitcode binary (`.bc`) |

All four currently return `QDMI_ERROR_NOTSUPPORTED` (see `test/qdmi/devices/dd/job_parameters_test.cpp`, `ProgramFormatSupport` test).

The goal of this issue is to implement support for **at minimum** `QDMI_PROGRAM_FORMAT_QIRBASESTRING` and `QDMI_PROGRAM_FORMAT_QIRBASEMODULE`, so that a QIR program can be submitted to and executed by the DD device just like an OpenQASM program.

MQT Core already includes a fully working QIR execution stack:
- **QIR Runtime** (`src/qir/runtime/`, `include/mqt-core/qir/runtime/`) — a DD-based implementation of QIR's `__quantum__qis__*` and `__quantum__rt__*` C API.
- **QIR Runner** (`src/qir/runner/Runner.cpp`) — a standalone CLI tool using LLVM OrcJIT to JIT-compile and execute `.ll`/`.bc` files against the runtime.

This issue is about integrating that execution stack into the QDMI device's `submit()` / `getResults()` pipeline.

---

### Background Reading

Before diving into implementation, make sure you understand the existing code:

1. **Familiarise yourself with the QDMI DD device** — go through `include/mqt-core/qdmi/devices/dd/Device.hpp` and `src/qdmi/devices/dd/Device.cpp`. Pay close attention to `MQT_DDSIM_QDMI_Device_Job_impl_d::setParameter()`, `submit()`, and `getResults()` — these are the three entry points you will be extending.

2. **Study the existing QIR runner** — skim through `src/qir/runner/Runner.cpp`. Understand how it uses `llvm::parseIRFile`, `llvm::orc::LLJIT`, and `llvm::sys::DynamicLibrary::AddSymbol` to JIT-compile a module and resolve QIR intrinsic symbols to the runtime. Note that it reads from a _file path_; for the device, the program arrives as an in-memory string or byte buffer, respectively.

3. **Study the QIR Runtime** — go through `include/mqt-core/qir/runtime/Runtime.hpp` and `src/qir/runtime/Runtime.cpp`. Understand the `qir::Runtime` singleton: how it manages qubit state (`qState`, `dd`), how measurement results are stored in `rRegister`, and what `Runtime::reset()` does. Also look at `include/mqt-core/qir/runtime/QIR.h` for the full list of C-level QIR intrinsics.

4. **Run the existing tests** — build and run `test/qdmi/devices/dd/job_parameters_test.cpp` and `test/qdmi/devices/dd/results_sampling_test.cpp` to confirm the baseline. Try running the QIR runner binary on `test/qir/BellPairStatic.ll` to see the existing execution path in action. For more information on how to run the test suite, see the [developer documentation](https://mqt.readthedocs.io/projects/core/en/latest/contributing.html#working-on-the-c-library). For how to build and use the QIR runner the [corresponding page in the documentation](https://mqt.readthedocs.io/projects/core/en/latest/qir/index.html).

5. **Read the QIR Base Profile spec** (optional but helpful) — https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Base_Profile.md

---

### Tasks

6. **In-memory module parsing** — The runner uses `llvm::parseIRFile(path, ...)` which reads from disk. For the QDMI device, the QIR program arrives as a `std::string` (LLVM IR text for `QIRBASESTRING`) or raw bytes (bitcode for `QIRBASEMODULE`). LLVM provides `llvm::parseIR(llvm::MemoryBufferRef, ...)` for parsing from a buffer — verify this works for both text and bitcode and decide how to handle the two format variants.

7. **Result collection from JIT-executed QIR** — After `main()` returns, measurement results reside in `qir::Runtime`'s private `rRegister` (`std::unordered_map<Result*, ResultStruct>`). There is currently no public accessor. Add a public `getResults()` / `getMeasurementBitstring()` accessor to `qir::Runtime`.

8. **Sampling (multiple shots)** — For `numShots_ > 0`, the QIR program must be re-executed `numShots_` times and results aggregated.

9. **CMake / LLVM dependency gating** — Linking `MQT::CoreQIRRuntime` and LLVM OrcJIT libs into `mqt-core-qdmi-ddsim-device` significantly increases the build footprint. Add a CMake option (e.g. `BUILD_MQT_CORE_QDMI_WITH_QIR`) so that QIR format support is opt-in.

10. **Accept QIR formats in `setParameter()`** — In `src/qdmi/devices/dd/Device.cpp`, extend `MQT_DDSIM_QDMI_Device_Job_impl_d::setParameter()` to allow `QDMI_PROGRAM_FORMAT_QIRBASESTRING` and `QDMI_PROGRAM_FORMAT_QIRBASEMODULE` (remove the `QDMI_ERROR_NOTSUPPORTED` guard for those two). For `QIRBASEMODULE`, the program parameter carries raw bytes rather than a null-terminated string — handle `program_` storage accordingly (the current `size - 1` substring logic assumes a C string; bitcode is binary).

12. **Implement QIR execution in `submit()`** — In `submit()`, add a branch alongside the existing OpenQASM path.

---

### Testing

11. **Update `job_parameters_test.cpp`** — Move `QDMI_PROGRAM_FORMAT_QIRBASESTRING` and `QDMI_PROGRAM_FORMAT_QIRBASEMODULE` from the "Unsupported" list to the "Supported" list in the `ProgramFormatSupport` test.

12. **Add QIR sampling tests** — Add new test cases in `test/qdmi/devices/dd/` (mirroring `results_sampling_test.cpp`) that submit a simple QIR Base Profile program (e.g. a Bell pair from `test/qir/BellPairStatic.ll`) as `QIRBASESTRING`, run with `N` shots, and assert the resulting histogram has the correct support (`"00"` and `"11"` states only) and sums to `N`. Repeat for `QIRBASEMODULE` using a pre-compiled `.bc` file.

13. **Guard tests with CMake option** — Ensure the new tests are only compiled and run when that option is enabled.

rturrado:
## Description

First contribution to UnitaryHACK for me! 🎉 And man, I've enjoyed working on this one! 😀

> [!NOTE]
> The current PR still does not cover all the functionality proposed in the #1695 issue. In particular, the support for QIR adaptive base modules and strings is not yet implemented. But the current solution is ready for review.

There are two main pieces of work:
1. The introduction of an MQT Core QIR JIT library (`MQT::CoreQIRJIT`, `src/qir/jit`). This library contains almost all of the previous `Runner.cpp` code: LLVM JIT initialization, QIR runtime symbols registration, and program execution. Compared to that previous Runner code, it also lets you parse an IR directly from a string, and not only from a file. The main purpose of the QIR JIT library is to encapsulate the LLVM JIT code so that it can also be used from the QDMI device. The interface is a `qir::jit::Session` type that can be constructed from an input file or from a byte string (text for LLVM assembly or binary for LLVM bitcode), and then run with or without arguments (e.g., the Runner CLI can run a JIT program with arguments, and the QDMI device without arguments).
2. The job submission in `Device.cpp`. By using the MQT Core QIR JIT library, the device can just run a JIT program, get the results, and process them (filter zero and one addresses, sort the addresses, and transform the measurements into a bit string).

The new functionality is tested via two tests in `results_sampling_test.cpp`, one that exercises the QIR base module path, and the other that exercises the QIR base string path.

For this PR I have used Claude Opus 4.7, mainly from a CLion terminal, and for different purposes: from understanding existing code, to drafting code snippets, catching bugs and typos, running clang-format and clang-tidy and proposing fixes for clang-tidy warnings, reviewing wording in commit messages and docs...

Fixes #1695

## Checklist

- [X] I have added appropriate tests that cover the new/changed functionality.
- [X] I have updated the documentation to reflect these changes.
- [X] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [X] I have added migration instructions to the upgrade guide (if needed).
- [X] The changes follow the project's style guidelines and introduce no new warnings.
- [X] The changes are fully tested and pass the CI checks.


docs/qir/index.md
```diff
@@ -36,3 +36,9 @@ The `mqt-core-qir-runner` can be used to execute a QIR file (typically with a `.
 
 This will simulate the circuit and print the measurement results to the console.
 The runner supports the QIR Base Profile.
+
+### QIR Support in the QDMI Device
+
+The QDMI Device accepts jobs in the following program formats: QASM2, QASM3, QIR base module (LLVM bitcode), and QIR base string (LLVM assembly).
```
ystade: To me, it is not immediately clear that "base" refers to QIR's Base Profile. Perhaps this sentence can be made more precise, similar to the sentence referring to the Base Profile in the paragraph above.

include/mqt-core/qdmi/devices/dd/Device.hpp
```diff
@@ -234,6 +234,14 @@ struct MQT_DDSIM_QDMI_Device_Job_impl_d {
   auto getProbabilities(size_t size, void* data, size_t* sizeRet)
       -> QDMI_STATUS;
 
+  /// Helper function to submit a QASM 2 or QASM 3 program
+  auto submitQASMProgram() -> QDMI_STATUS;
+
+#ifdef BUILD_MQT_CORE_QDMI_WITH_QIR
```
ystade: We have other QDMI Device implementations without a simulator backend in MQT Core. I am wondering whether this option is confusing and should rather refer to the DDSim QDMI Device explicitly by renaming this option to `BUILD_MQT_CORE_DDSIM_QDMI_WITH_QIR`. However, this option is exceptionally long then.
I would be interested in @burgholzer opinion whether we could even drop this option and always build the DDSim QDMI Device with QIR support whenever the MLIR build is enabled, i.e., `BUILD_MQT_CORE_MLIR`?

src/qdmi/devices/dd/CMakeLists.txt
```diff
@@ -40,6 +40,10 @@ if(NOT TARGET ${TARGET_NAME})
   # Add link libraries
   target_link_libraries(${TARGET_NAME} PRIVATE MQT::CoreDD MQT::CoreQASM MQT::CoreCircuitOptimizer
                                                MQT::CoreQDMICommon spdlog::spdlog)
+  if(BUILD_MQT_CORE_QDMI_WITH_QIR)
+    target_link_libraries(${TARGET_NAME} PRIVATE MQT::CoreQIRJIT MQT::CoreQIRRuntime)
```
ystade, line 44: Is it necessary to link against both, `MQT::CoreQIRJIT` and `MQT::CoreQIRRuntime`? I would expect that one is included in the other.

src/qdmi/devices/dd/Device.cpp
```diff
@@ -346,15 +358,20 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::setParameter(
...
+      ) {
         return QDMI_ERROR_NOTSUPPORTED;
       }
       format_ = format;
     }
     return QDMI_SUCCESS;
   case QDMI_DEVICE_JOB_PARAMETER_PROGRAM:
     if (value != nullptr) {
-      program_ = std::string(static_cast<const char*>(value), size - 1);
+      program_ = std::string(static_cast<const char*>(value), size);
```
ystade: Why this change? As far as I can see, the '-1' strips off the terminating null-character at the end.

src/qdmi/devices/dd/Device.cpp
```diff
@@ -425,6 +438,80 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::submit() -> QDMI_STATUS {
...
+      format_ == QDMI_PROGRAM_FORMAT_QASM3) {
+    return submitQASMProgram();
+  }
+#ifdef BUILD_MQT_CORE_QDMI_WITH_QIR
+  if (format_ == QDMI_PROGRAM_FORMAT_QIRBASEMODULE ||
+      format_ == QDMI_PROGRAM_FORMAT_QIRBASESTRING) {
+    return submitQIRProgram();
+  }
+#endif
+  return QDMI_ERROR_NOTSUPPORTED;
```
ystade: This line should not be reachable after all because I would expect the `NOTSUPPORTED` already when trying to set an unsupported program format. However, it feel slightly wrong to see `NOTSUPPORTED` here, `QDMI_ERROR_FATAL` feel more natural to me. A very nitpick...

src/qdmi/devices/dd/Device.cpp
```diff
@@ -425,6 +438,80 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::submit() -> QDMI_STATUS {
   }
   return QDMI_SUCCESS;
 }
+#ifdef BUILD_MQT_CORE_QDMI_WITH_QIR
+auto MQT_DDSIM_QDMI_Device_Job_impl_d::submitQIRProgram() -> QDMI_STATUS {
+  if (numShots_ == 0) {
+    return QDMI_ERROR_INVALIDARGUMENT;
```
ystade: Is this really not possible? In the QASM case, we take advantage of DDSim being a simulator rather than a real hardware device and when `numShots_` is set to `0`, one can query the state vector after the job has finished. Would something like that be possible for QIR as well? Sidenote: This obviously only works for the base profile, not any other QIR profile.

src/qir/jit/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+
+set(TARGET_NAME ${MQT_CORE_TARGET_NAME}-qir-jit)
+
+if(NOT TARGET ${TARGET_NAME})
+  # Add QIRJIT library
+  add_mqt_core_library(${TARGET_NAME} ALIAS_NAME QIRJIT)
+
+  # Add sources to target
+  target_sources(${TARGET_NAME} PRIVATE Session.cpp
+                                        ${MQT_CORE_INCLUDE_BUILD_DIR}/qir/jit/Session.hpp)
```
ystade: As far as I can see, the header should not be listed here because it is included as a public header in the lines below.

include/mqt-core/qir/jit/Session.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <llvm/ADT/StringRef.h>
+#include <llvm/ExecutionEngine/Orc/LLJIT.h>
+#include <llvm/ExecutionEngine/Orc/ThreadSafeModule.h>
+#include <llvm/IR/LLVMContext.h>
+#include <llvm/Support/Error.h>
+
+#include <memory>
+#include <string>
+
+namespace qir::jit {
```
ystade: We usually do not have super fine-grained namespaces. Just `qir` should be fine (if no conflict occurs).

CMakeLists.txt
```diff
@@ -121,6 +121,9 @@ endif()
 cmake_dependent_option(BUILD_MQT_CORE_QIR_RUNNER "Build the QIR runner of the MQT Core project" ON
                        "BUILD_MQT_CORE_MLIR" OFF)
 
+cmake_dependent_option(BUILD_MQT_CORE_QDMI_WITH_QIR "Enable QIR format support for QDMI" OFF
```
ystade: Additional to the discussion above, in case we keep this option and agnostic of its naming, at least the string after it should mention that this option is only relevant for the DDSim QDMI Device, not the other QDMI devices.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Add initial infrastructure for new QC and QCO MLIR dialects
   ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1542], [#1548], [#1550], [#1554], [#1567], [#1569], [#1570], [#1572], [#1573], [#1580], [#1602], [#1620], [#1623], [#1624], [#1626], [#1627], [#1635], [#1638], [#1673], [#1675], [#1700], [#1717], [#1728], [#1730], [#1749], [#1762], [#1765])
   ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**], [**@Ectras**], [**@MatthiasReumann**], [**@simon1hofmann**])
+- ✨ Add QIR program format support to the QDMI DDSim device ([#1766]) ([**@rturrado**])
```
ystade: We add new lines in the changelog to the top, not to the bottom.

test/qdmi/devices/dd/job_parameters_test.cpp
```diff
@@ -100,10 +106,12 @@ TEST(JobParameters, ProgramFormatSupport) {
...
   // Unsupported → NOTSUPPORTED
   for (QDMI_Program_Format fmt : {
-           QDMI_PROGRAM_FORMAT_QIRBASESTRING,
-           QDMI_PROGRAM_FORMAT_QIRBASEMODULE,
            QDMI_PROGRAM_FORMAT_QIRADAPTIVESTRING,
            QDMI_PROGRAM_FORMAT_QIRADAPTIVEMODULE,
+#ifndef BUILD_MQT_CORE_QDMI_WITH_QIR
+           QDMI_PROGRAM_FORMAT_QIRBASEMODULE,
+           QDMI_PROGRAM_FORMAT_QIRBASESTRING,
+#endif
```
ystade, line 116:
```suggestion
#ifndef BUILD_MQT_CORE_QDMI_WITH_QIR
           QDMI_PROGRAM_FORMAT_QIRBASEMODULE,
           QDMI_PROGRAM_FORMAT_QIRBASESTRING,
#endif
QDMI_PROGRAM_FORMAT_QIRADAPTIVESTRING,
QDMI_PROGRAM_FORMAT_QIRADAPTIVEMODULE,
```

Very minor nitpick. I would keep the ordering nonetheless.

test/qdmi/devices/dd/results_sampling_test.cpp
```diff
@@ -21,24 +21,70 @@
...
+    ASSERT_EQ(qdmi_test::setShots(j.job, shots), QDMI_SUCCESS);
+    ASSERT_EQ(qdmi_test::submitAndWait(j.job, 0), QDMI_SUCCESS);
+
+    auto [keys, vals] = qdmi_test::getHistogram(j.job);
+    ASSERT_EQ(keys.size(), vals.size());
+    size_t sum = 0U;
+    for (const auto& v : vals) {
+      sum += v;
+    }
+    EXPECT_EQ(sum, shots);
```
ystade: Is there a reason why this cannot be a simple function in the anonymous namespace? I would have a slight tendency not to intermingle the class and the function if there is no need for it, but I might overlook some dependency of GTest here.

test/qdmi/devices/dd/results_sampling_test.cpp
```diff
@@ -21,24 +21,70 @@
...
+  Run(format, program);
+}
+
+#ifdef BUILD_MQT_CORE_QDMI_WITH_QIR
+TEST_F(HistogramKeysAndValuesSumToShots, QIRBaseModule) {
+  const QDMI_Program_Format format = QDMI_PROGRAM_FORMAT_QIRBASEMODULE;
+  const std::string_view program = qdmi_test::QIR_BELL_PAIR_STATIC;
+  llvm::LLVMContext context;
+  llvm::SMDiagnostic err;
+  auto module = llvm::parseAssemblyString(program, err, context);
```
ystade: I am actually surprised this works with `module` as a variable name here. Since C++20, `module` is a reserved keyword. Since, we are using a C++20 in the project we should avoid it.

test/qdmi/devices/dd/helpers/test_utils.cpp
```diff
@@ -98,8 +98,7 @@ int setProgram(MQT_DDSIM_QDMI_Device_Job job, const QDMI_Program_Format fmt,
     return rc;
   }
   rc = MQT_DDSIM_QDMI_device_job_set_parameter(
-      job, QDMI_DEVICE_JOB_PARAMETER_PROGRAM, program.size() + 1,
-      program.data());
+      job, QDMI_DEVICE_JOB_PARAMETER_PROGRAM, program.size(), program.data());
```
ystade: For a string, also the terminating null-character must be given as a parameter, hence the `+1`. However, I think, there was a bug in the previous implementation, and it should actually be `program.c_str()` to ensure null-char termination.

include/mqt-core/qir/jit/Session.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+  static void registerRuntimeSymbols();
+  static void initNativeTargets();
+  llvm::Expected<llvm::orc::ThreadSafeModule>
+  loadModuleFromFile(llvm::StringRef irPath);
+  llvm::Expected<llvm::orc::ThreadSafeModule>
+  loadModuleFromMemory(llvm::StringRef irBytes, llvm::StringRef bufferName);
+  void initialize();
+  void deinitialize();
+};
```
ystade: This class is missing docstrings. Especially, because its name is quite generic, these would improve the understanding a lot.

ystade (changes requested):
Hi @rturrado,
Your contribution already looks great in this state. Thanks for taking part in the unitaryHACK. I am a big fan of the restructuring, i.e., having the JIT interpreter as its own library. This makes the code a lot better structured. It is actually also cool to see that this refactoring comprises the most part of this PR, and the remaining changes are quite local. The biggest action item that remains in my mind is the documentation of the Session class. In particular, I reviewed all files in detail except the session class because the documentation eases the review here a lot. So, in summary, my biggest request are comprehensive docstrings for the `Session` class. Overall, this contribution looks already very good and is very valuable.

include/mqt-core/qdmi/devices/dd/Device.hpp
burgholzer: I'd be in favor of building the DDSIM device with QIR support by default (as long as building MLIR is enabled).
Since we will soon (™️) also enable that flag for Python, MLIR will be available anywhere the next time we make a release.

So overall, I think it would be best to not add an extra option here.

test/qdmi/devices/dd/results_sampling_test.cpp
burgholzer: Just for context: I believe it is a reserved keyword only in certain contexts. Still best to avoid it altogether. Triggers weird syntax highlighting on GitHub 😅

burgholzer:
@rturrado Just came to say this is a great initial draft! Keep it up 💪🏻

rturrado:
> Hi @rturrado, Your contribution already looks great in this state. Thanks for taking part in the unitaryHACK. I am a big fan of the restructuring, i.e., having the JIT interpreter as its own library. This makes the code a lot better structured. It is actually also cool to see that this refactoring comprises the most part of this PR, and the remaining changes are quite local. The biggest action item that remains in my mind is the documentation of the Session class. In particular, I reviewed all files in detail except the session class because the documentation eases the review here a lot. So, in summary, my biggest request are comprehensive docstrings for the `Session` class. Overall, this contribution looks already very good and is very valuable.

Hi @ystade ! Many thanks for the comments and for the thorough review. I'll go over each of the items in the following days.

There are still a few things I want to tackle as part of this PR, on top of whatever you consider necessary:
- Runner is not try/catching the JIT session. ✔
- `submitQIRProgram` calls Runtime's `getResults` but then does a lot of post processing (filter zero and one addresses, order addresses, and convert to bit string). I would prefer part of that functionality to be done by the Runtime itself (filter and, optionally, order) and by a utility function (toBitString). ✔
- Outputting to an ostream, not just to standard output. ✔
- Dynamic modules. This is not tested in the current implementation. ✔
- Adaptive program formats. Same. Currently untested. ✔
- Output label at `result_record_output` depending on the output schema. ❓
- Code coverage errors.

rturrado:
A question for you, guys!

The last commits add support for QIR adaptive formats. These formats allow the client to record and output values other than `Result*`, e.g., `bool`, `int`, `float`... At the moment, the Runtime is *outputting* all these values, but __*recording* only bit values coming from measurement results__ (i.e., the Boolean `ResultStruct.r`). It just keeps those results as a bit string because the only consumer of recorded outputs is the QDMI DDsim device, via `submitQIRprogram`, which gets the bit string and updates the `counts_` map with it.

The question would be if we also want to record the other possible values, and if we want to do it as part of this PR. A possible solution would be to change that `std::string recordedBitResults` data structure into something like a `llvm::SmallVector<std::variant<bit, bool, int64_t, double, TupleHeader, ArrayHeader>>` with the required accessors. I don't think that the implementation would be very difficult; I just don't know if it would be necessary.

burgholzer:
> A question for you, guys!
> 
> 
> 
> The last commits add support for QIR adaptive formats. These formats allow the client to record and output values other than `Result*`, e.g., `bool`, `int`, `float`... At the moment, the Runtime is *outputting* all these values, but __*recording* only bit values coming from measurement results__ (i.e., the Boolean `ResultStruct.r`). It just keeps those results as a bit string because the only consumer of recorded outputs is the QDMI DDsim device, via `submitQIRprogram`, which gets the bit string and updates the `counts_` map with it.
> 
> 
> 
> The question would be if we also want to record the other possible values, and if we want to do it as part of this PR. A possible solution would be to change that `std::string recordedBitResults` data structure into something like a `llvm::SmallVector<std::variant<bit, bool, int64_t, double, TupleHeader, ArrayHeader>>` with the required accessors. I don't think that the implementation would be very difficult; I just don't know if it would be necessary.

Hm, very good question.
I believe the limiting factor for this at the moment is actually QDMI because it does not really expose any meaningful result types that would make sense except for measurement results.
The reasonable things are SHOTS and COUNTS.
I don't recall seeing many programs that output something else. Have you maybe come across some that seemed reasonable?
How does the official QIR runner handle output recording/results collection? 

I generally believe that for the sake of this PR, it is fine to focus on the things that can reasonably be piped through QDMI as results.
Technically, one could use a CUSTOM result type to expose the full output recording (which is probably what the QIR runner does). That might be a reasonable compromise here.

Does that make sense?

rturrado:
Thanks for the answer, @burgholzer.

> I don't recall seeing many programs that output something else. Have you maybe come across some that seemed reasonable?

No, not really. I asked Claude today to think about a realistic scenario for using a QIR program in adaptive format, and it came up with a Hamming weight, measuring 3 qubits, and calculating an int weight and a float mean, which I found nice.

But, back to my original question, I think that I was understanding wrongly the QIR concept of recording outputs of different types. I believe now that the spec only refers to logging labels and values somewhere (e.g., an output stream). And that, if we decide to use a data structure for keeping the measurements bit string, that is something completely apart. My confusion came from the idea that we were requested by the QIR spec to store bools, and ints, and records, and so on, into that data structure.

> How does the official QIR runner handle output recording/results collection?

The current QIR runner can only record `Result*`, and doesn't support [output schemas](https://github.com/qir-alliance/qir-spec/tree/main/specification/output_schemas). That means that it can only print measured qubits, and in a custom way, e.g., `r0: 0`. This PR is adding support both for [adaptive profiles](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md)  and output schemas (both to a certain extent), so it can record more data types and in a spec-compliant way, let's say.

ystade:
@rturrado, it is cool to see when the software components that we have, e.g., the runner, become more spec-compliant. In my opinion, to keep the PR focused and avoid overloading it, it should be fine to restrict it to recording results and leave other data types for future Issues/PRs.

ystade:
@rturrado Actually, I am not entirely sure whether I understood your last message to 100%. Is your initial question resolved? If not, could you rephrase it?

rturrado:
> @rturrado, it is cool to see when the software components that we have, e.g., the runner, become more spec-compliant. In my opinion, to keep the PR focused and avoid overloading it, it should be fine to restrict it to recording results and leave other data types for future Issues/PRs.

You're absolutely right here, guys. I think I got a bit overexcited with the coding and went beyond the scope of this PR. The current QIR runner functionality should be a good reference for what the QDMI device can do.

That means that some of the points from my previous [comment](https://github.com/munich-quantum-toolkit/core/pull/1766#issuecomment-4640498509) had better be addressed as follow-up PRs:

> * Outputting to an ostream, not just to standard output.

This is already implemented (commit `Make Runtime's output stream injectable`). At the moment, when a result is recorded, the QIR API prints to standard output. This piece of code allows you to define an output stream to write to.

> * Adaptive program formats.

This is also already implemented. It could be divided into 2 parts, one of which could go into this PR because the issue description asks for QDMI devices to support Adaptive program formats.
- That first part would let a QDMI device run a QIR program in Adaptive format (commit `Add Adaptive QIR profile support to the DDSIM QDMI device`) just like the current QIR runner does. The only requirements here are for a QIR program to be able to perform mid-circuit measurement instructions and branch conditionally on a measurement outcome (points 2 and 3 of the [spec](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md)). This is already supported by the current QIR JIT implementation.
- The second part is the family of QIR `*_record_output` functions for non-bit values (bools, ints, tuples...).

> * Output label at result_record_output depending on the output schema.

This is implemented only for the Labeled schema, not for the Ordered schema. It's a mandatory point in the QIR spec to support Adaptive profiles (point 4 of the [spec](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md))). It's basically a specification of how the output of a program should be written.

In summary:
- Make Runtime's output stream injectable: it can go in this PR or in a follow-up PR. Your call.
- Add QIR Adaptive Profile support: I would add the first part to this PR, and leave the second part for a follow-up PR.
- Add QIR Output Schemas support: definitely part of a follow-up PR.

rturrado:
> @rturrado Actually, I am not entirely sure whether I understood your last message to 100%. Is your initial question resolved? If not, could you rephrase it?

Thanks @ystade. It got resolved, yes. For the time being, 1) we only record `Result*` outputs (as the QIR runner does), and 2) we just keep measurement outcomes in the Runtime's data structure.

include/mqt-core/qdmi/devices/dd/Device.hpp
rturrado: This option was added as part of Task 9 in #1695. Gating directly on `BUILD_MQT_CORE_MLIR` would mean every MLIR-enabled build also links the QIR JIT (OrcJIT + native LLVM components) into the QDMI device, even for users who only submit QASM2/QASM3 jobs and just happen to have MLIR enabled for other reasons.

The current shape mirrors `BUILD_MQT_CORE_QIR_RUNNER`; both are `cmake_dependent_option`s under MLIR for the same reason: building MLIR is a prerequisite for the QIR JIT, but it shouldn't force every consumer of the JIT to be enabled.

In the top-level `CMakeLists.txt`:

```cmake
cmake_dependent_option(BUILD_MQT_CORE_QIR_RUNNER "Build the QIR runner of the MQT Core project" ON
                       "BUILD_MQT_CORE_MLIR" OFF)

cmake_dependent_option(BUILD_MQT_CORE_QDMI_WITH_QIR "Enable QIR format support for QDMI" OFF
                       "BUILD_MQT_CORE_MLIR" OFF)
```

And in `src/qir/CMakeLists.txt`, the JIT subdirectory is added only when either consumer needs it:

```cmake
if(BUILD_MQT_CORE_QIR_RUNNER OR BUILD_MQT_CORE_QDMI_WITH_QIR)
  add_subdirectory(jit)
endif()
```

Happy to change the default (currently OFF when MLIR is on; Runner defaults to ON), but I'd lean toward keeping the option itself.

If you'd prefer to keep the option but make it more specific to the device it actually affects, I'm happy to rename it; e.g., `BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR` (matching the `qdmi/devices/dd/` directory layout) or `BUILD_MQT_CORE_DDSIM_QDMI_WITH_QIR`. I'd lean toward the former since it follows the existing path hierarchy, but I don't have a strong preference.

src/qdmi/devices/dd/CMakeLists.txt
rturrado: `MQT::CoreQIRJIT` does link `MQT::CoreQIRRuntime`, but as `PRIVATE`. The Runtime is an implementation detail of how JIT executes QIR programs (it provides the `__quantum__*` symbols), not part of JIT's public API. Consumers that also need to talk to the Runtime directly (e.g., the DDSim QDMI device, which reads recorded outputs via `Runtime::getRecordedOutputs()` after each submission) declare that as their own explicit link dependency. Making it `PUBLIC` would transitively pull Runtime into every JIT consumer whether they touch it or not, and would couple Runtime's headers into JIT's public interface.

src/qdmi/devices/dd/Device.cpp
rturrado: It all boils down to programs, whether text strings or bitcodes, being stored in a `std::string program_`, and `std::string` always synthesizing a trailing `'\0'` at `data()[size()]` for `c_str()` compatibility (guaranteed since C++11, see [`std::basic_string::data`](https://en.cppreference.com/w/cpp/string/basic_string/data)).

That byte is not counted in `size()`, never transported over the wire, and never observed by any consumer in the chain:
- `qasm3::Importer::imports` reads `program_.size()` bytes.
- `submitQIRProgram` builds an `llvm::StringRef(program_.data(), program_.size())` that `parseIR` consumes in full via a `MemoryBuffer` with `RequiresNullTerminator=false`.

Sending `size + 1` would transport that bookkeeping byte as data, which is invisible for text but breaks bitcode with `Invalid bitcode signature`. The following [demo](https://godbolt.org/z/o1KhTePTP) replicates the producer/store/consumer chain in isolation.

docs/qir/index.md
rturrado: I changed this to QIR Base Profile Module and QIR Base Profile String. But please feel free to provide a better rewording!

src/qdmi/devices/dd/Device.cpp
rturrado: Cool. That trailing `return` is there because clang-tidy (and the compiler) flag the function otherwise. I'll change it to a `qdmi::unreachable` plus a comment.

src/qir/jit/CMakeLists.txt
rturrado: Good catch! I think I introduced this when adding the header in the CLion IDE. Removed!

include/mqt-core/qir/jit/Session.hpp
rturrado: Perfect, I'll remove the inner namespace, move global functions in `Session.cpp` to an anonymous namespace, and rename current `qir::jit::Session` to `qir::JitSession`.

CMakeLists.txt
rturrado: Understood. Let's wait first for a final decision on the discussion above.

CHANGELOG.md
rturrado: Thanks! What's the difference between ✨ and 🚸?

test/qdmi/devices/dd/job_parameters_test.cpp
rturrado: I have updated the code in the PR to use the original ordering (base string, base module, adaptive string, and adaptive module) in the 2 ifdef/ifndef blocks.

test/qdmi/devices/dd/results_sampling_test.cpp
rturrado: Fair point. However, the comment does not apply anymore because the class fixture uses now the `SetUp` and `TearDown` methods, besides a few member variables and constants.

test/qdmi/devices/dd/results_sampling_test.cpp
rturrado: Well spotted! Changed to `llvmModule`.

test/qdmi/devices/dd/helpers/test_utils.cpp
rturrado: See my [comment](https://github.com/munich-quantum-toolkit/core/pull/1766#discussion_r3382557822) above. This shouldn't be needed. The current code works for text based and binary inputs moving exactly `size` bytes and no null terminators.

include/mqt-core/qir/jit/Session.hpp
rturrado: Very good point! I'll add them. Class should be called `qir::JitSession` though after removing the inner `jit` workspace.

CHANGELOG.md
burgholzer: See https://gitmoji.dev for some general information on this 🙂

src/qdmi/devices/dd/Device.cpp
rturrado: Very good catch! I had completely overlooked this path.

I think something similar to the QASM case would be possible, yes. There, the final measurements are first removed from the quantum computation object (for the `numShots_ == 0` case). For a QIR Base Profile program, we could try rewriting the IR by removing the calls to the measure functions and to the result record output function.

I'll give it a go and be back to you.

include/mqt-core/qdmi/devices/dd/Device.hpp
ystade: @rturrado You are absolutely right. I have forgotten what I actually wrote in the issue. It's too long ago, I created it (or my brain is too bad :D ).

To resolve this thread, let's go with `BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR` to make the variable more specific and let's default to `ON`.

Feel free to resolve this thread yourself if this is implemented.

src/qdmi/devices/dd/CMakeLists.txt
ystade: @rturrado Yes, this makes absolutely sense. If I understand correctly, the `CoreQIRJIT` could be seen as an adaptor for the `CoreQIRRuntime`. The DDSim QDMI Device calls some functionality of the runtime through the adaptor and some directly. In particular, the adaptor exists to execute QIR files passed to the QDMI device.

If that is about right, feel free to resolve this thread.

src/qdmi/devices/dd/Device.cpp
ystade: @rturrado Thanks a lot for the comprehensive explanation. All of that makes sense.

I'm afraid you've discovered a flaw with QDMI. In QDMI, we assume that the trailing `\0` is counted towards the buffer size, see, e.g., https://munich-quantum-software-stack.github.io/QDMI/v1.3.1/group__device__session__interface.html#gaaf2fd0617860189cdbb2cc5c43d77c54

Do you agree, that this is a problem or is it different because QDMI is a C-level interface?

burgholzer: I think we have to be a bit careful here. Nothing in the QDMI spec mandates that the program coming in, or the program being stored in a device are stored as a std::string.
In the end, QDMI is a C-based interface.

What I take from the example above, is that there is a difference between binary data and text data. For binary data, we always want to allocate exactly as many bytes as are provided.
For text data, we have to account for the `\0` terminator, which is excluded in `std::string::size()`, otherwise a simple memcpy/strncpy on the device side would result in an unterminated `char*` string.
The alternative would be to require devices to manually account for this in their implementations (e.g., by using std::string, which automatically ensures the null termination).

I hope the rambling makes somewhat sense.
It's problematic that we are shaving off a byte of binary data. At the same time, it is unlikely that we are going to change the current version of the QDMI spec (we will definitely keep this in mind for the next major version).

What about the following solution: We distinguish on the format being binary or text and use `size-1` for text-based payloads, and `size` for binary payloads?

burgholzer:
Hey @rturrado 👋🏼 
Just checking in to see whether there is anything blocking you at the moment?
If you feel like this is ready for another round of review, then feel free to re-request one from Yannick 🙌🏼

src/qdmi/devices/dd/Device.cpp
rturrado: Agreed, guys. I think I've understood your point now. Basically, in `test_utils.cpp`, `setProgram`,

```cpp
  // Change this to use size+1 for text
  rc = MQT_DDSIM_QDMI_device_job_set_parameter(
      job, QDMI_DEVICE_JOB_PARAMETER_PROGRAM, program.size(), program.data());
```

we are simulating the device communication via a C-based interface (using `char*` types). The expected behaviour there would be for the device to send a null terminator for text and the exact byte count for binary (which is normal, because binary can already contain embedded `\0`s).

Now, the `setParameter` code at `Device.cpp` can strip that null terminator for text programs, because it's going to save it in a `std::string` (which already adds a `\0` at `data[size()]`):

```cpp
  case QDMI_DEVICE_JOB_PARAMETER_PROGRAM:
    if (value != nullptr) {
      // Change this to use size-1 for text
      program_ = std::string(static_cast<const char*>(value), size);
    }
```

That's the inbound way: device -> Device.cpp.

For the outbound way we don't have to do anything because we are already sending a `program_.c_str()`, i.e., we are already using the C-based interface (`queryProperty` function in `Device.cpp`):

```cpp
  ADD_STRING_PROPERTY(QDMI_DEVICE_JOB_PROPERTY_PROGRAM, program_.c_str(), prop,
                      size, value, sizeRet)
```

I'll do the changes accordingly!

src/qdmi/devices/dd/CMakeLists.txt
rturrado: @ystade That's it, exactly!

rturrado:
> Hey @rturrado 👋🏼 Just checking in to see whether there is anything blocking you at the moment? If you feel like this is ready for another round of review, then feel free to re-request one from Yannick 🙌🏼

Hi @burgholzer, thanks for checking in. All good on my end.

I think that there are only two pending fronts:

1. The `numShots_ == 0` case at `submitQIRProgram`. I have a proposal for this. I'll add more details in the review comment thread, and then we can all decide if we implement it for this PR or for a follow-up.
2. The code coverage errors. I'll address that tomorrow. It probably just needs tests for the uncovered code.

src/qdmi/devices/dd/Device.cpp
rturrado: So, a possible implementation for this would be:

- `JitSession` accepts an `Execution { Sampling, StateExtraction }` flag at construction time. This is quite symmetric to what we have for the DDSim (`sample` and `simulate`). In `StateExtraction` mode, measurement calls (and recordings) are stripped from the IR before JIT.
- The `Runtime` adds a `takeState` API to move the DD package and `qState` out of the singleton so the caller, the device, can own the post-execution state.
- `submitQIRProgram` implements the `numShots_ == 0` path (for Base Profile only): `JitSession(StateExtraction)`, then JIT run, then `takeState`.

Should we add it in this PR or in a follow-up?

ystade: I like you proposal a lot. Would you mind implementing this as part of this PR?

For the "stripping the measurements", I am wondering whether it is easier to change the definition of the function `...qis_m_...` to do nothing or only check that no gate comes after a measurement instead of actually stripping them.

src/qdmi/devices/dd/Device.cpp
burgholzer: Yeah, that sounds reasonable 👍🏼

src/qdmi/devices/dd/Device.cpp
burgholzer: For base profile programs, this shouldn't really make much of a difference, I'd be fine with either solution.
The tricky case is the adaptive profile, where determining the statevector is not necessarily possible (e.g. in case of mid-circuit measurements).
This is kind of similar to the regular DDSim methods `sample` and `simulate`. The `sample` method can handle arbitrary circuits, while the `simulate` method can only handle circuits where there are either no measurements or all measurements are at the end.

rturrado: Thanks for the comments, guys! Yes, happy to land it in this PR; the work is already on the branch.

The shape I went with:
- Reject the `numShots_ == 0` + Adaptive Profile combination at the device.
- Add an `Execution { Sampling, StateExtraction }` mode to `JitSession`.
- Rewrite the IR (strip the measurment and result-management calls) as part of session initialization when the mode is `StateExtraction`.

I considered making `__quantum__qis__m*__body` a no-op for the `StateExtraction` case, but that puts the "are we extracting state?" switch inside the Runtime (or the QIR runtime API), and the Runtime is otherwise format-agnostic. Keeping the switch on `JitSession` localizes the new state to the JIT layer and leaves Runtime untouched.

Treat this as a starting point for discussion. Happy to revisit if the runtime-mode angle has advantages I am not seeing.

ystade: This sounds like a very clean solution. Let's stick with it.

ystade:
@rturrado  let me know, when I should review this PR again. I'll try to get to it asap then.

rturrado:
> @rturrado let me know, when I should review this PR again. I'll try to get to it asap then.

@ystade Thanks! I think that the only pending points are the coverage errors, so if you want to review the last commits, and maybe check if some unresolved convesations can already be closed, you can do it already.

ystade:
@rturrado Thanks for the update, I am happy to wait until the coverage is also satisfied.

burgholzer:
BTW: Do not worry about the Windows CI errors. They are unrelated to your changes and a consequence of some GitHub runner updates for Windows systems not interacting well with our pre-built LLVM distributions. They will be fixed separately.

rturrado:
@ystade @burgholzer PR should be ready now for review. Functionality is there. Code coverage is happy. And tests are passing (except those 2 Windows ones that you mentioned we shouldn't worry about for this PR).

CMakeLists.txt
```diff
@@ -121,6 +121,10 @@ endif()
 cmake_dependent_option(BUILD_MQT_CORE_QIR_RUNNER "Build the QIR runner of the MQT Core project" ON
                        "BUILD_MQT_CORE_MLIR" OFF)
 
+cmake_dependent_option(
+  BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR "Enable QIR program format support for the DDSim QDMI device"
+  OFF "BUILD_MQT_CORE_MLIR" OFF)
```
ystade:
```suggestion
  ${BUILD_MQT_CORE_MLIR} "BUILD_MQT_CORE_MLIR" OFF)
```
Can we use the variable as default here to default to true when MLIR is enabled.

docs/qir/index.md
```diff
@@ -36,3 +36,9 @@ The `mqt-core-qir-runner` can be used to execute a QIR file (typically with a `.
 
 This will simulate the circuit and print the measurement results to the console.
 The runner supports the QIR Base Profile.
+
+### QIR Support in the QDMI Device
+
+The QDMI Device accepts jobs in the following program formats: QASM2, QASM3, QIR Base Profile Module (LLVM bitcode), and QIR Base Profile String (LLVM assembly).
+The QIR base formats are only supported when the `BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR` CMake option is enabled.
+It is disabled by default to avoid the cost of linking against the MQT Core QIR JIT (built on LLVM OrcJIT) and Runtime libraries.
```
ystade: This line must be adapted after adding the suggestion above.

include/mqt-core/qdmi/devices/dd/Device.hpp
```diff
@@ -234,6 +235,41 @@ struct MQT_DDSIM_QDMI_Device_Job_impl_d {
...
+  /// - increase the running-job count,
+  /// - set status to RUNNING,
+  /// - run @p body,
+  /// - set status to DONE or FAILED, and
+  /// - decrease the running-job count.
+  /// @p body is the format-specific work:
+  /// - parse the program,
+  /// - run or simulate it, and
+  /// - store the results in the job's output fields.
+  /// Returns @c QDMI_SUCCESS once the worker has been spawned.
```
ystade:
```suggestion
  /// @returns @c QDMI_SUCCESS once the worker has been spawned.
```
Can you doxygen syntax here.

include/mqt-core/qdmi/devices/dd/Device.hpp
```diff
@@ -234,6 +235,41 @@ struct MQT_DDSIM_QDMI_Device_Job_impl_d {
...
+  /// Run @p body on a worker thread with the standard job lifecycle:
+  /// - increase the running-job count,
+  /// - set status to RUNNING,
+  /// - run @p body,
+  /// - set status to DONE or FAILED, and
+  /// - decrease the running-job count.
+  /// @p body is the format-specific work:
+  /// - parse the program,
+  /// - run or simulate it, and
+  /// - store the results in the job's output fields.
```
ystade: To me this reads like a broken sentence. The bullet points require some context that says what the bullets describe.

include/mqt-core/qir/jit/Session.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  /// Parses LLVM IR (textual or bitcode) from @p irBytes using the session's
+  /// thread-safe context. @p bufferName is used in diagnostics.
+  llvm::Expected<llvm::orc::ThreadSafeModule>
+  loadModuleFromMemory(llvm::StringRef irBytes, llvm::StringRef bufferName);
+
+  /// Prepares the session to run the program:
+  /// - Validates the loaded module.
+  /// - Optionally strips measurement and result management calls
+  ///   (for @c Execution::StateExtraction).
```
ystade:
```suggestion
  ///   (for @ref Execution::StateExtraction).
```
Would that work as well?

include/mqt-core/qir/runtime/Runtime.hpp
```diff
@@ -356,5 +390,24 @@ class Runtime {
   auto deref(Result* result) -> ResultStruct&;
   auto rFree(Result* result) -> void;
   auto equal(Result* result1, Result* result2) -> bool;
+
+  /// Append the value referenced by `result` to the recorded outputs bit
+  /// string in record order.
+  auto recordOutput(Result* result) -> void;
+
+  /// Return the outputs declared by the program as a bit string in record
```
ystade:
```suggestion
  /// @returns the outputs declared by the program as a bit string in record
```
That's a very subjective nitpick. Aligns slightly better with the rest of the code base.

include/mqt-core/qir/runtime/Runtime.hpp
```diff
@@ -356,5 +390,24 @@ class Runtime {
...
+  /// order.
+  auto getRecordedOutputs() const -> const std::string&;
+
+  /// Move the quantum state out of the runtime.
+  /// Then reset the runtime to a clean state ready for the next job.
+  /// Intended for use after a @c JitSession constructed with
+  /// @c Execution::StateExtraction has finished running.
```
ystade, line 405:
```suggestion
  /// @c Execution::StateExtraction has finished running.
  /// @returns the moved @c QState from the runtime.
```

src/qdmi/devices/dd/Device.cpp
```diff
@@ -346,15 +357,29 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::setParameter(
...
     if (value != nullptr) {
-      program_ = std::string(static_cast<const char*>(value), size - 1);
+      // Text payloads include the trailing '\0' in `size`.
+      // Strip it so it is not counted in `program_.size()`.
+      // `std::string` re-synthesizes its own '\0' at `data()[size()]` for
+      // c_str() consumers.
+      // Binary payloads are stored exactly as received.
+      const auto bytes =
+          qdmi::dd::isTextProgramFormat(format_) ? size - 1 : size;
+      program_ = std::string(static_cast<const char*>(value), bytes);
```
ystade: Somehow, I did not realize it before. I personally find it a bit awkward to store binary data in a string and would rather use a `std::vector<std::byte>>` for it, which would require transforming the `program_` member to a `std::variant<...>` type. This perhaps gets out of hand, and I am fine with keeping it as it is right now.

src/qir/jit/Session.cpp
```diff
@@ -0,0 +1,396 @@
...
+  if (auto err = enableDebuggerSupport(jit)) {
+    [[maybe_unused]] const std::string errMsg = toString(std::move(err));
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-do-while)
+    LLVM_DEBUG(llvm::dbgs() << DEBUG_TYPE ": " << errMsg << "\n");
+  }
+  return llvm::Error::success();
+}
+
+static llvm::Expected<llvm::orc::ThreadSafeModule>
+getThreadSafeModuleOrError(std::unique_ptr<llvm::Module> module,
```
ystade:
```suggestion
getThreadSafeModuleOrError(std::unique_ptr<llvm::Module> m,
```
I am not sure whether this is old code that you only moved or whether this is new code. In anyways, would you mind changing the paramter name `module` to something else mosty to avoid the highlighting of `module` as a keyword? I would even be fine with just `m`.

test/qdmi/devices/dd/results_sampling_test.cpp
```diff
@@ -18,27 +18,146 @@
...
+
+class QIRHistogramTestString : public HistogramTest {
+protected:
+  static std::string getProgram(const std::string_view file) {
+    const std::filesystem::path path =
+        std::filesystem::path(QIR_FILES_DIR) / file;
+    std::ifstream ifs(path);
+    EXPECT_TRUE(ifs.is_open()) << "Failed to open " << path.string();
+    return {std::istreambuf_iterator<char>{ifs}, {}};
+  }
```
ystade: I see this function redundantly across multiple test files. Would it make sense to outsource this function to one header and source file and then use this single instance in all test files?

test/qir/CMakeLists.txt
```diff
@@ -6,5 +6,12 @@
 #
 # Licensed under the MIT License
 
+if(BUILD_MQT_CORE_QIR_RUNNER OR BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR)
+  add_subdirectory(jit)
+endif()
+
 add_subdirectory(runtime)
```
ystade, line 11: I am bit confused why this line is (and was not) guarded on the value of a variable, i.e., should it not be in the upper if-branch? Probably there is a good reason not to that I am overlokking right now.

ystade (commented):
@rturrado Thanks for this very cool PR extending the QDMI DDSIM device to support QIR in the base and adaptive profile. Your code is very well structured and I like also the few refactoring of the old code base. This is a very valauble contribution.

I have left some minor comments inline. Other then these, you get a green light from my end. I purposefully do not approve the PR yet and leave this to @burgholzer because I do not know whether he also wants to have an eye on the PR before it gets merged.

include/mqt-core/qdmi/devices/dd/ProgramFormat.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * - text formats (QASM, QIR Base/Adaptive String) are shipped with a trailing
+ *   '\0' counted in the buffer size.
+ * - binary formats (QIR Base/Adaptive Module bitcode) are shipped as exact byte
+ *   counts since '\0' may appear inside the payload.
+ */
+inline bool isTextProgramFormat(QDMI_Program_Format fmt) {
+  return fmt == QDMI_PROGRAM_FORMAT_QASM2 || fmt == QDMI_PROGRAM_FORMAT_QASM3 ||
+         fmt == QDMI_PROGRAM_FORMAT_QIRBASESTRING ||
+         fmt == QDMI_PROGRAM_FORMAT_QIRADAPTIVESTRING;
+}
```
burgholzer: It feels a bit over the top to add a dedicated file for a function that is essentially a one-liner and only used in two places (once in the library itself, once in the tests).
I'd personally favor inlining this into the call sites and removing this file.

CMakeLists.txt
rturrado: Sure! I left it to OFF because that was my understanding of task 9:

> 9. **CMake / LLVM dependency gating** — Linking `MQT::CoreQIRRuntime` and `LLVM OrcJIT` libs into `mqt-core-qdmi-ddsim-device` significantly increases the build footprint. Add a CMake option (e.g. `BUILD_MQT_CORE_QDMI_WITH_QIR`) so that QIR format support is opt-in.

 I apply the suggestion, but please confirm whether you finally want it this way (defaulting to `BUILD_MQT_CORE_MLIR`) or just disabled by default.

docs/qir/index.md
rturrado: Good catch!

include/mqt-core/qdmi/devices/dd/Device.hpp
rturrado: Perfect!

include/mqt-core/qdmi/devices/dd/Device.hpp
rturrado: Fair enough!

include/mqt-core/qir/jit/Session.hpp
rturrado: Yes, `@ref` works for any documented symbol and is strictly better here; it renders as monospace and turns into a clickable link to `Execution::StateExtraction`'s docs. `@c` only does the monospace.

The only caveats:
- `Execution::StateExtraction` must be a documented entity in the same Doxygen run (it is, since it's the project's own enum).
- `StateExtraction are unambiguous`, is a distinctive name (CamelCase, not a regular English word), so Doxygen won't confuse it with another symbol or accidentally link it from normal text.

include/mqt-core/qir/runtime/Runtime.hpp
rturrado: Good!

include/mqt-core/qir/runtime/Runtime.hpp
rturrado: Nice catch!

src/qir/jit/Session.cpp
rturrado: Absolutely! I'll leave it as `llvmModule` for consistency with other parts of the codebase though.

test/qir/CMakeLists.txt
rturrado: Well spotted.

This change isn't strictly needed, because the inner `test/qir/runner/CMakeLists.txt` already starts with `if(TARGET MQT::CoreQIRRunner)`. I added it for consistency with `src/qir/CMakeLists.txt`.

We could also drop the guard around `add_subdirectory(jit)` and leave the file as:

```
add_subdirectory(jit)
add_subdirectory(runtime)
add_subdirectory(runner)
```

Please let me know what version you prefer (guards/no guards), and I'll apply the change.

CMakeLists.txt
ystade: I am sorry for the confusion. Yes, considering the current setup as it is right now, please default to ON, deviating from the initial instructions. Sorry.

test/qir/CMakeLists.txt
ystade: If the guards are already subsumed by other ones, I am in favor of the more minimalistic version.

burgholzer:
I'll give this a proper review tomorrow. Didn't quite get to it today unfortunately.

CMakeLists.txt
rturrado: No problem at all!

test/qir/CMakeLists.txt
rturrado: Perfect!

test/qdmi/devices/dd/results_sampling_test.cpp
rturrado: I've added the new test helper at `test/qir/helpers/test_utils.{h,c}pp`, similar to `test/qdmi/devices/dd/helpers/test_utils.{h,c}pp`.

Here we define `qir_test::getProgram`; the qdmi helper defines its own functions under `qdmi_test`.

The difference is in the CMake setup. In the qdmi case, the `helpers/` folder is purely for file organization, so it doesn't need its own `CMakeLists.txt`; the `.cpp` is added directly from the parent's `package_add_test` call.

The new qir helper, however, is consumed from different subdirectories (`test/qir/jit/` and `test/qdmi/devices/dd/`), so the folder has its own `CMakeLists.txt` that declares an `mqt-core-qir-test-utils` static library that both consumers link against.

src/qdmi/devices/dd/Device.cpp
rturrado: @ystade Addressed. It was not a big change. It looks more like a proper solution now.

docs/qir/index.md
```diff
@@ -36,3 +37,9 @@ The `mqt-core-qir-runner` can be used to execute a QIR file (typically with a `.
 
 This will simulate the circuit and print the measurement results to the console.
 The runner supports the QIR Base Profile.
+
+### QIR Support in the QDMI Device
+
+The QDMI Device accepts jobs in the following program formats: QASM2, QASM3, QIR Base Profile Module (LLVM bitcode), and QIR Base Profile String (LLVM assembly).
+The QIR base formats are only supported when the `BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR` CMake option is enabled.
+It is enabled by default, but depends on `BUILD_MQT_CORE_MLIR` being set.
```
burgholzer: The device also supports the QIR adaptive profile, which is not reflected here.

docs/qir/index.md
```diff
@@ -36,3 +37,9 @@ The `mqt-core-qir-runner` can be used to execute a QIR file (typically with a `.
 
 This will simulate the circuit and print the measurement results to the console.
 The runner supports the QIR Base Profile.
+
+### QIR Support in the QDMI Device
+
```
burgholzer: This should probably spell "in the DDSIM QDMI Device"

src/qdmi/devices/dd/Device.cpp
```diff
@@ -380,50 +411,130 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::queryProperty(
   ADD_SINGLE_VALUE_PROPERTY(QDMI_DEVICE_JOB_PROPERTY_PROGRAMFORMAT,
                             QDMI_Program_Format, format_, prop, size, value,
                             sizeRet)
-  ADD_STRING_PROPERTY(QDMI_DEVICE_JOB_PROPERTY_PROGRAM, program_.c_str(), prop,
-                      size, value, sizeRet)
+  if (const auto* text = std::get_if<std::string>(&program_)) {
+    ADD_STRING_PROPERTY(QDMI_DEVICE_JOB_PROPERTY_PROGRAM, text->c_str(), prop,
+                        size, value, sizeRet)
+  }
```
burgholzer: I would argue there should also be a way to get back the program if it is binary based.

test/qdmi/devices/dd/results_sampling_test.cpp
```diff
@@ -18,27 +18,133 @@
...
+#ifdef BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR
+TEST_F(QIRHistogramTestModule, BaseStatic) {
+  constexpr auto format = QDMI_PROGRAM_FORMAT_QIRBASEMODULE;
+  checkHistogram(runProgram(format, getProgram("BellPairStatic.ll")));
+}
+
+TEST_F(QIRHistogramTestString, BaseStatic) {
+  constexpr auto format = QDMI_PROGRAM_FORMAT_QIRBASESTRING;
+  checkHistogram(runProgram(format, qir_test::getProgram("BellPairStatic.ll")));
+}
```
burgholzer, line 124: Just to double check that I got this correctly: Why are all of the tests duplicated here? One is string based and one is file based?

burgholzer (changes requested):
This looks really clean. Great contribution @rturrado 🙌🏼 
I have one small round of feedback that I'd like to see addressed. Afterwards, we can happily merge this 🚀

include/mqt-core/qdmi/devices/dd/ProgramFormat.hpp
rturrado: Fair enough, yes.

docs/qir/index.md
rturrado: Good catch, thanks!

docs/qir/index.md
rturrado: Indeed. I'll also fix two other text references in this pass:
- in the top `CMakeLists.txt`: `DDSim QDMI device` to `DDSIM QDMI Device`.
- in the `CHANGELOG.md`: `QDMI DDSim device` to `DDSIM QDMI Device`. This latter form is used several times in that file.

src/qdmi/devices/dd/Device.cpp
rturrado: Wow, this was a bug! Thanks!

test/qdmi/devices/dd/results_sampling_test.cpp
rturrado: Yes, exactly. For the module based tests, `getProgram` loads the file, which contains an LLVM assembly string, and **assembles it to bitcode**. For the string based tests, `qir_test::getProgram` just returns the LLVM assembly string. So, in the end, you either `runProgram(MODULE, bytes)` or `runProgram(STRING, text)`.

burgholzer: Thanks for the clarification. That makes sense 👍🏼

rturrado: @burgholzer We could also parameterize tests here. It's not a shorter code, but probably less confusing:

```cpp
#ifdef BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR
struct QIRHistogramParam {
  QDMI_Program_Format format;
  std::string_view file;
  std::string_view label;
};

class QIRHistogramTest : public HistogramTest,
                         public testing::WithParamInterface<QIRHistogramParam> {
protected:
  /// Parse @p text as an LLVM assembly module and return its bitcode encoding.
  static std::string toBitcode(const std::string& text) {
    llvm::LLVMContext context;
    llvm::SMDiagnostic err;
    auto llvmModule = llvm::parseAssemblyString(text, err, context);
    EXPECT_NE(llvmModule, nullptr)
        << "parseAssemblyString failed: " << err.getMessage().str();
    if (llvmModule == nullptr) {
      return {};
    }
    std::string bitcodeBuffer;
    llvm::raw_string_ostream os(bitcodeBuffer);
    llvm::WriteBitcodeToFile(*llvmModule, os);
    os.flush();
    return bitcodeBuffer;
  }

  static std::string getProgram(QDMI_Program_Format format,
                                std::string_view file) {
    auto text = qir_test::getProgram(file);
    if (format == QDMI_PROGRAM_FORMAT_QIRBASEMODULE ||
        format == QDMI_PROGRAM_FORMAT_QIRADAPTIVEMODULE) {
      return toBitcode(text);
    }
    return text;
  }
};
#endif

} // namespace

TEST_F(HistogramTest, QASM3Program) {
  constexpr QDMI_Program_Format format = QDMI_PROGRAM_FORMAT_QASM3;
  constexpr std::string_view program = qdmi_test::QASM3_BELL_SAMPLING;
  checkHistogram(runProgram(format, program));
}

#ifdef BUILD_MQT_CORE_QDMI_DDSIM_WITH_QIR
TEST_P(QIRHistogramTest, RunBellPair) {
  const auto& [format, file, label] = GetParam();
  checkHistogram(runProgram(format, getProgram(format, file)));
}

INSTANTIATE_TEST_SUITE_P(
    QIRProgramFormat, QIRHistogramTest,
    testing::Values(
        QIRHistogramParam(QDMI_PROGRAM_FORMAT_QIRBASEMODULE,
                          "BellPairStatic.ll", "BaseModuleStatic"),
        QIRHistogramParam(QDMI_PROGRAM_FORMAT_QIRBASESTRING,
                          "BellPairStatic.ll", "BaseStringStatic"),
        QIRHistogramParam(QDMI_PROGRAM_FORMAT_QIRBASEMODULE,
                          "BellPairDynamic.ll", "BaseModuleDynamic"),
        QIRHistogramParam(QDMI_PROGRAM_FORMAT_QIRBASESTRING,
                          "BellPairDynamic.ll", "BaseStringDynamic"),
        QIRHistogramParam(QDMI_PROGRAM_FORMAT_QIRADAPTIVEMODULE,
                          "BellPairAdaptive.ll", "AdaptiveModule"),
        QIRHistogramParam(QDMI_PROGRAM_FORMAT_QIRADAPTIVESTRING,
                          "BellPairAdaptive.ll", "AdaptiveString")),
    [](const testing::TestParamInfo<QIRHistogramTest::ParamType>& info) {
      return std::string(info.param.label);
    });
#endif
```

ystade (approved):
@rturrado This PR was truly a very pleasant interaction. This is exactly how we imagine a unitaryHACK contribution to be. Thanks for your contribution!
If you are interested, you are very welcome to contribute to our projects again in the future, either as part of the next unitaryHACK or independently.

rturrado:
> @rturrado This PR was truly a very pleasant interaction. This is exactly how we imagine a unitaryHACK contribution to be. Thanks for your contribution!

@ystade Thanks so much! I really appreciate the words. I can also say it was very pleasant for me. You were very quick to answer my questions and provide reviews, and I found the comments very constructive. Not to mention that for me, being able to work in (and learn from) a C++ codebase in this quantum/LLVM/IR world is kind of an ideal job.

> If you are interested, you are very welcome to contribute to our projects again in the future, either as part of the next unitaryHACK or independently.

Definitely. I have at least a couple of pending follow-ups: Add QIR Output Schemas Support, and Add QIR Adaptive Profile Support. Do you prefer me to open the PRs directly, or should I create an issue first?

test/qdmi/devices/dd/results_sampling_test.cpp
burgholzer: Nah, I think it is fine to keep the tests as-is for now 👍🏼

burgholzer (approved):
Wonderful 🙌🏼 

Before we merge this, could you please add a comment to the original issue here so that we can assign it to you and you get your bounty? 💸 

As for tackling further issues: Typically, it's not a bad idea to create tracking issues. Not an absolute must-have, but it kind of gives us the opportunity to discuss some design decisions before any implementation is done.

rturrado:
> Wonderful 🙌🏼 

🥳

> Before we merge this, could you please add a comment to the original issue here so that we can assign it to you and you get your bounty? 💸 

 Roger that. 

> As for tackling further issues: Typically, it's not a bad idea to create tracking issues. Not an absolute must-have, but it kind of gives us the opportunity to discuss some design decisions before any implementation is done.

Fair enough!

PR #1774, opened 2026-06-08
🔥 Remove Eigen dependency from QCO

simon1hofmann:
## Description

- Drop the Eigen dependency from the QCO MLIR dialect and replace it with a small, hand-crafted `UnitaryMatrix` library (`Matrix1x1`, `Matrix2`, `Matrix4`, `DynamicMatrix`).
- Migrate all gate and modifier `getUnitaryMatrix()` implementations off `Eigen::Matrix*` types.
- Remove Eigen from `FetchContent` / `MLIRQCODialect` link dependencies.

This removes the need for `unsupported/Eigen/KroneckerProduct` and other Eigen headers in QCO IR. 

## Motivation

Eigen was introduced primarily for matrix types and a Kronecker-product helper for controlled gates. For QCO's current needs that is heavy:
- `unsupported/Eigen/KroneckerProduct` has no stability guarantee.
- Actual usage is limited to 1×1 / 2×2 / 4×4 gate unitaries and simple modifier composition.
- A small custom library is easier to reason about, has no external dependency, and can be optimized incrementally (e.g. SIMD) as needed.
- `CtrlOp` now builds `I_{2^n} ⊗ U` via `DynamicMatrix::identity(dim)` + `setBottomRightCorner()`. `InvOp` uses in-place adjoint. 

## AI Assistance

Used Composer via Cursor for parts of this change. I reviewed the full diff and take responsibility for everything in this PR.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Removed
 
+- 🔥 Remove the Eigen dependency from the QCO MLIR dialect ([#1774]) ([**@simon1hofmann**])
 - 🔥 Remove the density matrix support from the MQT Core DD package ([#1466]) ([**@burgholzer**])
 - 🔥 Remove `datastructures` (`ds`) (sub)library from MQT Core ([#1458]) ([**@burgholzer**])
 
@@ -402,6 +403,7 @@ _📚 Refer to the [GitHub Release Notes](https://github.com/munich-quantum-tool
 
 <!-- PR links -->
 
+[#1774]: https://github.com/munich-quantum-toolkit/core/pull/1774
 [#1765]: https://github.com/munich-quantum-toolkit/core/pull/1765
 [#1762]: https://github.com/munich-quantum-toolkit/core/pull/1762
 [#1749]: https://github.com/munich-quantum-toolkit/core/pull/1749
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 - ✨ Add conversions between `jeff` and QCO ([#1479], [#1548], [#1565], [#1637], [#1676], [#1706]) ([**@denialhaag**], [**@burgholzer**])
 - ✨ Add a `place-and-route` pass for mapping circuits to architectures with restricted topologies ([#1537], [#1547], [#1568], [#1581], [#1583], [#1588], [#1600], [#1664], [#1709], [#1716], [#1748]) ([**@MatthiasReumann**], [**@burgholzer**])
 - ✨ Add initial infrastructure for new QC and QCO MLIR dialects
-  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1542], [#1548], [#1550], [#1554], [#1567], [#1569], [#1570], [#1572], [#1573], [#1580], [#1602], [#1620], [#1623], [#1624], [#1626], [#1627], [#1635], [#1638], [#1673], [#1675], [#1700], [#1717], [#1728], [#1730], [#1749], [#1762], [#1765])
+  ([#1264], [#1330], [#1402], [#1428], [#1430], [#1436], [#1443], [#1446], [#1464], [#1465], [#1470], [#1471], [#1472], [#1474], [#1475], [#1506], [#1510], [#1513], [#1521], [#1542], [#1548], [#1550], [#1554], [#1567], [#1569], [#1570], [#1572], [#1573], [#1580], [#1602], [#1620], [#1623], [#1624], [#1626], [#1627], [#1635], [#1638], [#1673], [#1675], [#1700], [#1717], [#1728], [#1730], [#1749], [#1762], [#1765], [#1774])
   ([**@burgholzer**], [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**], [**@Ectras**], [**@MatthiasReumann**], [**@simon1hofmann**])
 
 ### Changed
@@ -29,7 +29,6 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Removed
 
-- 🔥 Remove the Eigen dependency from the QCO MLIR dialect ([#1774]) ([**@simon1hofmann**])
 - 🔥 Remove the density matrix support from the MQT Core DD package ([#1466]) ([**@burgholzer**])
 - 🔥 Remove `datastructures` (`ds`) (sub)library from MQT Core ([#1458]) ([**@burgholzer**])
 
--- a/mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
+++ b/mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
@@ -93,8 +93,10 @@ struct Matrix2x2 {
    * @param m11 Element at row 1, column 1.
    * @return A new `Matrix2x2` with the given elements.
    */
-  [[nodiscard]] static Matrix2x2 fromElements(Complex m00, Complex m01,
-                                              Complex m10, Complex m11);
+  [[nodiscard]] static Matrix2x2 fromElements(const Complex& m00,
+                                              const Complex& m01,
+                                              const Complex& m10,
+                                              const Complex& m11);
 
   /**
    * @brief Returns the 2x2 identity matrix.
@@ -195,10 +197,12 @@ struct Matrix4x4 {
    * @return A new `Matrix4x4` with the given elements.
    */
   [[nodiscard]] static Matrix4x4
-  fromElements(Complex m00, Complex m01, Complex m02, Complex m03, Complex m10,
-               Complex m11, Complex m12, Complex m13, Complex m20, Complex m21,
-               Complex m22, Complex m23, Complex m30, Complex m31, Complex m32,
-               Complex m33);
+  fromElements(const Complex& m00, const Complex& m01, const Complex& m02,
+               const Complex& m03, const Complex& m10, const Complex& m11,
+               const Complex& m12, const Complex& m13, const Complex& m20,
+               const Complex& m21, const Complex& m22, const Complex& m23,
+               const Complex& m30, const Complex& m31, const Complex& m32,
+               const Complex& m33);
 
   /**
    * @brief Returns the 4x4 identity matrix.
--- a/mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
+++ b/mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
@@ -19,21 +19,21 @@
 
 namespace mlir::qco {
 
-/// Complex scalar type used for unitary matrix entries.
+/// Complex scalar type used for matrix entries.
 using Complex = std::complex<double>;
 
-/// Default absolute tolerance for unitary matrix comparisons.
-inline constexpr double UNITARY_MATRIX_TOLERANCE = 1e-14;
+/// Default absolute tolerance for matrix comparisons.
+inline constexpr double MATRIX_TOLERANCE = 1e-14;
 
 /**
- * @brief 1x1 unitary matrix for global-phase gates.
+ * @brief 1x1 matrix for global-phase gates.
  *
  * Wraps a single complex scalar. Used by operations such as `GPhaseOp` whose
  * unitary is a global phase factor.
  */
 struct Matrix1x1 {
   /// The sole matrix entry.
-  Complex value{1.0, 0.0};
+  Complex value{0.0, 0.0};
 
   /**
    * @brief Constructs a matrix from its single entry.
@@ -65,13 +65,13 @@ struct Matrix1x1 {
    * @return True if the difference is within @p tol.
    */
   [[nodiscard]] bool isApprox(const Matrix1x1& other,
-                              double tol = UNITARY_MATRIX_TOLERANCE) const;
+                              double tol = MATRIX_TOLERANCE) const;
 };
 
 /**
- * @brief Fixed-size 2x2 unitary matrix in row-major layout.
+ * @brief Fixed-size 2x2 matrix in row-major layout.
  *
- * Represents single-qubit gate unitaries. Elements are stored in a flat array
+ * Used to represent single-qubit unitaries. Elements are stored in a flat array
  * with index `(row * K_COLS) + col`.
  */
 struct Matrix2x2 {
@@ -154,14 +154,14 @@ struct Matrix2x2 {
    * @return True if every entry differs by at most @p tol.
    */
   [[nodiscard]] bool isApprox(const Matrix2x2& other,
-                              double tol = UNITARY_MATRIX_TOLERANCE) const;
+                              double tol = MATRIX_TOLERANCE) const;
 };
 
 /**
- * @brief Fixed-size 4x4 unitary matrix in row-major layout.
+ * @brief Fixed-size 4x4 matrix in row-major layout.
  *
- * Represents two-qubit gate unitaries. Elements are stored in a flat array with
- * index `(row * K_COLS) + col`.
+ * Used to represent two-qubit gate unitaries. Elements are stored in a flat
+ * array with index `(row * K_COLS) + col`.
  */
 struct Matrix4x4 {
   /// Number of rows.
@@ -263,13 +263,13 @@ struct Matrix4x4 {
    * @return True if every entry differs by at most @p tol.
    */
   [[nodiscard]] bool isApprox(const Matrix4x4& other,
-                              double tol = UNITARY_MATRIX_TOLERANCE) const;
+                              double tol = MATRIX_TOLERANCE) const;
 };
 
 /**
- * @brief Square unitary matrix with runtime dimension.
+ * @brief Square matrix with runtime dimension.
  *
- * Used when the Hilbert-space dimension depends on the operation, for example
+ * Used when the Hilbert-space dimension depends on the operation, for example,
  * in controlled gates (`CtrlOp`) and inverses (`InvOp`). Storage is row-major
  * and held behind a private implementation pointer.
  */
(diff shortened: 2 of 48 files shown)


simon1hofmann:
@coderabbitai review

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning], with the exception that minor rel
 
 ### Removed
 
+- 🔥 Remove the Eigen dependency from the QCO MLIR dialect ([#1774]) ([**@simon1hofmann**])
```
burgholzer: I think this does not really deserve its own changelog entry because the PR that introduced this has not been released yet. As such, I would rather add this to the general Changelog entry for the dialect rewrite.

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/DCXOp.cpp
```diff
@@ -43,9 +43,7 @@ void DCXOp::getCanonicalizationPatterns(RewritePatternSet& results,
...
 }
 
-Eigen::Matrix4cd DCXOp::getUnitaryMatrix() {
-  return Eigen::Matrix4cd{{1, 0, 0, 0},  // row 0
-                          {0, 0, 1, 0},  // row 1
-                          {0, 0, 0, 1},  // row 2
-                          {0, 1, 0, 0}}; // row 3
+Matrix4x4 DCXOp::getUnitaryMatrix() {
+  return Matrix4x4::fromElements(1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 1, 0,
+                                 0);
```
burgholzer: Can we somehow preserve the nicer formatting from before?
I don't actually mind this being a 1D list, but the presentation could be improved (or match the existing one)

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/RXOp.cpp
```diff
@@ -65,13 +65,15 @@ void RXOp::getCanonicalizationPatterns(RewritePatternSet& results,
...
-std::optional<Eigen::Matrix2cd> RXOp::getUnitaryMatrix() {
+std::optional<Matrix2x2> RXOp::getUnitaryMatrix() {
   using namespace std::complex_literals;
 
   if (const auto theta = valueToDouble(getTheta())) {
     const auto m00 = std::cos(*theta / 2.0) + 0i;
     const auto m01 = -1i * std::sin(*theta / 2.0);
-    return Eigen::Matrix2cd{{m00, m01}, {m01, m00}};
+    const auto m10 = m01;
+    const auto m11 = m00;
```
burgholzer: Does it not work if the same value is used twice (as before)?

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+
+/**
+ * @brief 1x1 unitary matrix for global-phase gates.
+ *
+ * Stores a single complex scalar in row-major layout. Used by operations such
+ * as `GPhaseOp` whose unitary is a global phase factor.
+ */
+struct Matrix1x1 {
+  /// Flat storage of the single matrix entry.
+  std::array<Complex, 1> data{Complex{1.0, 0.0}};
```
burgholzer: Why would this even use an array when the entry is just a single number?
Feels like an abstraction too many to me.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+
+  /**
+   * @brief Constructs a matrix from its single entry.
+   * @param m00 Element at row 0, column 0.
+   * @return A new `Matrix1x1` with the given element.
+   */
+  [[nodiscard]] static Matrix1x1 fromElements(Complex m00) {
+    Matrix1x1 m{};
+    m(0, 0) = m00;
+    return m;
```
burgholzer:
```suggestion
    return {m00};
```
Should also work just fine I'd suppose. Might hold for other places.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+
+  /**
+   * @brief Mutable element access with `(row, col)` indexing.
+   * @param row Row index (ignored; always 0).
+   * @param col Column index (ignored; always 0).
+   * @return Reference to the sole matrix entry.
+   */
+  [[nodiscard]] Complex& operator()(std::size_t row, std::size_t col) {
+    (void)row;
+    (void)col;
```
burgholzer: Not a fan of C-style casts. There's C++'s maybe_unused for that.
But even more so: shouldn't these be error checks?
Or is this unchecked on purpose to be fast?

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+    Matrix2x2 m{};
+    m.data = {m00, m01, m10, m11};
+    return m;
+  }
+
+  /**
+   * @brief Returns the 2x2 identity matrix.
+   * @return Identity matrix `[[1, 0], [0, 1]]`.
+   */
+  [[nodiscard]] static Matrix2x2 identity() { return fromElements(1, 0, 0, 1); }
```
burgholzer: I would believe that this could almost be constexpr. Couldn't it?

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+   */
+  [[nodiscard]] Matrix2x2 operator*(const Matrix2x2& rhs) const {
+    Matrix2x2 out{};
+    for (std::size_t i = 0; i < K_ROWS; ++i) {
+      for (std::size_t j = 0; j < K_COLS; ++j) {
+        out(i, j) = (*this)(i, 0) * rhs(0, j) + (*this)(i, 1) * rhs(1, j);
+      }
+    }
+    return out;
+  }
```
burgholzer: This is 2*2 matrix multiplication. Shouldn't this simply be written out?
This feels like a place where a lot of the low level optimization may come in.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+
+  /**
+   * @brief Returns the conjugate transpose (adjoint) of this matrix.
+   * @return Adjoint matrix `A^\dagger`.
+   */
+  [[nodiscard]] Matrix2x2 adjoint() const {
+    return fromElements(std::conj((*this)(0, 0)), std::conj((*this)(1, 0)),
+                        std::conj((*this)(0, 1)), std::conj((*this)(1, 1)));
```
burgholzer: Instead of the kind of awkward this and operator() usage, I'd probably just directly reference the right array elements throughout this file.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+   * @return True if every entry differs by at most @p tol.
+   */
+  [[nodiscard]] bool isApprox(const Matrix2x2& other,
+                              double tol = UNITARY_MATRIX_TOLERANCE) const {
+    for (std::size_t i = 0; i < data.size(); ++i) {
+      if (std::abs(data[i] - other.data[i]) > tol) {
+        return false;
+      }
+    }
+    return true;
```
burgholzer: Can be realized with std::all_of or similar.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+
+  /**
+   * @brief Returns the 4x4 identity matrix.
+   * @return Identity matrix with ones on the diagonal.
+   */
+  [[nodiscard]] static Matrix4x4 identity() {
+    Matrix4x4 m{};
+    for (std::size_t i = 0; i < K_ROWS; ++i) {
+      m(i, i) = 1.0;
+    }
```
burgholzer: Only one more line and all the calls are explicitly written out without a loop.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+    for (std::size_t i = 0; i < K_ROWS; ++i) {
+      for (std::size_t j = 0; j < K_COLS; ++j) {
+        Complex sum{0.0, 0.0};
+        for (std::size_t k = 0; k < K_COLS; ++k) {
+          sum += (*this)(i, k) * rhs(k, j);
+        }
+        out(i, j) = sum;
+      }
+    }
+    return out;
```
burgholzer: Here it's a bit trickier to say how to optimize this well (maybe the current code already compiles down to very efficient assembly).
Might be worth exploring in CompilerExplorer and compare against the things that Eigen or the Nalgene a library do here.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+  /**
+   * @brief Returns the trace of this matrix.
+   * @return Sum of diagonal entries.
+   */
+  [[nodiscard]] Complex trace() const {
+    Complex t{0.0, 0.0};
+    for (std::size_t i = 0; i < K_ROWS; ++i) {
+      t += (*this)(i, i);
+    }
+    return t;
```
burgholzer: Can be condensed to a single line that adds four values that are directly indexed.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+  void adjointInPlace() {
+    for (std::int64_t i = 0; i < dim_; ++i) {
+      for (std::int64_t j = i + 1; j < dim_; ++j) {
+        const Complex tmp = (*this)(i, j);
+        (*this)(i, j) = std::conj((*this)(j, i));
+        (*this)(j, i) = std::conj(tmp);
+      }
+      (*this)(i, i) = std::conj((*this)(i, i));
+    }
+  }
```
burgholzer: This one is suddenly in place while the others aren't.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -0,0 +1,642 @@
...
+  /**
+   * @brief Checks approximate equality against a fixed 4x4 matrix.
+   *
+   * Returns false if this matrix is not 4x4.
+   *
+   * @param other Fixed-size matrix to compare against.
+   * @param tol Maximum allowed absolute difference per entry.
+   * @return True if dimensions match and every entry differs by at most @p tol.
+   */
+  [[nodiscard]] bool isApprox(const Matrix4x4& other,
```
burgholzer, line 410: Definitions for 4x4 and dynamic but not for 2x2. Feels inconsistent.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+template <> struct IsUnitaryMatrix<Matrix4x4> : std::true_type {};
+/// @brief Specialization for `DynamicMatrix`.
+template <> struct IsUnitaryMatrix<DynamicMatrix> : std::true_type {};
+
+/**
+ * @brief Convenience variable template for `IsUnitaryMatrix<T>::value`.
+ * @tparam T Candidate type.
+ */
+template <typename T>
+inline constexpr bool IS_UNITARY_MATRIX_V = IsUnitaryMatrix<T>::value;
```
burgholzer: This is probably due to our clang tidy config, but I'd actually rather follow the standard library's case style for this and ignore the error here.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
```diff
@@ -0,0 +1,642 @@
...
+ * @tparam Src Source matrix type.
+ * @param out Destination matrix.
+ * @param src Source matrix.
+ * @return Always `true` on success.
+ */
+template <typename Src>
+inline bool assignToDynamic(DynamicMatrix& out, const Src& src) {
+  copyInto(out, src);
+  return true;
+}
```
burgholzer: These feel slightly weird. Why can't these be class methods of the dynamic class?
Would feel more like OOP and proper interfaces.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
burgholzer, line 1: Two general comments: one of the main downsides of Eigen was that it was Header-only and thus needed to be compiled over and over again.
This is not solved here (yet). I would propose to turn the header into a proper library and move as many definitions as reasonable possible to the source file.
Given how practically none of the code here is templated, this should work quite nicely.
The header itself should include as little as possible; using forward declarations wherever that makes sense.
Only then this becomes a lightweight addition.

A lot of the code here feels duplicated between the different classes (probably still the case even after applying the inline suggestions).
Some of this could be mitigated by defining a common Interface for all of these matrix classes.
We are still dealing with 4 classes here that all implement matrices. Might as well give them a common API.
But maybe that is also overkill. So feel free to disagree; just some food for thought. 💭

burgholzer (changes requested):
Really like the direction here @simon1hofmann!
Feels good to see that dependency gone.
I have a bit of cosmetic feedback and a bunch of comments on the matrix library that should hopefully be fairly easy to address. Overall I believe the current implementation is a tiny bit too raw and needs a bit of polish. 
No major blockers though.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
simon1hofmann: Did some experiments locally on arm64 (Apple clang, `-O3`), comparing several scalar reimplementations against Eigen/nalgebra-style approaches. 

### Benchmarks (4×4 `std::complex<double>` multiply)
| Implementation | ns/op |
|---|---:|
| Eigen `Matrix4cd` | **11.9** |
| row-cache + unroll (current) | 48.3 |
| triple loop (original) | 58.5 |
| full source unroll (no row cache) | **47.0** |
| eigen-style coeff loop (col-major) | 54.4 |
| eigen-style coeff full unroll | 56.0 |
| nalgebra-style gemm loop | 59.2 |
| nalgebra-style rank-1 gemv | 64.6 |

### Assembly size (emitted `-O3` function bodies, arm64)
| Implementation | asm lines |
|---|---:|
| Eigen `Matrix4cd` (wrapper + SIMD kernel) | **278** (258 lines of `fcmla` NEON) |
| full source unroll | 1555 |
| row-cache + unroll / `Matrix4x4::operator*` | **599** |
| triple loop | 164 |
| eigen-style coeff loop | 157 |

All in all, I think row-cache + unroll implementation is the best option without Eigen. 

Eigen is faster because it uses CPU SIMD instructions (`fcmla` on ARM). If 4x4 matrix multiplication ever becomes the performance bottleneck, we would either have to rely on some external library again or introduce a small hand-written 4×4 complex SIMD kernel per platform 🤔

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/RXOp.cpp
simon1hofmann: Was flagged by `clang-tidy`:
3rd argument 'm01' (passed to 'm10') looks like it might be swapped with the 4th, 'm00' (passed to 'm11'). Happy to change it back, but does it hurt to make it that explicit?

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
simon1hofmann: Agree on the split into header + source file, should be more efficient now. 

For common interface, I do not really see the benefit and it's probably more of an overkill..

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

mlir/lib/Dialect/QCO/IR/Operations/StandardGates/RXOp.cpp
burgholzer: nah, does not hurt. highly likely this boils down to the same assembly anyway.

mlir/include/mlir/Dialect/QCO/Utils/UnitaryMatrix.h
burgholzer: Thanks for the microbenchmarking 👍🏼 
Fairly interesting that this makes a 4x difference in speed.
This is nothing we need to worry about now for the first implementation as it would only hold us up.
But I would imagine that we would adapt some of the Eigen code ourselves for the most popular platforms. On large programs, quite a bit of matrix manipulation is happening; and we definitely want it to be fast.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
burgholzer: Makes sense to me 👍🏼

burgholzer:
Doing a full review of this now. I'll fix the open clang-tidy comments while I am at it.

burgholzer (approved):
Alright. Finished with my round of cleanups.
Please have a look at the series of commits and check if they make sense.

I am happy with this now. Maybe you could do a last sanity check and/or CodeRabbit run before we merge this 😌

denialhaag:
@coderabbitai review

denialhaag (approved):
@burgholzer asked me to have a brief look at this so we can get #1751 ready as soon as possible. I pushed two very minor improvements just now. Given that CodeRabbit is happy as well, I'll set this PR to auto-merge. 🎉

simon1hofmann:
@denialhaag @burgholzer Thanks a lot for the final changes and finishing this PR 🙏

PR #1671, opened 2026-04-27
✨ Add OpenQASM-to-QC translation

J4MMlE:
## Description

This PR adds a direct OpenQASM-to-QC translation that bypasses the `qc::QuantumComputation`.

The new flow reuses the existing parser and walks the AST directly to emit QC dialect ops. Skipping `QuantumComputation` removes an intermediate step and enables previously unsupported QASM3 features, such as gate modifiers, classical computations, and control flow structures.

### What works

- All standard gates plus Qiskit-style MCX variants
- Gate modifiers: `ctrl @`, `negctrl @`, `inv @`, nested combinations
- Register declarations, qubit/bit allocation
- Measure, reset, barrier
- `if`/`else` over quantum statements
- Hardware qubits convert to `qc.static`
- Broadcasting (register-width gate calls)

### Current limitations

- Gate parameters must be compile-time constants
- `pow` modifier is unsupported (but will/should be once #1603 is merged)
- Layout pragmas are unsupported (they have no equivalent in `QC`)

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] ~~I have updated the documentation to reflect these changes.~~
- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] ~~I have added migration instructions to the upgrade guide (if needed).~~
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

PR #1603, opened 2026-04-01
✨ Add power modifier to MLIR

J4MMlE:
## Description
Adds a power modifier (`pow`) to both MLIR-dialects `QC` and `QCO`.
The power modifier raises an operation `U` by an integer/floating number `k` `pow(k) { U } == U^k` in other words the corresponding unitary matrix gets raised by factor `k` via eigendecomposition: `U = V D V† => U^p = V D^p V†`.
When `k` is an integer `U^k` could also be _unrolled_ ( `U` is then applied `k`-times to the input qubit - but this modifier does **NOT** implement any unrolling - this is left for (possibly) later stages in the pipeline).

### QC
``` mlir
qc.pow (2.000000e+00){
        qc.s %q: !qc.qubit
}
```
### QCO
``` mlir
%out = qco.pow (2.000000e+00) (%a = %in) {
        %a_1 = qco.rz(%c_0) %a : !qco.qubit -> !qco.qubit
        qco.yield %a_1
      } : {!qco.qubit} -> {!qco.qubit}
```

## Canonicalizations
The modifier contains the following canonicalizations in both dialects:
- Nested power flattening: `pow(a) { pow(b) { U } } → pow(a*b) { U }`
- Identity power: `pow(1) { U } → U`
- Zero power: `pow(0) { U } → remove (becomes identity, then eliminated)`
- Negative power: `pow(-r) { U } → pow(r) { inv { U } }`
- Canonical ordering: `pow { ctrl { U } } → ctrl { pow { U } }`
- Canonical ordering for `inv`: `inv { pow { U } } -> pow { inf { U } }`

## Gate specific canonicalizations
- Rotation gates - pow scales θ:
`pow(r) { G(θ, ...) }  →  G(r·θ, ...)`
`G in {GPhase, RX, RY, RZ, P, RXX, RYY, RZX, RZZ, R(θ,φ), XX+YY(θ,β), XX-YY(θ,β)}`
- Diagonal / `P`-Gates: `Z=P(π)`, `S=P(π/2)`, `Sdg=P(-π/2)`, `T=P(π/4)`, `Tdg=P(-π/4)`:
`pow(r) { G }`  →  named gate for normalized `r*base_angle`, else `P(r*base_angle)`
[...]

Issue #1726, opened 2026-05-20
✨ Add support for classical registers


denialhaag:
#### Description

Currently, QC and QCO do not have native support for classical registers. Their builders do have methods to allocate classical registers, but these only return collections of `Bit` structures that store some metadata. We should adapt these methods to allocate an MLIR object that can store and retrieve measurement results.

#### Open questions

- Should classical registers be represented as `tensor` or `memref` types? For reference, `jeff` uses `tensor` types. How are classical registers represented in [openqasm/qe-compiler](https://github.com/openqasm/qe-compiler)?
- Should the same representation be used in QC and QCO? Using the same representation would significantly simplify internal conversions.

#### Acceptance criteria

- `QCProgramBuilder::allocClassicalBitRegister()` and `QCOProgramBuilder::allocClassicalBitRegister()` create some sort of MLIR object. It likely makes sense to mimic the behavior of the `allocQubitRegister()` methods and return a custom structure containing the loaded entries.
- The `register_name`, `register_size`, and `register_index` attributes are removed from `qc::MeasureOp` and `qco::MeasureOp`.
- The conversion of `qc::MeasureOp`s to QIR is improved. The original register information is retained during conversion to the Adaptive Profile. During conversion to the Base Profile, all bits are converted to statically allocated ones. Implement the output recording as suggested in https://github.com/munich-quantum-toolkit/core/pull/1710#discussion_r3290853101.
- `translateQuantumComputationToQC()` can translate measurements inside `qc::IfElseOperation`s (see https://github.com/munich-quantum-toolkit/core/pull/1717#pullrequestreview-4335539780 for context).

Issue #1707, opened 2026-05-13
✨ Convert `scf.while` to `jeff`


denialhaag:
https://github.com/munich-quantum-toolkit/core/pull/1706 did not add the conversion between `scf.while` and `jeff.while`.

This is because `jeff.while`, in its current state, does not allow for the modification of the state in the `condition` region: https://github.com/unitaryfoundation/jeff/blob/5559ee12aec810ff1264e1bf1d079637529f7365/impl/capnp/jeff.capnp#L1450-L1455

Once https://github.com/unitaryfoundation/jeff/issues/4 is addressed, we should add a conversion between `scf.while` and (the new) `jeff.while`.

PR #1780, opened 2026-06-11
✨ Add OpenQASM-to-QC translation

denialhaag:
## Description

This is a rebased version of #1671 after #1671 was automatically closed when #1751 was merged.

Original description by @J4MMlE:

> ## Description
>
> This PR adds a direct OpenQASM-to-QC translation that bypasses the `qc::QuantumComputation`.
> 
> The new flow reuses the existing parser and walks the AST directly to emit QC dialect ops. Skipping `QuantumComputation` removes an intermediate step and enables previously unsupported QASM3 features, such as gate modifiers, classical computations, and control flow structures.
> 
> ### What works
> 
> - All standard gates plus Qiskit-style MCX variants
> - Gate modifiers: `ctrl @`, `negctrl @`, `inv @`, nested combinations
> - Register declarations, qubit/bit allocation
> - Measure, reset, barrier
> - `if`/`else` over quantum statements
> - Hardware qubits convert to `qc.static`
> - Broadcasting (register-width gate calls)
> 
> ### Current limitations
> 
> - Gate parameters must be compile-time constants
> - `pow` modifier is unsupported (but will/should be once [#1603](https://github.com/munich-quantum-toolkit/core/pull/1603) is merged)
> - Layout pragmas are unsupported (they have no equivalent in `QC`)

## Checklist


denialhaag:
@coderabbitai full review

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    if (!expr) {
+      return defaultValue;
+    }
+    const auto constVal = std::dynamic_pointer_cast<qasm3::Constant>(expr);
+    if (!constVal) {
+      throw qasm3::CompilerError("Expected a constant integer expression.",
+                                 debugInfo);
+    }
+    return static_cast<size_t>(constVal->getUInt());
+  }
```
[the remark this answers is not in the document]
denialhaag: This is a conscious decision. Support for classical computations will be added in the future.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    }
+
+    if (target->indices.size() > 1) {
+      throw qasm3::CompilerError("Only single-index expressions are supported.",
+                                 debugInfo);
+    }
+    const auto& indexExpression = target->indices[0]->indexExpressions[0];
+    const auto index = evaluatePositiveConstant(indexExpression, debugInfo);
+    bits.push_back(creg[static_cast<int64_t>(index)]);
+    return bits;
```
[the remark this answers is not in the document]
denialhaag: This is a conscious decision. Support for this may be added in the future.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    for (const auto& stmt : statements) {
+      if (const auto gateCall =
+              std::dynamic_pointer_cast<qasm3::GateCallStatement>(stmt)) {
+        applyGateCallStatement(gateCall, qubitRegisters);
+        continue;
+      }
+      throw qasm3::CompilerError(
+          "If statements with non-quantum statements are not supported.",
+          debugInfo);
+    }
```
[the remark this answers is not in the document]
denialhaag: This is a conscious decision. Support for this may be added in the future.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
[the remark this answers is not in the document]
denialhaag: Sorry, I responded to the wrong comment. My most recent commit should fix this, right?

mlir/unittests/Dialect/QC/Translation/test_qasm3_translation.cpp
```diff
@@ -0,0 +1,394 @@
...
+            "ResetMultipleQubitsAfterSingleOp",
+            qasm::resetMultipleQubitsAfterSingleOp,
+            MQT_NAMED_BUILDER(qc::resetMultipleQubitsAfterSingleOp)},
+        QASM3TranslationTestCase{
+            "RepeatedResetAfterSingleOp", qasm::repeatedResetAfterSingleOp,
+            MQT_NAMED_BUILDER(qc::repeatedResetAfterSingleOp)},
+        QASM3TranslationTestCase{"GlobalPhase", qasm::globalPhase,
+                                 MQT_NAMED_BUILDER(qc::globalPhase)},
+        QASM3TranslationTestCase{"Identity", qasm::identity,
+                                 MQT_NAMED_BUILDER(qc::identity)},
```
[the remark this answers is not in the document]
denialhaag: Thanks for catching that! My most recent commit should fix this, right?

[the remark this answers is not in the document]
denialhaag: Can you check again? It's in lines 165 and 166.

denialhaag:
@coderabbitai review

denialhaag:
@coderabbitai review

denialhaag:
@coderabbitai review

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1055 @@
...
+      throw qasm3::CompilerError("No classical bit of register '" + regName +
+                                     "' has been measured yet.",
+                                 debugInfo);
+    }
+    const auto& regBits = it->second;
+
+    if (id->indices.empty()) {
+      assert(regBits.size() == 1);
+      return regBits[0];
+    }
```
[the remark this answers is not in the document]
denialhaag: The `lookupBitValue` function is only ever entered in case of a single bit. 🤔

mlir/tools/mqt-cc/mqt-cc.cpp
burgholzer, line 1: Since none of this depends on the old MQT Core IR anymore, I am wondering whether we should split the translation target into to separate ones, where the OpenQASM3 one no longer links against MQT::CoreIR.
This would mean that `mqt-cc` as an executable has no direct connection to the old IR anymore.
This is something that can be done here.

If we manage to (somehow) find a way to interface with Qiskit without the MQT Core IR, we may finally be able to disable RTTI and Exception handling for the MLIR part of the code base.
I suppose the easiest way to get close to feature parity would be to expose the QCProgramBuilder to Python and replicate the existing Qiskit conversion through that API.
This is something to be done in a separate PR 😌

mlir/include/mlir/Dialect/QC/Translation/TranslateQASM3ToQC.h
```diff
@@ -0,0 +1,39 @@
...
+translateQASM3ToQC(MLIRContext* context, const std::string& filename);
+
+/**
+ * @brief Translate an OpenQASM3 program to the QC dialect.
+ *
+ * @param context MLIRContext to create the module in.
+ * @param input Stream containing the OpenQASM3 program.
+ */
+[[nodiscard]] OwningOpRef<ModuleOp> translateQASM3ToQC(MLIRContext* context,
+                                                       std::istream& input);
```
burgholzer: IIRC LLVM ships its own streams and recommends against using the stdlib ones. Worth checking.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1055 @@
...
+  visitInitialLayout(std::shared_ptr<qasm3::InitialLayout> layout) override {
+    throw qasm3::CompilerError("InitialLayout pragmas are not supported.",
+                               layout->debugInfo);
+  }
+
+  void visitOutputPermutation(
+      std::shared_ptr<qasm3::OutputPermutation> perm) override {
+    throw qasm3::CompilerError("OutputPermutation pragmas are not supported.",
+                               perm->debugInfo);
+  }
```
burgholzer: I think these should not be errors. The pragmas are informational for the most part. They are otherwise simply comments.
Just skipping them is fine.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1055 @@
...
+      }
+    }
+  }
+
+  void visitIfStatement(std::shared_ptr<qasm3::IfStatement> stmt) override {
+    if (stmt->thenStatements.empty()) {
+      throw qasm3::CompilerError(
+          "If statements with empty then blocks are not supported.",
+          stmt->debugInfo);
+    }
```
burgholzer: Why not?

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
```diff
@@ -0,0 +1,1055 @@
...
+        resolvedId = resolvedId.substr(1);
+        ++numCompatControls;
+      }
+      if (numCompatControls > 0) {
+        it = gates.find(resolvedId);
+      }
+    }
+
+    if (it == gates.end()) {
+      throw qasm3::CompilerError("No OpenQASM 3 definition found for gate '" +
```
burgholzer:
```suggestion
      throw qasm3::CompilerError("No definition found for gate '" +
```

burgholzer (changes requested):
This generally looks very good to me as a direct port of the QuantumComputation QASM Importer 🙌
I only have a couple of inline comments or requests for changes.
We should be able to merge this quickly.

A thought that came up while reviewing this: how much is the current architecture of the parser and the visitor implementation limiting us in terms of expanding the capabilities?
There are quite a few things from the OQ3 spec that we would want to support (loops being one of the most prominent ones).
Would it be interesting to define our own OpenQASM3 dialect (inspired by the qe-compiler one) that can directly parse in QASM files. And then we would write a conversion to MLIR standard dialects plus QC. Does that make sense? Is that overkill? To me it feels like this could generally make sense. What do you think?

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
denialhaag: Good point! Done in https://github.com/munich-quantum-toolkit/core/pull/1780/commits/e8133cedad91efb4fdfd69712ce95e1c046cf002.

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
denialhaag: They are now! See https://github.com/munich-quantum-toolkit/core/pull/1780/commits/cce81349e4bbf62fd49ca6b735c53e9e457562a2.

mlir/include/mlir/Dialect/QC/Translation/TranslateQASM3ToQC.h
denialhaag: I went with `llvm::SourceMgr` in https://github.com/munich-quantum-toolkit/core/pull/1780/commits/a7a81b9b1f6b32bc52b974b986991abad2ea9e7e. This is also used in [`Translation.h`](https://mlir.llvm.org/doxygen/Translation_8h_source.html), which we should look into at some point.

Let me know what you think!

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
denialhaag: This is actually semi-intentional to be in contrast with this error message:

https://github.com/munich-quantum-toolkit/core/blob/77043ba95add0ef175f73b792436d47b1e287a36/mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp#L682-L683

mlir/tools/mqt-cc/mqt-cc.cpp
denialhaag: > Since none of this depends on the old MQT Core IR anymore, I am wondering whether we should split the translation target into to separate ones, where the OpenQASM3 one no longer links against MQT::CoreIR.
> This would mean that `mqt-cc` as an executable has no direct connection to the old IR anymore.
> This is something that can be done here.

Good point! Done in https://github.com/munich-quantum-toolkit/core/pull/1780/commits/86bbf4bfdc523b6649b572fccb7109b0778940f8.

> If we manage to (somehow) find a way to interface with Qiskit without the MQT Core IR, we may finally be able to disable RTTI and Exception handling for the MLIR part of the code base.
> I suppose the easiest way to get close to feature parity would be to expose the QCProgramBuilder to Python and replicate the existing Qiskit conversion through that API.
> This is something to be done in a separate PR 😌 

I'll keep that in mind!

denialhaag:
> A thought that came up while reviewing this: how much is the current architecture of the parser and the visitor implementation limiting us in terms of expanding the capabilities?
> There are quite a few things from the OQ3 spec that we would want to support (loops being one of the most prominent ones).
> Would it be interesting to define our own OpenQASM3 dialect (inspired by the qe-compiler one) that can directly parse in QASM files. And then we would write a conversion to MLIR standard dialects plus QC. Does that make sense? Is that overkill? To me it feels like this could generally make sense. What do you think?

I will think about this a bit and respond tomorrow!

denialhaag:
@coderabbitai review

mlir/include/mlir/Dialect/QC/Translation/TranslateQASM3ToQC.h
```diff
@@ -0,0 +1,28 @@
...
+namespace mlir::qc {
+
+/**
+ * @brief Translate an OpenQASM 3 program to a QC program.
+ *
+ * @param context MLIRContext to create the module in.
+ * @param sourceMgr Source manager containing the OpenQASM3 program.
+ */
+[[nodiscard]] OwningOpRef<ModuleOp>
+translateQASM3ToQC(MLIRContext* context, llvm::SourceMgr& sourceMgr);
```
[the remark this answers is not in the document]
denialhaag: Done in https://github.com/munich-quantum-toolkit/core/pull/1780/commits/c06be11507c818f9f6a84bbcfd2053e3e79b6bf6. Does the commit look good to you?

mlir/lib/Dialect/QC/Translation/TranslateQASM3ToQC.cpp
burgholzer: Ah, that makes sense.
In that case, maybe drop the "3" because it could also be an OpenQASM 2 file.

mlir/include/mlir/Dialect/QC/Translation/TranslateQASM3ToQC.h
burgholzer: Hm. The current solution feels somewhat incomplete. Previously one could directly pass a string or file. Now one has to do a bit of extra work.

In principle, I am fine with this for this PR, but I have a feeling that we should really be looking into the official translation interface, which seems quite convenient.

mlir/tools/mqt-cc/CMakeLists.txt
```diff
@@ -17,7 +17,6 @@ target_link_libraries(
           MLIRSupport
           # Required for OpenQASM parsing
           MQT::CoreQASM
-          MQT::CoreIR
```
burgholzer, line 20: How does that work? This is still one target that also contains the QuantumComputation translation. I suppose the QASM target also links the Core IR?
That might be something we want to disentangle (depending on what you think about the bigger plan for OQ3 support).

burgholzer (changes requested):
This is very close to being mergable 🙂

mlir/tools/mqt-cc/CMakeLists.txt
denialhaag: Sorry, I misunderstood https://github.com/munich-quantum-toolkit/core/pull/1780#discussion_r3431362518. I think this still works because `MLIRQCTranslation` still links against `MQT::CoreIR`. By that logic, we could also remove the explicit link against `MQT::CoreQASM` here. I'll do that for now.

mlir/include/mlir/Dialect/QC/Translation/TranslateQASM3ToQC.h
denialhaag: Okay, I have (hopefully) made some improvements to this now:

- In https://github.com/munich-quantum-toolkit/core/pull/1780/changes/19f5d957bfead6c4b0fd944c81330982443cfb4f, I have added an overload that directly accepts a string containing an OpenQASM program.
- In https://github.com/munich-quantum-toolkit/core/pull/1780/changes/683f558bcde142a754e0770a39d5dd3c0cead3e2, I have added a `mqt-cc-translate` tool. It can be invoked via `./build/release/mlir/tools/mqt-cc-translate/mqt-cc-translate --qasm3-to-qc test.qasm`. The `SourceMgr` signature of our `translateQASM3ToQC` function made the implementation very straightforward.

In general, I'm not sure if this made things better or worse, though. Let me know what you think!

burgholzer: The overload commit LGTM.
I do not really know what to make of the other one. What would we be using that tool for?
I think I'd actually prefer to revert that last one and only keep the first one. Then this should be ready to go in

denialhaag: > I do not really know what to make of the other one. What would we be using that tool for?
> I think I'd actually prefer to revert that last one and only keep the first one.

Yeah, I don't really know why one wouldn't use `mqt-cc` to translate an OpenQASM program into a QC program. I have removed the `mqt-cc-translate` tool again now. 😌

burgholzer (approved):

denialhaag:
Thanks a lot for getting this started, @J4MMlE! Your initial version was a great foundation! 🙂

denialhaag:
> A thought that came up while reviewing this: how much is the current architecture of the parser and the visitor implementation limiting us in terms of expanding the capabilities? There are quite a few things from the OQ3 spec that we would want to support (loops being one of the most prominent ones).

The current implementation is certainly limiting. It's probably not wise to keep updating MQT Core's `qasm3` package if are anyway transition away from the `ir` package.

> Would it be interesting to define our own OpenQASM3 dialect (inspired by the qe-compiler one) that can directly parse in QASM files. And then we would write a conversion to MLIR standard dialects plus QC. Does that make sense? Is that overkill? To me it feels like this could generally make sense. What do you think?

I also like the sound of the idea, but I'm not sure if we are unnecessarily creating additional work for ourselves. If such an OpenQASM 3 dialect has a chance of becoming established, I see the point in going for it. However, if it were to be an internal "tool" only, I feel like we'd be creating another hoop to jump through. For example, in my opinion, the current translation from OpenQASM 3 to QC benefits from the `QCProgramBuilder`. If we were to define the translation to a new OpenQASM 3 dialect instead, do we also want an `OpenQASM3ProgramBuilder`? If yes, any infrastructure changes would need to be implemented in two _additional_ locations (`OpenQASM3ProgramBuilder` and conversion from the OpenQASM 3 dialect to QC). Thought differently, if we were to define an OpenQASM 3 dialect with a builder, do we still want or need the QC dialect and its builder? Could we not just convert directly from OpenQASM 3 to QCO and QIR? The QC dialect would be losing its position as the input (i.e., translation) dialect anyway. Also, if we define an OpenQASM 3 dialect, should we also define a Qiskit dialect to ease the translation from Qiskit? (I'll admit that I'm playing devil's advocate here. The questions are more nuanced than that.)

In short, if we plan to establish the OpenQASM 3 dialect as a standard used by others (much like `jeff-mlir`), I like the idea. If not, I think we'd end up with one dialect (either OpenQASM 3 or QC) that would be a potentially pointless hoop in the pipeline.

Let me know what you think! One way or another, we need to ensure that we can translate arbitrary OpenQASM 3 programs into `mqt-cc`.

burgholzer:
> > A thought that came up while reviewing this: how much is the current architecture of the parser and the visitor implementation limiting us in terms of expanding the capabilities? There are quite a few things from the OQ3 spec that we would want to support (loops being one of the most prominent ones).
> 
> The current implementation is certainly limiting. It's probably not wise to keep updating MQT Core's `qasm3` package if are anyway transition away from the `ir` package.

Yeah, I figured as much. The current parser implementation still ties quite heavily to the `ir` package and it is also shaped by quite a few of the limitations of the legacy MQT Core IR. Support for loops as well as general classical computation would be really nice to have and should actually match quite naturally to concept in MLIR.

> > Would it be interesting to define our own OpenQASM3 dialect (inspired by the qe-compiler one) that can directly parse in QASM files. And then we would write a conversion to MLIR standard dialects plus QC. Does that make sense? Is that overkill? To me it feels like this could generally make sense. What do you think?
> 
> I also like the sound of the idea, but I'm not sure if we are unnecessarily creating additional work for ourselves. If such an OpenQASM 3 dialect has a chance of becoming established, I see the point in going for it. However, if it were to be an internal "tool" only, I feel like we'd be creating another hoop to jump through. For example, in my opinion, the current translation from OpenQASM 3 to QC benefits from the `QCProgramBuilder`. If we were to define the translation to a new OpenQASM 3 dialect instead, do we also want an `OpenQASM3ProgramBuilder`? If yes, any infrastructure changes would need to be implemented in two _additional_ locations (`OpenQASM3ProgramBuilder` and conversion from the OpenQASM 3 dialect to QC). Thought differently, if we were to define an OpenQASM 3 dialect with a builder, do we still want or need the QC dialect and its builder? Could we not just convert directly from OpenQASM 3 to QCO and QIR? The QC dialect would be losing its position as the input (i.e., translation) dialect anyway. Also, if we define an OpenQASM 3 dialect, should we also define a Qiskit dialect to ease the translation from Qiskit? (I'll admit that I'm playing devil's advocate here. The questions are more nuanced than that.)
> 
> In short, if we plan to establish the OpenQASM 3 dialect as a standard used by others (much like `jeff-mlir`), I like the idea. If not, I think we'd end up with one dialect (either OpenQASM 3 or QC) that would be a potentially pointless hoop in the pipeline.
> 
> Let me know what you think! One way or another, we need to ensure that we can translate arbitrary OpenQASM 3 programs into `mqt-cc`.

I would not like to create an `OpenQASM3ProgramBuilder`. On the one hand for the overhead reasons you mention. On the other hand, because OpenQASM is a well spec'd out language, where we can simply consume QASM files (or strings) as input. Furthermore, OpenQASM is one of the potential entry points into `mqt-cc`. It does not quite feel right to write our own "producer" for these files. That's the job of a frontend. The more frontends or program formats we connect, the less relevant the QCProgramBuilder becomes in my opinion; although it will probably remain relevant for quite some while because we will likely be able to represent and manipulate more complex hybrid program constructs in `mqt-cc` than we could create through translating from any frontend.

That being said, I would generally view an OpenQASM3 dialect as something fairly similar to `jeff-mlir`. A project that we maintain for the community and that can be consumed by the community. In the best case, we would define the dialect in a way that would allow us to directly read in OpenQASM3 programs (probably through custom parsers) and that allows us to emit an OpenQASM3 program from an MLIR module that only contains OpenQASM3 operations (probably through custom printers).
Similarly to `jeff-mlir`, we would provide some passes to map OQ3 constructs to standard dialects and back.
I believe this should be doable.
I am not yet quite sure whether we would want to maintain such a dialect out-of-tree (like `jeff-mlir`) or in-tree to keep it closer to the `mqt-cc` source code.

On the point whether we'd actually still need the QC dialect: I think we'd still need it. OQ3 does not define individual gates (X, Y, Z, S, etc.); it also mixes classical and quantum constructs. Translating that to clearly separated dialects (QC + Standard Dialects) seems valuable to me.

How does all of that sound?

denialhaag:
> I would not like to create an `OpenQASM3ProgramBuilder`. On the one hand for the overhead reasons you mention. On the other hand, because OpenQASM is a well spec'd out language, where we can simply consume QASM files (or strings) as input. Furthermore, OpenQASM is one of the potential entry points into `mqt-cc`. It does not quite feel right to write our own "producer" for these files. That's the job of a frontend. The more frontends or program formats we connect, the less relevant the QCProgramBuilder becomes in my opinion; although it will probably remain relevant for quite some while because we will likely be able to represent and manipulate more complex hybrid program constructs in `mqt-cc` than we could create through translating from any frontend.
> 
> That being said, I would generally view an OpenQASM3 dialect as something fairly similar to `jeff-mlir`. A project that we maintain for the community and that can be consumed by the community. In the best case, we would define the dialect in a way that would allow us to directly read in OpenQASM3 programs (probably through custom parsers) and that allows us to emit an OpenQASM3 program from an MLIR module that only contains OpenQASM3 operations (probably through custom printers).
> Similarly to `jeff-mlir`, we would provide some passes to map OQ3 constructs to standard dialects and back. I believe this should be doable.
> I am not yet quite sure whether we would want to maintain such a dialect out-of-tree (like `jeff-mlir`) or in-tree to keep it closer to the `mqt-cc` source code.
> 
> On the point whether we'd actually still need the QC dialect: I think we'd still need it. OQ3 does not define individual gates (X, Y, Z, S, etc.); it also mixes classical and quantum constructs. Translating that to clearly separated dialects (QC + Standard Dialects) seems valuable to me.
> 
> How does all of that sound?

That all sounds good to me! I would be happy to get started on this rather soon, maybe after #1707 and potentially #1726. I don't have a super strong opinion on where the dialect should live. A repsoitory that contains the dialect, bi-directional translations, and bi-directional conversions (of everything but the gates) sounds neat, but it could also make sense within MQT Core. 🤔

burgholzer:
> > I would not like to create an `OpenQASM3ProgramBuilder`. On the one hand for the overhead reasons you mention. On the other hand, because OpenQASM is a well spec'd out language, where we can simply consume QASM files (or strings) as input. Furthermore, OpenQASM is one of the potential entry points into `mqt-cc`. It does not quite feel right to write our own "producer" for these files. That's the job of a frontend. The more frontends or program formats we connect, the less relevant the QCProgramBuilder becomes in my opinion; although it will probably remain relevant for quite some while because we will likely be able to represent and manipulate more complex hybrid program constructs in `mqt-cc` than we could create through translating from any frontend.
> 
> > 
> 
> > That being said, I would generally view an OpenQASM3 dialect as something fairly similar to `jeff-mlir`. A project that we maintain for the community and that can be consumed by the community. In the best case, we would define the dialect in a way that would allow us to directly read in OpenQASM3 programs (probably through custom parsers) and that allows us to emit an OpenQASM3 program from an MLIR module that only contains OpenQASM3 operations (probably through custom printers).
> 
> > Similarly to `jeff-mlir`, we would provide some passes to map OQ3 constructs to standard dialects and back. I believe this should be doable.
> 
> > I am not yet quite sure whether we would want to maintain such a dialect out-of-tree (like `jeff-mlir`) or in-tree to keep it closer to the `mqt-cc` source code.
> 
> > 
> 
> > On the point whether we'd actually still need the QC dialect: I think we'd still need it. OQ3 does not define individual gates (X, Y, Z, S, etc.); it also mixes classical and quantum constructs. Translating that to clearly separated dialects (QC + Standard Dialects) seems valuable to me.
> 
> > 
> 
> > How does all of that sound?
> 
> 
> 
> That all sounds good to me! I would be happy to get started on this rather soon, maybe after #1707 and potentially #1726. I don't have a super strong opinion on where the dialect should live. A repsoitory that contains the dialect, bi-directional translations, and bi-directional conversions (of everything but the gates) sounds neat, but it could also make sense within MQT Core. 🤔 

Sounds like a plan to me!
It's probably best to get any internal dialect changes in before adding another intergration point that has to be updated constantly. I'd add the power modifier to the list of things that should be merged before the introduction of the dialect.

Since there is no obvious best choice for where the dialect should live, I'd argue let's keep it in MQT Core for now. We can always pull it out afterwards if that should proof beneficial.

PR #1781, opened 2026-06-11
♻️ Refactor IR Verification Logic

### LINKED ISSUE #1752: 🐛 MLIR Module equivalence verifier fails for equivalent circuits with different insertion order
### System and Environment Information

- mlir/llvm version 22.1.5
- OS: Ubuntu 22.04

### Bug Description

The current module equivalence checker in the MLIR tests fail if the insertion order of qubits to the registers are different at the end. The `areModulesEquivalentWithPermutations` function should still succeed even if the insertion order at the end is different, as the resulting circuit is still equivalent.

### Steps to Reproduce

The following two modules are not equivalent according to the verifier:
```
 module {                                                                                                             
     %0 = qtensor.alloc(%c2) : tensor<2x!qco.qubit>                                                                   
     %out_tensor, %result = qtensor.extract %0[%c0] : tensor<2x!qco.qubit>                                            
     %out_tensor_0, %result_1 = qtensor.extract %out_tensor[%c1] : tensor<2x!qco.qubit>                               
     %1 = qco.alloc : !qco.qubit                                                                                      
     %2 = qco.h %1 : !qco.qubit -> !qco.qubit                                                                         
     %controls_out, %targets_out = qco.ctrl(%2) targets (%arg0 = %result) {                                           
       %5 = qco.x %arg0 : !qco.qubit -> !qco.qubit                                                                    
       qco.yield %5 : !qco.qubit                                                                                      
     } : ({!qco.qubit}, {!qco.qubit}) -> ({!qco.qubit}, {!qco.qubit})                                                 
     %controls_out_2, %targets_out_3 = qco.ctrl(%controls_out) targets (%arg0 = %result_1) {                          
       %5 = qco.x %arg0 : !qco.qubit -> !qco.qubit                                                                    
       qco.yield %5 : !qco.qubit                                                                                      
     } : ({!qco.qubit}, {!qco.qubit}) -> ({!qco.qubit}, {!qco.qubit})                                                 
     qco.sink %controls_out_2 : !qco.qubit                                                                            
     %3 = qtensor.insert %targets_out_3 into %out_tensor_0[%c1] : tensor<2x!qco.qubit>                                
     %4 = qtensor.insert %targets_out into %3[%c0] : tensor<2x!qco.qubit>                                             
     qtensor.dealloc %4 : tensor<2x!qco.qubit>                                                                        
     return %c0_i64 : i64                                                                                             
   }                                                                                                                  
 } 
``` 
```
module {                                                                                                             
     %0 = qtensor.alloc(%c2) : tensor<2x!qco.qubit>                                                                   
     %out_tensor, %result = qtensor.extract %0[%c0] : tensor<2x!qco.qubit>                                            
     %out_tensor_0, %result_1 = qtensor.extract %out_tensor[%c1] : tensor<2x!qco.qubit>                               
     %1 = qco.alloc : !qco.qubit                                                                                      
     %2 = qco.h %1 : !qco.qubit -> !qco.qubit                                                                         
     %controls_out, %targets_out = qco.ctrl(%2) targets (%arg0 = %result) {                                           
       %5 = qco.x %arg0 : !qco.qubit -> !qco.qubit                                                                    
       qco.yield %5 : !qco.qubit                                                                                      
     } : ({!qco.qubit}, {!qco.qubit}) -> ({!qco.qubit}, {!qco.qubit})                                                 
     %controls_out_2, %targets_out_3 = qco.ctrl(%controls_out) targets (%arg0 = %result_1) {                          
       %5 = qco.x %arg0 : !qco.qubit -> !qco.qubit                                                                    
       qco.yield %5 : !qco.qubit                                                                                      
     } : ({!qco.qubit}, {!qco.qubit}) -> ({!qco.qubit}, {!qco.qubit})                                                 
     qco.sink %controls_out_2 : !qco.qubit                                                                            
     %3 = qtensor.insert %targets_out into %out_tensor_0[%c0] : tensor<2x!qco.qubit>                                  
     %4 = qtensor.insert %targets_out_3 into %3[%c1] : tensor<2x!qco.qubit>                                           
     qtensor.dealloc %4 : tensor<2x!qco.qubit>                                                                        
     return %c0_i64 : i64                                                                                             
   }                                                                                                                  
 } 
```

MatthiasReumann:
## Description

Among other things, resolves #1752. The goal of this pull request is to simplify the verifier, while improving its performance, and make it somewhat bulletproof by explicitly stating its assumptions. 

## Checklist


mlir/unittests/programs/qc_programs.cpp
```diff
@@ -277,11 +277,10 @@ void trivialControlledX(QCProgramBuilder& b) {
...
 void repeatedControlledX(QCProgramBuilder& b) {
-  auto control = b.allocQubit();
-  b.h(control);
-  for (auto i = 0; i < 50; i++) {
-    auto qubit = b.allocQubit();
-    b.cx(control, qubit);
+  auto q = b.allocQubitRegister(2);
+  b.h(q[0]);
+  for (auto i = 0; i < 2; i++) {
+    b.cx(q[0], q[1]);
```
burgholzer: This has become a bit of a pointless test now.
We probably do not do this yet, but the CNOTs being produced here would always cancel, which makes the program useless.
The previous test at least generated a GHZ state. Could we restore that somehow?

mlir/lib/Support/IRVerification.cpp
```diff
@@ -11,407 +11,251 @@
...
-  if (lhsChains.size() != rhsChains.size()) {
-    return false;
+  workset.insert_range(rhs.getTargets());
+  for (const auto& trgt : lhs.getTargets()) {
+    const auto& v = m.lookup(trgt);
+    if (!workset.contains(v)) {
+      return false;
+    }
+    workset.erase(v);
   }
```
burgholzer, line 219: This feels dangerous. Targets typically have a semantic value. DCX q0, q1 is not the same as DCX q1, q0.
This feels like it would result in false positives. Same holds below

burgholzer (changes requested):
Thanks @MatthiasReumann for working on this! 🙏🏼 much appreciated.
To be honest, I did not fully think through all of the verifier rewrites, which are quite substantial. But they do look convincing.

I just pushed some fairly minor cosmetic changes (things that CLion showed as warnings).
Other than that, I have two smaller comments inline and one bigger one here:
It is nice that the existing test suite is still running. However, that was the status quo also before this PR.
To gain some certainty that this addresses the underlying problems we have been trying to solve, it would be helpful to add at least one new test case that would have failed before this PR.

mlir/unittests/programs/qc_programs.cpp
MatthiasReumann: Good catch! This is definitely a mistake. Updated it to a GHZ state again.

mlir/lib/Support/IRVerification.cpp
MatthiasReumann: Exactly! But the modifier does only specify which targets are used and not in which order right? There is even such an example in the `qco-to-qc` unit-test suite:

```mlir
module {
    %c3 = arith.constant 3 : index
    %alloc = memref.alloc() : memref<4x!qc.qubit>
    %0 = memref.load %alloc[%c0] : memref<4x!qc.qubit>
    %1 = memref.load %alloc[%c1] : memref<4x!qc.qubit>
    %2 = memref.load %alloc[%c2] : memref<4x!qc.qubit>
    %3 = memref.load %alloc[%c3] : memref<4x!qc.qubit>
    qc.ctrl(%0, %1) targets (%arg0 = %3, %arg1 = %2) { // ← (3, 2)
      qc.dcx %arg1, %arg0 : !qc.qubit, !qc.qubit // ← (arg1, arg0)
      qc.yield
    } : {!qc.qubit, !qc.qubit}, {!qc.qubit, !qc.qubit}
    memref.dealloc %alloc : memref<4x!qc.qubit>
    return %c0_i64 : i64
  }
}
```
```mlir
module {
    %c3 = arith.constant 3 : index
    %alloc = memref.alloc() : memref<4x!qc.qubit>
    %0 = memref.load %alloc[%c0] : memref<4x!qc.qubit>
    %1 = memref.load %alloc[%c1] : memref<4x!qc.qubit>
    %2 = memref.load %alloc[%c2] : memref<4x!qc.qubit>
    %3 = memref.load %alloc[%c3] : memref<4x!qc.qubit>
    qc.ctrl(%0, %1) targets (%arg0 = %2, %arg1 = %3) { // ← (2, 3)
      qc.dcx %arg0, %arg1 : !qc.qubit, !qc.qubit // ← (arg0, arg1)
      qc.yield
    } : {!qc.qubit, !qc.qubit}, {!qc.qubit, !qc.qubit}
    memref.dealloc %alloc : memref<4x!qc.qubit>
    return %c0_i64 : i64
  }
} 
```

The new verifier now first verifies if the `ctrl` is equivalent and then proceeds to recurse into the nested block structure to verify if the `dcx` is equivalent (which is the case here).

MatthiasReumann:
@burgholzer Thanks for the review! 🙏

> To gain some certainty that this addresses the underlying problems we have been trying to solve, it would be helpful to add at least one new test case that would have failed before this PR.

I've added an additional test to the `qtensor-ir` suite. The test fails on `main` but succeeds on this PR.

mlir/lib/Support/IRVerification.cpp
burgholzer: Fair. I think somewhere along the way we dropped the condition that the unitary in the modifier region must use the qubits in the same order that they are declared as block arguments.
Dropping this makes sense. In the past, this has led to some problems with the DCX inverse simplification though as it flips the targets and invalidates that invariant.
If everything works as expected now without that invariant, I am happy 😌

mlir/unittests/programs/qco_programs.h
```diff
@@ -1122,4 +1122,13 @@ void qtensorInsertExtractIndexMismatch(QCOProgramBuilder& b);
 /// Inserts a qubit into a tensor and extracts it immediately at the same index.
 void qtensorInsertExtractSameIndex(QCOProgramBuilder& b);
 
+/// Extracts three qubits with ascending index (0, 1, 2), performs a
+/// computation, and finally inserts the qubits in ascending order (0, 1, 2).
+void qtensorChain(QCOProgramBuilder& b);
+
+/// Performs the same computation as the `qtensorChain` function, but uses
+/// qubits immediately after the extract and inserts the qubits in descending
+/// order (0, 1, 2).
```
burgholzer:
```suggestion
/// order (2, 1, 0).
```
c&p error

burgholzer (approved):
LGTM modulo one typo.
Feel free to resolve and merge after CI is green.

PR #1782, opened 2026-06-12
⚡️ Improve canonicalizations for gate fusion and inverse cancellation

simon1hofmann:
## Description

Adds a few merge patterns and consolidates shared logic in `QCOUtils.h`.

Additional patterns:
For `r`: merge when `phi` matches (sum `theta`).
For `xx_plus_yy` and `xx_minus_yy`: sum `theta` with **swapped** target wires.

## AI Assistance

Used Composer via Cursor for parts of this change. I reviewed the full diff and take responsibility for everything in this PR.

## Checklist


simon1hofmann:
@coderabbitai full review

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -479,7 +482,8 @@ INSTANTIATE_TEST_SUITE_P(
         QCOTestCase{"CanonicalizeRToRx", MQT_NAMED_BUILDER(canonicalizeRToRx),
                     MQT_NAMED_BUILDER(rx)},
         QCOTestCase{"CanonicalizeRToRy", MQT_NAMED_BUILDER(canonicalizeRToRy),
-                    MQT_NAMED_BUILDER(ry)}));
+                    MQT_NAMED_BUILDER(ry)},
+        QCOTestCase{"TwoR", MQT_NAMED_BUILDER(twoR), MQT_NAMED_BUILDER(r)}));
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai Declined. `valuesMatchWithinTolerance` is a thin wrapper around `valueToDouble + std::abs` vs `utils::TOLERANCE`; a dedicated near-equal fixture would mostly re-test that arithmetic. `TwoR` already covers the merge path with matching compile-time phi, and mismatch is covered by the beta case below.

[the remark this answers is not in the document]
simon1hofmann: @coderabbitai Won't add test for that now.

simon1hofmann:
@coderabbitai review

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -10,12 +10,95 @@
...
+ * @return Value The sum, as a constant or `arith.addf` result.
+ */
+static inline Value addFloatParameters(PatternRewriter& rewriter, Location loc,
+                                       Value lhs, Value rhs) {
+  const auto lhsVal = utils::valueToDouble(lhs);
+  const auto rhsVal = utils::valueToDouble(rhs);
+  if (lhsVal.has_value() && rhsVal.has_value()) {
+    return utils::constantFromScalar(rewriter, loc, *lhsVal + *rhsVal);
+  }
+  return arith::AddFOp::create(rewriter, loc, lhs, rhs).getResult();
```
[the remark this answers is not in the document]
simon1hofmann: @coderabbitai Addressed by removing `addFloatParameters` and restoring `main`’s unconditional `arith.addf` at merge sites. The comment targets code that no longer exists. Dominance behavior is unchanged from `main`; the tests use literal parameters only.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -166,45 +229,154 @@ LogicalResult mergeOneTargetOneParameter(OpType op, PatternRewriter& rewriter) {
...
  *
  * @tparam OpType The type of the operation to be merged.
  * @param op The operation instance.
  * @param rewriter The pattern rewriter.
  * @return LogicalResult Success or failure of the merge.
  */
 template <typename OpType>
-LogicalResult mergeTwoTargetOneParameter(OpType op, PatternRewriter& rewriter) {
-  // Check if the successor is the same operation
+LogicalResult mergeOneTargetTwoParameter(OpType op, PatternRewriter& rewriter) {
```
denialhaag: I'm wondering whether this function should get some other form of input. The implementation feels specific and general at the same time. Maybe something like `operand_index_whose_angels_should_be_within_tolerance` and `operand_index_whose_angels_are_added` (but way better names)? Right now, we are saying that `phi` needs to be within range, while `1` (which is `theta` for `ROp`) is added. For a (potentially imaginary) different operation, this might not hold up. 🤔

denialhaag: Maybe `primaryParameter` and `secondaryParameter` (like you write in the docstring) is a nice idea.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -10,12 +10,75 @@
...
+ * @details
+ * Returns failure when no partner exists or when the partner is directly
+ * adjacent on the same wire (zero `ctrl` hops).
+ *
+ * @tparam OpType The type of the operation to be merged.
+ * @param op The first operation instance.
+ * @return FailureOr<OpType> The partner operation, or failure.
+ */
+template <typename OpType>
+static FailureOr<OpType> findPartnerThroughCtrlControlChain(OpType op) {
```
denialhaag: Is there maybe a less verbose suffix than "ThroughCtrlControlChain" for all the new functions? Maybe "OnControlWire", or does that not convey all the aspects? 🤔

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -166,45 +229,154 @@ LogicalResult mergeOneTargetOneParameter(OpType op, PatternRewriter& rewriter) {
...
   }
 
-  // Confirm operations act on the same qubits
-  if (op.getOutputQubit(1) != nextOp.getInputQubit(1)) {
+  if (!valuesMatchWithinTolerance(op.getPhi(), nextOp.getPhi())) {
+    return failure();
+  }
+
+  auto newTheta = arith::AddFOp::create(rewriter, op.getLoc(), op.getOperand(1),
+                                        nextOp.getOperand(1));
```
denialhaag: Related to my point above, this should probably be called `newPrimaryParameter` or something like that.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -166,45 +229,154 @@ LogicalResult mergeOneTargetOneParameter(OpType op, PatternRewriter& rewriter) {
...
+  auto partner = findPartnerThroughCtrlControlChain(op);
+  if (failed(partner)) {
+    return failure();
+  }
+
+  rewriter.setInsertionPoint(op);
+  auto newTheta = arith::AddFOp::create(rewriter, op.getLoc(), op.getTheta(),
+                                        partner->getTheta());
+  rewriter.modifyOpInPlace(
+      op, [&] { op.getThetaMutable().assign(newTheta.getResult()); });
```
denialhaag: We should probably decide on whether we want to consistently use `op.getThetaMutable` or `op->setOperand`, at least for this file. 🤔

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -223,27 +395,93 @@ template <typename OpType>
...
-  op->setOperand(2, newParameter.getResult());
-
-  // nextOp results correspond to swapped operands, so swap replacements too
-  rewriter.replaceOp(nextOp, {op.getOutputQubit(1), op.getOutputQubit(0)});
+  if (!valuesMatchWithinTolerance(op.getBeta(), nextOp.getBeta())) {
+    return failure();
+  }
 
+  auto newTheta = arith::AddFOp::create(rewriter, op.getLoc(), op.getOperand(2),
+                                        nextOp.getOperand(2));
```
denialhaag: Also here, I don't _love_ that we are hardcoding `Beta` and `2` in a file as general as `QCOUtils.h`. We could also call the function something like `mergeXXPlusMinusYY`. I'll leave the final decision up to you!

mlir/include/mlir/Dialect/QCO/QCOUtils.h
denialhaag: Then again, if this pattern is only used for `ROp` at the moment, we might as well put it in `ROp.cpp`. I'll leave the final decision up to you, but I wouldn't put functions that are this specialized into a file that is as general as `QCOUtils.h` (especially while giving it a general name).

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -166,45 +229,154 @@ LogicalResult mergeOneTargetOneParameter(OpType op, PatternRewriter& rewriter) {
...
+ * @tparam OpType The type of the operation to be merged.
+ * @param op The first operation instance.
+ * @param rewriter The pattern rewriter.
+ * @param swappedTargets Whether the successor consumes swapped target wires.
+ * @return LogicalResult Success or failure of the merge.
+ */
+template <typename OpType>
+static LogicalResult mergeTwoTargetOneParameterImpl(OpType op,
+                                                    PatternRewriter& rewriter,
+                                                    bool swappedTargets) {
```
denialhaag: I like the idea of combining the two functions! There is an argument to be made that this could also be the only function; we could just let the `swappedTargets` default to `false` and specify it when necessary. 🤔 I'll leave this up to you.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -223,27 +395,93 @@ template <typename OpType>
...
+ * @tparam OpType The type of the operation to be merged.
+ * @param op The first operation instance.
+ * @param rewriter The pattern rewriter.
+ * @param swappedTargets Whether the successor consumes swapped target wires.
+ * @return LogicalResult Success or failure of the merge.
+ */
+template <typename OpType>
+static LogicalResult mergeTwoTargetTwoParameterImpl(OpType op,
+                                                    PatternRewriter& rewriter,
+                                                    bool swappedTargets) {
```
denialhaag: Same comment here; this _could_ also be the only function.

denialhaag (changes requested):
Thanks a lot for your work on this, @simon1hofmann! 🙂 This already looks really good to me. I just have some points on consistency ✨ below.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
simon1hofmann: Makes sense, also applied that concept to other functions.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
simon1hofmann: Good idea! Created a `mergeXXPlusMinusYY` function using a shared implementation with `mergeTwoTargetOneParameter`.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
simon1hofmann: Made it consistent using `op ->setOperand` everywhere.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
simon1hofmann: Did some refactoring and renamed it to `newParameter`.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
simon1hofmann: `OnControlWire` makes sense.

mlir/include/mlir/Dialect/QCO/QCOUtils.h
simon1hofmann: Good point, moved the implementation to `ROp.cpp`.

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

denialhaag (commented):
Thanks for addressing my comments! 🙂

Apart from the open CodeRabbit comment, this looks good to me now. I'll leave the final judgment to @burgholzer.

denialhaag (approved):

simon1hofmann:
@coderabbitai review

burgholzer (approved):
Thanks for another PR @simon1hofmann 🙏🏼 
I also edited this one a bit.
In particular, I further improved the existing canonicalization patterns. I really liked the direction here!

On the other hand, I reverted the changes for for the control-wire canonicalization for now. I am not convinced that this is the way to go here. I believe that a more promising way to go about this would be to introduce commutation rules that push diagonal gates through control qubits. I would argue this should not be part of a canonicalization pipeline though. Rather I'd put this into some kind of optimization pass similar to Qiskit's commutation-based optimizer. I would not put this on high priority though.

Since I really believe the other changes are a genuine improvement, I figured this is the quickest way to get this in.
I'll change the title and the PR description a little bit to reflect this.

PR #1799, opened 2026-06-20
✨ Add non-bit record output functions to the QIR runtime

### LINKED ISSUE #1797: ✨ Add QIR Adaptive Profile non-bit output recording to the QIR runtime
### Problem Statement

PR #1766 added QIR Adaptive Profile format routing to the DDSIM QDMI Device. The program-format gate accepts `QIRADAPTIVE{MODULE,STRING}` and `submitQIRProgram` runs them through the JIT. The runtime already handles `__quantum__rt__read_result + br i1`, so an Adaptive program that only branches on measurement results works end-to-end.

What is still missing is the set of recording functions Adaptive programs use to emit classical compute outputs alongside measurement results. The spec defines five additional `__quantum__rt__*_record_output` symbols (`bool`, `int`, `float`, `tuple`, `array`) that are not currently implemented in the runtime, so any Adaptive program that emits a classical bool, int, or float value, a tuple, or an array fails to link at JIT time.

QIR Adaptive Profile spec: https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md

**Blocks**: #1798

#### Scope

1. **Five new runtime recording functions** declared in `QIR.h` and implemented in `QIR.cpp`: `__quantum__rt__bool_record_output`, `__quantum__rt__int_record_output`, `__quantum__rt__float_record_output`, `__quantum__rt__tuple_record_output`, `__quantum__rt__array_record_output`.
2. **JIT symbol registration**: register the five new symbols in `Session::registerRuntimeSymbols` so JIT-compiled Adaptive programs can resolve them.
3. **Test circuit and tests**: a new test circuit exercises every new function in one run. Runtime-level and QDMI-level tests assert the exact emitted format.

#### Out of scope

- Switching the runtime to the spec-mandated `HEADER / START / END` + `<TYPE>\t<value>\t<label>` output schema. That work is tracked separately under "Add QIR Output Schemas support" and depends on this PR.

rturrado:
## Description

This PR completes the runtime side of QIR Adaptive Profile support. PR #1766 already added the format routing for `QIRADAPTIVE{MODULE,STRING}` and the bit-recording infrastructure. This follow-up fills the remaining gap: the five spec-defined `__quantum__rt__*_record_output` entry points for non-bit values (`bool`, `int`, `float`, `tuple`, `array`). Without them, any Adaptive program emitting classical compute outputs alongside measurement results would fail to link at JIT time.

The new entry points emit in MQT Core's current custom `label: value` format. Switching to the spec-mandated `HEADER / START / END` output schemas (Labeled / Ordered) is tracked separately as a follow-up issue.

Fixes #1797


burgholzer:
Hey @rturrado 👋🏻 
This looks pretty good already! Is this ready for a review?

rturrado:
> Hey @rturrado 👋🏻 
> This looks pretty good already! Is this ready for a review?

@burgholzer Hello! Yes, it is. And if you consider that the issue description needs refining and that that can affect the subsequent PR, no problem with that also.

burgholzer:
> > Hey @rturrado 👋🏻 
> > This looks pretty good already! Is this ready for a review?
> 
> @burgholzer Hello! Yes, it is. And if you consider that the issue description needs refining and that that can affect the subsequent PR, no problem with that also.
> 
> 

Nice! The issue descriptions read just fine as they are 🙂👍🏻
I'll try to get a review in here in the coming days.

rturrado:
> Nice! The issue descriptions read just fine as they are 🙂👍🏻
> I'll try to get a review in here in the coming days.

Thanks so much. No rush at all!

include/mqt-core/qir/runtime/QIR.h
```diff
@@ -167,6 +167,31 @@ bool __quantum__rt__read_result(Result*);
...
+/// included in the output or omitted.
+void __quantum__rt__int_record_output(int64_t, const char*);
+
+/// Adds a floating-point value to the generated output. The second parameter
+/// defines a string label for the value. Depending on the output schema, the
+/// label is included in the output or omitted.
+void __quantum__rt__float_record_output(double, const char*);
+
+/// Inserts a marker in the generated output indicating that the next
+/// `elementCount` recorded values form the contents of a tuple. The second
```
denialhaag: Could be nice to make use of [Doxygen's `\p` command](https://www.doxygen.nl/manual/commands.html#cmdp) here. 🤔 

```suggestion
/// \p elementCount recorded values form the contents of a tuple. The second
```

include/mqt-core/qir/runtime/QIR.h
```diff
@@ -167,6 +167,31 @@ bool __quantum__rt__read_result(Result*);
...
+/// parameter defines a string label for the tuple.
+void __quantum__rt__tuple_record_output(int64_t elementCount, const char*);
+
+/// Inserts a marker in the generated output indicating that the next `size`
```
denialhaag:
```suggestion
/// Inserts a marker in the generated output indicating that the next \p size
```

src/qir/runtime/QIR.cpp
```diff
@@ -381,10 +383,34 @@ bool __quantum__rt__read_result(Result* result) {
...
-  runtime.getOstream() << label << ": "
-                       << (__quantum__rt__read_result(result) ? 1 : 0) << "\n";
+  runtime.outputValue(label, bit ? "1" : "0");
+  // Accumulate new measurement bit.
+  runtime.appendMeasurementBit(bit);
+}
+
+void __quantum__rt__bool_record_output(bool value, const char* label) {
+  qir::Runtime::getInstance().outputValue(label, value ? "1" : "0");
+}
```
denialhaag, line 394: Given that we always have to swap the order of the arguments here, could it make to change the signature of `outputValue` to accept `value` first and `label` second? 🤔

test/qdmi/devices/dd/results_sampling_test.cpp
```diff
@@ -71,13 +72,30 @@ class HistogramTest : public ::testing::Test {
...
         keys, [](const auto& k) { return k == "00" || k == "11"; }));
   }
+
+  /// Smoke check: used for circuits whose distribution we don't pin down (e.g.
+  /// multi-output adaptive programs).
+  static void checkSmokeHistogram(const Histogram& hist) {
+    const auto& [keys, vals] = hist;
+    // Both vectors have the same size.
+    ASSERT_EQ(keys.size(), vals.size());
+    // Values sum up to the number of SHOTS.
```
denialhaag: Feels a bit more consistent with the renaming of the variable. There's also a few more instances of this comment that could be changed.

```suggestion
    // Values sum up to NUM_SHOTS.
```

denialhaag (changes requested):
Thanks a lot for your contribution, @rturrado! 🙂 

This looks really good to me! I only left a few minor comments, which you can find below.

include/mqt-core/qir/runtime/QIR.h
rturrado: Better!

src/qir/runtime/QIR.cpp
rturrado: Sure, more consistent!

test/qdmi/devices/dd/results_sampling_test.cpp
rturrado: Good catch!

denialhaag (approved):
This LGTM now! I'll leave the final approval to @burgholzer. 🙂

src/qir/runtime/QIR.cpp
```diff
@@ -381,10 +383,34 @@ bool __quantum__rt__read_result(Result* result) {
...
+  qir::Runtime::getInstance().outputValue(value ? "1" : "0", label);
+}
+
+void __quantum__rt__int_record_output(int64_t value, const char* label) {
+  qir::Runtime::getInstance().outputValue(std::to_string(value), label);
+}
+
+void __quantum__rt__float_record_output(double value, const char* label) {
+  std::ostringstream oss;
+  oss << value;
```
burgholzer, line 407: I'd like to avoid the stringstream include if possible (all stream/io in the STL is pretty costly to include). Why not use `std::to_string` here as well?

src/qir/runtime/Runtime.cpp
```diff
@@ -52,13 +53,14 @@ auto Runtime::reset() -> void {
...
   rRegister.clear();
+  measurements.clear();
   // NOLINTBEGIN(performance-no-int-to-ptr)
   rRegister.emplace(reinterpret_cast<Result*>(RESULT_ZERO_ADDRESS),
                     ResultStruct{.refcount = 0, .r = false});
   rRegister.emplace(reinterpret_cast<Result*>(RESULT_ONE_ADDRESS),
                     ResultStruct{.refcount = 0, .r = true});
   // NOLINTEND(performance-no-int-to-ptr)
-  recordedOutputs.clear();
+  measurements.clear();
```
burgholzer, line 62: This is already reset above, right?
```suggestion
```

src/qir/runtime/Runtime.cpp
```diff
@@ -164,12 +166,12 @@ auto Runtime::equal(Result* result1, Result* result2) -> bool {
   return deref(result1).r == deref(result2).r;
 }
 
-auto Runtime::recordOutput(Result* result) -> void {
-  recordedOutputs.push_back(deref(result).r ? '1' : '0');
+auto Runtime::appendMeasurementBit(bool result) -> void {
+  measurements.push_back(result ? '1' : '0');
```
burgholzer, line 169:
```suggestion
  measurements.emplace_back(result ? '1' : '0');
```

src/qir/runtime/Runtime.cpp
```diff
@@ -178,7 +180,17 @@ auto Runtime::takeState() -> QState {
   return ret;
 }
 
-auto Runtime::getOstream() -> std::ostream& { return *os; }
+auto Runtime::outputContainer(int64_t /* elementCount */,
+                              const char* label) const -> void {
+  *os << (label != nullptr ? label : "") << ":\n";
+}
```
burgholzer, line 185: I do not quite know what it is, but something here feels wrong to me. Why does the element count have no effect/use here? Shouldn't this change something about the output?

src/qir/runtime/Runtime.cpp
```diff
@@ -178,7 +180,17 @@ auto Runtime::takeState() -> QState {
...
+
+auto Runtime::outputValue(std::string_view valueStr, const char* label) const
+    -> void {
+  *os << (label != nullptr ? label : "") << ": " << valueStr << "\n";
+}
```
burgholzer, line 190: Is the default of an empty label really what the spec implies here? To me, this seems like it would print a broken output recording if the label were empty as it would just print the colon.
Should there be a default/dummy label?

test/circuits/AdaptiveRecordOutputs.ll
```diff
@@ -0,0 +1,102 @@
...
+
+declare void @__quantum__rt__result_update_reference_count(%Result*, i32)
+
+attributes #0 = { "entry_point" "output_labeling_schema" "qir_profiles"="custom" "required_num_qubits"="3" "required_num_results"="3" }
+attributes #1 = { "irreversible" }
+
+!llvm.module.flags = !{!0, !1, !2, !3}
+
+!0 = !{i32 1, !"qir_major_version", i32 1}
+!1 = !{i32 7, !"qir_minor_version", i32 0}
```
burgholzer:
```suggestion
!0 = !{i32 1, !"qir_major_version", i32 2}
!1 = !{i32 7, !"qir_minor_version", i32 1}
```
Should probably reflect the latest spec version

burgholzer (changes requested):
This is looking great! Thanks for another nice contribution @rturrado 🙌
I have a couple of nitpick comments and maybe some stupid questions. Then this can go in 🚀

src/qir/runtime/QIR.cpp
rturrado: Fair enough. We can use `std::to_string` for the time being, for keeping it simple and clear and consistent. In case more control is needed later on, another alternative can be used.

I think that the main differences between both options are:

1. `std::to_string` will always print 6 digits after the decimal, e.g., 0.543000 while `std::ostringstream` (in the default mode) would print up to 6 significant digits, e.g., 1.23456.
2. `std::to_string` will print 0.000000 for very small numbers while `std::ostringstream` would use scientific notation, e.g., 1.23457e+08.

See the _Notes_ and _Example_ sections in [std::to_string](https://en.cppreference.com/cpp/string/basic_string/to_string) for more details. In that same page they also talk about `std::to_chars` as a more performant version (since C++17). It's also much more verbose though: char array buffer definition, call, if/else block for error checking and conversion. See [std::to_chars](https://en.cppreference.com/cpp/utility/to_chars).

src/qir/runtime/Runtime.cpp
rturrado: Good catch, indeed!

src/qir/runtime/Runtime.cpp
rturrado: OK. I understand this is a solid general rule: prefer `emplace_back` to `push_back`. Although there's no real gain for this particular case.

test/circuits/AdaptiveRecordOutputs.ll
rturrado: Another good catch. Should we apply it also to the other circuit files as part of this PR? BellPairStatic.ll, BellPairDynamic.ll, BellPairAdaptive.ll, and GHZ4Dynamic.ll.

src/qir/runtime/Runtime.cpp
rturrado: Oh, wait. `measurements` is a just a `std::string` (of 0s and 1s), not a `std::vector`.

burgholzer: Ah, my bad in that case. Spoke too soon. Let's resolve this.

test/circuits/AdaptiveRecordOutputs.ll
burgholzer: Yeah. Good to be consistent.

src/qir/runtime/Runtime.cpp
rturrado: You're very right to point this here, the same as your next comment.

Both points are addressed by the follow-up Output Schemas PR, which introduces the spec-compliant Ordered and Labeled formats (`OUTPUT\t<TYPE>\t<value>[\t<label>]`), including element counts for tuple and array.

`outputValue` and `outputContainer` are a small refactor: they hold the print shared by the six `*_record_output` functions.

src/qir/runtime/Runtime.cpp
rturrado: Following above's answer:

`__quantum__rt__result_record_output`'s output is unchanged from `main`. Upstream prints:

```cpp
runtime.getOstream() << label << ": "
                     << (__quantum__rt__read_result(result) ? 1 : 0) << "\n";
```

This PR produces the same `label: value\n`, with the same behavior when the label is null or empty.

The spec-compliant emitter (with element counts and proper framing) lands in the follow-up PR.

src/qir/runtime/QIR.cpp
burgholzer: Ah. Precision is a valid point. Maybe that alone justifies the use of the stringstream.

rturrado:
Force-pushed to add DCO sign-off to previously unsigned commits. No code change.

rturrado:
> This is looking great! Thanks for another nice contribution @rturrado 🙌 I have a couple of nitpick comments and maybe some stupid questions. Then this can go in 🚀

@burgholzer Many thanks! Very much OK with nitpicking. And no stupid questions at all.

src/qir/runtime/Runtime.cpp
burgholzer: That makes a lot of sense. If this is tackled in a follow-up, let's resolve the conversation here. 👍🏻

src/qir/runtime/Runtime.cpp
burgholzer: Looking forward to the follow-up. Let's resolve here

rturrado:
@burgholzer @denialhaag Hi guys! Can you check if there was anything else to address in this PR? Thanks!

src/qir/runtime/Runtime.cpp
rturrado: I left that 2nd `measurements.clear()` but removed the 1st appearance, a bit earlier. That way we follow the order of definition of  fields, i.e., we completely deal with `rRegister` before dealing with `measurements`.

burgholzer (approved):
Looks all good to me 🙌 let's get this in 🚀

PR #1802, opened 2026-06-22
✨ Enhance Matrix functionality with new methods and optimizations

simon1hofmann:
## Description

Extends `mlir::qco` matrix utilities with fixed-size helpers, n-qubit embedding/reordering, and a real symmetric 4×4 eigensolver. These are building blocks for native two-qubit gate synthesis (Weyl decomposition and related passes).

### New APIs
- **`symmetricEigen4`** — eigendecomposition of a real symmetric 4×4 matrix via EISPACK `tred2`/`tql2` (adapted from John Burkardt’s MIT-licensed C port); includes a `Matrix4x4` overload that uses `realPart()`
- **`kron`** — Kronecker product for single-qubit factors (MSB-first wire order)
- **`reorderTwoQubitMatrix`** — conjugate a 4×4 gate to act on wires `{0,1}`
- **`embedSingleQubitInNqubit` / `embedTwoQubitInNqubit`** — embed fixed gates into larger Hilbert spaces

### `Matrix4x4` / `DynamicMatrix` improvements
- `transpose`, `diagonal`, `fromDiagonal`, `column`, `setColumn`, `realPart`, `imagPart`, `isIdentity`
- `DynamicMatrix::operator*=` and scalar `operator*` overloads for 2×2 / 4×4 / dynamic matrices
- Fast paths for 2×2 and 4×4 `DynamicMatrix` multiply; single-pass `isIdentity`
- Shared internal helpers (`multiply2x2`/`multiply4x4`, `isIdentityEntries`)

## AI Assistance

## Checklist


mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -334,6 +353,53 @@ struct Matrix4x4 {
...
+   * @param tol Maximum allowed complex modulus of each entry difference.
+   * @return True if every entry is within @p tol of the identity.
+   */
+  [[nodiscard]] bool isIdentity(double tol = MATRIX_TOLERANCE) const;
+
+  /**
+   * @brief Returns the four diagonal entries `(m00, m11, m22, m33)`.
+   * @return Array of diagonal entries.
+   */
+  [[nodiscard]] std::array<Complex, K_ROWS> diagonal() const;
```
burgholzer, line 396: Shouldn't this return an ArrayRef? Or is it intended that the ownership of the returned values lives with the caller?

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -334,6 +353,53 @@ struct Matrix4x4 {
...
+
+  /**
+   * @brief Builds a diagonal matrix from four diagonal entries.
+   * @param diagonalEntries Diagonal entries `(m00, m11, m22, m33)`.
+   * @return Diagonal matrix with the given entries.
+   */
+  [[nodiscard]] static Matrix4x4
+  fromDiagonal(const std::array<Complex, K_ROWS>& diagonalEntries);
```
burgholzer: The argument here should likely be an ArrayRef

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -334,6 +353,53 @@ struct Matrix4x4 {
...
+   * @return Array of the four column entries.
+   */
+  [[nodiscard]] std::array<Complex, K_ROWS> column(std::size_t col) const;
+
+  /**
+   * @brief Overwrites column @p col with @p values.
+   * @param col Column index in `[0, K_COLS)`.
+   * @param values New column entries, top to bottom.
+   */
+  void setColumn(std::size_t col, const std::array<Complex, K_ROWS>& values);
```
burgholzer: Also here, `ArrayRef` should likely be used.

I suppose there is a reason why there's only a `column` but no `row` method?

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -334,6 +353,53 @@ struct Matrix4x4 {
...
+   * @brief Returns the element-wise real parts in row-major order.
+   * @return Real parts of all entries.
+   */
+  [[nodiscard]] std::array<double, K_SIZE_AT_COMPILE_TIME> realPart() const;
+
+  /**
+   * @brief Returns the element-wise imaginary parts in row-major order.
+   * @return Imaginary parts of all entries.
+   */
+  [[nodiscard]] std::array<double, K_SIZE_AT_COMPILE_TIME> imagPart() const;
```
burgholzer, line 458: I suppose these could be ArrayRef's as well, but I am not 100% sure here.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -558,4 +659,99 @@ inline constexpr bool
...
+ * @brief Kronecker product `lhs (x) rhs` of two single-qubit matrices.
+ *
+ * Uses the computational-basis bit order where the first operand labels the
+ * high bit, matching `UnitaryOpInterface::getUnitaryMatrix4x4`.
+ *
+ * @param lhs Left factor (acts on the high bit / qubit 0).
+ * @param rhs Right factor (acts on the low bit / qubit 1).
+ * @return The `4x4` Kronecker product.
+ */
+[[nodiscard]] Matrix4x4 kron(const Matrix2x2& lhs, const Matrix2x2& rhs);
```
burgholzer: Should this potentially be a static method of the `Matrix4x4` class?

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -558,4 +659,99 @@ inline constexpr bool
...
+ * @brief Embed a two-qubit matrix into an @p numQubits-qubit Hilbert space.
+ *
+ * Operand 0 labels the high bit of the pair and acts on @p q0Index; operand 1
+ * labels the low bit and acts on @p q1Index. For each basis pair whose other
+ * wires match, copies @p matrix at the packed two-qubit row/column indices.
+ */
+[[nodiscard]] DynamicMatrix embedTwoQubitInNqubit(const Matrix4x4& matrix,
+                                                  std::size_t numQubits,
+                                                  std::size_t q0Index,
+                                                  std::size_t q1Index);
```
burgholzer: I suppose these should be member functions of the 2x2 and 4x4 matrix classes.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
```diff
@@ -558,4 +659,99 @@ inline constexpr bool
...
+[[nodiscard]] SymmetricEigen4 symmetricEigen4(const Matrix4x4& symmetric);
+
+/**
+ * @brief Reorder a two-qubit matrix to act on qubits `{0, 1}`.
+ *
+ * @param q0Index Wire index of operand 0; @p q1Index wire index of operand 1.
+ */
+[[nodiscard]] Matrix4x4 reorderTwoQubitMatrix(const Matrix4x4& matrix,
+                                              std::size_t q0Index,
+                                              std::size_t q1Index);
```
burgholzer: Also feels like these need not be free functions but could be added to the 4x4 matrix class.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
```diff
@@ -81,21 +82,32 @@ assignFromDynamicImpl(const DynamicMatrix& src,
...
-/// Writes the row-major product `lhs * rhs` into @p out (4x4, unrolled rows).
 static void
-multiply4x4(const std::array<Complex, Matrix4x4::K_SIZE_AT_COMPILE_TIME>& lhs,
-            const std::array<Complex, Matrix4x4::K_SIZE_AT_COMPILE_TIME>& rhs,
-            std::array<Complex, Matrix4x4::K_SIZE_AT_COMPILE_TIME>& out) {
+multiply2x2(const std::array<Complex, Matrix2x2::K_SIZE_AT_COMPILE_TIME>& lhs,
+            const std::array<Complex, Matrix2x2::K_SIZE_AT_COMPILE_TIME>& rhs,
+            std::array<Complex, Matrix2x2::K_SIZE_AT_COMPILE_TIME>& out) {
+  multiply2x2(ArrayRef(lhs), ArrayRef(rhs), MutableArrayRef(out));
+}
```
burgholzer: Is this function really necessary? Is there no implicit conversion to ArrayRef's?

Same holds below.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
```diff
@@ -453,4 +553,493 @@ bool DynamicMatrix::isApprox(const DynamicMatrix& other,
...
+
+  std::array<double, 16> z{};
+  std::array<double, 4> diag{};
+  std::array<double, 4> subdiag{};
+  symmetricTred24(symmetric, z, diag, subdiag);
+  symmetricTql24(diag, subdiag, z);
+
+  SymmetricEigen4 result;
+  for (std::size_t col = 0; col < n; ++col) {
+    result.eigenvalues[col] = diag[col];
```
burgholzer: This logic could be simplified right? This is just `result.eigenvalues = diag`.
At that point, one can actually allocate the result structure up front and directly pass `result.eigenvalues` instead of the `diag` array.
I suppose with some trickery, something similar could also be possible for avoiding the temporary `z`. But it's maybe too much effort to go there.
One can feel a little that this is coming from a pure C implementation.

burgholzer (changes requested):
Had some time, so I thought I'd give this a first review already 😌

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
simon1hofmann: Not sure if that works here, as this function returns `{data[0], data[5], data[10], data[15]}`, which is not a slice of `data` and not contiguous, so we have to copy them into a `std::array`right? Same for the `column()` method.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
simon1hofmann: Here it works, same for `setColumn`.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
simon1hofmann: For `column()` I think it's the same reasoning as for `diagonal()`. For `column()` it's possible. 
Weyl decomposition only needs the `column()` and `setColumn()` methods, but does not hurt to also add `row()` and `setRow()`, as they might be usable elsewhere.

For `row()`, `ArrayRef` should be possible as we're slicing through `data` 🤔

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
simon1hofmann: I think it's also not possible here as we have to walk through `data` to collect the real and imaginary parts into an `std::array`.

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
simon1hofmann: Reasonable either way, moved it to `Matrix4x4`.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
simon1hofmann: You're right, they're redundant and can be removed 👍

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
simon1hofmann: Dropped the separate diag buffer and wrote eigenvalues directly into `result.eigenvalues`, avoiding the temporary `z` is a little bit more difficult and probably not worth the effort (for now).

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
> Had some time, so I thought I'd give this a first review already 😌

Thanks a lot for the quick review 🙏 
I went through your comments and also applied `CodeRabbit`'s suggestions.
Would be ready for another review now :)

mlir/include/mlir/Dialect/QCO/Utils/Matrix.h
burgholzer: Yeah. Just read up on this; ArrayRef requires contiguous storage.
I think it's safe to say that all my suggestions in that regard can be disregarded.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
```diff
@@ -317,6 +431,79 @@ Complex Matrix4x4::determinant() const {
...
+  return {data[0], data[5], data[10], data[15]};
+}
+
+Matrix4x4 Matrix4x4::fromDiagonal(const ArrayRef<Complex> diagonalEntries) {
+  assert(diagonalEntries.size() == K_ROWS &&
+         "fromDiagonal requires exactly K_ROWS entries");
+  Matrix4x4 out{};
+  for (std::size_t i = 0; i < K_ROWS; ++i) {
+    out.data[(i * K_COLS) + i] = diagonalEntries[i];
+  }
```
burgholzer: I suppose this is just me being picky, but the rest of the code often directly spells out the computations that need to be performed. 
To be consistent, one would probably also unroll this loop and simply write out the four assignments.

Might apply to a couple more places.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
```diff
@@ -325,6 +512,310 @@ bool Matrix4x4::assignFrom(const DynamicMatrix& src) {
...
+
+  SymmetricEigen4 result;
+  std::array<double, 16> z{};
+  std::array<double, 4> subdiag{};
+  symmetricTred24(symmetric, z, result.eigenvalues, subdiag);
+  symmetricTql24(result.eigenvalues, subdiag, z);
+
+  for (std::size_t col = 0; col < n; ++col) {
+    for (std::size_t row = 0; row < n; ++row) {
+      result.eigenvectors(row, col) = Complex{z[row + (col * n)], 0.0};
```
burgholzer:
```suggestion
      result.eigenvectors(row, col) = z[row + (col * n);
```
I would have guessed this world as well and is less verbose.

burgholzer (approved):
LGTM. Just two minor comments/suggestions left. Feel free to address and resolve these and merge this afterwards.

One overall comment on the whole PR iteration process: it would be good if you wouldn't resolve review comments that  you replied to as part of incorporating feedback. People using the GitHub App/Notifications (like me) do not see these comments without going through all resolved comments, opening them up, and checking whether there have been any replies. Most of the time, it is best for the person who wrote the last comment in a discussion thread to leave it to someone else to actually resolve the conversation. Makes the whole process a little smoother. When I get a request for a re-review, I will typically first go through all open conversations of the previous review before going through the actual code.
I assume you are resolving the comments to indicate that you have addressed the comments and to keep a clean slate. Hopefully the above is only a slight adaptation to your workflow.
Let me know if any of that does not make sense.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
burgholzer: Seems like I was wrong 😑

simon1hofmann:
> LGTM. Just two minor comments/suggestions left. Feel free to address and resolve these and merge this afterwards.
> 
> One overall comment on the whole PR iteration process: it would be good if you wouldn't resolve review comments that you replied to as part of incorporating feedback. People using the GitHub App/Notifications (like me) do not see these comments without going through all resolved comments, opening them up, and checking whether there have been any replies. Most of the time, it is best for the person who wrote the last comment in a discussion thread to leave it to someone else to actually resolve the conversation. Makes the whole process a little smoother. When I get a request for a re-review, I will typically first go through all open conversations of the previous review before going through the actual code. I assume you are resolving the comments to indicate that you have addressed the comments and to keep a clean slate. Hopefully the above is only a slight adaptation to your workflow. Let me know if any of that does not make sense.

I usually resolve comments where I do no think a re-review is necessary, but makes total sense to let the other person resolve that comment. Will keep that in mind in the following PR's!

I addressed the remaining comments, should be good to go in after all tests are green.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
simon1hofmann: Should work, but applied the suggestion too quickly, was missing a `]` at the end.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
simon1hofmann: Added three more explicit unrolls.

PR #1803, opened 2026-06-22
✨ Add two-qubit Weyl (KAK) decomposition library

simon1hofmann:
## Description

Adds a two-qubit Weyl (KAK) decomposition on top of `Matrix4x4`. A 4×4 unitary is factored as `(K1l ⊗ K1r) · U_canon(a,b,c) · (K2l ⊗ K2r)` up to global phase, with optional canonical-gate specialization and basis-gate (CX/CZ) synthesis.
This PR is intentionally scoped to the **decomposition library + tests**. It does **not** wire an MLIR pass, emit gate sequences, or add `NativeProfile` / fuse-pass integration — those are planned follow-ups.

### New API (`mlir/Dialect/QCO/Transforms/Decomposition/Weyl.h`)
- **`TwoQubitWeylDecomposition`** — KAK decomposition with Weyl-chamber reduction and 10 canonical-gate specializations (identity, SWAP, partial-SWAP ±, controlled, mirror-controlled, fSim variants)
- **`TwoQubitBasisDecomposer`** — decomposes a Weyl result into single-qubit factors + basis-gate count (0–3 uses)

### Supporting changes
- **`Euler.h` / `Euler.cpp`** — promote `EulerAngles` and `anglesFromUnitary()` to public API (used by Weyl specialization for controlled / fSim cases)

### AI Assistance

## Checklist


simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
burgholzer, line 1: I don't really like that we are duplicating the matrix definitions here and in the individual operation files.

How about moving these functions to the individual operation files as static methods that are then used in the individual operation's implementation of the getUnitaryMatrix method?
That gives them a clear home and makes results in a single place of definition.

mlir/include/mlir/Dialect/QCO/Transforms/Decomposition/Weyl.h
```diff
@@ -0,0 +1,214 @@
...
+
+#include <array>
+#include <complex>
+#include <cstdint>
+#include <optional>
+
+namespace mlir::qco::decomposition {
+
+/** Numeric tolerance for Weyl internal matrix checks. */
+inline constexpr double WEYL_TOLERANCE = 100 * MATRIX_TOLERANCE;
```
burgholzer: Is there a reason for being less precise here?

mlir/include/mlir/Dialect/QCO/Transforms/Decomposition/Weyl.h
```diff
@@ -0,0 +1,214 @@
...
+    Matrix2x2 u3r;
+    Matrix2x2 q0l;
+    Matrix2x2 q0r;
+    Matrix2x2 q1la;
+    Matrix2x2 q1lb;
+    Matrix2x2 q1ra;
+    Matrix2x2 q1rb;
+    Matrix2x2 q2l;
+    Matrix2x2 q2r;
+  };
```
burgholzer, line 308: Would it make sense to add a docstring to this explaining the structure and how these matrices are combined together?
Like a Qiskit circuit diagram with boxes.

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+#include <limits>
+#include <numbers>
+#include <optional>
+#include <utility>
+
+namespace mlir::qco::decomposition {
+
+using namespace std::complex_literals;
+
+static constexpr double DEFAULT_WEYL_FIDELITY = 1.0 - 1e-12;
```
burgholzer: Should this be tied to some of the tolerance values instead of a magic number?

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+static constexpr double INV_SQRT2 = 1.0 / std::numbers::sqrt2;
+
+static const Matrix2x2 K12_R_ARR = Matrix2x2::fromElements(
+    1i * INV_SQRT2, INV_SQRT2, -INV_SQRT2, -1i * INV_SQRT2);
+static const Matrix2x2 K12_L_ARR =
+    Matrix2x2::fromElements(Complex{0.5, 0.5}, Complex{0.5, 0.5},
+                            Complex{-0.5, 0.5}, Complex{0.5, -0.5});
+static const Matrix2x2 K22_L =
+    Matrix2x2::fromElements(INV_SQRT2, -INV_SQRT2, INV_SQRT2, INV_SQRT2);
+static const Matrix2x2 K22_R = Matrix2x2::fromElements(0, 1, -1, 0);
```
burgholzer: Truly minor and I don't know if it makes a difference (more my curiosity): could these be made constexpr as well?

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+  const auto absRhs = std::abs(rhs);
+  if (absRhs > absLhs) {
+    return absDiff <= absRhs * maxRelative;
+  }
+  return absDiff <= absLhs * maxRelative;
+}
+
+static double traceToFidelity(const Complex& trace) {
+  const auto traceAbs = std::abs(trace);
+  return (4.0 + (traceAbs * traceAbs)) / 20.0;
```
burgholzer: This computation feels entirely arbitrary 😅 where is it coming from and/pr what is this doing?

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+        "TwoQubitBasisDecomposer: basisFidelity must be finite and in [0, 1] "
+        "(got {0})",
+        basisFidelity));
+  }
+
+  const auto basisWeyl =
+      TwoQubitWeylDecomposition::create(basisMatrix, DEFAULT_WEYL_FIDELITY);
+  const auto isSuperControlled =
+      relativeEq(basisWeyl.a(), PI_OVER_4, 1e-13, 1e-09) &&
+      relativeEq(basisWeyl.c(), 0.0, 1e-13, 1e-09);
```
burgholzer: Magic numbers should probably be replaced by solid defaults in the respective methods (and be inferred from other constant values)

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+      temp * (-1i * std::exp(-2i * b)), temp * std::exp(-2i * b),
+      temp * (1i * std::exp(2i * b)), temp * std::exp(2i * b));
+  const Matrix2x2 k31l = Matrix2x2::fromElements(
+      INV_SQRT2 * std::exp(-1i * b), INV_SQRT2 * std::exp(-1i * b),
+      INV_SQRT2 * -std::exp(1i * b), INV_SQRT2 * std::exp(1i * b));
+  const Matrix2x2 k31r = Matrix2x2::fromElements(1i * std::exp(1i * b), 0, 0,
+                                                 -1i * std::exp(-1i * b));
+  const Matrix2x2 k32r = Matrix2x2::fromElements(
+      temp * std::exp(1i * b), temp * -std::exp(-1i * b),
+      temp * (-1i * std::exp(1i * b)), temp * (-1i * std::exp(-1i * b)));
```
burgholzer: Not sure if the compiler is clever enough to only compute the repeated std::exp only once.
Might be worth to extract into a temporary variable for all the computations that are repeated here.

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+      .u3l = k2lDagger * K12_L_ARR,
+      .u3r = k2rDagger * K12_R_ARR,
+      .q0l = K12_L_ARR.adjoint() * k1lDagger,
+      .q0r = K12_R_ARR.adjoint() * iPauliZ() * k1rDagger,
+      .q1la = k2lDagger * k11l.adjoint(),
+      .q1lb = k11l * k1lDagger,
+      .q1ra = k2rDagger * iPauliZ() * k11r.adjoint(),
+      .q1rb = k11r * k1rDagger,
+      .q2l = k2lDagger * K12_L_ARR,
+      .q2r = k2rDagger * K12_R_ARR,
```
burgholzer: Some matrices are computed more than once here. This feels wasteful.
The compiler might be clever enough to eliminate the redundant computations, but we don't know that without checking.

To throw the more general question in the room: if some of those matrices are the same as part of the decomposition, are some of the members redundant and could be eliminated? (Likely that I am missing something)

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+      TwoQubitWeylDecomposition::create(targetUnitary, DEFAULT_WEYL_FIDELITY);
+  return twoQubitDecompose(targetWeyl, numBasisGateUses);
+}
+
+std::optional<TwoQubitNativeDecomposition>
+TwoQubitBasisDecomposer::twoQubitDecompose(
+    const TwoQubitWeylDecomposition& targetDecomposition,
+    std::optional<std::uint8_t> numBasisGateUses) const {
+  const auto traceValues = traces(targetDecomposition);
+  auto getDefaultNbasis = [&]() -> std::uint8_t {
```
burgholzer: Feels a bit weird to define this lambda which is then only used once (as far as I can tell). May as well avoid that overhead.

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,312 @@
...
+  }
+  return SmallVector<Matrix2x2, 8>{
+      smb.q2r * target.k2r(),
+      smb.q2l * target.k2l(),
+      smb.q1ra * rzMatrix(2. * target.b()) * smb.q1rb,
+      smb.q1la * rzMatrix(-2. * target.a()) * smb.q1lb,
+      target.k1r() * smb.q0r,
+      target.k1l() * smb.q0l,
+  };
+}
```
burgholzer: The 8 somewhat feels like a waste here given how we know that the decomposition will be smaller in all of these cases.
Can this be handled a little more dynamically/efficiently?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+  Matrix2x2 k2r;
+  double globalPhase{};
+  double a{};
+  double b{};
+  double c{};
+};
+
+} // namespace
+
+static constexpr double DIAGONALIZATION_PRECISION = 1e-13;
```
burgholzer: So we need yet another precision value?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+static constexpr double PI = std::numbers::pi;
+static constexpr double PI_OVER_4 = PI / 4.0;
+
+static double remEuclid(const double a, const double b) {
+  if (b == 0.0) {
+    llvm::reportFatalInternalError("remEuclid expects non-zero divisor");
+  }
+  const auto r = std::fmod(a, b);
+  return (r < 0.0) ? r + std::abs(b) : r;
+}
```
burgholzer: Feels like I saw this somewhere else as well. Unify?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+/** Maps `|tr(U^dag V)|` to average two-qubit gate fidelity. */
+static double traceToFidelity(const Complex& trace) {
+  const auto traceAbs = std::abs(trace);
+  return (4.0 + (traceAbs * traceAbs)) / 20.0;
+}
```
burgholzer: Also saw this somewhere before an had a comment on it.

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    1, 1i, 0, 0,                                                             //
+    0, 0, 1i, 1,                                                             //
+    0, 0, 1i, -1,                                                            //
+    1, -1i, 0, 0);
+static const Matrix4x4 MAGIC_BASIS_NON_NORMALIZED_DAGGER =
+    Matrix4x4::fromElements(                          //
+        0.5, 0, 0, 0.5,                               //
+        Complex{0.0, -0.5}, 0, 0, Complex{0.0, 0.5},  //
+        0, Complex{0.0, -0.5}, Complex{0.0, -0.5}, 0, //
+        0, 0.5, -0.5, 0);
```
burgholzer: Related to the other comment: should/can this be constexpr?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+           MAGIC_BASIS_NON_NORMALIZED;
+  }
+  return MAGIC_BASIS_NON_NORMALIZED * unitary *
+         MAGIC_BASIS_NON_NORMALIZED_DAGGER;
+}
+
+static double closestPartialSwap(double a, double b, double c) {
+  const auto m = (a + b + c) / 3.;
+  const auto [am, bm, cm] = std::array{a - m, b - m, c - m};
+  const auto [ab, bc, ca] = std::array{a - b, b - c, c - a};
```
burgholzer: Strange constellation; why the array here if it is immediately dissolved?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+static std::pair<Matrix4x4, std::array<Complex, 4>>
+diagonalizeComplexSymmetric(const Matrix4x4& m, double precision) {
+  auto state = std::mt19937{2023};
+  std::normal_distribution<double> dist;
+
+  const auto mReal = m.realPart();
+  const auto mImag = m.imagPart();
+
+  double bestErr = 1e300;
```
burgholzer: Use numeric limits?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+static double closestPartialSwap(double a, double b, double c) {
+  const auto m = (a + b + c) / 3.;
+  const auto [am, bm, cm] = std::array{a - m, b - m, c - m};
+  const auto [ab, bc, ca] = std::array{a - b, b - c, c - a};
+  return m + (am * bm * cm * (6. + (ab * ab) + (bc * bc) + (ca * ca)) / 18.);
+}
+
+static std::pair<Matrix4x4, std::array<Complex, 4>>
+diagonalizeComplexSymmetric(const Matrix4x4& m, double precision) {
+  auto state = std::mt19937{2023};
```
burgholzer, line 125: Just thinking out loud here: as far as ai understand, the fixed seed is to ensure determinism.
However, will this actually guarantee that all implementations/OSes will produce the same sequence of random values here? That is, is this truly reproducible?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    const auto compare = p * Matrix4x4::fromDiagonal(d) * p.transpose();
+    {
+      double err = 0.0;
+      for (std::size_t r = 0; r < 4; ++r) {
+        for (std::size_t cc = 0; cc < 4; ++cc) {
+          err = std::max(err, std::abs(compare(r, cc) - m(r, cc)));
+        }
+      }
+      bestErr = std::min(bestErr, err);
+    }
```
burgholzer: Redundant braces (or I am missing something)

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    if (compare.isApprox(m, precision)) {
+      assert((p.transpose() * p).isIdentity(WEYL_TOLERANCE));
+      assert(std::abs(Matrix4x4::fromDiagonal(d).determinant() - 1.0) <
+             WEYL_TOLERANCE);
+      return std::make_pair(p, d);
```
burgholzer:
```suggestion
      return {p, d};
```
More convenient syntax

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    if (requestedFidelity) {
+      return traceToFidelity(tr) >= *requestedFidelity;
+    }
+    return false;
+  };
+
+  const auto closestAbc = closestPartialSwap(
+      decomposition.a(), decomposition.b(), decomposition.c());
+  const auto closestAbMinusC = closestPartialSwap(
+      decomposition.a(), decomposition.b(), -decomposition.c());
```
burgholzer: Wasteful to compile these upfront before they are needed in the if chain below. Rather interleave them at the latest possible point.

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+  u *= std::pow(detU, -0.25);
+  return {u, std::arg(detU) / 4.0};
+}
+
+static std::tuple<Matrix4x4, Matrix4x4, std::array<double, 3>,
+                  std::array<double, 4>>
+computeOrderedWeylCoordinates(const Matrix4x4& u) {
+  const auto uP = magicBasisTransform(u, /*outOfMagicBasis=*/true);
+  const Matrix4x4 m2 = uP.transpose() * uP;
+  auto [p, d] = diagonalizeComplexSymmetric(m2, DIAGONALIZATION_PRECISION);
```
burgholzer: Shouldn't the precision be the default argument?

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    k1l = k1l * iPauliZ();
+    k2r = iPauliZ() * k2r;
+    globalPhase += (PI / 2.0);
+  }
+  if (cs[2] > PI_OVER_4) {
+    cs[2] -= (PI / 2.0);
+    k1l = k1l * iPauliZ();
+    k1r = k1r * iPauliZ();
+    globalPhase -= (PI / 2.0);
+  }
```
burgholzer: Potentially worth avoiding the repeated Pauli computations? And only computing them once.
Again something I am not sure the compiler does by default.

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    double preSpecializationB, double preSpecializationC,
+    const std::optional<double>& fidelity) {
+  const auto trace =
+      flippedFromOriginal
+          ? getTrace((PI / 2.0) - preSpecializationA, preSpecializationB,
+                     -preSpecializationC, a_, b_, c_)
+          : getTrace(preSpecializationA, preSpecializationB, preSpecializationC,
+                     a_, b_, c_);
+  const double calculatedFidelity = traceToFidelity(trace);
+  if (fidelity && calculatedFidelity + 1.0e-13 < *fidelity) {
```
burgholzer: Random magic number here

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+  const auto reconstructed =
+      Matrix4x4::kron(decomposition.k1l_, decomposition.k1r_) *
+      decomposition.getCanonicalMatrix() *
+      Matrix4x4::kron(decomposition.k2l_, decomposition.k2r_) *
+      std::exp(Complex{0.0, 1.0} * decomposition.globalPhase_);
+  if (!reconstructed.isApprox(unitaryMatrix, WEYL_TOLERANCE)) {
+    llvm::reportFatalInternalError(
+        "TwoQubitWeylDecomposition: failed to reconstruct unitary after "
+        "specialization");
+  }
```
burgholzer: Feels like this could be an assertion as well. Feels kind of costly for running as a regular check.

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+  return decomposition;
+}
+
+Matrix4x4 TwoQubitWeylDecomposition::getCanonicalMatrix(double a, double b,
+                                                        double c) {
+  return rzzMatrix(-2.0 * c) * ryyMatrix(-2.0 * b) * rxxMatrix(-2.0 * a);
```
burgholzer: Hypothetically speaking, one could write this computation out explicitly without requiring matrix computation, right?
Don't know if that brings much efficiency, but it could.

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    k2r_ = ryMatrix(k2rtheta) * rzMatrix(k2rlambda);
+    break;
+  }
+  case Specialization::FSimaabEquiv: {
+    const auto [k2ltheta, k2lphi, k2llambda, k2lphase] =
+        anglesFromUnitary(k2l_, EulerBasis::ZYZ);
+    const auto ab = (a_ + b_) / 2.;
+    a_ = ab;
+    b_ = ab;
+    globalPhase_ = globalPhase_ + k2lphase;
```
burgholzer:
```suggestion
    globalPhase_ *= k2lphase;
```

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    k2r_ = rzMatrix(-k2lphi) * k2r_;
+    break;
+  }
+  case Specialization::FSimabbEquiv: {
+    const auto [k2ltheta, k2lphi, k2llambda, k2lphase] =
+        anglesFromUnitary(k2l_, EulerBasis::XYX);
+    const auto bc = (b_ + c_) / 2.;
+    b_ = bc;
+    c_ = bc;
+    globalPhase_ = globalPhase_ + k2lphase;
```
burgholzer:
```suggestion
    globalPhase_ *= k2lphase;
```

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
```diff
@@ -0,0 +1,635 @@
...
+    k2r_ = rxMatrix(-k2lphi) * k2r_;
+    break;
+  }
+  case Specialization::FSimabmbEquiv: {
+    const auto [k2ltheta, k2lphi, k2llambda, k2lphase] =
+        anglesFromUnitary(k2l_, EulerBasis::XYX);
+    const auto bc = (b_ - c_) / 2.;
+    b_ = bc;
+    c_ = -bc;
+    globalPhase_ = globalPhase_ + k2lphase;
```
burgholzer:
```suggestion
    globalPhase_ *= k2lphase;
```

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_weyl_decomposition.cpp
```diff
@@ -0,0 +1,409 @@
...
+#include <optional>
+#include <random>
+#include <tuple>
+#include <vector>
+
+using namespace mlir::qco;
+using namespace mlir::qco::decomposition;
+
+static Complex globalPhaseFactor(const double phase) {
+  return std::exp(Complex{0.0, 1.0} * phase);
```
burgholzer: This is just std::polar(phase) isn't it?

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_weyl_decomposition.cpp
```diff
@@ -0,0 +1,409 @@
...
+  return MATRIX;
+}
+
+static const Matrix4x4& twoQubitControlledX10() {
+  static const Matrix4x4 MATRIX = Matrix4x4::fromElements(1.0, 0.0, 0.0, 0.0, //
+                                                          0.0, 0.0, 0.0, 1.0, //
+                                                          0.0, 0.0, 1.0, 0.0, //
+                                                          0.0, 1.0, 0.0, 0.0);
+  return MATRIX;
+}
```
burgholzer: Why define these as functions with static members when they can simply be constexpr globals here?

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_weyl_decomposition.cpp
```diff
@@ -0,0 +1,409 @@
...
+    for (std::size_t i = 0; i < 4; ++i) {
+      columns[j][i] /= norm;
+    }
+  }
+  const Matrix4x4 unitary = Matrix4x4::fromElements(
+      columns[0][0], columns[1][0], columns[2][0], columns[3][0], columns[0][1],
+      columns[1][1], columns[2][1], columns[3][1], columns[0][2], columns[1][2],
+      columns[2][2], columns[3][2], columns[0][3], columns[1][3], columns[2][3],
+      columns[3][3]);
+  assert((unitary.adjoint() * unitary).isIdentity(WEYL_TOLERANCE));
```
burgholzer: The function right above this one defines a utility function for exactly this, doesn't it?

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_weyl_decomposition.cpp
```diff
@@ -0,0 +1,409 @@
...
+  return unitary;
+}
+
+static Matrix4x4 restoreWeyl(const TwoQubitWeylDecomposition& decomposition) {
+  return Matrix4x4::kron(decomposition.k1l(), decomposition.k1r()) *
+         decomposition.getCanonicalMatrix() *
+         Matrix4x4::kron(decomposition.k2l(), decomposition.k2r()) *
+         globalPhaseFactor(decomposition.globalPhase());
+}
```
burgholzer: Almost feels like this could be a utility method of the decomposer. I saw this code somewhere else as well.
Maybe holds for more of the functions here.

burgholzer (changes requested):
Thanks @simon1hofmann for another big one. ☝️ 
Made it through the PR now. I did not question the main logic, which seems to be working just fine. 

So my main focus was on things that either seemed redundant or inefficient. I hope most of these make sense. Shouldn't take too long to get all the changes in. Practically all of these are nitpicks. The main implementation seems really convincing.

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
simon1hofmann: Yes, also made the `MatrixXxX::fromElements` functions in `Matrix.h` constexpr.

mlir/lib/Dialect/QCO/Utils/Matrix.cpp
simon1hofmann: This was the solution I was looking for, thanks a lot for the suggestion :)

mlir/include/mlir/Dialect/QCO/Transforms/Decomposition/Weyl.h
simon1hofmann: It's looser than `MATRIX_TOLERANCE` because the pipeline chains many 4x4 operations (magic-basis transforms, approximate complex symmetric diagonalization at `~1e-13`, Kronecker factorization, canonical RXX/RYY/RZZ synthesis). Residual error at `1e-14` is not realistic; `1e-12` matches the scale observed in reconstruction tests and internal debug checks.

Also compared to the precision in Qiskit:
Complex symmetric diagonalization: `1e-13`
Angle / simplification / “zero” checks: `1e-12`
Kronecker product-gate validation: `1e-13`
Default Weyl fidelity: `1.0-1e-9`

I added the for precision values in `Weyl.h` to make that more consistent.

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
simon1hofmann: Horodecki et al., Phys. Rev. A 60, 1888 (1999):
`F_avg = (d + |Tr(U^dag V)|^2) / (d(d+1))`. With `d = 4` this is `(4 + |trace|²) / 20` (adapted from Qiskit's `trace_to_fid`).

I added a docstring and rewrote the calculation to explain that.

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
simon1hofmann: From https://stackoverflow.com/a/48730423: `std::mt19937{2023}` should be standardized, so the raw 32-bit output sequence is the same across different Oses.
`std::normal_distribution` is not guaranteed to be portable, as libc++, libstdc++, and MSVC can map the same engine stream to different normal samples, so I switched to a portable Box-Muller transform on raw mt19937 output.

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
> Thanks @simon1hofmann for another big one. ☝️ Made it through the PR now. I did not question the main logic, which seems to be working just fine.
> 
> So my main focus was on things that either seemed redundant or inefficient. I hope most of these make sense. Shouldn't take too long to get all the changes in. Practically all of these are nitpicks. The main implementation seems really convincing.

Thanks a lot for the feedback, most of it was straightforward to fix and I directly resolved those comments.
The few remaining ones might be worth another look.

mlir/include/mlir/Dialect/QCO/Transforms/Decomposition/Weyl.h
burgholzer: That makes sense to me 👍🏼

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
burgholzer: Thanks for the explanation! Now it makes sense to me!

mlir/lib/Dialect/QCO/Transforms/Decomposition/Weyl.cpp
burgholzer: Very interesting. Learned something new today 😄

mlir/lib/Dialect/QCO/Transforms/Decomposition/BasisDecomposer.cpp
```diff
@@ -0,0 +1,307 @@
...
+    llvm::reportFatalInternalError(
+        "Basis gate of TwoQubitBasisDecomposer is not super-controlled "
+        "- no guarantee for exact decomposition with two basis gates");
+  }
+  out.push_back(smb.u3r * target.k2r());
+  out.push_back(smb.u3l * target.k2l());
+  out.push_back(smb.q1ra * RZOp::unitaryMatrix(2. * target.b()) * smb.u2rb);
+  out.push_back(smb.q1la * RZOp::unitaryMatrix(-2. * target.a()) * smb.u2lb);
+  out.push_back(target.k1r() * smb.q0r);
+  out.push_back(target.k1l() * smb.q0l);
```
burgholzer: All of the `push_back` calls here should likely be `emplace_back` calls to avoid a potential copy and rather perform inplace construction.

burgholzer (approved):
Thanks @simon1hofmann for another round of changes!
Just browsed through and I believe this looks great now.
Once the conflicts are resolved and the last comment is addressed, this can be merged 🚀

PR #1805, opened 2026-06-23
✨Map `scf.for` operations by restoration using the 4-approximation token swap algorithm

MatthiasReumann:
## Changes 

- ✨Add functionality to handle operation with nested operations.
- ✨ Implement the 4-Approximation Algorithm described in arXiv:1602.05150v3.
- ✨Add two unit tests for quantum programs using `scf.for` operations and one unrolling unit test.
- 📝 Use LLVM codebase style function description (consistent with the IRVerifier and some other parts of the cc). 
- 🚛 Move `Layout` class to separate file. 

## Checklist


MatthiasReumann:
@coderabbitai full review

mlir/include/mlir/Dialect/QCO/Utils/Graph.h
```diff
@@ -0,0 +1,218 @@
...
+    for (const auto& [u, nbrs] : adj_) {
+      for (const auto& v : nbrs) {
+        edges.emplace_back(u, v);
+      }
+    }
+
+    Matrix<size_t> dist(n, Vector<size_t>(n, UINT64_MAX));
+    for (const auto& [u, v] : edges) {
+      dist[u][v] = 1;
+    }
```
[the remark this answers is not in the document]
MatthiasReumann: How would you enforce that for `n == 2` the edge `(3, 4)` should be `(0, 1)`?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -715,85 +689,145 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+
+              const auto prog0 = infos.lookupProgram(i0);
+              const auto prog1 = infos.lookupProgram(i1);
 
-            const auto p0 = progs[0];
-            const auto p1 = progs[1];
-            window.emplace_back(p0, p1);
-            if (window.size() == 1 + nlookahead) {
-              return WalkResult::interrupt();
+              window.emplace_back(prog0, prog1);
```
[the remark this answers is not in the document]
MatthiasReumann: The `walkProgramGraph` function never sets one-qubit gates into the ready set.

MatthiasReumann:
@coderabbitai full review

MatthiasReumann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -715,85 +690,149 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+      // correct location) to one.
+
+      for (const auto u : f.getNodes()) {
+        if (f.getDegree(u) != 0) {
+          for (const auto v : f.getNeighbours(u)) {
+            if (f.getDegree(v) == 0) {
+              curr.swap(u, v);
+              swaps.emplace_back(u, v);
+              break;
+            }
```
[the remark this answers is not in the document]
MatthiasReumann: Good catch. It should break out of both loops.

burgholzer:
@MatthiasReumann Just to double check given how there have been fairly recent pushes with changes here and the CI is failing: Is this PR ready for review?

MatthiasReumann:
> @MatthiasReumann Just to double check given how there have been fairly recent pushes with changes here and the CI is failing: Is this PR ready for review?

Sorry! Didn't check the pipeline yesterday anymore. Workflows seem to fail unexpectedly quite frequently recently. Anyway, this pull request is reviewable; yes 🙌

mlir/include/mlir/Dialect/QCO/Utils/Layout.h
```diff
@@ -0,0 +1,79 @@
...
+
+protected:
+  /// Maps a program qubit index to its hardware index.
+  SmallVector<size_t> programToHardware_;
+  /// Maps a hardware qubit index to its program index.
+  SmallVector<size_t> hardwareToProgram_;
+
+private:
+  explicit Layout(const size_t nqubits)
+      : programToHardware_(nqubits), hardwareToProgram_(nqubits) {}
```
burgholzer, line 77: Just a general observation: it may make sense to have two numbers of qubits here: the number of program qubits and the number of hardware qubits. In most cases, these will not be the same and there will be more hardware than program qubits.
A layout is generally not a bijective mapping. It is only injective.

mlir/lib/Dialect/QCO/Utils/Graph.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+ArrayRef<size_t> Graph::getNeighbours(size_t id) const { return adj_.at(id); }
+SmallVector<size_t> Graph::getNodes() const { return to_vector(adj_.keys()); }
+
+size_t Graph::getMaxDegree() const {
+  size_t deg = 0;
+  for (const auto& [u, nbrs] : adj_) {
+    deg = std::max(deg, nbrs.size());
+  }
+  return deg;
```
burgholzer, line 40: I am pretty sure there is a ranges version of `max` that could be used here.

mlir/lib/Dialect/QCO/Utils/Graph.cpp
```diff
@@ -0,0 +1,145 @@
...
+}
+
+void Graph::clearEdges() {
+  for_each(adj_, [](auto& kv) { kv.second.clear(); });
+}
+
+Graph::DistanceMatrix Graph::getDistMatrix() const {
+  const auto n = getNumNodes();
+
+  Graph::DistanceMatrix dist(n, UINT64_MAX);
```
burgholzer: I believe it is preferable to use the C++ `limits` header here instead of these constants.

mlir/include/mlir/Dialect/QCO/Utils/Graph.h
```diff
@@ -0,0 +1,106 @@
...
+#include <optional>
+#include <utility>
+
+namespace mlir::qco {
+
+/// A directed graph.
+class Graph {
+public:
+  class DistanceMatrix {
+    SmallVector<size_t> data_;
```
burgholzer, line 31: Just an efficiency question, but do we really need 64Bit here? Wouldn't 32 or even 16 also do the trick?
Holds throughout the PR.

mlir/lib/Dialect/QCO/Utils/Layout.cpp
```diff
@@ -0,0 +1,70 @@
...
+  assert(hwA < hardwareToProgram_.size() && "hardware index out of bounds");
+  assert(hwB < hardwareToProgram_.size() && "hardware index out of bounds");
+  const auto progA = hardwareToProgram_[hwA];
+  const auto progB = hardwareToProgram_[hwB];
+
+  std::swap(hardwareToProgram_[hwA], hardwareToProgram_[hwB]);
+  std::swap(programToHardware_[progA], programToHardware_[progB]);
+}
+
+size_t Layout::nqubits() const { return programToHardware_.size(); }
```
burgholzer, line 64: Related to the other comment, this should probably be two functions.

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
```diff
@@ -25,7 +24,8 @@ namespace mlir::qco {
  * @brief Create a mapping pass instance with the given target architecture.
  * @returns a pass object.
  */
-std::unique_ptr<Pass> createMappingPass(size_t nqubits, const Edges& coupling,
-                                        MappingPassOptions options);
+std::unique_ptr<Pass>
+createMappingPass(const llvm::DenseSet<std::pair<size_t, size_t>>&,
+                  MappingPassOptions);
```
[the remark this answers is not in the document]
burgholzer: @MatthiasReumann doesn't @coderabbitai have a point here?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -64,235 +70,107 @@ namespace {
...
-    Matrix dist_;
-    Edges coupling_;
-    Neighbours neighbours_;
+    Graph coupling_;
+    Graph::DistanceMatrix dist_;
   };
 
-  struct [[nodiscard]] Trial {
-    explicit Trial(Layout layout) : layout(std::move(layout)) {}
+  struct WireInfos {
```
burgholzer, line 121: This whole class kind of feels like a layout. Isn't it redundant?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -496,36 +389,37 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
         }
       }
     }
-    return wires;
+
+    return std::make_pair(wires, infos);
```
burgholzer, line 393:
```suggestion
    return {wires, infos};
```
Maybe this is just me, but I hate writing out the construction explicitly.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -557,84 +452,138 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+              qubits.erase(op.getQubitIn());
+            })
+            .Case<AllocOp, qtensor::AllocOp>([&](auto) {
+              llvm::reportFatalInternalError("unexpected dynamic qubit alloc");
+            });
       }
     }
 
-    return best;
+    return std::make_pair(wires, infos);
```
```suggestion
    return {wires, infos};
```

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -715,85 +664,159 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
-   * @details Traverses the pair of wire iterators in tandem until a two-qubit
-   * operation is found. If the two-qubit operation is equivalent, continue.
-   * Otherwise stop.
-   */
-  template <WireDirection Direction>
-  static void skipQubitPairBlock(WireIterator& w0, WireIterator& w1) {
-    using Traits = WireTraversalTraits<Direction>;
+  /// Return the sequence of SWAPs to move from one layout to another.
+  /// Implements the 4-Approximation algorithm described in arXiv:1602.05150v3.
+  SmallVector<IndexPairType> restore(const Layout& from, const Layout& to) {
```
burgholzer, line 669: Just out of curiosity:
We solve a similar problem in our DD package, where we track a permutation, then compare it to an expected one at the end of the computation and restore it when necessary. Would that restoration potentially also benefit from an improved algorithm like this? The major difference is that in the DD package the graph is always all-to-all.

The current code for this is here: https://github.com/munich-quantum-toolkit/core/blob/main/include/mqt-core/dd/Operations.hpp#L232-L304

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -802,145 +825,243 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+  // Verify the assumption that the coupling set is symmetric:
+  // For every edge (u, v) in the set, (v, u) must also be present.
+
+  for (const auto& [u, v] : couplingSet) {
+    if (u == v) {
+      llvm::reportFatalUsageError("Found an invalid (u, u) edge.");
+      return nullptr;
+    }
+
+    if (!couplingSet.contains(std::make_pair(v, u))) {
```
```suggestion
    if (!couplingSet.contains({v, u})) {
```

burgholzer (changes requested):
This generally looks really clean to me! I'm only have a couple of nitpicks and questions.

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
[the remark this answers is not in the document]
burgholzer: Maybe this is an artificial concern as an architecture with a single isolated qubit that can't interact with others is not practically relevant; in which case the design here is fine as it is.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Almost. The `WireInfos` class maps an index of a `Vector<WireIterator>` to a program index, whereas a layout maps a hardware index to a program index.

The current implementation in `main` assumes that the i-th index of a `Vector<WireIterator>` is the i-th program qubit. However, with nested structures (such as `scf.for`) which may only use a subset of qubit wires as input, this assumption does not hold anymore. Consequently, it requires some kind of additional data structure with as little overhead as possible: `WireInfos`.

mlir/include/mlir/Dialect/QCO/Utils/Graph.h
MatthiasReumann: I keep wondering this too. Especially regarding performance gain for larger circuits. Maybe something to try and benchmark in a follow-up pull request.

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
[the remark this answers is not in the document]
MatthiasReumann: Yeah. However, given that the signature will change anyways once QDMI comes in, I thought it isn't worthwhile spending time on it.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Mhm. I think the all-to-all part is the interesting bit here. Given that, I don't think the algorithm makes much sense as the current DD package implementation requires only `O(sizeof(to:permutation))` swaps.

mlir/include/mlir/Dialect/QCO/Utils/Layout.h
MatthiasReumann: Yes. I think at some point we also discussed that the layout qubit indices may not be consecutive. So a follow-up pull request resolving these issues would be appropriate.

burgholzer: Agreed. Can we track this somewhere?

mlir/include/mlir/Dialect/QCO/Utils/Graph.h
burgholzer: Benchmarking this would really be nice. Obviously requires a good enough benchmarking harness to actually see the difference.
Can you track this somewhere?

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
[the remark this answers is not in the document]
burgholzer: > Yeah. However, given that the signature will change anyways once QDMI comes in, I thought it isn't worthwhile spending time on it. 

Agreed! 👍🏻

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Yeah, I get the need for _a_ data structure here. I am mostly wondering whether the layout class is generic enough to fill the role. In the end, it's also just a mapping between indices, isn't it?
But maybe I am also just simplifying things too much. This obviously works and should be fine as-is.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Alright. Thanks for your assessment.
I also had the impression that the DD version is pretty efficient already, but you never know 🙂 feel free to resolve here

burgholzer (approved):
Once the open conversations are resolved, feel free to merge this 🚀

mlir/include/mlir/Dialect/QCO/Utils/Graph.h
MatthiasReumann: I will create an issue with sub-issues for this and the above point!

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Ahh. Got it. Interesting point. I think if it the class wasn't called layout, I could live with it.

burgholzer: Not within this PR (rather one of the subissues), but maybe `QubitMapping` would be a suitable name that would cover both use cases.

MatthiasReumann: I mean, we could also just make it a `BidirectionalMap<T>` and let the user decide its meaning...? (Keeping the issue of injective mappings open)

mlir/include/mlir/Dialect/QCO/Utils/Layout.h
MatthiasReumann: #1867

mlir/include/mlir/Dialect/QCO/Utils/Graph.h
MatthiasReumann: https://github.com/munich-quantum-toolkit/core/issues/1868

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: That would certainly also be an option. 
However, let's not hold up this PR for longer than it needs to and tackle that in a follow up. Feel free to resolve here and hit the merge button 😌 🚀

PR #1806, opened 2026-06-23
:sparkles: Add SCF support to `WireIterator`

MatthiasReumann:
## Changes

- ✨Add SCF op switch cases in `WireIterator.cpp` and add unit tests

(Extracted from #1805 to decrease that PR size and ease reviewing) 

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).
- [x] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

PR #1808, opened 2026-06-24
✨ Add SCF Support to `walkProgramGraph`

MatthiasReumann:
## Changes

- ✨Add SCF op switch cases in `Drivers.h` and update unit tests

(Extracted from #1805 to decrease that PR size and ease reviewing) 
(Depends on #1806)
## Checklist


MatthiasReumann:
@coderabbitai full review

mlir/include/mlir/Dialect/QCO/Utils/Drivers.h
```diff
@@ -88,13 +91,33 @@ LogicalResult walkProgram(Region& region, const WalkProgramFn& fn) {
...
+      .Case<qco::IfOp>([&](qco::IfOp op) { return op.getQubits().size(); })
+      .Case<AllocOp, StaticOp, ResetOp, MeasureOp, SinkOp, YieldOp,
+            qtensor::ExtractOp, qtensor::InsertOp, scf::YieldOp>(
+          [&](auto) { return 1; })
+      .Default([&](Operation* op) {
+        const auto name = op->getName().getStringRef();
+        reportFatalInternalError("unknown op: " + name);
+        return 0;
+      });
+}
```
burgholzer: This feels wrong to me on some levels.
- Why would all of the inits of a for or while loop be qubits? That's an assumption not being stated here.
- Why would a yield operation only ever yield one value? Also seems like a limiting assumption.

mlir/include/mlir/Dialect/QCO/Utils/Drivers.h
```diff
@@ -134,67 +157,55 @@ LogicalResult walkProgramGraph(MutableArrayRef<WireIterator> wires,
...
+        const auto nqubits = impl::getNumQubits(it.operation());
+
+        assert(nqubits != 0);
+        if (nqubits == 1) {
+          std::ranges::advance(it, Traits::stride());
+          continue;
         }
 
-        if (res.wasInterrupted()) {
+        assert(nqubits >= 2);
```
burgholzer: Kind of a useless assertion. Either the 0-one triggers already or this is trivially fulfilled.

burgholzer (changes requested):
Likely just me being too dense, but there are two things that do not quite make sense to me here.

mlir/include/mlir/Dialect/QCO/Utils/Drivers.h
MatthiasReumann: You are absolutely right. 

- Yes, this is an implicit assumption, taken over from the existing code base. I will adapt it to filter for qubit values for now. Maybe this function can at some point also walk the tensor chain of tensors to find qubit values and by doing so iterate the program graph. 
- Yeah, this is literally just to match the `if (nqubits == 1)` result below. I have actually no idea why I thought that this was a good idea yesterday 🙈 This case simply matches non-unitary-like operations (source-like ops and sink-like ops). I will update the code to reflect the intent more clearly.

mlir/include/mlir/Dialect/QCO/Utils/Drivers.h
MatthiasReumann: Yup

burgholzer (approved):
This looks much more reasonable now 👍🏻

Issue #1760, opened 2026-06-02
✨ Implement `getUnitaryMatrix()` for multi-operation modifiers


denialhaag:
#1751 has dropped the condition that modifier bodies can only contain a single unitary operation followed by a `yield` operation.

However, the `getUnitaryMatrix()` methods of `qco::CtrlOp`, `qco::InvlOp`, and `qco::PowOp` currently do not support such multi-operation modifiers. Such support should be added. To do so, we have to decide on some form of convention for the basis in which the matrices are represented.

PR #1809, opened 2026-06-24
✨ Compose multi-gate unitary matrices in CtrlOp and InvOp bodies

simon1hofmann:
## Description

- Add shared `composeSingleQubitBodyMatrix` in `QCOUtils` to compose compile-time single-qubit unitaries inside modifier bodies (in program order).
- Extend `CtrlOp::getUnitaryMatrix()` to use composition when the body is not a single unitary gate (e.g. `ctrl { x; h }`).
- Refactor `InvOp::getUnitaryMatrix()` to reuse the same helper and take the adjoint, removing duplicated logic.
- Add matrix tests for controlled modifier bodies, including `inv` inside `ctrl`.

## Problem

`CtrlOp::getUnitaryMatrix()` only handled bodies with exactly one unitary op via `getSoleBodyUnitary`. Multi-gate bodies like `ctrl { x; h }` returned `nullopt`.

## Solution

Extract `composeSingleQubitBodyMatrix(Block&)` into `mlir/lib/Dialect/QCO/IR/QCOUtils.cpp` and use it from both modifiers:

- **`CtrlOp`**: sole body unitary → existing fast path; otherwise (1 target) compose body gates and embed as controlled block.
- **`InvOp`**: sole body unitary → adjoint of target matrix; otherwise compose body and return `fromAdjoint(composed)`.

## Checklist


simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

denialhaag:
We can probably mark this as closing #1760, but we should create a new issue for `PowOp` if we do not already implement the matrices in #1603.

denialhaag:
> We can probably mark this as closing #1760, but we should create a new issue for `PowOp` if we do not already implement the matrices in #1603.

Ahh, I just realized that this PR is only targeting single-target modifiers. I think this is a perfectly mergeable improvement on the status quo, but it might be better to adapt the wording in #1760 rather than marking this PR as closing it. At some point, we probably want to compose the unitary matrices of arbitrary modifier operations, which will add some more complexity. Sorry for the confusion!

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -11,15 +11,29 @@
...
+#include "mlir/Dialect/QCO/Utils/Matrix.h"
 
 #include <mlir/Dialect/Arith/IR/Arith.h>
 #include <mlir/Dialect/Utils/Utils.h>
+#include <mlir/IR/Block.h>
 #include <mlir/IR/PatternMatch.h>
 #include <mlir/Support/LLVM.h>
 #include <mlir/Support/LogicalResult.h>
 
+#include <cmath>
```
burgholzer: Why is this include necessary here?

```suggestion
```

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -301,11 +302,15 @@ bool CtrlOp::hasCompileTimeKnownUnitaryMatrix() {
...
-  auto bodyUnitary = utils::getSoleBodyUnitary<UnitaryOpInterface>(*getBody());
-  if (!bodyUnitary) {
-    return std::nullopt;
+  std::optional<DynamicMatrix> targetMatrix;
+  if (auto bodyUnitary =
+          utils::getSoleBodyUnitary<UnitaryOpInterface>(*getBody())) {
+    targetMatrix = bodyUnitary.getUnitaryMatrix<DynamicMatrix>();
+  } else if (getNumTargets() == 1) {
+    if (const auto composed = composeSingleQubitBodyMatrix(*getBody())) {
+      targetMatrix = DynamicMatrix(*composed);
```
burgholzer: I somewhat dislike that this imposes a copy of the matrix. Can we get rid of that?

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
```diff
@@ -0,0 +1,72 @@
...
+                 return false;
+               }
+               global *= (*matrix)(0, 0);
+               return true;
+             })
+             .Case<UnitaryOpInterface>([&](UnitaryOpInterface unitary) {
+               Matrix2x2 matrix;
+               if (!unitary.getUnitaryMatrix2x2(matrix)) {
+                 return false;
+               }
```
burgholzer, line 53: Shouldn't this use the `isSingleQubit()` and `hasStaticUnitary` checks to be more efficient?

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
```diff
@@ -0,0 +1,72 @@
...
+               };
+               return !llvm::any_of(operation->getOperands(), usesQubit) &&
+                      !llvm::any_of(operation->getResults(), usesQubit);
+             })) {
+      return std::nullopt;
+    }
+  }
+  if (!found && std::abs(global - Complex{1.0, 0.0}) <= utils::TOLERANCE) {
+    return std::nullopt;
+  }
```
burgholzer: I do not quite think that this condition is right.
If `found` is false, how could the phase not be 1.0?
Probably I am missing something.

mlir/unittests/Dialect/QCO/IR/test_qco_ir_matrix.cpp
```diff
@@ -93,6 +93,59 @@ TEST_F(QCOMatrixTest, CXOpMatrix) {
...
+  ASSERT_TRUE(moduleOp);
+
+  auto funcOp = *moduleOp->getBody()->getOps<func::FuncOp>().begin();
+  auto ctrlOp = *funcOp.getBody().getOps<CtrlOp>().begin();
+  auto matrix = ctrlOp.getUnitaryMatrix();
+  ASSERT_TRUE(matrix);
+
+  DynamicMatrix expected = DynamicMatrix::identity(4);
+  expected.setBottomRightCorner(HOp::getUnitaryMatrix() *
+                                XOp::getUnitaryMatrix());
```
burgholzer: Given how we rely on the DD package for the test above and some others already, we may as well use the DD package here consistently, right?

mlir/unittests/programs/qco_programs.h
```diff
@@ -173,6 +173,14 @@ void twoX(QCOProgramBuilder& b);
 /// Creates a circuit with a control modifier applied to two subsequent X gates.
 void controlledTwoX(QCOProgramBuilder& b);
 
+/// Creates a circuit with a control modifier applied to an X-H sequence in the
+/// ctrl body.
+void controlledXH(QCOProgramBuilder& b);
+
+/// Creates a circuit with a control modifier applied to an inverse H-T sequence
+/// in the ctrl body.
+void controlledInverseHT(QCOProgramBuilder& b);
```
burgholzer: Given how these are not used as part of the overall pipeline e2e tests, I'd rather move these to the single test cases where they are used.

burgholzer (changes requested):
This generally LGTM with some small exceptions that are probably worth revisiting.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
simon1hofmann: Yes that should be possible, refactored the function a little bit.

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
simon1hofmann: Yes that is possible 👍

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
simon1hofmann: That is basically this test:
```
void inverseGphaseBarrier(QCOProgramBuilder& b) {
  auto q = b.allocQubitRegister(1);
  b.inv(q[0], [&](ValueRange qubits) -> SmallVector<Value> {
    b.gphase(0.123);
    return {b.barrier({qubits[0]})[0]};
  });
}
```
Here `found` stays `false`, but `global = e^{i*0.123}`. The function should return a phased identity (`global*I₂`), which is what `InverseGphaseBarrierOpMatrix` tests.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -301,33 +302,40 @@ bool CtrlOp::hasCompileTimeKnownUnitaryMatrix() {
...
-  // define dimensions and type of output matrix
-  const auto dim = static_cast<int64_t>((1ULL << getNumControls()) * targetDim);
-
-  // initialize result with identity
-  auto matrix = DynamicMatrix::identity(dim);
+  // Build `I_{2^controls} ⊗ U` by placing the target block in the bottom-right
+  // corner of a `2^controls * targetDim` identity.
+  const auto controlledMatrix =
+      [this](const std::int64_t targetDim,
+             const auto& targetBlock) -> std::optional<DynamicMatrix> {
```
burgholzer:
```suggestion
             const auto& targetBlock) -> DynamicMatrix {
```
Should just work, right? This will always return a concrete matrix.

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -301,33 +302,40 @@ bool CtrlOp::hasCompileTimeKnownUnitaryMatrix() {
...
 
-  // define dimensions and type of output matrix
-  const auto dim = static_cast<int64_t>((1ULL << getNumControls()) * targetDim);
-
-  // initialize result with identity
-  auto matrix = DynamicMatrix::identity(dim);
+  // Build `I_{2^controls} ⊗ U` by placing the target block in the bottom-right
+  // corner of a `2^controls * targetDim` identity.
+  const auto controlledMatrix =
+      [this](const std::int64_t targetDim,
```
burgholzer: One could remove the `this` capture here by passing on the number of controls into the lambda.

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
burgholzer: Ah. Now I see. I overlooked that gphase does not set `found` to true.
Wouldn't that be cleaner? In the end, it is also a unitary operation.

mlir/unittests/Dialect/QCO/IR/test_qco_ir_matrix.cpp
```diff
@@ -93,6 +116,65 @@ TEST_F(QCOMatrixTest, CXOpMatrix) {
...
+  ASSERT_TRUE(moduleOp);
+
+  auto funcOp = *moduleOp->getBody()->getOps<func::FuncOp>().begin();
+  auto ctrlOp = *funcOp.getBody().getOps<CtrlOp>().begin();
+  auto matrix = ctrlOp.getUnitaryMatrix();
+  ASSERT_TRUE(matrix);
+
+  const auto ch = qc::StandardOperation(1, 0, qc::OpType::H);
+  const auto dd = std::make_unique<dd::Package>(2);
+  const auto chDD = dd::getDD(ch, *dd);
```
burgholzer: Now that I think about this a bit more: if we still rely on the old IR here for these tests, then we may as well build a QuantumComputation here, which has a way more convenient builder interface and then call the buildFunctionality method on the quantum computation. Should be way less verbose.

burgholzer (changes requested):
One more round of changes until we can get this in 😊

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
simon1hofmann: Yes, overlooked that.

mlir/unittests/Dialect/QCO/IR/test_qco_ir_matrix.cpp
simon1hofmann: Switched to QuantumComputation, just looks a little bit odd as control and target have to be switched 🤔

burgholzer (commented):
Hm. The code here generally looks good to me now; except for the endianness thing you brought up.

I think we should strive to use a consistent endianness representation throughout the MQT. Currently, the interpretation between the MQT Core IR/the DD Package and the MLIR code base seem to differ. This sounds like it could lead to problems down the way.

In "old" MQT Core, the higher valued qubit index is always the more significant entry. DDs are always constructed with highest index on top. Thus, `CX(1, 0)` (1 as control, 0 as target) has the textbook CNOT matrix $diag(I, X)$, while `CX(0, 1)` has the "less usual matrix representation.
This is also compatible with Qiskit: https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.library.CXGate#cxgate

Any thoughts on how to best handle this?

simon1hofmann:
> Hm. The code here generally looks good to me now; except for the endianness thing you brought up.
> 
> I think we should strive to use a consistent endianness representation throughout the MQT. Currently, the interpretation between the MQT Core IR/the DD Package and the MLIR code base seem to differ. This sounds like it could lead to problems down the way.
> 
> In "old" MQT Core, the higher valued qubit index is always the more significant entry. DDs are always constructed with highest index on top. Thus, `CX(1, 0)` (1 as control, 0 as target) has the textbook CNOT matrix d i a g ( I , X ) , while `CX(0, 1)` has the "less usual matrix representation. This is also compatible with Qiskit: https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.library.CXGate#cxgate
> 
> Any thoughts on how to best handle this?

So there was basically an inconsistency before that already, as this was the test for CNOT:
```
TEST_F(QCOMatrixTest, CXOpMatrix) {
  auto moduleOp = QCOProgramBuilder::build(context.get(), singleControlledX);
  ASSERT_TRUE(moduleOp);

  // Get the operation from the module
  auto funcOp = *moduleOp->getBody()->getOps<func::FuncOp>().begin();
  auto ctrlOp = *funcOp.getBody().getOps<CtrlOp>().begin();
  auto matrix = ctrlOp.getUnitaryMatrix();

  // Get the definition of the matrix from the DD library
  const auto cx = qc::StandardOperation(1, 0, qc::OpType::X);
  const auto dd = std::make_unique<dd::Package>(2);
  const auto cxDD = dd::getDD(cx, *dd);
  const auto definition = cxDD.getMatrix(2);
```

So what we would have to change:
- for e.g. `CX(0, 1)`, instead of embedding X in the bottom right corner, we would have to map control/target operands to the right indices to get the same matrix representation as in Qiskit.
- for asymmetric gates (e.g. `DCX`) we have to reorder the definition of its unitary matrix.

The code won't be as nice anymore, as it's not as straightforward as creating an identity matrix and setting the bottom right corner by the body matrix anymore.

Should we do that as part of this PR? 
As far as I can tell, no parts in the code base yet rely on the concrete ordering besides `Matrix.cpp` and the corresponding tests. Two qubit synthesis will be affected, which should be easy to fix.

burgholzer:
> > Hm. The code here generally looks good to me now; except for the endianness thing you brought up.
> > I think we should strive to use a consistent endianness representation throughout the MQT. Currently, the interpretation between the MQT Core IR/the DD Package and the MLIR code base seem to differ. This sounds like it could lead to problems down the way.
> > In "old" MQT Core, the higher valued qubit index is always the more significant entry. DDs are always constructed with highest index on top. Thus, `CX(1, 0)` (1 as control, 0 as target) has the textbook CNOT matrix d i a g ( I , X ) , while `CX(0, 1)` has the "less usual matrix representation. This is also compatible with Qiskit: https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.library.CXGate#cxgate
> > Any thoughts on how to best handle this?
> 
> So there was basically an inconsistency before that already, as this was the test for CNOT:
> 
> ```
> TEST_F(QCOMatrixTest, CXOpMatrix) {
>   auto moduleOp = QCOProgramBuilder::build(context.get(), singleControlledX);
>   ASSERT_TRUE(moduleOp);
> 
>   // Get the operation from the module
>   auto funcOp = *moduleOp->getBody()->getOps<func::FuncOp>().begin();
>   auto ctrlOp = *funcOp.getBody().getOps<CtrlOp>().begin();
>   auto matrix = ctrlOp.getUnitaryMatrix();
> 
>   // Get the definition of the matrix from the DD library
>   const auto cx = qc::StandardOperation(1, 0, qc::OpType::X);
>   const auto dd = std::make_unique<dd::Package>(2);
>   const auto cxDD = dd::getDD(cx, *dd);
>   const auto definition = cxDD.getMatrix(2);
> ```
> 
> So what we would have to change:
> 
> * for e.g. `CX(0, 1)`, instead of embedding X in the bottom right corner, we would have to map control/target operands to the right indices to get the same matrix representation as in Qiskit.
> * for asymmetric gates (e.g. `DCX`) we have to reorder the definition of its unitary matrix.
> 
> Should we do that as part of this PR? As far as I can tell, no parts in the code base yet rely on the concrete ordering besides `Matrix.cpp` and the corresponding tests. Two qubit synthesis will be affected, which should be easy to fix.

Hm. The thing is that the two-target gates should be fine already. We run unit tests that compare the generated matrix with the DD Package one. And I believe these already use the same ordering of the qubits. Worth checking though.

I would guess that the only "mistake" here is in the control modifier. But then again, I am not even sure if this is an error in the matrix computation itself. The current code simply puts the matrix in the bottom right corner of the larger identity matrix. This is correct in one case (`cx(1, 0)`) but wrong in the other (`cx(0, 1)`). If you now flip the logic for the matrix generation, then it will just flip which case works correctly and which one does not.

I think the problem is coming from the fact that the current implementation does not dictate any kind of endianness, but rather works as "control is higher, target is lower"-convention.
I suppose there is no way in actually changing that except if we would impose some kind of qubit order as an input to the operation. That does not seem to make too much sense to me though.

Okay.. lots of loud thinking. Overall this has led me to believe that there is nothing actually wrong here. As long as the unitary synthesis rewires the matrices accordingly so that they are assembled and combined "correctly" with the assumption on the qubit ordering from above, we should be good.

Does my rambling make any sense? Might also be worth to fact check with an LLM. @coderabbitai can you do that?

simon1hofmann:
> > > Hm. The code here generally looks good to me now; except for the endianness thing you brought up.
> > > I think we should strive to use a consistent endianness representation throughout the MQT. Currently, the interpretation between the MQT Core IR/the DD Package and the MLIR code base seem to differ. This sounds like it could lead to problems down the way.
> > > In "old" MQT Core, the higher valued qubit index is always the more significant entry. DDs are always constructed with highest index on top. Thus, `CX(1, 0)` (1 as control, 0 as target) has the textbook CNOT matrix d i a g ( I , X ) , while `CX(0, 1)` has the "less usual matrix representation. This is also compatible with Qiskit: https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.library.CXGate#cxgate
> > > Any thoughts on how to best handle this?
> > 
> > 
> > So there was basically an inconsistency before that already, as this was the test for CNOT:
> > ```
> > TEST_F(QCOMatrixTest, CXOpMatrix) {
> >   auto moduleOp = QCOProgramBuilder::build(context.get(), singleControlledX);
> >   ASSERT_TRUE(moduleOp);
> > 
> >   // Get the operation from the module
> >   auto funcOp = *moduleOp->getBody()->getOps<func::FuncOp>().begin();
> >   auto ctrlOp = *funcOp.getBody().getOps<CtrlOp>().begin();
> >   auto matrix = ctrlOp.getUnitaryMatrix();
> > 
> >   // Get the definition of the matrix from the DD library
> >   const auto cx = qc::StandardOperation(1, 0, qc::OpType::X);
> >   const auto dd = std::make_unique<dd::Package>(2);
> >   const auto cxDD = dd::getDD(cx, *dd);
> >   const auto definition = cxDD.getMatrix(2);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So what we would have to change:
> > 
> > * for e.g. `CX(0, 1)`, instead of embedding X in the bottom right corner, we would have to map control/target operands to the right indices to get the same matrix representation as in Qiskit.
> > * for asymmetric gates (e.g. `DCX`) we have to reorder the definition of its unitary matrix.
> > 
> > Should we do that as part of this PR? As far as I can tell, no parts in the code base yet rely on the concrete ordering besides `Matrix.cpp` and the corresponding tests. Two qubit synthesis will be affected, which should be easy to fix.
> 
> Hm. The thing is that the two-target gates should be fine already. We run unit tests that compare the generated matrix with the DD Package one. And I believe these already use the same ordering of the qubits. Worth checking though.
> 
> I would guess that the only "mistake" here is in the control modifier. But then again, I am not even sure if this is an error in the matrix computation itself. The current code simply puts the matrix in the bottom right corner of the larger identity matrix. This is correct in one case (`cx(1, 0)`) but wrong in the other (`cx(0, 1)`). If you now flip the logic for the matrix generation, then it will just flip which case works correctly and which one does not.
> 
> I think the problem is coming from the fact that the current implementation does not dictate any kind of endianness, but rather works as "control is higher, target is lower"-convention. I suppose there is no way in actually changing that except if we would impose some kind of qubit order as an input to the operation. That does not seem to make too much sense to me though.
> 
> Okay.. lots of loud thinking. Overall this has led me to believe that there is nothing actually wrong here. As long as the unitary synthesis rewires the matrices accordingly so that they are assembled and combined "correctly" with the assumption on the qubit ordering from above, we should be good.
> 
> Does my rambling make any sense? Might also be worth to fact check with an LLM. @coderabbitai can you do that?

`CtrlOp::getUnitaryMatrix()` always embeds in the bottom right corner, I compared the unitary matrix of `b.cx(q[0], q[1])` and `b.cx(q[1], q[0])`, which are exactly the same.

Should we merge this PR and then fix `CtrlOp::getUnitaryMatrix()` in a follow-up?

burgholzer (approved):
Following the suggestion by the rabbit, let's merge this for now, and resolve the underlying issue in a follow up.

I think it might also be worth checking what Qiskit does here. As far as I can tell, they implement CX as a singleton gate with a fixed matrix. As such, I am not sure that they have the concept of "give me the matrix for this particar instatiation of an operation applied to these qubits". So it has to do this rewiring somewhere as well.
But let's maybe take this discussion to a follow-up issue or pull request. 
It does not affect the functionality of this PR.

simon1hofmann:
> Following the suggestion by the rabbit, let's merge this for now, and resolve the underlying issue in a follow up.
> 
> I think it might also be worth checking what Qiskit does here. As far as I can tell, they implement CX as a singleton gate with a fixed matrix. As such, I am not sure that they have the concept of "give me the matrix for this particar instatiation of an operation applied to these qubits". So it has to do this rewiring somewhere as well. But let's maybe take this discussion to a follow-up issue or pull request. It does not affect the functionality of this PR.

Sounds good, created a follow-up issue: #1821

PR #1810, opened 2026-06-24
✨ Add `decompose-multi-controlled` pass for multi-controlled X/Z/phase decomposition

simon1hofmann:
## Description

Add a `decompose-multi-controlled` QCO pass that decomposes controlled X/Z/phase
gates (`qco.ctrl` with a single `qco.x`, `qco.z`, or constant-angle `qco.p` body)
and `qco.rccx` with at least `min-controls` controls (default 2; must be ≥ 2).

Synthesis is paper-derived via a Plan→Emit pipeline:

- **X/Z**
  - `k = 2, 3`: elementary CCX/CCZ and CCCX/CCCZ; `rccx` when `min-controls ≤ 2`
  - `k = 4`: ancilla-free relative-phase `C^4(Z)` (Barenco √Z peels + Maslov
    relative-phase toggles); MCX = `H · core · H`
  - `k = 5`: Vale MCP(π) hybrid residual (expanded NestedMCX)
  - `k ≥ 6`: Huang–Palsberg (HP24) borrowed-helper MCZ core with a compile-time
    CX policy table; MCX = `H · core · H`
- **Phase**
  - `k = 2`: optimized ancilla-free `C^2(P)`
  - `k = 3, 4`: Vale Fig. 7 (optimized `C^2(P)` residual at 3; Barenco-relative
    / RCCX residual at 4)
  - `k ≥ 5`: da Silva–Park 2022 (SP22) linear-depth LDD (`CP` / `CRX`)
  - compile-time `±π` is routed through the Z / MCZ path

The greedy rewriter lowers intermediate `qco.ctrl` / `qco.rccx` building blocks
when `min-controls` allows (default 2 fully expands).

## Usage

```bash
# Fully lower CCX/CCZ/CCP/RCCX and larger MCX/MCZ/MCP (default min-controls=2)
mqt-cc --decompose-multi-controlled input.mlir -o output.mlir
# Leave CCX/RCCX/CCP; decompose k≥3
mqt-cc --decompose-multi-controlled \
       --decompose-multi-controlled-min-controls=3 \
       input.mlir -o output.mlir
# Keep Toffoli/CCZ native; decompose only k≥4
mqt-cc --decompose-multi-controlled \
       --decompose-multi-controlled-min-controls=4 \
       input.mlir -o output.mlir
```
### AI Assistance

## Checklist


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

mlir/include/mlir/Compiler/CompilerPipeline.h
```diff
@@ -49,6 +50,13 @@ struct QuantumCompilerConfig {
 
   /// Enable Hadamard lifting
   bool enableHadamardLifting = false;
+
+  /// Decompose multi-controlled X/Z gates into elementary one- and two-qubit
+  /// gates.
+  bool enableDecomposeMultiControlled = false;
```
burgholzer: Just a comment: it's likely that we will add this to the pass pipeline per default at some point, at which point we would likely flip the default here.

mlir/lib/Compiler/CompilerPipeline.cpp
```diff
@@ -77,6 +77,13 @@ QuantumCompilerPipeline::runPipeline(ModuleOp module,
                     "enabled and the record pointer to be non-null.\n";
     return failure();
   }
+  if (config_.enableDecomposeMultiControlled &&
+      config_.decomposeMultiControlledMinControls < 2) {
+    llvm::errs()
+        << "decomposeMultiControlledMinControls must be at least 2 when "
+           "enableDecomposeMultiControlled is enabled.\n";
```
burgholzer: This limitation should be spelled out explicitly in the pass documentation and the option description for the compiler pipeline.

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_multi_controlled_decomposition.cpp
```diff
@@ -0,0 +1,462 @@
...
+  return buildControlledPauliModule(context, numControls, ControlledPauli::X);
+}
+
+[[nodiscard]] static OwningOpRef<ModuleOp>
+buildMczModule(MLIRContext* context, std::size_t numControls) {
+  return buildControlledPauliModule(context, numControls, ControlledPauli::Z);
+}
+
+[[nodiscard]] static qc::QuantumComputation
+funcOpToQuantumComputation(func::FuncOp funcOp, std::size_t& numQubits) {
```
burgholzer: I generally really like the verification via the DD package. As such, this kind of translation is necessary at the moment. 
I see at least two potential directions for follow-ups here:
- make this a more general feature that we can convert programs to quantum computations. Might require too much work for too little benefit.
- generalize the DD package so that it can also work with the MLIR code base. While it might not be completely possible to detangle the DD package from the old IR module, a large part of the functionality is actually independent. We could potentially port the simulate and buildFunctionality functions over to MLIR programs with only moderate effort. That would avoid going through the QuantumComputation class for tests and computations like these.

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_multi_controlled_decomposition.cpp
```diff
@@ -0,0 +1,462 @@
...
+  if (pauli == ControlledPauli::X) {
+    referenceQc.mcx(controls, target);
+  } else {
+    referenceQc.mcz(controls, target);
+  }
+
+  const dd::MatrixDD decomposedDD = dd::buildFunctionality(decomposedQc, *dd);
+  const dd::MatrixDD referenceDD = dd::buildFunctionality(referenceQc, *dd);
+  EXPECT_TRUE(dd->multiply(decomposedDD, dd->conjugateTranspose(referenceDD))
+                  .isIdentity(/*upToGlobalPhase=*/false));
```
burgholzer: I suppose that numerical accuracy is a problem here so that. Simple comparison of the resulting DDs (EXPECT_EQ) is not sufficient here and fails, right?

mlir/include/mlir/Dialect/QCO/Transforms/Decomposition/MultiControlled.h
burgholzer, line 1: This might once more just be me, but is there really any value in having a dedicated header file for these functions? Couldn't these simply be part of the respective pass CPP file?
It feels like these are very particular function that do not have any public API usage or at least none that I could foresee.
Would you agree?

Edit: having gone through more of the PR, I think I just don't quite like the idea that so much of the pass logic is actively decoupled from MLIR. For the most part, we have the MLIR pass logic and that then builds some auxiliary data structures, which are passed to a collection of methods that doesn't even really care about MLIR.
I would have thought/hoped that we could integrate the respective functionality a little tighter with MLIR itself and avoid auxiliary boilerplate as much as possible.
However, I can also see that the current design also has its pro's; decoupling the infrastructure from the actual pass logic.
Maybe this is just me leaning towards the tightly integrated solution.
Tagging @munich-quantum-toolkit/mqt-cc here for visibility. Maybe someone has a comment or a different perspective

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+  void p(std::size_t q, double theta) {
+    setWire(q, POp::create(*builder_, loc_, wire(q), theta).getOutputQubit(0));
+  }
+
+  void cx(std::size_t control, std::size_t target) {
+    auto ctrlOp = CtrlOp::create(
+        *builder_, loc_, ValueRange{wire(control)}, ValueRange{wire(target)},
+        [&](ValueRange args) -> SmallVector<Value> {
+          return {XOp::create(*builder_, loc_, args[0]).getOutputQubit(0)};
+        });
```
burgholzer: I have noticed this now in multiple cases: the usage for single-qubit controlled gates is really not ergonomic. The fact that a small vector has to be constructed here is unfortunate and cumbersome. This could be written way more compactly if the CtrlOp (and potentially other modifiers) would expose another builder that only works for single-qubit bodies and eliminates the indirection through the small vector.
This should be easy to add and make the code in various places much cleaner. Could you look into that in a separate PR?

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+    setWire(target, ctrlOp.getTargetsOut()[0]);
+  }
+
+  void cp(std::size_t control, std::size_t target, double theta) {
+    p(control, theta / 2.0);
+    cx(control, target);
+    p(target, -theta / 2.0);
+    cx(control, target);
+    p(target, theta / 2.0);
+  }
```
burgholzer: Technically, controlled phase gates could be left in the decomposition so that they are synthesized efficiently by the 2qb synthesis pass.
That sounds beneficial to me. 
If implemented, the docs of the pass also need to be updated.

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+    tdg(target);
+    cx(c0, target);
+    t(c1);
+    t(target);
+    h(target);
+    cx(c0, c1);
+    t(c0);
+    tdg(c1);
+    cx(c0, c1);
+  }
```
burgholzer: I am wondering whether it makes sense to leave this undecomposed as part of the decompositions and simply have either another pattern or another decomposition function as part of the pass that handles the two-control case.
It feels like there could still be some value in leaving the 3qb gates in the decomposition.

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+    h(target);
+    t(target);
+    cx(c1, target);
+    tdg(target);
+    cx(c0, target);
+    t(target);
+    cx(c1, target);
+    tdg(target);
+    h(target);
+  }
```
burgholzer: We can't really model this as a "proper" gate in our IR, can we?
RCCX gates are rather helpful and quite frequently used in these kind of decompositions.
I suppose we could introduce these as a new UnitaryOp. Would that be beneficial?

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+    p(3, K_PI8);
+    cx(2, 3);
+    p(3, -K_PI8);
+    cx(1, 3);
+    p(3, K_PI8);
+    cx(2, 3);
+    p(3, -K_PI8);
+    cx(0, 3);
+    h(3);
+  }
```
burgholzer: This is just a fixed decomposition for a three-controlled X gate, right?
Maybe this also fits well with the other suggestion of making this an explicit special case of the pattern instead of hard-coding it here.

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+  void setWire(std::size_t local, Value value) {
+    (*wires_)[wireIndex(local)] = value;
+  }
+
+  OpBuilder* builder_;
+  Location loc_;
+  SmallVector<Value>* wires_;
+  ArrayRef<std::size_t> remap_;
+};
```
burgholzer: These feel like generally useful helpers. Is there any good place we could put these? Are there other places where this could turn out to become useful (synthesis for example)?

It's totally fine to keep this here for now. I'd just like to avoid a situation where every pass invents its own helper routines.

Edit: this kind of feels like our program builder to some degree. Could it be generalized so that it is also genuinely useful here?

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+}
+
+static void addResetGadget(GateEmitter& builder, std::size_t q0, std::size_t q1,
+                           std::size_t q2) {
+  builder.cx(q1, q2);
+  builder.t(q2);
+  builder.cx(q0, q2);
+  builder.tdg(q2);
+  builder.h(q2);
+}
```
burgholzer: Does it make sense to model this as `inv(action)`? Now that we allow multiple unitaries in a modifier, this feels like quite a natural way to express this.

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+  const std::size_t num3 = numControls / 3;
+  const std::size_t num2 = (numControls - num3) / 2;
+  const std::size_t num1 = numControls - num3 - num2;
+  const std::size_t block2Begin = num1;
+  const std::size_t block3Begin = num1 + num2;
+  const std::size_t controlsEnd = numControls;
+
+  SmallVector<std::size_t, 16> map;
+  const auto relativeStep = [&](const double sign, const std::size_t begin,
```
burgholzer: Kind of feels wrong that `sign` is not a bool here.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -201,4 +201,46 @@ def HadamardLifting : Pass<"hadamard-lifting", "mlir::ModuleOp"> {
...
+    and `cx` gates.
+
+    By default (`min-controls=2`), gates with at least two controls (e.g. `CCX`,
+    `CCZ`, and larger `MCX`/`MCZ`) are decomposed, leaving single-control gates
+    (e.g. `CX`, `CZ`) and bare one-qubit gates intact. Increase `min-controls`
+    to keep smaller multi-controlled gates intact (e.g., when `CCX` is a native
+    gate of the target).
+  }];
+  let options = [Option<
+      "minControls", "min-controls", "uint64_t", "2",
```
burgholzer, line 286: IIRC tablegen also allows to directly place constraints on options/parameters. Maybe it is worth looking into whether this would simplify some of the error handling code throughout this PR.

burgholzer (changes requested):
Finally found some time to go through this.
Really cool to see this implemented and have it as part of our compiler collection!

I have a couple of comments here that are more questions on the general design. The actual implementation seems to be fine as confirmed by the tests.

I really like these kinds of tests because this also means that we can heavily rely on agentic AI to actually improve the implementation efficiency as long as the output is still correct 🙌

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_multi_controlled_decomposition.cpp
simon1hofmann: EXPECT_EQ checks structural equivalence right? In our case, we compare the decomposed mcx/mcz (`decomposedDD`) with an undecomposed reference mcx/mcz (`referenceDD`).

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
simon1hofmann: Seems reasonable. Should that also be applied to the QC dialect and the QC{O}ProgrammBuilder?

simon1hofmann: Created a first draft: #1869

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_multi_controlled_decomposition.cpp
burgholzer: Yeah, but since DDs are canonical, the respective pointers and weights should actually be equivalent (in the literal sense).
In many practical cases, this doesn't hold because of numerical inaccuracies that get in the way. Then, something like the computation that is in the code at the moment is necessary.

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
burgholzer: Yeah. I'd argue it should apply to all builders and dialect constructs 👍🏼 
Feel free to tag me for a review once the CI is green on that other PR.

mlir/include/mlir/Compiler/CompilerPipeline.h
simon1hofmann: Will be enabled once we have all puzzle pieces of the decomposition/synthesis pipeline 👍

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
simon1hofmann: Would be beneficial and I can add a short PR that introduces `RCCXOp`. That gate would only exist in QCO or also in QC with the respective transformations to and from QCO?

burgholzer: I suppose it makes sense to introduce the gate in both dialects as it also natively exists in Qiskit and we would potentially import it from there to QC. IIRC it also exists in OpenQASM's library, so we would potentially also read it from there and output it, if we ever support emitting OpenQASM 3.

simon1hofmann: Created a PR to introduce `RCCX`: #1886

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
simon1hofmann: I changed the structure of the pass such that 3- and 4-Qubit gates are only decomposed when needed.

mlir/lib/Dialect/QCO/Transforms/Decomposition/MultiControlled.cpp
simon1hofmann: Not really sure if it makes sense to add those to the program builder. If another decomposition pass needs the same pattern, I’d rather extract a small wire-bank/remap helper then.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
simon1hofmann: Looked into TableGen/option constraints. There’s no ODS-style IntMinValue for pass options; a custom llvm::cl::parser can enforce it at parse time, but it’s heavier than the current C++ checks and doesn’t cover the options-struct API. 
Seems like it's only possible for enums, e.g. https://github.com/llvm/llvm-project/blob/a255c1ed36a1d06f79bd2633ba9f8d900153007c/mlir/include/mlir/Conversion/Passes.td#L1222.

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_multi_controlled_decomposition.cpp
simon1hofmann: Generalizing the DD package to work with the MLIR code base sounds like a good follow-up, created an issue to track that idea: #1913

mlir/include/mlir/Dialect/QCO/Transforms/Decomposition/MultiControlled.h
simon1hofmann: Agreed on the dedicated header, those helpers aren’t a public API. I’ve folded them into the pass.

On decoupling from MLIR: the HP24 routines are easier to port/verify as index-based sequences, so GateEmitter is a thin adapter that turns those into qco ops via OpBuilder. I don’t think a more “MLIR-native” rewrite of the core would buy much right now (same ops, more boilerplate), but I’m open if somebody sees a cleaner pattern we should follow for similar passes.

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

burgholzer:
🤖 *AI text below* 🤖

### Licensing follow-up

I compared the implementation with the upstream sources named in the code. Although the underlying decomposition algorithms and methods described by the papers are not themselves protected as source-code expression, the current implementation is not merely based on the publications:

- the HP24 helpers closely follow Qiskit's [`mcx.rs`](https://github.com/Qiskit/qiskit/blob/884e39ed8222b65fd66a2fe5b1b46ae92ff8b881/crates/synthesis/src/multi_controlled/mcx.rs), including its helper split, thresholds, loop structure, wire mappings, and gate ordering;
- the phase decomposition specializes Qiskit's [`_mcsu2_real_diagonal` and `MCPhaseGate._define`](https://github.com/Qiskit/qiskit/blob/884e39ed8222b65fd66a2fe5b1b46ae92ff8b881/qiskit/synthesis/multi_controlled/multi_control_rotation_gates.py); and
- the SP22 step is a direct translation of qclib's [`Ldmcu._c1c2`](https://github.com/qclib/qclib/blob/cf069817b0ea226aec93b59a1c4307f750673ef4/qclib/gates/ldmcu.py).

Consequently, replacing the source attribution with publication citations would not be sufficient for this version. The publication references should be added for academic provenance, but the relevant Apache-2.0 source obligations remain.

The following steps are needed to make the distribution properly compliant:

1. Retain the relevant Qiskit/qclib copyright and attribution notices in the affected source, together with a clear notice that MQT modified the original code.
2. Ship a complete copy of the Apache License 2.0 with source and binary/package distributions. The current Qiskit comment points to `LICENSE.txt`, but that file does not exist in MQT Core.
3. Correct the file-level and package-level licensing metadata so the affected code is not presented as MIT-only. In particular, decide on and consistently apply an SPDX expression/file policy for the combined work, and include the Apache license/third-party notice files through `pyproject.toml`'s `license-files` configuration and other distribution manifests.
4. Add a third-party notice that identifies the affected implementation sections, upstream projects, pinned source revisions/URLs, and Apache-2.0 license. Neither checked upstream repository has a root `NOTICE` file at the revisions reviewed, but all relevant source notices still need to be retained.
5. Add the HP24, Vale et al., and da Silva/Park publication references in the documentation/comments as supplementary scholarly attribution.

An alternative is a genuinely independent, paper-based reimplementation that does not carry over protectable source expression. That would need to be treated as a real clean-room rewrite rather than a cosmetic restructuring of the current translation.

This is a source-provenance and license-compliance assessment, not legal advice.

burgholzer:
@simon1hofmann I had a lengthy discussion with GPT-5.6 on the topic of licensing and proper attribution. The above is the consequence of that. I also asked it to evaluate the potential impact and effort of a clean-room rewrite of these methods. With the respective tools at our disposal, I would guess this is doable fairly efficiently and is likely worth considering over the added complexity redistributing the proper license attribution for the very-close-to-source implementation.

simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

burgholzer (approved):
Just gave this another quick read after the fixes that have already been pushed. Let's not hold this up for longer and get it in 🚀 Great stuff!

PR #1822, opened 2026-06-29
✨ Implement controlled unitary embedding

### LINKED ISSUE #1821: ✨ Correct register-aware unitary embedding for `CtrlOp::getUnitaryMatrix`
### Problem Statement

`CtrlOp::getUnitaryMatrix()` currently embeds the inner gate unitary via `setBottomRightCorner`, which implements `I_{2^c} ⊗ U`. That only corresponds to the intended controlled operator when control qubits sit on **more significant** register positions than the targets (e.g. `cx(1, 0)` in little-endian indexing).

For the common case `cx(0, 1)` — control on qubit 0, target on qubit 1 — the returned matrix does not match `QuantumComputation` with the same register indices. Operand order in the `ctrl` op is not reflected correctly; `cx(0, 1)` and `cx(1, 0)` are not distinguished as they should be.

This limits reliable use of `getUnitaryMatrix()` for controlled operations in optimizations, verification, and cross-checks against `QuantumComputation` or Qiskit. An existing CX test inadvertently compared `cx(q[0], q[1])` against `comp.cx(1, 0)` rather than `comp.cx(0, 1)`, hiding the mismatch.

Example:

```mlir
// QCO: control q[0], target q[1]
ctrl(%c0, %t1) { cx ... }

// Should match qc::QuantumComputation::cx(0, 1), not cx(1, 0)
```

### Proposed Solution

Introduce **register-aware controlled embedding** for `CtrlOp::getUnitaryMatrix()`.

simon1hofmann:
## Description

Fixes incorrect controlled-gate unitary matrices in QCO and unifies matrix wire indexing on the `QuantumComputation` convention.

Problem: `CtrlOp::getUnitaryMatrix()` embedded the inner unitary via `DynamicMatrix::setBottomRightCorner (I_{2^c} ⊗ U)`, which is only correct when controls sit on the most significant wires. For general wire placement (e.g. `cx(q[0], q[1])`), the result did not match `QuantumComputation`.

Solution:
- Add `embedControlledUnitary()` to place a multi-controlled unitary using actual control/target register indices (QC little-endian: qubit `i` = bit `i`).
- Resolve compile-time wire indices in CtrlOp from `qco.static` and constant `qtensor.extract` operands.
- Unify `embedInNqubit`, `kron`, and related helpers on the same bit ordering (replacing the old MSB-first embedding).
- Remove unused `DynamicMatrix::setBottomRightCorner` API.

Fixes #1821 

## Checklist


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

burgholzer:
Just gave this some thought as discussed offline today.
My conclusion is that this is not the way to go. In its current form this will not really work as the inference of the qubit index doesn't really work in my opinion.
Since qubits in QCO are linear types in all but very select cases, their defining operation will be another quantum operation and not a static qubit op or an extract.
As such, you would have to traverse the entire qubit value chain back to such an operation to potentially (!) infer the index. That's a big no-go complexity-wise.

As a result, I believe the other direction we discussed is a little more promising. That is, the methods of the individual operations just report a matrix representation in a canonical order. That's order is the one where the the control qubits are the MSQ's and the target qubits are the LSQ's, where we typically use big endian notation to write down bitstrings (q_n-1 ... q0).
It is then the responsibility of synthesis passes to construct and combine the functionality of matrices appropriately/in the right order.

The DD package always considers global/cirxuit-wide qubit indices; so it makes a distinction between cx(0,1) and cx(1,0) when constructing the DDs for these operations. In mqt-cc, the matrix would always be the cx(1,0) one. This also fits well with how Qiskit handles this: https://github.com/Qiskit/qiskit/blob/9ff63334c989c8c6e0ddd303a224ddbd907f0b6e/crates/circuit/src/gate_matrix.rs#L87

simon1hofmann:
> Just gave this some thought as discussed offline today. My conclusion is that this is not the way to go. In its current form this will not really work as the inference of the qubit index doesn't really work in my opinion. Since qubits in QCO are linear types in all but very select cases, their defining operation will be another quantum operation and not a static qubit op or an extract. As such, you would have to traverse the entire qubit value chain back to such an operation to potentially (!) infer the index. That's a big no-go complexity-wise.
> 
> As a result, I believe the other direction we discussed is a little more promising. That is, the methods of the individual operations just report a matrix representation in a canonical order. That's order is the one where the the control qubits are the MSQ's and the target qubits are the LSQ's, where we typically use big endian notation to write down bitstrings (q_n-1 ... q0). It is then the responsibility of synthesis passes to construct and combine the functionality of matrices appropriately/in the right order.
> 
> The DD package always considers global/cirxuit-wide qubit indices; so it makes a distinction between cx(0,1) and cx(1,0) when constructing the DDs for these operations. In mqt-cc, the matrix would always be the cx(1,0) one. This also fits well with how Qiskit handles this: https://github.com/Qiskit/qiskit/blob/9ff63334c989c8c6e0ddd303a224ddbd907f0b6e/crates/circuit/src/gate_matrix.rs#L87

I also believe that is the way to go, closing this PR and the related Issue #1821.

PR #1824, opened 2026-06-30
✨ Add helper functions to `qco.if`

### LINKED ISSUE #1746: ✨ Add helper functions to `qco.if`
## Description

MLIR's built-in SCF operations have useful operations to get inputs from outputs and vice versa. For example, the `LoopLikeInterface` implements the methods `getTiedLoopInit` and `getTiedLoopResult`. It would be quite useful to have something similar for the custom `qco.if` operations. 

The implementation-logic could simply be adapted from the source code of the above methods.

MatthiasReumann:
## Description

Inspired by [MLIR](https://github.com/turing-db/llvm/blob/main/mlir/include/mlir/Interfaces/LoopLikeInterface.td), this pull request adds a few helpful utilities to the `qco.if` operation. Resolves #1746.

## Checklist


MatthiasReumann:
@coderabbitai full review

MatthiasReumann:
@burgholzer 

FYI: The codecov check is failing because of failures I can't enforce.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1266,13 +1266,80 @@ def IfOp
...
+          return arguments[std::distance(qubits.begin(), it)];
+        }
+
+        /// Return the yielded value that corresponds to the given argument
+        /// for the then-block, or `nullptr` on failure.
+        OpOperand* getTiedThenYieldedValue(BlockArgument bbArg);
+
+        /// Return the yielded value that corresponds to the given argument
+        /// for the else-block, or `nullptr` on failure.
+        OpOperand* getTiedElseYieldedValue(BlockArgument bbArg);
```
burgholzer, line 1321: Why aren't more of these split between declaration and definition? AFAICT none of these methods use templates so their defintion could be moved to the source file.

Follow-up question to that: would it pay off to put these into the actual IfOp source file so that one can benefit from syntax highlighting and linting?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1266,13 +1266,80 @@ def IfOp
...
+          auto results = getResults();
+          if (results.empty()) {
+            return {};
+          }
+          auto qubits = getQubitsMutable();
+          auto it = ::llvm::find(qubits, *qubit);
+          if (it == qubits.end()) {
+            return {};
+          }
+          return results[std::distance(qubits.begin(), it)];
```
burgholzer: This looks overly complicated given how there is https://mlir.llvm.org/doxygen/classmlir_1_1OpOperand.html#aaefd8f43cab0db90326616e741bee604 which allows you to directly get the OperandNumber. Wouldn't that work?

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1266,13 +1266,80 @@ def IfOp
...
+          auto results = getResults();
+          if (results.empty()) {
+            return nullptr;
+          }
+          auto it = ::llvm::find(results, result);
+          if (it == results.end()) {
+            return nullptr;
+          }
+          return &getQubitsMutable()[std::distance(results.begin(), it)];
+        }
```
burgholzer: Same here with https://mlir.llvm.org/doxygen/classmlir_1_1OpResult.html#ad8714eb4af7b6172d88a25bd6442b00b

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1266,13 +1266,80 @@ def IfOp
...
+          if (arguments.empty()) {
+            return {};
+          }
+          auto qubits = getQubitsMutable();
+          auto it = ::llvm::find(qubits, *qubit);
+          if (it == qubits.end()) {
+            return {};
+          }
+          return arguments[std::distance(qubits.begin(), it)];
+        }
```
burgholzer: Same potential simplification applies here.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
```diff
@@ -1282,6 +1349,32 @@ def IfOp
...
+          if (arguments.empty()) {
+            return nullptr;
+          }
+          auto it = ::llvm::find(arguments, bbArg);
+          if (it == arguments.end()) {
+            return nullptr;
+          }
+          auto yield = elseYield();
+          return &yield.getTargetsMutable()[std::distance(arguments.begin(), it)];
+        }
```
burgholzer: And here via https://mlir.llvm.org/doxygen/classmlir_1_1BlockArgument.html#a5396ce59c00cd3ef7a8a500c59af295c

burgholzer (changes requested):
Really like the direction! It feels like there is some room for further improvement here.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
MatthiasReumann: Good point. I was thinking about it, but didn't do it. I think moving these to the `.cpp` file sounds reasonable.

mlir/include/mlir/Dialect/QCO/IR/QCOOps.td
MatthiasReumann: I still remember that I encountered some issues a while back when I was using the operand numbers. Nonetheless, will give it a try 👍🏻

MatthiasReumann: Although, I am then wondering why the MLIR codebase doesn't implement it this way 🤔

MatthiasReumann: ~~Ahh. Got it. There is no way of telling if the provided `OpOperand` is actually belonging to the operation. If we simply use the operand number, we might use invalid SSA values to retrieve the results (for example) without failure. To verify if this is the case one always has to do a linear search - which is essentially the current implementation.~~

~~Any thoughts on this?~~

Edit: You can simply check the operation pointer...

MatthiasReumann:
@coderabbitai full review

MatthiasReumann:
@burgholzer 

Applied the request changes. Thanks for the feedback; much appreciated! 

I am not sure what to do about the codecov. Not sure if adding unit tests adds a lot of benefit in this particular case. Any thoughts?

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -280,3 +280,48 @@ LogicalResult IfOp::verify() {
...
   return success();
 }
+
+OpResult IfOp::getTiedResult(OpOperand* qubit) {
+  if (qubit->getOwner() != getOperation()) {
+    return {};
+  }
+  // Because the first operand is the if-condition, subtract one.
+  return getResults()[qubit->getOperandNumber() - 1];
+}
```
burgholzer, line 290: `qubit` must not be null here, so this should likely just be an `OpOperand&` or `OpOperand` (don't know at the top of my head how "heavy" a copy of an OpOperand is).

Edit: Applies below as well.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -280,3 +280,48 @@ LogicalResult IfOp::verify() {
...
+
+OpOperand* IfOp::getTiedQubit(OpResult result) {
+  if (result.getDefiningOp() != getOperation()) {
+    return nullptr;
+  }
+  return &getQubitsMutable()[result.getResultNumber()];
+}
```
burgholzer, line 297: Does the result here really need to be a pointer? Can't we avoid that level of indirection?

burgholzer (changes requested):
One more round of potential refinements. Then this should be good to go.
I wouldn't mind the codecov check here.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
MatthiasReumann: Using a `OpOperand*` was inspired by the MLIR implementation. 

The only real alternative instead of an `OpOperand*` would be to use an `OpOperand&`. Then we must either: 
- assume (and trust the caller) that the `result` is in fact belonging to the current `IfOp`
- or report a fatal error.  

I really dislike the first option as it will probably cause us some headaches down the line.  

I haven't seen a function return a copy of `OpOperand`s, yet. `Value`s are usually copied. I would assume for "reporting errors; but failing gracefully"[^1] they simply use a pointer. Plus, semantically an `OpOperand` is an operand of an operation, hence, using a pointer to point at that value somewhat makes sense as it "belongs" to the operation. 

I don't have a strong opinion on that, but I rather prefer aligning with the "gold standard" (in some regards; others maybe not) MLIR implementation. Maybe you could throw in an argument for consistency also. But I'll let you decide as I am somewhat agnostic. 

[^1]: The Duck.ai GPT agrees.

burgholzer: Fair arguments. I think I was just wondering whether one could actually return an "empty" OpOperand somehow.
But if MLIR itself is also doing it this way, I am all for following that 👍🏼

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer: Based on the discussion below, this is also fine in principle. I'll leave it up to you to adapt this or not.

MatthiasReumann: Okay! Then I would say we keep the current implementation and adapt it, if any issues should arise.

burgholzer (approved):
Alright, then let's resolve the conversations, resolve the conflicts and merge 🚀

PR #1835, opened 2026-07-01
🔧 Maintenance round

denialhaag:
## Description

This PR backports (hopefully) all maintenance-related changes. For more information, please refer to the cherry-picked commits.

## Checklist


denialhaag:
@burgholzer, I'm thinking of adding a `target-branch` argument to the templating action so that we can simply ignore these files here. Technically speaking, https://github.com/munich-quantum-toolkit/core/pull/1835/commits/699d3d83a3ea469e663d61d52261dd10690c1e97 is incorrect, and we should still be using the `c++-python` project type here. However, `c++-mlir-python` is more consistent with the current state of the rendered templates (because I didn't want to manually adapt them during the rebase). It would be nice if the templating action handled this. What do you think? 🤔

denialhaag:
I went through everything again, and I think this should be good to go in. 😌

burgholzer:
> @burgholzer, I'm thinking of adding a `target-branch` argument to the templating action so that we can simply ignore these files here. Technically speaking, [699d3d8](https://github.com/munich-quantum-toolkit/core/commit/699d3d83a3ea469e663d61d52261dd10690c1e97) is incorrect, and we should still be using the `c++-python` project type here. However, `c++-mlir-python` is more consistent with the current state of the rendered templates (because I didn't want to manually adapt them during the rebase). It would be nice if the templating action handled this. What do you think? 🤔

That feels reasonable 👍🏼 Feel free to make a concrete proposal for that 😌

.devcontainer/Dockerfile
burgholzer, line 1: That container will not actually work, will it? The MLIR code base in the v3 series is not yet compatible with 22.x
One could potentially simply drop the devcontainer here.

pyproject.toml
```diff
@@ -8,7 +8,7 @@
 
 [build-system]
 requires = [
-  "nanobind~=2.12.0",
+  "nanobind~=2.13.0",
```
burgholzer, line 11: Given this change, this definitely needs to be a minor version bump when we release.

burgholzer (changes requested):
Mostly looks good! Thanks a lot! 🙏🏼 
Only one small detail and a comment.

pyproject.toml
denialhaag: Good point! I've added the label. 😌

.devcontainer/Dockerfile
denialhaag: I've reverted the respective commit. 👍

burgholzer (approved):
LGTM! Nice 😌 Might be worth getting a release out soonish to benefit from the nanobind updates?

denialhaag:
> Might be worth getting a release out soonish to benefit from the nanobind updates?

I'd say so! I can take care of this in the coming days. 😌

burgholzer:
> > Might be worth getting a release out soonish to benefit from the nanobind updates?
> 
> I'd say so! I can take care of this in the coming days. 😌

Nice! Not the most pressing issue, but still nice!

PR #1849, opened 2026-07-03
♻️ Refactor FoMaC classes

MatthiasReumann:
## Description

Proposed FoMaC rewrite: 

- Use `unique_ptr` instead of custom destructor, copy, and move semantics.
- Remove trailing return types as the return type does not depend on the arguments. 
- Remove "Tokens". Use private constructors and friendship semantics. 
- Consistency improvements (`std::size_t` vs `size_t`, etc.)

## Checklist


include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -175,560 +174,609 @@ struct SessionConfig {
...
-     */
-    [[nodiscard]] auto getSparseProbabilities() const
-        -> std::map<std::string, double>;
-  };
+  [[nodiscard]] static Device fromQDMIDevice(QDMI_Device device) {
+    return Device(device);
+  }
+
+  /// @returns the underlying QDMI_Device object.
+  [[nodiscard]] QDMI_Device getQDMIDevice() const { return device_; }
```
burgholzer: Would `getHandle` potentially be a better name? one could also generically call these private members `handle_` because that's what they essentially are. The above method would/could then also become `fromHandle`.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -175,560 +174,609 @@ struct SessionConfig {
...
+  [[nodiscard]] QDMI_Device_Status getStatus() const;
+
+  /// @see QDMI_DEVICE_PROPERTY_LIBRARYVERSION
+  [[nodiscard]] std::string getLibraryVersion() const;
+
+  /// @see QDMI_DEVICE_PROPERTY_QUBITSNUM
+  [[nodiscard]] size_t getQubitsNum() const;
+
+  /// @see QDMI_DEVICE_PROPERTY_SITES
+  [[nodiscard]] std::vector<Site> getSites() const;
```
burgholzer, line 267: This has come up elsewhere as kind of urgent and I thought I'd bring it up here: 
Do we see any kind of way to expose the `CUSTOM` properties that QDMI devices may implement?
Maybe also a question for @ystade in general.
The big "problem" is that we do not necessarily know the type of these properties up front.
I haven't thought too deeply about that, but we could at least provide `(u)int[X], float, double, string` getters that would return `nullopt` if the reported size does not match to the datatype.
Maybe this is also something for a follow-up, but an immediate one, if so.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -175,560 +174,609 @@ struct SessionConfig {
...
+ * @brief Class representing a submitted job.
+ * @details
+ * This class provides methods to query job status and retrieve
+ * results.
+ *
+ * The class can only be constructed by Device instances.
+ *
+ * @see QDMI_Job
+ */
+class Job {
```
burgholzer, line 408: WHy does this class not implement the starship operator (`<=>`)?

burgholzer (changes requested):
This generally looks good to me and we can happily move forward with this, if you @MatthiasReumann feel confident in the changes.

I have two very small comments and a medium one that we could also tackle after this PR to not stall it for too long.

In any case, this still needs labels and a changelog entry.
We should also backport these changes to the `v3.x` series so that the changes can go into the next release there.

@denialhaag for awareness.

include/mqt-core/fomac/FoMaC.hpp
MatthiasReumann: Yes, I like that 👍🏻 

Because you mention `fromHandle` (or `fromQDMIDevice`): Could you maybe elaborate a bit on what the rationale of the `Token`s was? Because as far as I can tell `fromQDMIDevice` short-circuits these guardrails anyway (only to make the Python bindings work...). 
A question that I've asked myself is: Would it be that bad to be able to construct a "Site" class from a site QDMI pointer?

include/mqt-core/fomac/FoMaC.hpp
MatthiasReumann: Good question. The current implementation in `main` does also not have one.

include/mqt-core/fomac/FoMaC.hpp
MatthiasReumann: ~~This is a good question. Given that the caller must know the type of the results of a specific device to use it, couldn't we do templating:~~

```cpp
template <class T0 = void, class T1 = void>
class Device {
  T0 getCustomProperty0() { ... }
  T1 getCustomProperty1() { ... }
  // Rest of the device implementation.
};
```

~~Not sure if this even compiles, but the idea generally?~~

Edit: Ah no. How would the session which queries the devices know these types...

include/mqt-core/fomac/FoMaC.hpp
MatthiasReumann: Added.

include/mqt-core/fomac/FoMaC.hpp
ystade: @MatthiasReumann What you are pointing out is an issue that while compiling the Session/FoMaC (whatever it is called) the types are not known, only after loading the QDMI device library at runtime, correct? Hence, the tamplate arguments must be resolved during compile time this does not work. Is this the right explanation?

ystade: Towards @burgholzer, the only option I currently see to make the custom properties/parameters accessible is to expose via `std::byte` buffers aka `std::array<std::byte, N>` or `std::vector<std::byte>` whose length are given by the caller (since hopefully at least the caller knows the type and hence the required size of the buffer).

include/mqt-core/fomac/FoMaC.hpp
MatthiasReumann: Interestingly, I just re-googled the definition of "handle". Apparently, a "handle" is the object holding the raw pointer (basically our FoMaC classes!). So `ptr_` and `fromPointer` would probably be semantically even more correct. Analogously to `unique_ptr` the method `get` could then return the underlying pointer. 

Let me know what you think!

ystade: > Yes, I like that 👍🏻
> 
> Because you mention `fromHandle` (or `fromQDMIDevice`): Could you maybe elaborate a bit on what the rationale of the `Token`s was? Because as far as I can tell `fromQDMIDevice` short-circuits these guardrails anyway (only to make the Python bindings work...). A question that I've asked myself is: Would it be that bad to be able to construct a "Site" class from a site QDMI pointer?

I am on that recovering my thoughts about the `Token`s. I do remember that I also considered `friend class`es but deliberately decided against them. Give me some minutes to recover my rationale there.

include/mqt-core/fomac/FoMaC.hpp
MatthiasReumann: > What you are pointing out is an issue that while compiling the Session/FoMaC (whatever it is called) the types are not known, only after loading the QDMI device library at runtime, correct?

Yes, exactly. There is no (clean) way of constructing devices correctly in `session.getDevices()`.

include/mqt-core/fomac/FoMaC.hpp
ystade: I'll answer in the main thread of this PR as I deem it as fundamental for this PR.

ystade:
@MatthiasReumann Continuing the discussion in https://github.com/munich-quantum-toolkit/core/pull/1849#discussion_r3534256061:

The initial motivation of the `FoMaC` implementation was to mirror the QDMI behavior on a C++ level. This meant not only to provide C++ wrappers around the C handles but also to mirror the access pattern, i.e., a`Site` can only be retrieved from a `Device` or an `Operation` or a `Device` can only be retrieved from an active `Session` with a `Driver`. As you, @MatthiasReumann correctly mentioned, `from QDMIdevice` short-circuits this intended behavior.

To implement the described behavior, I decided to use `Token`s instead of using `friend class`es to not expose all other private functions to the outside world and have a more fine-granular way to steer what can be accessed from outside the class.

The AI also mentioned that the Python bindings do not understand  friendship. However, I do not recall, how this influenced the decision but maybe a point to check/keep in mind.

include/mqt-core/fomac/FoMaC.hpp
burgholzer: > Towards @burgholzer, the only option I currently see to make the custom properties/parameters accessible is to expose via `std::byte` buffers aka `std::array<std::byte, N>` or `std::vector<std::byte>` whose length are given by the caller (since hopefully at least the caller knows the type and hence the required size of the buffer).

That is one potential solution.
However, I also believe the one I tried to outline above might also work (in addition to the byte fallback). We could offer getters for these properties like `get_custom1_str() -> std::optional<std::string>` that would return nullptr when the property is not supported and, otherwise, would interpret the data as a string. Similar for int, etc.

All of this makes me think that the concept of QInfo might have turned out to be rather helpful here. It would provide named and typed elements which are not limited to 5 per quantity. Might it be worth to revisit that?

burgholzer:
> @MatthiasReumann Continuing the discussion in https://github.com/munich-quantum-toolkit/core/pull/1849#discussion_r3534256061:
> 
> 
> 
> The initial motivation of the `FoMaC` implementation was to mirror the QDMI behavior on a C++ level. This meant not only to provide C++ wrappers around the C handles but also to mirror the access pattern, i.e., a`Site` can only be retrieved from a `Device` or an `Operation` or a `Device` can only be retrieved from an active `Session` with a `Driver`. As you, @MatthiasReumann correctly mentioned, `from QDMIdevice` short-circuits this intended behavior.
> 
> 
> 
> To implement the described behavior, I decided to use `Token`s instead of using `friend class`es to not expose all other private functions to the outside world and have a more fine-granular way to steer what can be accessed from outside the class.
> 
> 
> 
> The AI also mentioned that the Python bindings do not understand  friendship. However, I do not recall, how this influenced the decision but maybe a point to check/keep in mind.

The reason for the token solution was exactly to avoid friend semantics. I am (personally) not the biggest fan of them because it always feels like bad design to me.
However, it feels like the solution here works just fine 🙂 and it feels moderately clean.

Regarding the overall abstraction: I think we could allow ourselves some freedom and do not necessarily stick to the C design. However, one has to be careful that most of the opaque pointers (site, operation, job, etc.) are tied to a particular session with a device. 
So passing around the handles without the corresponding "owner" is dangerous.

include/mqt-core/fomac/FoMaC.hpp
ystade: > That is one potential solution. However, I also believe the one I tried to outline above might also work (in addition to the byte fallback). We could offer getters for these properties like `get_custom1_str() -> std::optional<std::string>` that would return nullptr when the property is not supported and, otherwise, would interpret the data as a string. Similar for int, etc.
> 
> All of this makes me think that the concept of QInfo might have turned out to be rather helpful here. It would provide named and typed elements which are not limited to 5 per quantity. Might it be worth to revisit that?

Uh, you are opening a big box here :D Actually, yes, this would have made sense based only on a very quick thought. I am asking myself how much we should still change v1, when we plan to do it entirely different in v2. Perhaps that's something for our discussion in Wednesday.

burgholzer: Yeah. I suppose this will be a big discussion point. It's not quite clear to me how to strike the best balance here.
The most important factor to me is how quickly we can move everything to v2.

MatthiasReumann:
> I am (personally) not the biggest fan of them because it always feels like bad design to me. 

I think there is only one way to use friendships correctly in C++: A friend class `A` constructs an object `B` with private constructor. 

> So passing around the handles without the corresponding "owner" is dangerous. 

This is what bugs me about the Python bindings using the `fromQDMIDevice` function.

burgholzer:
> > I am (personally) not the biggest fan of them because it always feels like bad design to me. 
> 
> 
> 
> I think there is only one way to use friendships correctly in C++: A friend class `A` constructs an object `B` with private constructor.


Yeah. This feels fine 👍🏻
> 
> 
> 
> > So passing around the handles without the corresponding "owner" is dangerous. 
> 
> 
> 
> This is what bugs me about the Python bindings using the `fromQDMIDevice` function. 

Yeah. There are a couple more things that bother me about certain ergonomics at the moment. Most importantly the way we inject external QDMI device libraries at runtime, which circumvents the QDMI client interface. But it was the only way to pass authentication information to the actual device.

MatthiasReumann:
@burgholzer @ystade 

Okay. Maybe let's summarize the discussion and action items up to this point to get this current PR merged. 

- ⚡️I'll think a bit about the QDMI C structure in C++ (or FoMaC classes). Particularly, the ownership-semantics and change the implementation, if necessary. 
- 💬 You discuss `QInfo`s. A follow-up pull request will then deal with the custom property getters.
- 💭 We'll leave the Python Bindings as they are for now. 

Sounds good?

burgholzer:
> @burgholzer @ystade 
> 
> 
> 
> Okay. Maybe let's summarize the discussion and action items up to this point to get this current PR merged. 
> 
> 
> 
> - ⚡️I'll think a bit about the QDMI C structure in C++ (or FoMaC classes). Particularly, the ownership-semantics and change the implementation, if necessary. 
> 
> - 💬 You discuss `QInfo`s. A follow-up pull request will then deal with the custom property getters.
> 
> - 💭 We'll leave the Python Bindings as they are for now. 
> 
> 
> 
> Sounds good?

Sounds good to me 👍🏻
Leaving both the bindings as well as the custom properties to follow up PRs is reasonable because these do not need to block the PR here.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -175,560 +174,611 @@ struct SessionConfig {
...
-    qdmi::throwIfError(result, msg);
-    remove_optional_t<T> value(
-        size / sizeof(typename remove_optional_t<T>::value_type));
-    result = QDMI_session_query_session_property(session_, prop, size,
-                                                 value.data(), nullptr);
-    qdmi::throwIfError(result, msg);
-    return value;
+    if constexpr (string_or_optional_string<T>) {
+      size_t size = 0;
+      auto result = QDMI_device_query_site_property(device_->get(), site_, prop,
```
burgholzer: Are these calls to `get()` actually necessary throughout this file?
Given the implicit conversion routines, it should be entirely possible to simply pass the class object without explicitly calling `get()`, right?
Feels a bit cleaner to me and actually makes use of the implicit converters.

MatthiasReumann: That is actually a pretty great idea.

MatthiasReumann:
@burgholzer 

This one should be reviewable also. 

---

Regarding ownership semantics: I think the only real alternative to the current implementation would be to store the QDMI data structures in the FoMaC classes and then construct the subclasses using const double pointers in the FoMaC methods (`getCouplingMap`). Not sure if I like this.

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
   qdmi::unreachable();
 }
 
-auto Session::Job::cancel() const -> void {
-  qdmi::throwIfError(QDMI_job_cancel(job_), "Cancelling job");
+void Job::cancel() const {
+  qdmi::throwIfError(QDMI_job_cancel(job_.get()), "Cancelling job");
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_cancel(job_), "Cancelling job");
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
 }
 
-auto Session::Job::getId() const -> std::string {
+std::string Job::getId() const {
   size_t size = 0;
-  qdmi::throwIfError(
-      QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_ID, 0, nullptr, &size),
-      "Querying job ID size");
+  qdmi::throwIfError(QDMI_job_query_property(job_.get(), QDMI_JOB_PROPERTY_ID,
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_ID,
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
+                                             0, nullptr, &size),
+                     "Querying job ID size");
   std::string id(size - 1, '\0');
-  qdmi::throwIfError(QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_ID, size,
-                                             id.data(), nullptr),
+  qdmi::throwIfError(QDMI_job_query_property(job_.get(), QDMI_JOB_PROPERTY_ID,
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_ID,
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
+                                             size, id.data(), nullptr),
                      "Querying job ID");
   return id;
 }
 
-auto Session::Job::getProgramFormat() const -> QDMI_Program_Format {
+QDMI_Program_Format Job::getProgramFormat() const {
   QDMI_Program_Format format{};
-  qdmi::throwIfError(QDMI_job_query_property(job_,
+  qdmi::throwIfError(QDMI_job_query_property(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_query_property(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
                                              sizeof(format), &format, nullptr),
                      "Querying program format");
   return format;
 }
 
-auto Session::Job::getProgram() const -> std::string {
+std::string Job::getProgram() const {
   size_t size = 0;
-  qdmi::throwIfError(QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_PROGRAM, 0,
+  qdmi::throwIfError(QDMI_job_query_property(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_query_property(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
+                                             QDMI_JOB_PROPERTY_PROGRAM, 0,
                                              nullptr, &size),
                      "Querying program size");
 
   std::string program(size - 1, '\0');
-  qdmi::throwIfError(QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_PROGRAM,
-                                             size, program.data(), nullptr),
+  qdmi::throwIfError(QDMI_job_query_property(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_query_property(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -305,14 +314,15 @@ auto Session::Device::submitJob(const std::string& program,
...
-  qdmi::throwIfError(QDMI_job_check(job_, &status), "Checking job status");
+  qdmi::throwIfError(QDMI_job_check(job_.get(), &status),
+                     "Checking job status");
   return status;
 }
 
-auto Session::Job::wait(const size_t timeout) const -> bool {
-  const auto ret = QDMI_job_wait(job_, timeout);
+bool Job::wait(const size_t timeout) const {
+  const auto ret = QDMI_job_wait(job_.get(), timeout);
```
burgholzer:
```suggestion
  const auto ret = QDMI_job_wait(job_, timeout);
```

src/fomac/FoMaC.cpp
```diff
@@ -305,14 +314,15 @@ auto Session::Device::submitJob(const std::string& program,
   return jobWrapper;
 }
 
-auto Session::Job::check() const -> QDMI_Job_Status {
+QDMI_Job_Status Job::check() const {
   QDMI_Job_Status status{};
-  qdmi::throwIfError(QDMI_job_check(job_, &status), "Checking job status");
+  qdmi::throwIfError(QDMI_job_check(job_.get(), &status),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_check(job_, &status),
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
 
-auto Session::Job::getNumShots() const -> size_t {
+size_t Job::getNumShots() const {
   size_t numShots = 0;
-  qdmi::throwIfError(QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_SHOTSNUM,
-                                             sizeof(numShots), &numShots,
-                                             nullptr),
-                     "Querying number of shots");
+  qdmi::throwIfError(
+      QDMI_job_query_property(job_.get(), QDMI_JOB_PROPERTY_SHOTSNUM,
```
burgholzer:
```suggestion
      QDMI_job_query_property(job_, QDMI_JOB_PROPERTY_SHOTSNUM,
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
   return numShots;
 }
 
-auto Session::Job::getShots() const -> std::vector<std::string> {
+std::vector<std::string> Job::getShots() const {
   size_t shotsSize = 0;
-  qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_SHOTS, 0, nullptr, &shotsSize),
-      "Querying shots size");
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(), QDMI_JOB_RESULT_SHOTS, 0,
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_SHOTS, 0,
```

src/fomac/FoMaC.cpp
```diff
@@ -323,65 +333,67 @@ auto Session::Job::wait(const size_t timeout) const -> bool {
...
+                                          nullptr, &shotsSize),
+                     "Querying shots size");
 
   if (shotsSize == 0) {
     return {};
   }
 
   std::string shots(shotsSize - 1, '\0');
-  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_SHOTS,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(), QDMI_JOB_RESULT_SHOTS,
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_SHOTS,
```

src/fomac/FoMaC.cpp
```diff
@@ -401,25 +413,26 @@ auto Session::Job::getShots() const -> std::vector<std::string> {
   return shotsVec;
 }
 
-auto Session::Job::getCounts() const -> std::map<std::string, size_t> {
+std::map<std::string, size_t> Job::getCounts() const {
   // Get the histogram keys
   size_t keysSize = 0;
-  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_HIST_KEYS, 0,
-                                          nullptr, &keysSize),
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(), QDMI_JOB_RESULT_HIST_KEYS,
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_HIST_KEYS,
```

src/fomac/FoMaC.cpp
```diff
@@ -401,25 +413,26 @@ auto Session::Job::getShots() const -> std::vector<std::string> {
...
+                                          0, nullptr, &keysSize),
                      "Querying histogram keys size");
 
   if (keysSize == 0) {
     return {}; // Empty histogram
   }
 
   std::string keys(keysSize - 1, '\0');
-  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_HIST_KEYS,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(), QDMI_JOB_RESULT_HIST_KEYS,
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_HIST_KEYS,
```

src/fomac/FoMaC.cpp
```diff
@@ -401,25 +413,26 @@ auto Session::Job::getShots() const -> std::vector<std::string> {
...
                                           keysSize, keys.data(), nullptr),
                      "Querying histogram keys");
 
   // Get the histogram values
   size_t valuesSize = 0;
-  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_HIST_VALUES, 0,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -429,7 +442,8 @@ auto Session::Job::getCounts() const -> std::map<std::string, size_t> {
   }
 
   std::vector<size_t> values(valuesSize / sizeof(size_t));
-  qdmi::throwIfError(QDMI_job_get_results(job_, QDMI_JOB_RESULT_HIST_VALUES,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -452,10 +466,9 @@ auto Session::Job::getCounts() const -> std::map<std::string, size_t> {
   return counts;
 }
 
-auto Session::Job::getDenseStateVector() const
-    -> std::vector<std::complex<double>> {
+std::vector<std::complex<double>> Job::getDenseStateVector() const {
   size_t size = 0;
-  qdmi::throwIfError(QDMI_job_get_results(job_,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -467,16 +480,16 @@ auto Session::Job::getDenseStateVector() const
 
   std::vector<std::complex<double>> stateVector(size /
                                                 sizeof(std::complex<double>));
-  qdmi::throwIfError(QDMI_job_get_results(job_,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -467,16 +480,16 @@ auto Session::Job::getDenseStateVector() const
...
                                           size, stateVector.data(), nullptr),
                      "Querying dense state vector");
   return stateVector;
 }
 
-auto Session::Job::getDenseProbabilities() const -> std::vector<double> {
+std::vector<double> Job::getDenseProbabilities() const {
   size_t size = 0;
-  qdmi::throwIfError(QDMI_job_get_results(job_,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -487,19 +500,18 @@ auto Session::Job::getDenseProbabilities() const -> std::vector<double> {
   }
 
   std::vector<double> probabilities(size / sizeof(double));
-  qdmi::throwIfError(QDMI_job_get_results(job_,
+  qdmi::throwIfError(QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
  qdmi::throwIfError(QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -487,19 +500,18 @@ auto Session::Job::getDenseProbabilities() const -> std::vector<double> {
...
 }
 
-auto Session::Job::getSparseStateVector() const
-    -> std::map<std::string, std::complex<double>> {
+std::map<std::string, std::complex<double>> Job::getSparseStateVector() const {
   size_t keysSize = 0;
   qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS, 0,
-                           nullptr, &keysSize),
+      QDMI_job_get_results(job_.get(), QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS,
```
burgholzer:
```suggestion
      QDMI_job_get_results(job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS,
```

src/fomac/FoMaC.cpp
```diff
@@ -508,15 +520,15 @@ auto Session::Job::getSparseStateVector() const
 
   std::string keys(keysSize - 1, '\0');
   qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS,
+      QDMI_job_get_results(job_.get(), QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS,
```
burgholzer:
```suggestion
      QDMI_job_get_results(job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS,
```

src/fomac/FoMaC.cpp
```diff
@@ -508,15 +520,15 @@ auto Session::Job::getSparseStateVector() const
...
                            keysSize, keys.data(), nullptr),
       "Querying sparse state vector keys");
 
   size_t valuesSize = 0;
-  qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES, 0,
-                           nullptr, &valuesSize),
-      "Querying sparse state vector values size");
+  qdmi::throwIfError(QDMI_job_get_results(
+                         job_.get(), QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES,
```
burgholzer:
```suggestion
                         job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES,
```

src/fomac/FoMaC.cpp
```diff
@@ -526,10 +538,10 @@ auto Session::Job::getSparseStateVector() const
 
   std::vector<std::complex<double>> values(valuesSize /
                                            sizeof(std::complex<double>));
-  qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES,
-                           valuesSize, values.data(), nullptr),
-      "Querying sparse state vector values");
+  qdmi::throwIfError(QDMI_job_get_results(
+                         job_.get(), QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES,
```
burgholzer:
```suggestion
                         job_, QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES,
```

src/fomac/FoMaC.cpp
```diff
@@ -550,27 +562,27 @@ auto Session::Job::getSparseStateVector() const
...
-auto Session::Job::getSparseProbabilities() const
-    -> std::map<std::string, double> {
+std::map<std::string, double> Job::getSparseProbabilities() const {
   size_t keysSize = 0;
-  qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS, 0,
-                           nullptr, &keysSize),
-      "Querying sparse probabilities keys size");
+  qdmi::throwIfError(QDMI_job_get_results(
+                         job_.get(), QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS,
```
burgholzer:
```suggestion
                         job_, QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS,
```

src/fomac/FoMaC.cpp
```diff
@@ -550,27 +562,27 @@ auto Session::Job::getSparseStateVector() const
...
     return {}; // Empty probabilities
   }
 
   std::string keys(keysSize - 1, '\0');
-  qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS,
-                           keysSize, keys.data(), nullptr),
-      "Querying sparse probabilities keys");
+  qdmi::throwIfError(QDMI_job_get_results(
+                         job_.get(), QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS,
```
burgholzer:
```suggestion
                         job_, QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS,
```

src/fomac/FoMaC.cpp
```diff
@@ -550,27 +562,27 @@ auto Session::Job::getSparseStateVector() const
...
+                         keysSize, keys.data(), nullptr),
+                     "Querying sparse probabilities keys");
 
   size_t valuesSize = 0;
   qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_PROBABILITIES_SPARSE_VALUES, 0,
+      QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
      QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -581,7 +593,8 @@ auto Session::Job::getSparseProbabilities() const
 
   std::vector<double> values(valuesSize / sizeof(double));
   qdmi::throwIfError(
-      QDMI_job_get_results(job_, QDMI_JOB_RESULT_PROBABILITIES_SPARSE_VALUES,
+      QDMI_job_get_results(job_.get(),
```
burgholzer:
```suggestion
      QDMI_job_get_results(job_,
```

src/fomac/FoMaC.cpp
```diff
@@ -603,23 +616,25 @@ auto Session::Job::getSparseProbabilities() const
...
+        session, QDMI_session_free);
+  }();
+
   // Helper to set session parameters
   const auto setParameter = [this](const std::optional<std::string>& value,
                                    QDMI_Session_Parameter param) -> void {
     if (value) {
       const auto status = static_cast<QDMI_STATUS>(QDMI_session_set_parameter(
-          session_, param, value->size() + 1, value->c_str()));
+          session_.get(), param, value->size() + 1, value->c_str()));
```
burgholzer:
```suggestion
          session_, param, value->size() + 1, value->c_str()));
```

src/fomac/FoMaC.cpp
```diff
@@ -690,43 +705,21 @@ Session::Session(const SessionConfig& config) {
     setParameter(config.custom5, QDMI_SESSION_PARAMETER_CUSTOM5);
 
     // Initialize the session
-    qdmi::throwIfError(QDMI_session_init(session_), "Initializing session");
+    qdmi::throwIfError(QDMI_session_init(session_.get()),
```
burgholzer:
```suggestion
    qdmi::throwIfError(QDMI_session_init(session_),
```

src/fomac/FoMaC.cpp
```diff
@@ -690,43 +705,21 @@ Session::Session(const SessionConfig& config) {
...
+                       "Initializing session");
   } catch (...) {
-    cleanup();
     throw;
   }
```
burgholzer: With the removal of the cleanup call, the try catch block has become pointless as it simply rethrows. I'd argue it can be removed, which also reduces the level of indentation.

burgholzer (approved):
Nice 👍🏼 don't get distracted by the flood of comments below. It's just local removals of further `.get()` calls. I'll quickly apply those in a batch after the review.
Then, I believe there is only one comment left before this can go in.

src/fomac/FoMaC.cpp
MatthiasReumann: Good point. Will do!

MatthiasReumann:
> It's just local removals of further .get() calls. 

I don't think this will compile unfortunately (edit: it didn't locally). The implicit conversion converts it to the object; not the pointer.

burgholzer:
> > It's just local removals of further .get() calls. 
> 
> 
> 
> I don't think this will compile unfortunately (edit: it didn't locally). The implicit conversion converts it to the object; not the pointer. 

🥲 that is unfortunate. I suppose there is no (easy) way to make that work?

MatthiasReumann:
> that is unfortunate. I suppose there is no (easy) way to make that work?

Don't think so no :/ Unfortunately [`.get`](https://en.cppreference.com/cpp/memory/unique_ptr) is the only way to access the managed pointer 🥲

burgholzer:
> > that is unfortunate. I suppose there is no (easy) way to make that work?
> 
> 
> 
> Don't think so no :/ Unfortunately [`.get`](https://en.cppreference.com/cpp/memory/unique_ptr) is the only way to access the managed pointer 🥲

In that case, let's get this PR in. 🚀 
It is what it is.

Issue #1733, opened 2026-05-22
✨ Extend OpenQASM-to-QC translation


denialhaag:
#1671 has added an initial translation from OpenQASM to the QC dialect. The translation does not yet support all features of OpenQASM.

This issue collects all features we should support. It is expected to grow in the future.

- Add support for `duration` type
- Add support for `opaque` gate declarations
- Improve support for `include` statement (currently, only `stdgates.inc` and `qelib1.inc` are supported)

> [!IMPORTANT]
> Please break out sub-issues as needed when implementing certain bullet points.

MatthiasReumann:
Regarding the support of the `scf.while` operations, I just had a really interesting conversation with Gemini: 

Apparently, the go-to way to translate a `while` loop to the (rather complex) `scf.while` operation is to convert the `while` to the `cf` dialect and then use the [`lift-cf-to-scf`](https://github.com/llvm/llvm-project/blob/1bf1f13be94fa5f4994d10310ec349cc3450d70f/mlir/include/mlir/Conversion/Passes.td#L338) pass to create the `scf.while` operation. This lifting procedure will (probably) always work for frontends such as OpenQASM because their `while` loop specification is rather cleverly chosen. The person addressing this issue might want to investigate this further. Looking at [Polygeist](https://github.com/llvm/Polygeist) might be worthwhile.

burgholzer:
With the exception of #2289, none of these should be blockers for v4.0. Moving this to the 4.1 milestone.

Issue #1732, opened 2026-05-22
✨ Add support for register-based conditions


denialhaag:
The translations from OpenQASM and `qc::QuantumComputation` to the QC dialect currently do not support register-based conditions; an `scf.if` operation can only be controlled by a single classical bit resulting from a measurement. Once #1726 is implemented, we can also start supporting register-based conditions. Temporary support for this was removed in https://github.com/munich-quantum-toolkit/core/pull/1671/changes/3b8a22848d408586ae07b758705e991ef50fa5b0, making for a good starting point for implementing this.

burgholzer:
We should remember to keep the branch around that https://github.com/munich-quantum-toolkit/core/commit/3b8a22848d408586ae07b758705e991ef50fa5b0 is part of.

PR #1862, opened 2026-07-06
✨ Translate `for` and `while` from OpenQASM to QC

### LINKED ISSUE #1846: ✨ Translate `for` and `while` from OpenQASM to QC
The current version of the OpenQASM-to-QC translation can only support `if` statements. We should extend this to `for` and `while` so that we can translate to all SCF operations.

denialhaag:
## Description

This PR adds support for translating `for` and `while` from OpenQASM to QC. In doing so, this PR also completely refactors the translation. Instead of relying on `qasm3::InstVisitor`, the new version of the translation dispatches on `qasm3::Token`. While more manual, the new implementation is lighter and less tied to legacy decisions.

Fixes #1846

AI notice: I have used Opus 4.8 via Claude Code for a first version of the changes. I have then iterated on this version (largely manually, sometimes with further help of Opus 4.8).

## Checklist

- [x] ~~AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.~~


denialhaag:
@coderabbitai full review

denialhaag:
@coderabbitai review

include/mqt-core/qasm3/DebugInfo.hpp
```diff
@@ -0,0 +1,36 @@
...
+            std::shared_ptr<DebugInfo> parentDebugInfo = nullptr)
+      : line(l), column(c), filename(std::move(std::move(file))),
+        parent(std::move(parentDebugInfo)) {}
+
+  [[nodiscard]] std::string toString() const {
+    return filename + ":" + std::to_string(line) + ":" + std::to_string(column);
+  }
+};
+
+} // namespace qasm3
```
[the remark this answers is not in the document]
denialhaag: Isn't that fixed now?

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+  const GateInfo mcxInfo{
+      .nControls = 0, .nTargets = 0, .nParameters = 0, .type = ::qc::OpType::X};
+  t["mcx"] = mcxInfo;
+  t["mcx_gray"] = mcxInfo;
+  t["mcx_vchain"] = mcxInfo;
+  t["mcx_recursive"] = mcxInfo;
+
+  t["mcphase"] = GateInfo{
+      .nControls = 0, .nTargets = 0, .nParameters = 1, .type = ::qc::OpType::P};
```
burgholzer: I am not the biggest fan of this kind of special handling. Can we get rid of this somehow and find a "proper" solution?

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+ * `QCProgramBuilder` as it parses. Each `parse` function parses all tokens of a
+ * given statement. Validation helpers raise `CompilerError`s wherever needed.
+ */
+class QASM3Parser final {
+public:
+  explicit QASM3Parser(MLIRContext* ctx)
+      : builder(ctx), gates(buildGateTable()) {}
+
+  OwningOpRef<ModuleOp> parse(std::string_view source) {
+    input = std::make_unique<std::istringstream>(std::string(source));
```
burgholzer: I think we could be relying on LLVM's streams and string libraries here and completely eliminate iostream/stringstream.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+  llvm::StringSet<> declaredNames;
+
+  /// Map from a parameter constant to its `f64`-typed MLIR value.
+  qasm3::NestedEnvironment<Value> parameterConstants;
+
+  /// Map from an induction variable to its MLIR value.
+  qasm3::NestedEnvironment<Value> loopVariables;
+
+  /// Cache of dynamically loaded qubits.
+  qasm3::NestedEnvironment<Value> dynamicallyLoadedQubits;
```
burgholzer: The `NestedEnvironment` class could be using LLVM types for its internal data structures over STL ones. Or could be reimplemented here using these types.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+  /// Map from qubit-register name to `QubitScope`.
+  QubitScope qubitRegisters;
+
+  /// Map from classical-register name to `ClassicalRegister`.
+  llvm::StringMap<QCProgramBuilder::ClassicalRegister> classicalRegisters;
+
+  /// Map from classical-register name to measurement results.
+  llvm::StringMap<SmallVector<Value>> bitValues;
```
burgholzer: Just an observation: I suspect that some of the code here will have to change once we properly support classical registers, right?

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+  /// Map from gate identifier to its metadata.
+  llvm::StringMap<std::variant<GateInfo, ParsedCompoundGate>> gates;
+
+  bool openQASM2CompatMode{false};
```
burgholzer: I think this rewrite is a nice opportunity to drop this compatibility mode. Even MQT Core will nowadays emit OQ3 with control modifiers and no longer uses the `cccc...` syntax if not explicitly asked to do so.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+  /// Whether the entire input has been consumed.
+  [[nodiscard]] bool isAtEnd() const {
+    return currentToken.kind == Token::Kind::Eof;
+  }
+
+  /// Create a `DebugInfo` object from @p token.
+  [[nodiscard]] static std::shared_ptr<DebugInfo>
+  makeDebugInfo(const Token& token) {
+    return std::make_shared<DebugInfo>(token.line, token.col, "<input>");
+  }
```
burgholzer: Doesn't MLIR have something similar related to debugging information that we could simply use out-of-the-box? Feels like something we could simplify here.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+  /// Assert that the current token matches an expected kind, then advance.
+  Token expect(const Token::Kind expected) {
+    auto token = current();
+    if (token.kind != expected) {
+      error(token, "Expected '" + Token::kindToString(expected) + "', got '" +
+                       Token::kindToString(token.kind) + "'.");
+    }
+    advance();
+    return token;
+  }
```
burgholzer: Can we try to avoid Exceptions here and make use of more MLIR functionality for error handling?

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    case Token::Kind::CReg:
+      parseCregDeclaration();
+      return;
+    case Token::Kind::Int:
+    case Token::Kind::Uint:
+    case Token::Kind::Bool:
+    case Token::Kind::Float:
+    case Token::Kind::Angle:
+    case Token::Kind::Duration:
+      error(current(), "Declaration type is not supported yet.");
```
burgholzer: On the one hand, this error message feels wrong. On the other hand, I am wondering why all of these are not supported here. I get it for `Angle` and `Duration` because these are somewhat special. However, the others are just plain data types that have a simple equivalent in MLIR.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+      return;
+    case Token::Kind::Gate:
+      parseGateDeclaration();
+      return;
+    case Token::Kind::Opaque:
+      error(current(), "Opaque gate declarations are not supported.");
```
burgholzer: Would it make sense to create a tracking issue for this? It is unclear to me, how we would be supporting these gates, but it makes sense to me to track the general problem somewhere.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+  void parseInclude() {
+    const auto beginToken = expect(Token::Kind::Include);
+    const auto filename = expect(Token::Kind::StringLiteral).str;
+    expect(Token::Kind::Semicolon);
+    if (filename != "stdgates.inc" && filename != "qelib1.inc") {
+      error(beginToken, "Unsupported include '" + filename +
+                            "'. Only 'stdgates.inc' and 'qelib1.inc' are "
+                            "supported.");
+    }
```
burgholzer: Another potential tracking issue? Potentially aggregated under a larger tracking issue for OpenQASM feature support.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+  /// Parse `const float <id> = <expression>;`.
+  void parseConstantDeclaration() {
+    const auto constToken = expect(Token::Kind::Const);
+    const auto debugInfo = makeDebugInfo(constToken);
+
+    if (current().kind != Token::Kind::Float ||
+        peek().kind != Token::Kind::Identifier) {
+      error(*debugInfo, "Only `const float <id> = <expression>;` declarations "
+                        "are supported for now.");
+    }
```
burgholzer: This feels very limiting. Any particular reason this cannot be extended here (e.g. to all primitive data types)?

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    expect(Token::Kind::Float);
+    const auto id = expect(Token::Kind::Identifier).str;
+
+    expect(Token::Kind::Equals);
+    auto initExpr = parseExpression();
+    expect(Token::Kind::Semicolon);
+
+    registerDeclaredName(constToken, id);
+    parameterConstants.emplace(id, emitFloatExpression(initExpr, debugInfo));
```
burgholzer: The name of this map is a bit irritating as it generally stores constants not only parameters.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    }
+
+    expect(Token::Kind::Semicolon);
+
+    declareClassicalRegister(bit, id, size);
+
+    if (operand) {
+      emitMeasureAssignment(ParsedBitRef{.name = id, .index = std::nullopt},
+                            *operand, debugInfo);
+    }
```
burgholzer: Without thinking about this to deeply this feels like it wouldn't support something like
```qasm
qubit[3] q;
bit[3] c = measure q;
```
While I believe this is valid OpenQASM.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+    if (current().kind != Token::Kind::Equals) {
+      error(current(), "Classical computations are not supported yet.");
+    }
+    advance();
+
+    if (current().kind != Token::Kind::Measure) {
+      error(current(), "Classical computations are not supported yet.");
+    }
+    advance();
```
burgholzer: This feels like a bit of a regression to me. The previous parser supported compile-time classical computation nearly everywhere if I am not mistaken.
I would have thought this becomes easier when building on MLIR.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+      step = second;
+      stop = parseExpression();
+    } else {
+      stop = second;
+    }
+
+    if (step.intValue < 1) {
+      error(*debugInfo,
+            "For loops with a non-positive step are not supported.");
+    }
```
burgholzer: Is that a limit of OpenQASM or of our implementation here?
In the latter case, it feels a bit arbitrary.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    if (current().kind == Token::Kind::Identifier) {
+      const auto bit = parseBitRef();
+      switch (current().kind) {
+      case Token::Kind::DoubleEquals:
+      case Token::Kind::NotEquals:
+      case Token::Kind::LessThan:
+      case Token::Kind::GreaterThan:
+      case Token::Kind::LessThanEquals:
+      case Token::Kind::GreaterThanEquals:
+        error(*debugInfo, "Register comparisons are not supported.");
```
burgholzer: Another limitation that I would have thought we can avoid.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+      default:
+        break;
+      }
+      return lookupBitValue(bit, debugInfo);
+    }
+
+    error(*debugInfo, "Unsupported condition expression in if statement.");
```
burgholzer: I would have hoped that we can be way more lenient in the kinds of computations we support. A condition is fine as long as it evaluates to an `i1` value. In the previous code, the TypeChecker and ConstantFolder kind of ensured this for an already rather broad selection.
This kind of feels like a regression to me

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+      assert(registerBits.size() == 1);
+      return registerBits[0];
+    }
+
+    const auto index = evaluateNonNegativeConstant(*bit.index, debugInfo);
+    if (index >= registerBits.size() || !registerBits[index]) {
+      error(*debugInfo, "Bit " + std::to_string(index) + " of register '" +
+                            registerName + "' has been not measured yet.");
+    }
+    return registerBits[index];
```
burgholzer: Hm.. I am beginning to more and more think that this is something that would be perfectly suitable for an MLIR verifier on an OpenQASM3 dialect.. or at least some kind of semantic analysis.

I am also not 100% sure whether it is really so clever to mix semantic analysis with the actual program emission like it is currently being done here.
From past conversations with LLMs, they mostly suggested to parse to an AST, then perform semantic analysis, then emit code. The current implementation kind of combines all three.

Feels like it may be worth to revisit this design decision.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    if (it == gates.end()) {
+      error(*debugInfo, "No OpenQASM definition found for gate '" + id + "'.");
+    }
+
+    // Resolve parameters
+    SmallVector<Value> params;
+    params.reserve(call.parameters.size());
+    for (const auto& arg : call.parameters) {
+      params.push_back(emitFloatExpression(arg, debugInfo));
+    }
```
burgholzer: Parameters do not necessarily need to be floating point expressions, right? One could declare a custom gate that uses other kind or arguments for parameters, I suppose.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    // Handle non-broadcasted calls
+    const auto split = splitControlsAndTargets<Value>(
+        call.modifiers, numCompatControls, operands, debugInfo);
+
+    // Inline compound gate
+    if (const auto* compound = std::get_if<ParsedCompoundGate>(&it->second)) {
+      emitCompoundGate(*compound, params, split.targets, split.posControls,
+                       split.negControls, split.invert, debugInfo);
+      return;
+    }
```
burgholzer: I believe this is another opportunity to improve the parser. There is no need to inline custom gate definitions.
We can simply materialize the definitions as proper `func::func` functions in the MLIR code and call the respective methods.
Inlining is something that can then happen in MLIR.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+        operands.push_back(*value);
+      } else {
+        operandsBroadcasting.push_back(std::get<SmallVector<Value>>(resolved));
+      }
+    }
+
+    if (!operandsBroadcasting.empty() && !operands.empty()) {
+      error(*debugInfo, "Gate operands must be single qubits or quantum "
+                        "registers and not a mix of both.");
+    }
```
burgholzer: I am not quite sure this is correct according to the OpenQASM spec.
Something like
```qasm
qubit[3] q;
qubit r;
cnot q, r;
```
is valid and is expected to expand to
```qasm
qubit[3] q;
qubit r;
cnot q[0], r;
cnot q[1], r;
cnot q[2], r;
```

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
```diff
@@ -0,0 +1,1850 @@
...
+      error(*debugInfo, "Expected a constant integer expression.");
+    }
+  }
+
+  /// Statically evaluate `ParsedExpr` to an `size_t`.
+  [[nodiscard]] size_t evaluateNonNegativeConstant(
+      const ParsedExpr& expr,
+      const std::shared_ptr<DebugInfo>& debugInfo) const {
+    return static_cast<size_t>(evaluateIntegerConstant(expr, debugInfo));
+  }
```
burgholzer: This kind of feels like something that should already happen in MLIR. I would have assumed we can have the translation to rather plainly emit the respective constructs and MLIR constant propagation and folding do the job for us.

burgholzer (changes requested):
Thanks @denialhaag for working on this! 🙏🏼 

I pushed one small commit with some cosmetic fixed I stumbled over while reviewing.
As you will probably see through the reviews, I am not yet quite happy where this is at. Overall, the current solution almost feels more constrained than the previous solution and still does not quite naturally fit to MLIR. I believe we should be taking much more advantage of MLIR built-in functionality for this.

It's also in one of the comments, but I'd like to highlight this here again: I am not 100% certain the approach of directly emitting the program while parsing really creates a nice design.
It mixes a couple too many things for my taste.
Maybe you can take all of the thoughts from the comments in this review back to the drawing board and re-evaluate what a good approach for all of this looks like.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Yes, quite a few things will have to change when https://github.com/munich-quantum-toolkit/core/issues/1726 is implemented. It should be rather straightforward, though.  🤔

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Sure, I can also add support for those in this PR! That said, I think we should draw the line _somewhere_ for this PR and extend the support for classical computations in a follow-up PR.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: I've added a bullet point to https://github.com/munich-quantum-toolkit/core/issues/1733.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Not at all! My initial goal for this PR was to translate all of the existing test programs, but I see the point of extending this a bit already here.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: I've added a bullet point to https://github.com/munich-quantum-toolkit/core/issues/1733.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: `scf.for` does not allow negative steps. That said, we can of course work around this by swapping start and stop.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: We cannot do this before https://github.com/munich-quantum-toolkit/core/issues/1726 is implemented. This is already tracked in https://github.com/munich-quantum-toolkit/core/issues/1732.

denialhaag: We can, of course, already support "purely classical" comparisons. They just cannot be based on a classical register.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: Agreed. At the same point, this feels like the right opportunity to make changes that we would otherwise not be making. I would argue operations/types that have a natural and direct analogue in MLIR should be supported here (that also includes all the arithmetic operations that OpenQASM supports)

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: Yeah. My main motivation for going rather broad here is that we can then use this as a trigger for finding capability gaps in our infrastructure.
I want to be able to throw OQ3 programs at our compiler and see where it breaks.
If we grow the OpenQASM support from the concepts that we already support anyway outwards, we are limiting ourselves in my opinion. We have to start thinking from the "user"/"producer" side of things and try to support as much as we can from there in my opinion.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: Okay. If OpenQASM allows this, we should map this to MLIR accordingly. Your proposed fix seems simple enough.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: Ah yeah, that makes sense 👍🏼 
Kind of bumps the priority of #1726 by a bit.
Feels like one of those fundamental representation questions that we need to have figured out rather sooner than later.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: This function is only used in four cases:

1. Translating `q[<index>]`: I guess this could always be translated to `memref.load` operations. There are some limitations in `QCProgramBuilder` that prevent repeated loads. Plus, we currently do not allow `memref.load` operations within modifiers. All of these points have a solution, but it could get nasty. 
2. Translating `c[<index>]`: Cannot be changed before [#1726](https://github.com/munich-quantum-toolkit/core/issues/1726) is implemented.
3. Translating `ctrl(<number of controls>)`: I don't really think this can ever be a dynamic value.
4. Translating `pow<exponent>`: This can be computed dynamically. I will change it.

burgholzer: > This function is only used in four cases:
> 
> 1. Translating `q[<index>]`: I guess this could always be translated to `memref.load` operations. There are some limitations in `QCProgramBuilder` that prevent repeated loads. Plus, we currently do not allow `memref.load` operations within modifiers. All of these points have a solution, but it could get nasty.

I see the nastiness. But this is exactly one of the examples for the point I was trying to raise in the other comment: indexed accessors are a standard thing in OpenQASM; so people will use them and tools will produce them in all kinds of places. We need to be able to handle those cases and find ways how these constructs can be represented well in our dialect.
Given how gate statements are parsed "as a whole" including the modifiers, it should be moderately straight forward to first compute the required qubit values and then do all the modifier handling and wrapping.

> 2. Translating `c[<index>]`: Cannot be changed before [#1726](https://github.com/munich-quantum-toolkit/core/issues/1726) is implemented.

Another good reason for bumping the priority on that one 😌 

> 3. Translating `ctrl(<number of controls>)`: I don't really think this can ever be a dynamic value.

Yeah, probably not, otherwise the operand list would have to be dynamic as well.

> 4. Translating `pow<exponent>`: This can be computed dynamically. I will change it.

👍🏼

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: I'd prefer to do this as part of https://github.com/munich-quantum-toolkit/core/issues/1659.

1. This would definitely break equivalence checking because the `QCProgramBuilder` cannot build the respective reference programs.
2. I'm also pretty sure we'll need to implement additional canonicalization patterns. Consider, for example, this program:

   ```qasm
   OPENQASM 3.0;
   include "stdgates.inc";
   qubit[3] q;
   gate compound q0, q1 {
     x q0;
     x q0;
   }
   ctrl(2) @ compound q[0], q[1], q[2];
   ```

   If translated into a new function, it would likely be canonicalized to an empty function. Subsequently, the control modifier would need to check whether the called function is empty and then remove the call, unless this is already handled by built-in patterns.
3. I'm not 100 percent sure how our modifiers would handle function calls right now. They might be removed if their body does not contain a `UnitaryOpInterface`.
4. I'm all but certain that we could not convert a program with multiple functions to QCO (let alone QIR).

burgholzer: Yeah, that makes sense. Maybe one can add a couple of notes there or point the issue to this discussion

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done as part of https://github.com/munich-quantum-toolkit/core/pull/1862/commits/853879a5a50453950169bba90249d517588ae686.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done as part of https://github.com/munich-quantum-toolkit/core/pull/1862/commits/853879a5a50453950169bba90249d517588ae686. In particular, we are using `llvm::ScopedHashTable` now.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done as part of https://github.com/munich-quantum-toolkit/core/commit/853879a5a50453950169bba90249d517588ae686 (unfortunately, I did this in between two things and couldn't break it out anymore).

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done as part of https://github.com/munich-quantum-toolkit/core/pull/1862/commits/853879a5a50453950169bba90249d517588ae686. In particular, we are using `llvm::SMLoc` now.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done as part of https://github.com/munich-quantum-toolkit/core/commit/853879a5a50453950169bba90249d517588ae686.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done in https://github.com/munich-quantum-toolkit/core/pull/1862/commits/7069858bb59d3e0049f12cd356436b35f3e6f904 and cleaned up in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/90e61c525fe05efb2bfcd783cbad0d5e24add40f.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Makes sense!

Done in https://github.com/munich-quantum-toolkit/core/pull/1862/commits/7069858bb59d3e0049f12cd356436b35f3e6f904 and cleaned up in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/90e61c525fe05efb2bfcd783cbad0d5e24add40f.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/2d064f7f8a650c7e3d52d0cc0c82cd630c7ba5b1.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: I _think_ this was already working, but I have now added an explicit test case in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/f3c82078e74f9a95c5a8f40599aa6b70819b6530.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/d1d4bf2bdcf19ebb41edf3b2f02ea2938c88c28d and cleaned up in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/90e61c525fe05efb2bfcd783cbad0d5e24add40f.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: #1726 is at the top of my list! Once #1603, #1755, and this PR are merged, I'll start implementing it. 😌

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/5b1af5f030e89ad29a3fcf6bbe6a0c713dcc76ab and cleaned up in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/90e61c525fe05efb2bfcd783cbad0d5e24add40f.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Ahh, wow, I had misunderstood the spec...

Fixed across https://github.com/munich-quantum-toolkit/core/pull/1862/changes/86f057791f59d02a1d7dce0643693401d00f1a97, https://github.com/munich-quantum-toolkit/core/pull/1862/changes/90e61c525fe05efb2bfcd783cbad0d5e24add40f, and https://github.com/munich-quantum-toolkit/core/pull/1862/changes/0925bc35d29e127de6907256c4de2cbd14705524.

denialhaag:
> I pushed one small commit with some cosmetic fixed I stumbled over while reviewing.

Sorry, I forgot to pull, and `git pull --rebase` was no longer feasible... I manually cherry-picked your improvements in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/3b1e39b0fd828cf09cd4c22586452382727ad950.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: I just updated #1659. Let me know what you think!

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: To be honest, I haven't consciously thought about these so far. I'm assuming that the first iteration of the translation in #1671 had these because of the special treatment in the original importer:

https://github.com/munich-quantum-toolkit/core/blob/9a2fd7dadbf8a29b89638926b341c4b0a65e6fcf/src/qasm3/Importer.cpp#L499-L504

Is there any reason to have a special definition for these multi-controlled gates in the first place? One obvious solution is just removing these special cases outright. 🤔

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: If I'm understanding the spec correctly right now, it looks like the parameters are technically supposed to be `angle`s: 

> The parameters are identifiers that behave as `angle` type with unknown size.
> 
> [...]
>
> [...], and the variable parameters params can have any type that can promote to `angle` type.

https://openqasm.com/language/gates.html#hierarchical-gates-definitions

In our case, I think it's fine to resolve everything to an `f64`-typed MLIR value because that's what our gates accept. 🤔

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: These were hacked into the parser add some point to add "native" support for these gates out of the box without needing a definition.
For a long time (still true, I believe), the OpenQASM output in Qiskit has been broken and will emit broken definitions for multi controlled phase gates for example (see also one of the open MQT Bench issues).
And the custom "mcx" names were simply added so that we could efficiently read in those gates from a Qiskit OpenQASM 3 dump (such as those coming from MQT Bench).

I am fairly certain this was never really the cleanest solution, nor the most efficient one. So I am generally happy to drop these. We just have to keep in mind that we may have to add something similar back at some point.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: Sounds good to me. Feel free to resolve here for now.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Ahh, that makes sense! I'll keep it in mind!

I've removed the special treatment in https://github.com/munich-quantum-toolkit/core/pull/1862/commits/756f80301c9485a21524ca6fe362bb0912093a80.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: Done in https://github.com/munich-quantum-toolkit/core/pull/1862/changes/e154961fa249fc64098b82c20f566b995e36fcb2.

Claude pointed out that simply swapping start and stop does not work if the operations in the loop body do not commute. The corrected solution still seems simple enough.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
burgholzer: Ah. Thanks for pointing that out. Makes sense to rely on `f64` then for now until we may support the angle type.

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: `evaluateIntConstant` is now used only to

1. determine the size of a register (as the spec requires a `const int`),
2. determine the number of controls of a control modifier (as the spec requires a `const int`), and
3. to extract bits from a classical register (limited by our current implementation).

For qubit registers, `tryEvaluateIntConstant` is used to not emit `memref.load` operations every single time. If `tryEvaluateIntConstant` fails, a `memref.load` operation is emitted. Let me know how you feel about this! If we wanted to get rid of this logic, we would have to get rid of these checks in the `QCProgramBuilder`: 

https://github.com/munich-quantum-toolkit/core/blob/def69f1517602ba802d1401ab97e6fa331e99d2e/mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp#L162-L170

burgholzer: Without having looked at all the details yet, this sounds reasonable and I believe we can postpone further refinements to follow-ups.

I'd definitely be interested in seeing wether we can drop these builder assumption. I assume this mainly had to do with our conversion code not really handling it well when there are multiple extracts of the same value within a scope.
However, 
- I would have assumed that these can be canonicalized away (maybe we need to teach MLIR something additional for the respective operations)
- we could probably adjust the conversions to handle this directly.

None of that is a good fit for this PR though.

denialhaag:
@coderabbitai full review

mlir/lib/Dialect/QC/Translation/QASM3Parser.cpp
denialhaag: I just created https://github.com/munich-quantum-toolkit/core/issues/1893 to keep track of this.

denialhaag: I'm not entirely sure how we would convert repeated `memref.load`s to QCO because we would basically have each instance to an insert-extract pair to not have a dangling qubit value. 🤔

burgholzer: I think we would have to do the state tracking manually ourselves and basically check whether a particular qubit has already been loaded; if so, we'd just use the already loaded value instead. I would assume there is a bit of literature on aliasing and how to deal with it properly in MLIR.
Nothing for this issue though. Feel free to resolve here.

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
```diff
@@ -0,0 +1,1477 @@
...
+        return failure();
+      }
+      resolvedStart = *startValue;
+      resolvedStep = *stepValue;
+      // Add 1 to make the stop value inclusive, as OpenQASM 3's for loops are
+      // inclusive.
+      resolvedStop =
+          arith::AddIOp::create(builder, *stopValue, builder.indexConstant(1))
+              .getResult();
+    }
```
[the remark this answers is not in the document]
denialhaag: @coderabbitai, if the step value is truly dynamic, there is no way for us to check whether it's positive, no?

@burgholzer, how do we want to proceed here? If the step is positive and a static value, it's straightforward to translate to `scf.for`. If the step is negative or a dynamic value, this falls apart quickly. Claude now suggests rescaling the loop and using a step of 1, which should always work, I guess. I'd only fall back to this if the step is not positive or a dynamic value. 🤔

mlir/.clang-tidy
```diff
@@ -8,6 +8,7 @@ Checks: |
   -bugprone-throwing-static-initialization,
   -cppcoreguidelines-avoid-non-const-global-variables,
   -cppcoreguidelines-pro-bounds-avoid-unchecked-container-access,
+  -cppcoreguidelines-slicing,
```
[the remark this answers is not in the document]
denialhaag: The current solution seems more in line with how the MLIR project itself does things. It does not activate _any_ `cppcoreguidelines` rule: https://github.com/llvm/llvm-project/blob/main/mlir/.clang-tidy

mlir/unittests/Dialect/QC/Translation/test_qasm3_translation.cpp
```diff
@@ -432,13 +703,91 @@ INSTANTIATE_TEST_SUITE_P(
...
+        QASM3TranslationTestCase{"SimpleWhileReset", qasm::simpleWhileReset,
+                                 MQT_NAMED_BUILDER(qc::simpleWhileReset)},
+        // --- ForOp --- //
+        QASM3TranslationTestCase{"SimpleForLoop", qasm::simpleForLoop,
+                                 MQT_NAMED_BUILDER(qc::simpleForLoop)},
+        QASM3TranslationTestCase{"NestedForLoopIfOp", qasm::nestedForLoopIfOp,
+                                 MQT_NAMED_BUILDER(qc::nestedForLoopIfOp)},
+        QASM3TranslationTestCase{"NestedForLoopWhileOp",
+                                 qasm::nestedForLoopWhileOp,
+                                 MQT_NAMED_BUILDER(nestedForLoopWhileOp)},
```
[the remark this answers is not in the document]
denialhaag: My next push removes the `forLoopOffsetIndex` test case because the functionality is already covered by `expressionDynamicIntIndex`.

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
```diff
@@ -0,0 +1,1477 @@
...
+      }
+      intConstants.insert(id, *folded);
+      return success();
+    }
+    auto emitted = emitIndex(value);
+    if (failed(emitted)) {
+      return failure();
+    }
+    intValues.insert(id, *emitted);
+    return success();
```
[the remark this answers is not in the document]
denialhaag: You definitely have a point. However, we also cannot use the SSA value for caching because the `i + 1` in `q[i + 1]` is evaluated every single time. If the `i` doesn't change between two instances of `q[i + 1]`, how can we load from the cache for the second instance?

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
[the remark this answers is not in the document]
burgholzer: @denialhaag Without having looked too deeply the suggestion from @coderabbitai seems to make sense.
In most cases (practically all that I can think of), the step of a for loop will be a compile-time constant anyway. So let's not make this too complicated.

denialhaag:
@coderabbitai review

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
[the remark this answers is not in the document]
denialhaag: @coderabbitai, this is fixed now, right?

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
```diff
@@ -0,0 +1,1477 @@
...
+        return std::optional{**lhs % **rhs};
+      case Expr::Kind::Pow: {
+        if (**rhs < 0) {
+          return error(expr.loc, "negative exponent in constant expression");
+        }
+        int64_t result = 1;
+        for (int64_t i = 0; i < **rhs; ++i) {
+          result *= **lhs;
+        }
+        return std::optional{result};
```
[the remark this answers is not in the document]
denialhaag: @coderabbitai, this is fixed now, right?

mlir/include/mlir/Dialect/QC/Translation/qasm3/QASM3Parser.h
```diff
@@ -0,0 +1,1571 @@
...
+                        "cannot assign to an element of a scalar");
+    }
+    if (symbol.isConst) {
+      return sink.error(loc,
+                        "cannot assign to the 'const'-declared '" + id + "'");
+    }
+    if (!scopes.back().contains(id)) {
+      return sink.error(loc, "cannot assign to '" + id +
+                                 "' from inside a block; it is declared in an "
+                                 "enclosing scope");
```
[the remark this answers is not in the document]
denialhaag: @burgholzer, I would move the point @coderabbitai is raising here to a follow-up. I'm not sure how our conversions and the builders would deal with this at the moment. 🤔

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
```diff
@@ -0,0 +1,1520 @@
...
+
+  /// A guarded scope for declared scalar symbols and dynamically loaded qubits.
+  struct DeclarationScope {
+    explicit DeclarationScope(QCEmitter& emitter)
+        : bindings(emitter.bindings),
+          loadedQubits(emitter.dynamicallyLoadedQubits) {}
+
+    BindingScope bindings;
+    ValueScope loadedQubits;
+  };
```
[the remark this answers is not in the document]
denialhaag: For the same reason as laid out in https://github.com/munich-quantum-toolkit/core/pull/1862#discussion_r3582104191, I would also move this to a follow-up and forbid block-local register declarations for now, as @coderabbitai suggests as a second option. What's your opinion here, @burgholzer?

mlir/include/mlir/Dialect/QC/Translation/qasm3/QASM3Parser.h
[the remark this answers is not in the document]
burgholzer: @coderabbitai hm. I'll look into this as part of my review. at a quick glance, this feels a bit like an arbitrary limitation to be honest.

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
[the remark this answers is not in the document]
burgholzer: @denialhaag @coderabbitai the OpenQASM 3 spec states
> All qubits are global variables. Qubits cannot be declared within gates or subroutines. This simplifies OpenQASM significantly since there is no need for quantum memory management. However, it also means that users or compiler have to explicitly manage the quantum memory.

this should somewhat lead to a simpler solution here.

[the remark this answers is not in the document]
burgholzer: @coderabbitai I'll have a look as part of my review. it feels like we could conveniently realize both.

mlir/include/mlir/Dialect/QC/Translation/qasm3/QASM3Parser.h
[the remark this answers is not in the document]
denialhaag: It is a limitation, but I don't think it's a limitation of the translation. The `QCProgramBuilder` currently cannot create a while loop that modifies a classical variable. For this to be necessary, we would need to add block arguments and respective yields. In addition, all conversions that involve `scf` operations would not respect a classical block argument at the moment. This is all fixable, but I probably wouldn't do it as part of this PR.

mlir/lib/Dialect/QC/Translation/qasm3/QASM3Emitter.cpp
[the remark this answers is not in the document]
denialhaag: https://github.com/munich-quantum-toolkit/core/pull/1862/commits/eda435e47c4e0cd96d5286285c8f6af7867afaad should fix this. @coderabbitai, can you have a look?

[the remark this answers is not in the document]
denialhaag: @coderabbitai, sorry, I had a typo in your tag. Can you have a look at https://github.com/munich-quantum-toolkit/core/pull/1862/changes/eda435e47c4e0cd96d5286285c8f6af7867afaad and see if it fixes the issue?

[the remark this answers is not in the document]
denialhaag: @coderabbitai, perfect, thanks for confirming! Just to clarify: The implemented restriction on classical registers is not temporary, but what the spec requires, right? Classical registers may now be declared inside a block, but they are confined to the block.

And thanks, @burgholzer, for the hint!

mlir/include/mlir/Dialect/QC/Translation/qasm3/QASM3Parser.h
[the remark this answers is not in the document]
burgholzer: Would we? SCF while regions are not isolated from above, aren't they?
I am probably just missing something here. I'll look at this with fresh eyes in the coming days.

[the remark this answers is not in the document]
denialhaag: No, the regions are not isolated from above. That said, I think you need to thread any values you are updating through the region arguments anyway. For example, see this example from the [MLIR docs](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scffor-scfforop):

```mlir
func.func @reduce(%buffer: memref<1024xf32>, %lb: index,
                  %ub: index, %step: index) -> (f32) {
  // Initial sum set to 0.
  %sum_0 = arith.constant 0.0 : f32
  // iter_args binds initial values to the loop's region arguments.
  %sum = scf.for %iv = %lb to %ub step %step
      iter_args(%sum_iter = %sum_0) -> (f32) {
    %t = load %buffer[%iv] : memref<1024xf32>
    %sum_next = arith.addf %sum_iter, %t : f32
    // Yield current iteration sum to next iteration %sum_iter or to %sum
    // if final iteration.
    scf.yield %sum_next : f32
  }
  return %sum : f32
}
```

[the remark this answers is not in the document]
denialhaag: Okay, I'm definitely getting too tired. Reassignment obviously isn't a thing in MLIR. A reassignment in OpenQASM corresponds to the definition of a new variable in MLIR. If this new variable is defined within a region, it would definitely need to be yielded to be accessible outside.

burgholzer:
Hey @denialhaag 👋🏼 I let GPT-5.6-Codex review the PR and compare it against the OpenQASM 3 live specification. Below is the improvement plan it came up with. To me, this makes a lot of sense. If this direction is okay with you, I’d task Codex with implementing the stabilizations for this PR and then continue the broader work in follow-up PRs.

# Improve PR #1862 into a specification-driven OpenQASM 3 frontend

## Summary

PR [#1862](https://github.com/munich-quantum-toolkit/core/pull/1862) has a strong foundation: zero-copy lexing, `SourceMgr`/`SMLoc` diagnostics, LLVM containers, `LogicalResult`, and direct use of `arith`, `math`, `memref`, `scf`, `func`, and QC operations. Broadcasting and structured-control lowering are also conceptually sound.

However, it is not yet suitable for the stated “arbitrary OpenQASM 3” goal. The custom parser implements only a narrow subset of the [official grammar](https://openqasm.com/grammar/index.html), while syntax, semantic analysis, and MLIR emission remain interleaved despite the sink abstraction. This causes rejected valid programs, accepted invalid syntax, target-specific parser restrictions, and some paths that can produce invalid MLIR.

Use a staged delivery:

1. Make #1862 safe and honest to merge by fixing correctness blockers and documenting its supported subset.
2. Replace the streaming frontend with a specification-driven, versioned OpenQASM-facing MLIR stage.
3. Lower incrementally to builtin MLIR and QC dialects, using custom operations only where no suitable builtin representation exists.

Target stable OpenQASM 3.0 and 3.1 with version-gated behavior. Track later live-spec additions explicitly.

## Review Conclusions

### Merge-blocking correctness issues

- Classical state is not represented correctly across regions. Scalar assignment to an enclosing scope is deliberately rejected, but measurement assignment to an enclosing classical register is allowed. The emitter then records an SSA value defined inside an `scf.if`/`scf.for`/`scf.while` region and may use or return it outside that region, violating dominance. Until classical storage is implemented, reject these assignments and verify every produced module before returning it.
- `while` cannot implement ordinary mutable loop state. OpenQASM block scopes inherit mutable outer variables, while the current parser rejects their assignment. Proper lowering requires `scf.while` arguments/results or mutable storage, as required by [OpenQASM scoping semantics](https://openqasm.com/language/scope.html).
- Standard-gate handling is inconsistent. The legacy gate metadata table and the emitter dispatch table contain different names: several recognized gates such as `ccx`, `c3x`, `c4x`, `cu1`, and `iswapdg` have no emitter, while several dispatch aliases are unreachable. Moreover, standard gates are available even without `include "stdgates.inc"`, contrary to the [versioned standard-library contract](https://openqasm.com/language/standard_library.html).
- Gate definitions can recurse during eager expansion and exhaust the stack. Diagnose self-recursion/cycles and enforce declaration-before-use.
- Dynamic zero loop steps reach signed division by zero; statically positive loops can overflow when converting an inclusive maximum bound with `stop + 1`. Add checked lowering or diagnostics.
- The lexer/parser accepts and rejects the wrong language in several places: it lacks integer bases and separators, `**`, `%`, shifts, bitwise operators, type designators such as `int[32]`, trailing commas, and most official statements; it incorrectly accepts `pow(a,b)` as a function despite the specification requiring `**`. It also permits version declarations in arbitrary positions and ignores the parsed version.
- Input/output handling is not OpenQASM-compatible. Only `output bit[...]` is accepted, every classical register is returned when no output exists, and outputs are required to have been measured. Implement `input`/`output` as function arguments/results and preserve OpenQASM 2 compatibility only behind an explicit mode.
- `measure q;`, anonymous scopes, set/expression-based `for` iteration, and valid gate-body loops are missing despite appearing in the official grammar.
- Remove the PR-wide `.clang-tidy` suppression and address any slicing locally.

### Important coverage gaps

The current frontend is missing substantial language families documented by the [live language specification](https://openqasm.com/language/):

- Width-aware `bit`, `int`, `uint`, `float`, `angle`, `complex`, `duration`, and `stretch` types, casts, promotions, and complete expression precedence.
- Arrays, aliases, concatenation, slicing, index sets, bit strings, timing literals, and dynamic classical indexing.
- Comparisons, compound assignment, shifts, bitwise operations, complete builtins, and general `i1` conditions.
- `def`, `extern`, calls, returns, `break`, `continue`, `switch`, `end`, `nop`, pragmas, and annotations.
- `pow` modifiers, timing instructions, boxes, calibration grammar, `cal`, and `defcal`.
- Real include resolution, nested-source diagnostics, include cycles, and version-dependent `stdgates.inc`.

### Architecture assessment

Keep these strengths:

- LLVM source locations and diagnostics.
- Bump allocation and zero-copy source views where useful.
- Direct lowering to standard MLIR dialects after semantic analysis.
- Existing equivalence tests and MLIR verification.
- QC operations for quantum semantics and modifiers already modeled by QC.

Replace these weaknesses:

- The template parser/sink callbacks do not create a true phase boundary; parser symbol checks and emitter semantic checks are mixed.
- The 1,500-line template header increases compile cost for negligible runtime benefit.
- Eager custom-gate inlining can cause exponential IR growth and prevents reuse of `func.func`/`func.call`.
- General integer computation uses `index`, losing width and signedness. Reserve `index` for indexing and loop bounds.
- Classical registers are external metadata rather than MLIR state.
- Dynamic-qubit caching serializes expression trees into heap-allocated strings and uses opaque SSA pointer values.
- Gate metadata is duplicated across legacy maps, `std::function` dispatch, and QC builders.

## Implementation Changes

### 1. Stabilize #1862 before merge

- Add an explicit supported-feature document and stop describing the implementation as accepting arbitrary OpenQASM.
- Reject any enclosing-scope classical update that cannot be represented with valid dominance, including measurement assignments in control-flow regions.
- Run `mlir::verify` before returning a translated module; translation failure must produce a source-anchored diagnostic, never malformed IR.
- Make version handling stateful: allow at most one leading version statement, accept only supported 3.0/3.1 versions, and select the corresponding standard library.
- Populate standard gates only after `stdgates.inc`; generate metadata and emission from one registry and test every registered gate. Keep `qelib1.inc` behind compatibility mode.
- Diagnose recursive gates, invalid control counts, unsupported power modifiers, zero dynamic loop steps, and inclusive-bound overflow.
- Correct the easiest grammar issues immediately: numeric literals, `**`, `%`, valid strings, type designators, trailing commas, anonymous scopes, and targetless measurement.
- Remove the global lint suppression and add changelog coverage.

### 2. Introduce a specification-driven source stage

- Pin the official 3.0 and 3.1 ANTLR grammars as the syntax source of truth. Check generated C++ sources into a dedicated generated target so normal builds require only the pinned ANTLR C++ runtime, not Java. Use SLL parsing with LL fallback.
- Parse all syntactically valid constructs into an internal `oq3` MLIR dialect. Unsupported constructs must survive parsing and fail later with a precise “not lowerable yet” diagnostic rather than a misleading syntax error.
- Model source-level declarations, assignments, gates, calls, modifiers, IO, control flow, timing, calibration blocks, and source types. Attach exact file/range locations from `SourceMgr`, including included files.
- Add a semantic-analysis pass for scope, declaration-before-use, type inference, promotions, casts, constness, version rules, recursion, operand cardinality, and gate broadcasting.
- Keep the source dialect internal initially; preserve the existing `translateQASM3ToQC` API.

### 3. Lower typed OpenQASM IR using MLIR infrastructure

| OpenQASM concept | Lowering |
|---|---|
| Boolean and classical arithmetic | `arith`, `math`, builtin integer/float/complex types |
| Mutable `bit[n]` registers | `memref<nxi1>` with loads/stores; integrate with issue #1726 |
| Structured mutable scalars | `scf.if` results and `scf.for`/`scf.while` iterated arguments |
| Inputs, outputs, subroutines, gates | `func.func`, function arguments/results, `func.call`; integrate with issue #1659 |
| Quantum storage and operations | QC qubit types, memrefs, gates, measurement, reset, and modifiers |
| Switch and unstructured exits | `scf.index_switch` where possible; otherwise `cf` blocks and branches |
| Angle, duration, stretch, power | Small OpenQASM/QC-specific types or ops until target-independent lowering exists |

- Represent signedness and declared widths explicitly during semantic analysis; select signed/unsigned `arith` operations accordingly.
- Use `OperationFolder`, dialect folding, canonicalization, and CSE for ordinary constants. Retain a small `APInt`/`APFloat` evaluator only where OpenQASM requires a compile-time value before IR shape can be formed.
- Represent custom gates and subroutines as symbols and calls. Leave inlining to an explicit MLIR pass.
- Implement `pow` as a QC modifier operation capable of dynamic/fractional exponents; statically integral cases may later canonicalize to repetition/inversion.
- Resolve includes textually through `llvm::vfs::FileSystem`, include paths, and `SourceMgr`; detect cycles and preserve parent locations.
- Parse calibration bodies losslessly as opaque source operations until a pulse-level dialect is available.

### 4. Public interface additions

Keep both existing overloads and add an options overload without breaking callers:

- `OpenQASMVersion`: `V3_0` and `V3_1`.
- `OpenQASMTranslationOptions`:
  - optional version override used only when the source omits `OPENQASM`;
  - include search paths and virtual filesystem;
  - opt-in `qelib1.inc`/OpenQASM 2 compatibility;
  - diagnostic limit.
- Default omitted-version behavior to OpenQASM 3.0.
- Do not expose the internal `oq3` dialect as a stable public API until its operation set has completed one compatibility cycle.

### 5. Performance work

- Replace the templated sink parser and global `std::function` maps with generated parser code and a single TableGen/generated gate registry.
- Stop eagerly extracting every qubit in a register; materialize constant or dynamic loads on demand.
- Replace expression-string cache keys with dominance-aware MLIR values and standard CSE/canonicalization after issue #1893 removes repeated-load restrictions.
- Avoid rebuilding and dynamically casting the complete legacy standard-gate map for every translation.
- Deduplicate constants through folding and avoid repeated `arith.constant` emission for builtin constants.
- Preserve gates and subroutines as calls to prevent parse-time IR expansion.
- Add benchmarks for 10k/100k statements, large mostly-unused registers, nested control flow, repeated indexing, and deeply composed gates. Track wall time, peak memory, generated operation count, and compiler build time; enforce a 15% regression threshold after establishing the new frontend baseline.

## Test Plan

- Import the official OpenQASM example and grammar-validation corpora for both 3.0 and 3.1.
- Maintain a generated conformance matrix mapping every grammar statement, expression, type, and versioned standard gate to “parses”, “semantically validates”, “lowers”, or “diagnosed unsupported”.
- Add focused regressions for:
  - width-bearing declarations, numeric bases/separators, all precedence levels, casts, and trailing commas;
  - version placement, duplicate versions, unsupported versions, and versioned standard gates;
  - absent/present `stdgates.inc`, custom includes, cycles, and included-file diagnostics;
  - measurement with and without targets and measurement inside each control-flow region;
  - mutable scalars and registers across `if`, ascending/descending/empty `for`, and `while`;
  - dynamic zero steps, boundary overflow, non-divisible ranges, sets, and expression iteration;
  - block-local shadowing, anonymous scopes, IO, and undefined values;
  - every standard gate, broadcasting, modifiers, duplicate operands, and recursive definitions;
  - invalid programs producing diagnostics without crashes, hangs, or malformed modules.
- Verify every successful result before comparison, then run QC cleanup and the supported QC→QCO/QIR pipelines.
- Add parser fuzzing and differential grammar tests; arbitrary bytes and syntactically invalid inputs must terminate with bounded diagnostics.
- Preserve all current equivalence tests, but supplement them with structural tests that inspect region arguments/results and classical storage rather than comparing only happy-path circuits.

## Assumptions

- The unanswered architecture choices use the recommended defaults: staged OpenQASM MLIR frontend, merge blockers plus roadmap, and stable 3.0/3.1 version gating.
- #1862 remains primarily the loop-enablement PR; broad language completion lands in sequenced follow-ups.
- Issues #1726, #1659, #1732, #1733, and #1893 are incorporated into one frontend roadmap rather than treated as independent limitations.
- Full calibration execution is outside the initial lowering target, but calibration syntax must still parse and remain representable.

denialhaag:
The stability improvements generally look good to me. I only have two comments:

1. I'm curious to see how it rejects dynamic zeros. I had a first version of this, but it wouldn't have been lowerable to `jeff`.
2. The global `clang-tidy` ignore is intentional. This is how the MLIR project itself does it. We also do the same in `jeff-mlir`.

Now to the bigger part: If we really want to go with an `oq3` dialect, we might as well do it now. This PR feels too big to be purely temporary code that is refactored again completely soon. That said, I do not see how an `oq3` dialect would necessarily help us in not having two layers of errors, unless the operands of the operations are mostly typed as the MLIR equivalent of `Any`. I feel like the current errors in the emission are either temporary limitations or typing-related. But I might be missing something.

Generally speaking, I don't really know how to deal with all of this efficiently. Doing too many LLM-generated refactors is very annoying. Given the size this PR is already at, I spent way too long reading and streamlining LLM-generated code. I agree that the result is still not ideal, and we should improve upon it. I might not be having a point, but I guess I want to say that we should be _really_ sure what we want this to look like. Any LLM-generated refactor is another start from zero. I would really like to get to a point where we have a stable foundation we know and can easily work with.

burgholzer:
> Now to the bigger part: If we really want to go with an `oq3` dialect, we might as well do it now. This PR feels too big to be purely temporary code that is refactored again completely soon. That said, I do not see how an `oq3` dialect would necessarily help us in not having two layers of errors, unless the operands of the operations are mostly typed as the MLIR equivalent of `Any`. I feel like the current errors in the emission are either temporary limitations or typing-related. But I might be missing something.
> 
> Generally speaking, I don't really know how to deal with all of this efficiently. Doing too many LLM-generated refactors is very annoying. Given the size this PR is already at, I spent way too long reading and streamlining LLM-generated code. I agree that the result is still not ideal, and we should improve upon it. I might not be having a point, but I guess I want to say that we should be _really_ sure what we want this to look like. Any LLM-generated refactor is another start from zero. I would really like to get to a point where we have a stable foundation we know and can easily work with.

Let me try the following: I'll hand the full rewrite from the plan above to Codex and let it create a new draft PR showing how the result would look like. Based on that, we can (hopefully) better judge whether that is a direction worth pursuing.

denialhaag:
Two more things that just came to mind and can maybe be thrown into Codex as well:

1. In an `oq3` dialect, can the semantic analysis be done exclusively with operation-level MLIR verifiers, or do we need module-level verification passes as well? If we need module-level verification, is it really not better to do that during program creation (i.e., as it is done now)? We should avoid walking the same program more than once.
2. Is it feasible to report the original `SMLoc`s if semantic analysis fails? The MLIR locations are likely useless for a user.

denialhaag:
Closing this in favor of #1910.

PR #1887, opened 2026-07-10
✨ Add custom job parameters to FoMaC

flowerthrower:
## Description

This PR adds the custom1-5 job parameters to FoMaC.

During implementing the [CUDAQ QDMI Integration](https://github.com/NVIDIA/cuda-quantum/pull/4882), I found that currently FoMaC only exposes custom session parameters but not custom job parameters (which is, e.g., crucial for using the [Amazon Braket QDMI Device](https://github.com/munich-quantum-software/amazon-braket-qdmi-device)).

## Checklist


burgholzer:
Hey @flowerthrower 👋🏼 
This may be a bit of a limiting solution and not 100% what we want.
See https://github.com/munich-quantum-toolkit/core/pull/1849#discussion_r3529735613 for some very recent discussion on this.

The solution for session parameters is already quite hacky, but worked in all cases so far.
It's a fairly strict assumption to make that all of the parameters are strings. What about ints and floats and other data?

Looping in @ystade and @MatthiasReumann again from the other discussion.
For convenience sake, we could make that assumption for QDMI v1 that all custom properties are strings. But then we definitely need to find a cleaner solution for this in v2.
Opinions?

Edit: I suppose since we are talking about job parameters here and these need to be serialized somehow to form a payload for the machine, one could reasonably argue that strings might be correct in almost all the places where this would be used.

include/mqt-core/fomac/FoMaC.hpp
```diff
@@ -318,9 +318,14 @@ class Device {
...
-                              QDMI_Program_Format format,
-                              size_t numShots) const;
+  [[nodiscard]] Job
+  submitJob(const std::string& program, QDMI_Program_Format format,
+            size_t numShots,
+            const std::optional<std::string>& custom1 = std::nullopt,
+            const std::optional<std::string>& custom2 = std::nullopt,
+            const std::optional<std::string>& custom3 = std::nullopt,
+            const std::optional<std::string>& custom4 = std::nullopt,
+            const std::optional<std::string>& custom5 = std::nullopt) const;
```
MatthiasReumann: I would be inclined to try templating this function here with `template<class T0, class T1, ...>` and set the parameter according to the type (we may assume basic types such as `int`, `float`, `std::string` for now). 

Not sure if this would work with the Python bindings.

burgholzer: For the Python bindings, only instantiated templates can be bound unfortunately.
However, this would still buy us some flexibility on the C++ side, while we could fix the Python bindings to strings for now.

MatthiasReumann: So theoretically we could explicitly instantiate the function for the basic types and Python would be able to handle it?

burgholzer: Yeah. The bound functions just have to have different names in Python then, though. Which makes this kind of prohibitive. You'd essentially need to provide all permutations of primitive types for a function like `submitJob` above.
Not really a pretty solution.

MatthiasReumann: > The bound functions just have to have different names in Python then, though. 

Mhm. Then, I think the "best" solution would be

> [...to] fix the Python bindings to strings for now.

burgholzer: @flowerthrower can you try that out? (and fix the other CI failures 😌)

flowerthrower: Found an interesting approach using a combination of Python-side `object` and cpp `variant`. What you you think about that (see last commit)?

burgholzer: That was a nudge in the right direction. I found an even more elegant solution that I am happy with. 🙌🏼

burgholzer (approved):
I polished the implementation of this feature slightly to make it mergeable. For parameters, this feels like a very clean solution.
For queryable properties, I am not so sure yet how to best approach this.

PR #1901, opened 2026-07-14
♻️ Redesign QDMI device management

burgholzer:
🤖 *AI text below* 🤖

## Summary

This PR replaces the overlapping FoMaC and client-driver layers with one public
QDMI device-management flow:

`DeviceRegistry → DeviceManager → Device`

The same object model is used by C++, Python, Qiskit, and the neutral-atom
adapter. The branch now contains only this v4 redesign on top of current
`main`; the configurable-device foundation from #1912 is already merged.

## What changed

- `DeviceRegistry` discovers configured definitions and supports idempotent
  fallback registration without loading device code.
- `DeviceManager` owns an immutable registry snapshot, opens a fresh device
  session for every call, and isolates `openAll()` failures by stable device ID.
- `Device`, `Site`, `Operation`, and `Job` retain the runtime state required by
  their QDMI handles, so derived objects may safely outlive their manager and
  parent wrappers.
- A process-wide weak cache shares compatible live device libraries by
  canonical path and symbol prefix. Replacement waits for the prior generation
  to finish finalization before initializing again.
- Python `DeviceDefinition` and `SessionParameters` accept path-like library and
  authentication-file arguments.
- Text and binary program submission preserve the contracts introduced by
  #1957: textual formats include the terminating null byte, while binary
  formats and `program_bytes` preserve exact bytes.
- Qiskit and neutral-atom integrations use the unified manager and device
  objects.
- The optional bundled-device switches introduced by #1965 remain effective:
  the registry and object model build without bundled devices, while
  device-specific bindings, tests, and wheel contents follow their respective
  CMake options.
- The FoMaC and `qdmi::Driver` public APIs are removed; `UPGRADING.md` documents
  the v4 migration.

## Configuration and integration

The implementation reuses the merged #1912 configuration contracts:
`qdmi.json`, `[tool.qdmi]`, stable-ID merging, disabled-ID reservations,
relocatable CMake metadata, and runtime manifests. It does not add another
parser or discovery mechanism.

Integration work with QDMI-on-IQM, Amazon Braket, and the QDMI templates led to
an immutable manager, one registration boundary, per-open session overrides,
and direct runtime ownership by the returned object graph.

Independently installed device wheels remain explicit: they expose their
library path and register a fallback definition before constructing a manager.
User configuration and disabled definitions retain precedence.

## Validation

- complete Release build with LLVM/MLIR 22
- all 3,882 CTest cases passed; two device job-ID cases were intentionally
  skipped by their fixtures
- separate configuration with all bundled devices disabled: 12 registry tests
  passed
- Python 3.14: 397 passed, 3 upstream-Qiskit skips
- Python stub generation
- warning-as-error documentation build
- full `uvx nox -s lint`
- `git diff --check`


include/mqt-core/qdmi/DeviceManager.hpp
```diff
@@ -0,0 +1,110 @@
...
+/** A parsed, side-effect-free registration for one QDMI device. */
+struct DeviceDefinition {
+  std::string id;
+  std::filesystem::path library;
+  std::string abi = "qdmi-v1";
+  std::string prefix;
+  bool enabled = true;
+  SessionParameters session;
+  std::filesystem::path source;
+};
```
MatthiasReumann: Nitpick: These attributes should probably be sorted by bit-size (large to small) to decrease the overall `struct` size. For example, moving `enabled` to the very bottom decreases the size from 424 to 416 bytes.

A brief `/// <description>` would be helpful also.

This holds for all `structs` in this file.

include/mqt-core/qdmi/Device.hpp
```diff
@@ -0,0 +1,358 @@
...
+  [[nodiscard]] static auto siteHandles(const std::vector<Site>& sites)
+      -> std::vector<void*>;
+  Operation(std::shared_ptr<detail::DeviceState> state, void* handle)
+      : state_(std::move(state)), handle_(handle) {}
+  [[nodiscard]] auto
+  queryCustomPropertyBytes(CustomProperty property,
+                           const std::vector<Site>& sites,
+                           const std::vector<double>& params) const
+      -> std::optional<std::vector<std::byte>>;
+  std::shared_ptr<detail::DeviceState> state_;
```
MatthiasReumann: Nitpick: Could we again remove the trailing return types?

include/mqt-core/qdmi/Device.hpp
```diff
@@ -0,0 +1,358 @@
...
+  IqmJson = 8,
+  BatchJob = 9,
+  Custom1 = 999999995,
+  Custom2 = 999999996,
+  Custom3 = 999999997,
+  Custom4 = 999999998,
+  Custom5 = 999999999,
+};
+
+/** Identifies an implementation-defined custom property or result slot. */
```
MatthiasReumann: Nitpick: Use `///` or make it

```cpp
/**
* @brief ...
*
*/
```

src/qdmi/DeviceManager.cpp
```diff
@@ -0,0 +1,672 @@
...
+    library = impl_->libraries[key].lock();
+    if (!library) {
+      library =
+          detail::makeV1DeviceApi(definition->library, definition->prefix);
+      impl_->libraries[key] = library;
+    }
+  }
+  auto parameters = definition->session;
+  overlay(parameters, sessionOverrides);
+  return Device(std::make_shared<detail::DeviceState>(library, parameters));
```
MatthiasReumann: We might want to re-think using a `shared_ptr` here. The `DeviceManager` should probably hold `unique_ptr`s and share `const` pointers with the `Device` class. A `Device` should not be able to modify the `shared_ptr`.

src/qdmi/DeviceManager.cpp
```diff
@@ -0,0 +1,672 @@
...
+  std::shared_ptr<detail::DeviceApi> library;
+  {
+    const std::scoped_lock lock(impl_->mutex);
+    library = impl_->libraries[key].lock();
+    if (!library) {
+      library =
+          detail::makeV1DeviceApi(definition->library, definition->prefix);
+      impl_->libraries[key] = library;
+    }
+  }
```
MatthiasReumann: Something is fishy here. 

`impl_->libraries[key]` is a [`std::weak_ptr`](https://en.cppreference.com/cpp/memory/weak_ptr) - a non-owning reference to a `shared_ptr`. The `DeviceManager` (or the `DeviceManager::Impl`) class should be owning the device `shared_ptr`.  `.lock()`ing and assigning the newly created `shared_ptr` only works (as far as I can tell) because the `Device` class below holds the same `shared_ptr`. I am pretty sure this function wouldn't work anymore if the `Device` takes a const pointer instead of a smart pointer.

MatthiasReumann: The map in `DeviceManager::Impl` should probably store the `unique_ptr` mentioned above (or below).

MatthiasReumann: I am also not really sure if I like the `mutex` here. I get where this is coming from - thread safety - but I think one really has to think this through if this would be enough.

include/mqt-core/qdmi/DeviceManager.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  /** Registers a complete definition. */
+  void registerDevice(DeviceDefinition definition, bool replace = false);
+
+  /** Removes a definition, returning whether it existed. */
+  bool unregisterDevice(std::string_view id);
+
+private:
+  std::vector<DeviceDefinition> definitions_;
+};
```
MatthiasReumann: Could we move the `DeviceRegistry` to a separate `DeviceRegistry.h` (+ `DeviceRegistry.cpp`) file?

MatthiasReumann (changes requested):
This is a great start into the QDMI rewrite! 🚀

I've gone through the code a couple of times now and the following comments are the result of that. 

However, currently this can't construct devices at runtime, right? As of now this simply loads dynamic libraries, I think 🤔.

src/qdmi/DeviceManager.cpp
MatthiasReumann: A little bit like a `Device` is simply a view over the actual `DeviceState`.

src/qdmi/DeviceManager.cpp
MatthiasReumann: @coderabbitai Could you please evaluate why a `weak_ptr` is used here and if this is correct?

MatthiasReumann: @coderabbitai Okay. If the device manager would now own the APIs via `unique_ptr`s in the map and provide non-owning views to its devices, the disadvantage would be the automatic reference counting and deconstruction of devices, right?

MatthiasReumann: @coderabbitai could you please answer the question above?

MatthiasReumann: @burgholzer I think this is quite interesting. Maybe something to think about whether the DeviceManager should be the only instance _managing_ devices. Personally, I am not 100% sure if I would like to keep the shared_ptr; although the arguments are somewhat convincing. In the end it's a design decision.

burgholzer: I'll add some more reasoning from my iteration later. Right now, I am rather convinced this is a good design as it is.

include/mqt-core/qdmi/DeviceManager.hpp
burgholzer: Good point. I moved the trailing booleans in DeviceDefinition and ConfigOptions to padding-friendly positions and added /// documentation to every struct member. I kept the remaining equally aligned fields in their logical order where reordering would not reduce the size.

include/mqt-core/qdmi/Device.hpp
burgholzer: Done. The new public QDMI headers now use explicit leading return types wherever the return type is not intentionally deduced. I also removed the remaining trailing return type from the internal custom-value decoding template.

include/mqt-core/qdmi/Device.hpp
burgholzer: Done. The new QDMI headers now consistently use /// Doxygen comments for these declarations and their members.

src/qdmi/DeviceManager.cpp
burgholzer: I considered the manager-owned unique_ptr approach, but kept shared ownership deliberately because a Device is not intended to be a view into its manager. Devices, child devices, sites, operations, and jobs must remain valid after the manager is destroyed or the corresponding definition is unregistered. A manager-owned unique_ptr plus borrowed pointers would make those objects dangle.

The device-facing objects now hold shared_ptr<const DeviceState>, so they cannot mutate the shared device state through that pointer. The pointer itself is private as well. Tests explicitly cover devices and jobs outliving the manager and registrations.

src/qdmi/DeviceManager.cpp
burgholzer: @coderabbitai @MatthiasReumann The weak ownership is intentional: the cache is only a reuse index, while live device and job state owns the loaded library. This lets multiple sessions reuse an already loaded provider without forcing the library to remain loaded after the final live object disappears.

I did find a real problem in the original implementation: the cache was manager-local, so separate managers could initialize the same library independently. It is now process-wide, keyed by canonical library path and symbol prefix, and guarded during lookup and construction. If no live object remains, the weak reference expires and reopening the provider is the expected behavior. A regression test covers reuse across simultaneously live managers and equivalent library paths.

Following the lifetime reasoning above, I did not put manager-owned unique_ptrs into this map. Doing so would couple library lifetime to one manager and conflict with the requirement that returned devices and jobs can outlive it.

The map has instead moved out of DeviceManager into the process-wide library loader and contains weak references. Strong ownership remains exclusively with live device and job state, so the final such object controls unloading.

Agreed that the original mutex was not a complete synchronization design. The responsibilities are now separated:

- a manager-local registry mutex protects definition snapshots, registration, unregistration, and lookup;

- definitions() returns a value snapshot instead of exposing a reference;

- a separate process-wide cache mutex protects canonical library lookup and initialization across all managers.

Session creation remains independent after the shared provider library has been obtained. Tests cover concurrent opens, concurrent registry access, independent sessions, and process-wide library reuse. This defines the manager’s guarantees without attempting to serialize all calls made through independent provider sessions.

include/mqt-core/qdmi/DeviceManager.hpp
burgholzer: Done. Discovery and registration now live in DeviceRegistry.hpp and DeviceRegistry.cpp, while DeviceManager.hpp and DeviceManager.cpp contain only opening, session creation, and manager lifetime behavior. I used the repository’s .hpp convention.

burgholzer:
@MatthiasReumann I am through with another iteration here. Feel free to take another look.
Note that this does _not_ yet touch the SC and NA devices and makes them configurable. That would be a follow-up.
I still believe this is shaping up quite nicely.

I may try to split this into two changes, one that is v3.x compatible but adds the configurability and a separate one that contains the breaking change of dropping the client interface and combining driver and FoMaC.
That also might be a nice separation of concerns.

Also pinging @marcelwa @flowerthrower @ystade here for awareness.
If you want a very quick overview of the changes, read https://github.com/munich-quantum-toolkit/core/blob/agent/qdmi-integration-redesign/UPGRADING.md#qdmi-device-management-has-been-redesigned.

src/qdmi/DeviceManager.cpp
MatthiasReumann: I think the rationale is getting clearer for me now, thanks! I have some question / thoughts though:

 > The weak ownership is intentional: the cache is only a reuse index, while live device and job state owns the loaded library.

Now I get why the `DeviceApi` is being used as a `shared_ptr` - this makes totally sense! 

> I did find a real problem in the original implementation: the cache was manager-local, so separate managers could initialize the same library independently. It is now process-wide, keyed by canonical library path and symbol prefix, and guarded during lookup and construction. If no live object remains, the weak reference expires and reopening the provider is the expected behavior. A regression test covers reuse across simultaneously live managers and equivalent library paths.

Just to make sure that I understand you fully here: Alternatively we could have made the `DeviceManager` a singleton, right? Probably depends on if we ever want to create DeviceManager's using different registries. 

> Following the lifetime reasoning above, I did not put manager-owned unique_ptrs into this map. Doing so would couple library lifetime to one manager and conflict with the requirement that returned devices and jobs can outlive it.

I think we should then probably find a more suitable name than `DeviceManager`. I don't know, something like `DeviceExplorer`? Following we should document how to best use the `DeviceManager`. Currently, I imagine something like 

```cpp
auto findAllDevices(....) {
  DeviceManager m(...); 
  return m.openAll(...);
}
```

src/qdmi/DeviceApi.hpp
MatthiasReumann, line 1: `.hpp` file in `src` folder

src/qdmi/DeviceApi.cpp
MatthiasReumann, line 1: This file still uses trailing return types.

MatthiasReumann (changes requested):
Solid improvements! 🚀

I've left two minor requests and some questions / thoughts in the open discussion above. 

Otherwise, I've nothing to complain than we should probably add a bit more documentation on the caching, `shared_ptr`, lifetime logic to the respective classes. I am pretty sure Codex can help with that ;)

src/qdmi/DeviceManager.cpp
burgholzer: 🤖 *AI text below* 🤖

The rebased design now makes this split explicit. `DeviceManager` is an immutable, non-singleton snapshot of a `DeviceRegistry`; it contains no mutable library cache and opens a fresh session for each call. Returned devices and jobs own the shared session and library state they require, so they may outlive the manager.

This keeps explicit registries isolated and makes short-lived use such as the example above valid. The process-default catalog is exposed separately through free functions. We kept `DeviceManager` because it opens sessions and reports per-device bulk-open failures. The intended use and lifetime contract are documented in `DeviceManager.hpp`, `docs/qdmi/driver.md`, and `docs/qdmi/configuration.md`.

Tests cover registry isolation, snapshot stability, fresh sessions, wrapper-independent lifetimes, and concurrent access.

src/qdmi/DeviceApi.hpp
burgholzer: 🤖 *AI text below* 🤖

Fixed. The private implementation header is now `src/qdmi/DeviceApi.hpp`, and all includes use the `.hpp` name.

src/qdmi/DeviceApi.cpp
burgholzer: 🤖 *AI text below* 🤖

Fixed. `src/qdmi/DeviceApi.cpp` now uses conventional return declarations throughout. The complete non-unity build and Clang-Tidy run passed after the change.

src/qdmi/DeviceApi.hpp
burgholzer: LLMs at their best. This was obviously not the intended consequence here.
But I still think it is fine to have the private header in the source file here. Otherwise, people would get the idea that they can include this header.

burgholzer:
I am closing this PR for now. The discussions in the recent weeks have led to the conclusion that the client interface is indeed worth keeping. The evidence for that is #2226, #2229, #2230, #2231
The QDMI v1.4 release will enable truly exchangeable QDMI drivers, which requires a stable (C) interface; the client interface. We are still tweaking the wording around all of this.
But we are no longer conflating the Driver implementation and the C++ QDMI abstraction.

PR #1910, opened 2026-07-15
🧪 Establish a typed OpenQASM 3 frontend

### LINKED ISSUE #1846: ✨ Translate `for` and `while` from OpenQASM to QC

burgholzer:
🤖 *AI text below* 🤖

## Summary

This PR replaces MQT Core's OpenQASM-to-MLIR translation with a staged, typed OpenQASM frontend that emits verified QC IR directly.

The frontend is intentionally a supported-subset implementation rather than a full OpenQASM syntax implementation. It is designed so that parsing, source-language semantics, QC target capabilities, and downstream compiler limitations remain distinct:

1. an LLVM-native lexer and handwritten grammar parser build a persistent syntax representation from `llvm::SourceMgr` buffers;
2. a separate, MLIR-independent semantic analyzer resolves scopes, symbols, types, conversions, constants, initialization, gate calls, broadcasting, and control-flow state into a typed program;
3. a private QC emitter lowers that typed program to QC and standard MLIR dialects;
4. `translateQASM3ToQC(...)` remains the small public convenience entry point.

There is no OQ3 MLIR dialect and no OQ3-to-QC conversion pass. OpenQASM is an input language, not an intermediate representation. The legacy `QuantumComputation` importer remains independent.

The living implementation record is in [`.agent/plans/oq3-foundation.md`](.agent/plans/oq3-foundation.md), while [`docs/mlir/OpenQASM.md`](docs/mlir/OpenQASM.md) contains the maintained feature and target-support matrices.

## Resolved issues

Closes #1846.

## Implemented frontend

- Preserves filenames, line/column locations, nested include stacks, and stage-specific syntax, semantic, and QC-target diagnostics.
- Treats explicit `OPENQASM 3.0;`, explicit `OPENQASM 3.1;`, and versionless input as the maintained OpenQASM 3 profile.
- Retains OpenQASM 2 compatibility, including legacy declarations, measurement arrows, conditionals, gate spellings, register broadcasting, and `qelib1.inc`.
- Supports configurable include handling through `llvm::SourceMgr`, with distinct `stdgates.inc` and `qelib1.inc` catalogs and deliberate compatibility leniency between source modes.
- Supports `qubit`, `bit`, `bool`, `int`, `uint`, and `float` declarations; constants; assignment; explicit semantic conversions; lexical scope; definite initialization; explicit outputs; and compatibility-oriented implicit global classical outputs.
- Preserves output declaration order. Bit outputs use the QC classical-register representation, scalar outputs retain builtin MLIR scalar types, and non-output bits remain SSA values.
- Lowers scalar `qubit q;` to `qc.alloc`, while explicitly sized declarations such as `qubit[1] q;` remain register-backed.
- Supports primitive, compatibility, broadcast, and custom gates with cycle, arity, scope, and broadcast-width validation.
- Uses one canonical gate catalog for semantic lookup and QC lowering, including native `cu`, `cu3`, and `cu1` handling and the OpenQASM 2/3 U-family phase conventions.
- Supports ordered and nested `inv`, `ctrl`, `negctrl`, and `pow` modifiers. Valid power modifiers emit `qc.pow`, including dynamic exponents, while unsupported downstream power bodies fail at the owning conversion boundary.
- Supports measurement declarations and assignments, legacy measurement arrows, targetless measurement, reset, and barrier.
- Supports `if`, `while`, and inclusive range-based `for`, including mutable scalar, bit, and quantum state across structured regions.
- Supports static and dynamic qubit/bit indexing with source-level bounds checks and compact structured dispatch.
- Implements scalar arithmetic, comparisons, short-circuit Boolean operators, exponentiation, trigonometric and inverse-trigonometric functions, `exp`, `sqrt`, `ceiling`, and `floor`.
- Implements typed whole-register `popcount`, `rotl`, and `rotr`, including dynamic signed rotation distances and registers wider than 64 bits.
- Diagnoses bounded resource limits for expression, block, modifier, gate-dependency, register-storage, typed-statement, and projected-emission growth.

## QC and downstream behavior

The primary acceptance path is:

```text
OpenQASM -> QC -> QCO -> optimized QCO -> reconstructed QC
         -> Adaptive QIR -> LLVM IR / bitcode
```

Straight-line programs additionally exercise Base QIR. `jeff` serialization is an optional compatibility path rather than a prerequisite for accepting valid source:

- the standard corpus includes structured control flow, runtime and induction-variable indexing, dynamic ranges, checked integer arithmetic, loop-carried classical state, custom gates, and power modifiers;
- runtime source preconditions are retained as assertions and lower through the standard QIR path;
- a smaller positive corpus round-trips through `jeff`;
- known `jeff` limitations are tested as explicit failures at QCO-to-`jeff`, rather than being rejected by the frontend or QC emitter.

Mixed physical and declared qubits are accepted by source analysis but currently rejected at the QC target because QC cannot mix static and dynamic allocation modes. Arbitrary scalar QIR output recording and bit outputs computed from general classical expressions remain downstream ABI/capability follow-ups.

## Deliberately unsupported source families

The lexer reserves and diagnoses unsupported language families instead of silently ignoring them. The current subset does not yet implement:

- width-qualified integer and floating-point declarations;
- `input`, subroutines, `return`, or `extern`;
- arrays, complex declarations, source-level angle declarations, or aliases;
- `switch`, `break`, or `continue`;
- timing, boxes, delays, durations, stretches, annotations, pragmas, or calibration constructs;
- bit-string literals, bit-register declaration initializers, and the broader cast/initializer matrix.

Measurement is supported in its statement contexts, not as a general expression. Bitwise and shift operators that require currently unsupported explicitly sized operands are parsed and rejected semantically.

## Independent infrastructure fixes

Several general MLIR/QIR defects discovered while validating this frontend were extracted, reviewed, and merged independently. This branch now consumes their upstream implementations and no longer carries divergent copies:

- [#1975](https://github.com/munich-quantum-toolkit/core/pull/1975): preserve classical loop state in QCO-to-QC;
- [#1976](https://github.com/munich-quantum-toolkit/core/pull/1976): exclude classical loads from QC-to-QCO quantum-state tracking;
- [#1977](https://github.com/munich-quantum-toolkit/core/pull/1977): retain the intentional MLIR slicing lint policy;
- [#1978](https://github.com/munich-quantum-toolkit/core/pull/1978): lower Math operations in Base and Adaptive QIR and register the Math dialect in `mqt-cc`;
- [#1979](https://github.com/munich-quantum-toolkit/core/pull/1979): preserve returned classical-register order during QIR recording.

Qubit reuse remains separate from this PR.

## Validation

The latest post-merge local validation ran 1,148 affected tests:

- 265 QC translation tests;
- 144 parser, semantic-analysis, and direct-emission tests;
- 204 compiler-pipeline tests;
- 137 QCO-to-QC tests;
- 139 QC-to-QCO tests;
- 119 Base QIR tests;
- 140 Adaptive QIR tests.

The complete repository lint session, diff checks, and an `mqt-cc` Math-dialect parse/emit smoke test also pass. The frontend and emitter additionally have behavior-driven coverage for diagnostics, source locations, includes, precedence, conversions, scopes, initialization, modifiers, custom gates, control flow, dynamic indexing, outputs, resource bounds, and downstream target boundaries.

## AI assistance

Codex materially assisted with implementation, test generation, conflict resolution, and repeated independent review under maintainer direction. AI-assisted commits carry `Assisted-by` trailers, and the changes remain subject to human review and responsibility.


mlir/include/mlir/Dialect/OQ3/Transforms/Passes.h
```diff
@@ -0,0 +1,36 @@
...
+namespace mlir {
+class Pass;
+namespace oq3 {
+
+/**
+ * @brief Options controlling conversion of typed OpenQASM IR to QC.
+ */
+struct OpenQASMLoweringOptions {
+  /// Whether the selected target can diagnose a zero step at runtime.
+  bool supportsRuntimeAssertions = false;
```
denialhaag: This flag is not used anywhere. Might as well delete it.

mlir/lib/Dialect/OQ3/Transforms/LowerOQ3ToQC.cpp
denialhaag, line 1: Is there any reason that this is not a conversion? See, for example, `qc-to-qco` for more information on how to define a conversion.

mlir/unittests/Dialect/OQ3/test_oq3.cpp
```diff
@@ -0,0 +1,250 @@
...
+    PassManager manager(context.get());
+    manager.addPass(oq3::createLowerOQ3ToQCPass());
+    ASSERT_TRUE(succeeded(manager.run(module.get()))) << gate.name.str();
+    EXPECT_TRUE(succeeded(verify(module.get()))) << gate.name.str();
+
+    bool hasOQ3GateApplication = false;
+    module->walk([&](oq3::ApplyGateOp) { hasOQ3GateApplication = true; });
+    EXPECT_FALSE(hasOQ3GateApplication) << gate.name.str();
+  }
+}
```
denialhaag: This test wouldn't be necessary if the lowering were implemented as a conversion that marks the OQ3 dialect as illegal. There are more tests like this that wouldn't be necessary.

cmake/ExternalDependencies.cmake
```diff
@@ -29,6 +29,7 @@ if(BUILD_MQT_CORE_MLIR)
     GIT_REPOSITORY https://github.com/PennyLaneAI/jeff-mlir.git
     GIT_TAG v0.3.0)
   list(APPEND FETCH_PACKAGES jeff-mlir)
+
 endif()
```
denialhaag:
```suggestion
endif()
```

docs/mlir/OQ3.md
denialhaag, line 1: This doesn't look right. See, e.g., `QC.md` in the same directory for how to do it.

mlir/include/mlir/Dialect/OQ3/IR/OQ3Ops.td
```diff
@@ -0,0 +1,99 @@
...
+                          SingleBlockImplicitTerminator<"::mlir::oq3::YieldOp">,
+                          AllTypesMatch<["start", "stop", "step"]>]> {
+  let summary = "Inclusive OpenQASM integer range loop";
+  let arguments = (ins AnyInteger:$start, AnyInteger:$stop, AnyInteger:$step);
+  let regions = (region SizedRegion<1>:$body);
+  let assemblyFormat = [{
+    $start `to` $stop `step` $step $body attr-dict `:` type($start)
+  }];
+  let hasVerifier = 1;
+}
```
denialhaag: If I'm not missing anything, this definition exists solely to eventually reject the case of a non-constant step. Is there really no other way to achieve the same? It would be nice if we could use `scf::ForOp` directly.

mlir/include/mlir/Dialect/OQ3/Transforms/CMakeLists.txt
denialhaag, line 1: As per my other comment, the entire `OQ3/Transforms` directory should not be necessary. Try to define the lowering as a conversion instead.

mlir/include/mlir/Target/OpenQASM/Frontend.h
denialhaag, line 1: Is there any reason to prefer `std::vector` over `llvm::SmallVector` here?

mlir/lib/Dialect/OQ3/IR/OQ3Ops.cpp
denialhaag, line 1: Depending on how much this is expected to grow, it might make sense to create a file per operation, similarly to how we do it for QC and QCO.

mlir/lib/Dialect/OQ3/Transforms/LowerOQ3ToQC.cpp
denialhaag, line 1: This is a general comment, but I noticed it here:

- Never declare MLIR's core IR types as `const`: https://mlir.llvm.org/docs/Rationale/UsageOfConst/
- Prefer using `auto` over `Value`, `ValueRange`, etc. whenever possible

mlir/unittests/CMakeLists.txt
denialhaag, line 17: Could you copy over all the test cases added in #1862 and amend `test_qasm3_translation.cpp` accordingly?

mlir/lib/Target/OpenQASM/Frontend.cpp
denialhaag, line 1: This is relying on the legacy parser. Do we want that? For example, we do not get the benefit of the MLIR/LLVM error handling. This was one of the triggers for one of the refactors in #1862.

mlir/lib/Target/OpenQASM/Frontend.cpp
```diff
@@ -0,0 +1,996 @@
...
+
+} // namespace
+
+ParseResult parseOpenQASM(llvm::SourceMgr& sourceMgr) {
+  ParseResult result;
+  try {
+    const auto* buffer = sourceMgr.getMemoryBuffer(sourceMgr.getMainFileID());
+    const auto contents = buffer->getBuffer();
+    std::istringstream input(
+        std::string(std::string_view(contents.data(), contents.size())));
```
denialhaag: Same comment here: We wanted to avoid using `std::istringstream` in #1862 and instead went with LLVM concepts.

mlir/lib/Target/OpenQASM/Frontend.cpp
denialhaag, line 1: As implemented right now, I don't think the parser can handle for and while loops, right? This is another shortcoming of the legacy parser. The lowering of for and while loops in `LowerOQ3ToQC.cpp` would never be reached.

denialhaag (changes requested):
Very excited if we can get this to work nicely! 🙂 

The comments below are a brain dump and not necessarily a structured review. I'm aware that some of the points I'm raising might have been planned for future iterations, but I wanted to mention everything that came to mind.

mlir/lib/Dialect/OQ3/IR/OQ3Ops.cpp
```diff
@@ -0,0 +1,205 @@
...
+  if (!type) {
+    return emitOpError("requires a function type");
+  }
+  if (type.getNumResults() != 0) {
+    return emitOpError("gate declarations cannot return values");
+  }
+  return success();
+}
+
+LogicalResult ApplyGateOp::verify() {
```
denialhaag: One more thing I just noticed: What's the overlap of this function with `analyzeGateApplication` in `Frontend.cpp`? Are we (partially) doing semantic analysis twice now?

The same might apply to the other OQ3 operations.

denialhaag: If semantic analysis does not rely on MLIR but can solely be done by `SemanticAnalyzer` in `Frontend.cpp`, how necessary is the OQ3 dialect in the first place? Could the `SemanticAnalyzer` already reject everything we don't support right now? Loosely speaking, could the functions in `LowerOQ3ToQC.cpp` just be copied into `OpenQASM.cpp`, and we achieve the exact same goal?

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -0,0 +1,1901 @@
...
+  [[nodiscard]] Value extendRangeValue(Value value, Type targetType,
+                                       const bool isUnsigned) {
+    if (isUnsigned) {
+      return arith::ExtUIOp::create(builder, targetType, value);
+    }
+    return arith::ExtSIOp::create(builder, targetType, value);
+  }
+
+  void emitFor(const frontend::ForStatement& loop, ValueRange gateParameters,
+               ValueRange gateQubits) {
```
denialhaag, line 2629: It would be nice to have a simplified version of this if start, stop, and step are statically known. Otherwise, we cannot convert any for loop to `jeff` because it lacks an assert operation.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -0,0 +1,1901 @@
...
+      lhs = coerceScalar(lhs, lhsType, frontend::ScalarType::Float);
+      rhs = coerceScalar(rhs, rhsType, frontend::ScalarType::Float);
+      const auto predicate = [&] {
+        switch (condition.comparison) {
+        case frontend::ComparisonKind::Equal:
+          return arith::CmpFPredicate::OEQ;
+        case frontend::ComparisonKind::NotEqual:
+          return arith::CmpFPredicate::UNE;
+        case frontend::ComparisonKind::Less:
+          return arith::CmpFPredicate::OLT; // spellchecker:disable-line
```
denialhaag: I think it makes more sense to add this to `[tool.typos.default.extend-words]`.

docs/mlir/OpenQASM.md
```diff
@@ -0,0 +1,56 @@
...
+| `array`, `complex`, `angle`, and aliases                 | Recognized and rejected by the parser | Aggregate, complex, angle, and `let` alias declarations have no typed representation yet                 | `RejectsUnsupportedReservedWordsAsIdentifiers`         |
+| `switch`, `break`, and `continue`                        | Recognized and rejected by the parser | These control-flow forms are reserved but are not in the implemented grammar                             | `RejectsUnsupportedReservedWordsAsIdentifiers`         |
+| Bitwise and shift operators                              | Parsed and rejected semantically      | Explicitly sized `uint`, `bit`, or `angle` operands are required by the language and are not implemented | `RejectsInvalidProgramsAcrossSemanticFamilies`         |
+
+## Translation and compiler support
+
+“Adaptive plus Jeff” means the tested public path from QC through optimized QCO,
+Jeff byte serialization and deserialization, back to QC, and finally to Adaptive
+QIR. Base refers to direct production of the QIR Base Profile.
```
denialhaag: Please use "`jeff`" whenever "`" is recognized (e.g., in Markdown files and in Docstrings). Otherwise, use "jeff" (e.g., in error messages).

.agent/plans/oq3-foundation.md
```diff
@@ -0,0 +1,744 @@
...
+  reconstructed QC but Adaptive QIR reported that no entry point existed. The
+  native Jeff regression now proves that nonempty result types and the marker
+  survive together.
+
+- Observation: the Jeff representation cannot preserve the frontend's runtime
+  `cf.assert` bounds checks. Evidence: genuinely runtime-dynamic indexing
+  reaches verified QC and QCO but QCO-to-Jeff rejects `cf.assert`; an index
+  resolved by cleanup traverses the complete chain. The direct emitter now
+  accepts only indices proven resolvable by conservative scalar dataflow and
+  reports a source-located target diagnostic for the remainder.
```
denialhaag: I don't think this is a good solution. Dynamic indices should not be rejected at the time of QC emission. At most, reject the conversion to `jeff`.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -0,0 +1,1901 @@
...
+
+  [[nodiscard]] bool
+  reportRuntimeIntegerCheck(const oq3::frontend::SourceLocation& source) const {
+    llvm::errs()
+        << source.filename << ':' << source.line << ':' << source.column
+        << ": OpenQASM QC emission error: checked integer arithmetic "
+           "and ranges are not supported by the complete QC/QCO/Jeff/QIR "
+           "compiler path.\n";
+    return false;
+  }
```
denialhaag: As pointed out elsewhere, one boundary (in this case, `jeff`) should not reject the emission.

mlir/include/mlir/Target/OpenQASM/Detail/OpenQASMLexer.h
denialhaag, line 1: Can you move all header files from this directory to `mlir/include/mlir/Target/OpenQASM`?

mlir/include/mlir/Target/OpenQASM/Detail/OpenQASMUnicode.h
denialhaag, line 1: General comment: We usually use `size_t` instead of `std::size_t`. The same also holds for `int64_t` and so on.

denialhaag (dismissed):
I like where this is going! 🙂 

You can find some more quick feedback below. Again, this is not necessarily a structured review.

mlir/lib/Conversion/QCToQCO/QCToQCO.cpp
denialhaag, line 1: Could you explain these changes? Did you find a bug in the current version of the conversion? If so, it might be nice to create a separate PR for this fix so that we can merge it immediately.

burgholzer:
Alright, I believe this has come pretty far.
It puts our OpenQASM 3 support on a much stronger footing that is independent of the MQT Core IR and QASM sub-packages. As such, it lifts many of the legacy limitations and adds support for a broader range of OpenQASM 3 features.
I'll go ahead and merge this for now so that the potential for conflicts is reduced as much as possible.
There will definitely come quite a couple of follow-ups from this to extend support and clean up the implementation even further.

@munich-quantum-toolkit/mqt-cc for awareness 💡

PR #1912, opened 2026-07-15
✨ Add configurable QDMI device discovery

### LINKED ISSUE #1362: ✨ Authentication per QDMI device
### Problem Statement

#1355 introduces support for QDMI authentication via FoMaC. Currently, any provided method of authentication is applied to all available devices.

### Proposed Solution

We need a per-device authentication mechanism. This concerns FoMaC, as well as the Qiskit Provider.

### LINKED ISSUE #1363: ✨ `BASEURL` support through QDMI authentication

burgholzer:
🤖 *AI text below* 🤖

## Summary

This PR extracts the additive, MQT Core 3.x-compatible QDMI configuration
foundation from #1901. It adds stable device identities and configurable
discovery while retaining the current client interface, `qdmi::Driver`, FoMaC,
Qiskit, and neutral-atom APIs.

The breaking object-model refactor remains in #1901 and will be rebased onto
this foundation.

## What changed

- Discover versioned definitions from `qdmi.json`, `[tool.qdmi]`, environment
  configuration, packaged fragments, and C++ runtime overrides.
- Merge definitions by stable device ID with field-wise precedence, explicit
  disabled-ID reservation, strict validation, relative-path resolution, and
  explicit or isolated configuration modes.
- Register and open devices by stable ID through C++, FoMaC, and Python. Fresh
  opens create independent sessions and apply per-call overrides without
  changing the configured client catalog.
- Generate relocatable fragments for built-in devices and export neutral
  `QDMI_DEVICE_ID`, `QDMI_DEVICE_PREFIX`, and `QDMI_MANIFEST_NAME` CMake target
  metadata. Installed consumers can copy selected device runtimes without
  project-specific loader code.
- Vendor the current toml++ single header used by the native TOML parser.
- Document administration, project, environment, C++, Python, relocation, and
  static-consumer workflows.

## Compatibility and impact

The public QDMI client C interface, `MQT::CoreQDMIDriver`, the FoMaC object
model, and the existing Qiskit and neutral-atom APIs remain available. The
unstable C++ `Driver::addDynamicDeviceLibrary` and Python
`add_dynamic_device_library` entry points are replaced by `DeviceDefinition`
registration and stable-ID opening; `UPGRADING.md` documents that migration.

Configuration discovery and registration do not load device code. A library is
loaded when a stable ID is explicitly opened or when the configured client
catalog is first materialized. Separately installed device wheels are not
scanned automatically; they can register a fallback definition or have their
runtime and generated fragment colocated by the consuming build.

## Validation

- Release configure and complete 471-step build with LLVM/MLIR 22.1.3
- QDMI Driver suite: 109 passed
- Registry suite: 12 passed
- FoMaC suite: 174 passed
- Neutral-atom FoMaC suite: 2 passed
- Imported-device consumer tests: 2 passed
- DDSIM suite: 45 passed
- QIR JIT, runtime, and runner suites: 9, 63, and 5 passed
- Python stub generation through the aggregate wheel target
- `uvx nox -s lint`
- `git diff --check`

Fixes #1362
Fixes #1363


include/mqt-core/qdmi/DeviceRegistry.hpp
```diff
@@ -0,0 +1,108 @@
...
+  /// Runtime definitions layered at the highest precedence.
+  std::vector<DeviceDefinition> runtimeOverrides;
+  /// Whether discovery excludes packaged built-in manifests.
+  bool isolated = false;
+};
+
+/// Discovers and merges QDMI device definitions without loading libraries.
+class DeviceRegistry {
+public:
+  explicit DeviceRegistry(const ConfigOptions& options = {});
```
MatthiasReumann: Nitpick: Because the `DeviceRegistry` is only instantiated without parameters in the private constructor of the `Driver`, we might want to drop the options logic here.

MatthiasReumann (dismissed):
Thanks for the effort 🙏 

I've left one small remark, otherwise I think this one looks good 👍🏻

include/mqt-core/qdmi/DeviceRegistry.hpp
burgholzer: Took this for another spin. I think this is all looking quite clean now.

PR #1914, opened 2026-07-16
🎨 Improve `mqt-cc` Python bindings

denialhaag:
## Description

This PR makes largely cosmetic improvements to #1815.

## Checklist

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -56,14 +56,6 @@ Its functionality has only ever been used in [MQT QMAP] since its inception. As
 a consequence, the code shall be moved to [MQT QMAP] once QMAP adopts an MQT
 Core version that includes this change.
 
-### QDMI child devices
-
-The QDMI driver now translates device-library-specific `QDMI_Child_Device`
-handles into client-facing `QDMI_Device` handles backed by dedicated child
-sessions. Direct child devices can be queried through
-`fomac::Device::getChildDevices()` in C++ and `Device.child_devices()` in
-Python. Devices without child-device support continue to behave unchanged.
-
 ### Dev container
 
 A [dev container](https://containers.dev/) configuration is available to provide
@@ -72,6 +64,14 @@ a consistent local development environment. Common IDEs like
 and [VS Code](https://code.visualstudio.com/docs/devcontainers/containers) can
 open the repository directly inside the container. If you are on Windows, we
 recommend using Docker Desktop with the WSL 2 backend.
+
+### QDMI child devices
+
+The QDMI driver now translates device-library-specific `QDMI_Child_Device`
+handles into client-facing `QDMI_Device` handles backed by dedicated child
+sessions. Direct child devices can be queried through
+`fomac::Device::getChildDevices()` in C++ and `Device.child_devices()` in
+Python. Devices without child-device support continue to behave unchanged.
 
 ## [3.7.0]
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -56,6 +56,14 @@ Its functionality has only ever been used in [MQT QMAP] since its inception. As
 a consequence, the code shall be moved to [MQT QMAP] once QMAP adopts an MQT
 Core version that includes this change.
 
+### QDMI child devices
+
+The QDMI driver now translates device-library-specific `QDMI_Child_Device`
+handles into client-facing `QDMI_Device` handles backed by dedicated child
+sessions. Direct child devices can be queried through
+`fomac::Device::getChildDevices()` in C++ and `Device.child_devices()` in
+Python. Devices without child-device support continue to behave unchanged.
+
 ### Dev container
 
 A [dev container](https://containers.dev/) configuration is available to provide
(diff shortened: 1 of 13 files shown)


denialhaag:
@burgholzer, I finally had the time to carefully go through #1815. This PR comprises largely cosmetic changes and shouldn't be too controversial.

I noticed two more things: 

1. Do we really want to establish `.qco` files as a thing? 🤔 https://github.com/munich-quantum-toolkit/core/blob/559fde4b245a01a958d85ba7d28e13f1c4eb99a4/mlir/tools/mqt-cc/mqt-cc.cpp#L64-L67
2. It would be nice to look into whether we can reduce some code duplication between `Programs.cpp` and `mqt-cc.cpp`. In an ideal world, `mqt-cc.cpp` could have little to no logic (apart from all the `llvm::cl` stuff). This would also allow us to test the logic that `mqt-cc` runs without needing to write CLI tests. I'd look into this in a follow-up PR so that this one can hopefully go in immediately. 😌

bindings/mlir/register_mlir.cpp
```diff
@@ -331,7 +326,7 @@ before conversion to QCO.)pb");
           "circuit"_a,
           nb::sig("def from_qiskit(circuit: qiskit.circuit.QuantumCircuit) "
                   "-> QCProgram"),
-          "Translate a Qiskit QuantumCircuit to QC MLIR.")
+          "Translate a Qiskit ``QuantumCircuit`` to QC MLIR.")
```
burgholzer: We could use actual Sphinx cross-references for these if we want. Probably overkill as the function parameters carry type hints in the docs already, but I'll let you be the judge of that.

bindings/mlir/register_mlir.cpp
```diff
@@ -302,11 +297,11 @@ before conversion to QCO.)pb");
       .def_static(
           "from_mlir_str",
           &OptionalFunctionAdapter<&mlir::QCProgram::fromMLIRString>::call,
-          "source"_a, "Parse a QC-dialect MLIR source string.")
+          "source"_a, "Parse a QC MLIR source string.")
```
burgholzer, line 300:
```suggestion
          "source"_a, "Parse a QC dialect MLIR source string.")
```
To be consistent with the edit below.

bindings/mlir/register_mlir.cpp
```diff
@@ -302,11 +297,11 @@ before conversion to QCO.)pb");
...
       .def_static(
           "from_mlir_file",
           &OptionalFunctionAdapter<&mlir::QCProgram::fromMLIRFile>::call,
-          "path"_a, "Parse QC-dialect MLIR from a file.")
+          "path"_a, "Parse QC MLIR from a file.")
```
burgholzer, line 304:
```suggestion
          "path"_a, "Parse QC dialect MLIR from a file.")
```
To be consistent with the edit below.

UPGRADING.md
burgholzer, line 1: Any particular reason for this change? Both of these are part of the [Unreleased] section.
(We might have forgotten to actually move the Dev container release note to a release. I am pretty sure we back ported and released this already)

burgholzer (changes requested):
Nice. Thanks @denialhaag 🙏🏼 
I only have minor nitpicks inline.

> I noticed two more things:
> 
> 1. Do we really want to establish `.qco` files as a thing? 🤔 https://github.com/munich-quantum-toolkit/core/blob/559fde4b245a01a958d85ba7d28e13f1c4eb99a4/mlir/tools/mqt-cc/mqt-cc.cpp#L64-L67

Hm. Not necessarily I suppose. My main thought was that this allows us to distinguish `.mlir` (which will typically mean QC-dialect programs) from QCO-dialect programs. I am not hard set on this though. Feel free to drop the `.qco` support. In that case, I'd probably add some auto-detection for `.mlir` files that would go over the module and scan for either QC or QCO and construct the program correspondingly.  
  
> 2. It would be nice to look into whether we can reduce some code duplication between `Programs.cpp` and `mqt-cc.cpp`. In an ideal world, `mqt-cc.cpp` could have little to no logic (apart from all the `llvm::cl` stuff). This would also allow us to test the logic that `mqt-cc` runs without needing to write CLI tests. I'd look into this in a follow-up PR so that this one can hopefully go in immediately. 😌

Yeah. Avoiding this kind of duplication would be nice.
I am not 100% sure if it is feasible though.
`mqt-cc` can directly use the whole MLIR concepts and pipelines.
The "Programs" abstraction is largely necessary for the bindings and the plug and play in Python.
If we were to switch to the official MLIR Python bindings at some point, it could even be that we entirely drop the "Programs" abstraction again and just add some useful helpers on top of the official bindings.

UPGRADING.md
denialhaag: Not really, I just figured the dev containers should be at the bottom because they change the developer experience and not the user experience.

We did not backport the dev containers because the configuration would have been different for `v3.x`.

bindings/mlir/register_mlir.cpp
denialhaag: Thanks for catching that! I adjusted the QCO docstrings below instead. Now, they are all consistent, also with `QCProgram.from_mlir_file()`, `QCProgram.from_qasm_str()`,  `QCProgram.from_qasm_file()`, `QCProgram.from_quantum_computation()`, and `QCProgram.from_qiskit()`. If we prefer having "dialect," I can add it everywhere. 🤔

bindings/mlir/register_mlir.cpp
denialhaag: See above.

bindings/mlir/register_mlir.cpp
denialhaag: Good point! Done.

denialhaag:
> Feel free to drop the `.qco` support. In that case, I'd probably add some auto-detection for `.mlir` files that would go over the module and scan for either QC or QCO and construct the program correspondingly.

Done! Let me know what you think!

UPGRADING.md
burgholzer: Ah, now I remember 👍🏼 
It doesn't really matter too much, because the child device PR is already back ported and will be part of the next minor release of MQT Core; at which point we will rewrite the upgrade guide on main anyway.

denialhaag: Ahh, good point! I'll revert the change then!

burgholzer (approved):
LGTM now 👍🏼

PR #1935, opened 2026-07-24
🐛 Preserve structured state in QC-to-QCO conversion

burgholzer:
🤖 *AI text below* 🤖

## Summary

- Preserve existing classical iteration arguments and results when the QC-to-QCO conversion appends explicit QCO and QTensor state to `scf.for` and `scf.while`.
- Preserve the distinct input and result signatures of type-changing `scf.while` operations, including reordered `scf.condition` arguments.
- Lower affected structured terminators in a second conversion phase so their operands use the final region state instead of depending on dialect-conversion traversal order.
- Teach QCO mapping to distinguish classical loop state from qubit state, preserve classical terminator operands during hot routing, and extend type-changing `scf.while` regions with the correct before/after signatures.
- Add focused pure-QC conversion regressions and mixed-state mapping regressions for `for` and `while`.
- Add this general conversion fix to the existing MLIR infrastructure changelog entry.

## Context

This conversion gap was found while working on OpenQASM control flow in #1910. The fix is independent of that frontend: any QC producer can create valid `scf.for` or `scf.while` operations whose existing classical values must survive conversion while quantum state is made explicit.

## Scope

This PR is limited to stable `scf.for` and `scf.while` state preservation. It intentionally does not add classical results to `qco.if` or `qco.index_switch`, and it introduces no scratch allocations, stores, or loads. Classical `qco.if` results are being designed separately as an SSA-based extension.

## Validation

- QC-to-QCO conversion tests: 132 passed
- QCO mapping tests: 13 passed
- Focused structured-control-flow conversion regressions: 4 passed
- Focused mixed-state mapping regressions: 2 passed
- `uvx nox -s lint`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,7 @@ releases may include breaking changes.
   [#1626], [#1627], [#1635], [#1638], [#1673], [#1675], [#1700], [#1717],
   [#1728], [#1730], [#1749], [#1751], [#1762], [#1765], [#1780], [#1781],
   [#1782], [#1806], [#1807], [#1815], [#1808], [#1824], [#1869], [#1872],
-  [#1886], [#1914], [#1925]) ([**@burgholzer**], [**@denialhaag**],
+  [#1886], [#1914], [#1925], [#1935]) ([**@burgholzer**], [**@denialhaag**],
   [**@taminob**], [**@DRovara**], [**@li-mingbao**], [**@Ectras**],
   [**@MatthiasReumann**], [**@simon1hofmann**], [**@J4MMlE**])
 
@@ -651,6 +651,7 @@ changelogs._
 
 <!-- PR links -->
 
+[#1935]: https://github.com/munich-quantum-toolkit/core/pull/1935
 [#1933]: https://github.com/munich-quantum-toolkit/core/pull/1933
 [#1925]: https://github.com/munich-quantum-toolkit/core/pull/1925
 [#1924]: https://github.com/munich-quantum-toolkit/core/pull/1924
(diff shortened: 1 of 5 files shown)


burgholzer:
@MatthiasReumann @denialhaag for awareness. Follow-up for IfOp is coming.
If you spot any refinements, feel free to follow-up 😌

denialhaag:
> - Lower affected structured terminators in a second conversion phase so their operands use the final region state instead of depending on dialect-conversion traversal order.

Opus 4.8 brought up a similar point while I was working on #1927. I still have to understand whether there is really no form of a guaranteed conversion order. If that were the case, the QC-to-QCO conversion at large could fall apart at any moment. We _heavily_ rely on the fact that each program is converted from top to bottom. 🤔

burgholzer:
> > - Lower affected structured terminators in a second conversion phase so their operands use the final region state instead of depending on dialect-conversion traversal order.
> 
> 
> 
> Opus 4.8 brought up a similar point while I was working on #1927. I still have to understand whether there is really no form of a guaranteed conversion order. If that were the case, the QC-to-QCO conversion at large could fall apart at any moment. We _heavily_ rely on the fact that each program is converted from top to bottom. 🤔 

Yeah, me too. I was skeptical at first. But the regression tests in this PR did not pass without the changes to the pass, and the changes themselves feel sound.
Feel free to keep the investigation for that on some todo list 👍🏻

denialhaag:
> > > - Lower affected structured terminators in a second conversion phase so their operands use the final region state instead of depending on dialect-conversion traversal order.
> > 
> > 
> > 
> > Opus 4.8 brought up a similar point while I was working on #1927. I still have to understand whether there is really no form of a guaranteed conversion order. If that were the case, the QC-to-QCO conversion at large could fall apart at any moment. We _heavily_ rely on the fact that each program is converted from top to bottom. 🤔 
> 
> Yeah, me too. I was skeptical at first. But the regression tests in this PR did not pass without the changes to the pass, and the changes themselves feel sound.
> Feel free to keep the investigation for that on some todo list 👍🏻

No no, the changes also look good to me, and the MLIR docs also don't suggest that there is a conversion order. At this point, I'm wondering more why the rest of the QC-to-QCO conversion is working as intended. But yeah, I'll investigate a bit more and then report back! 👍

PR #1938, opened 2026-07-24
✨ Support classical results in QCO conditionals

burgholzer:
🤖 *AI text below* 🤖

## Summary

- Allow `qco.if` and `qco.index_switch` to return ordinary classical SSA values
  before their existing linear quantum results.
- Preserve result-bearing `scf.if` and `scf.index_switch` operations through
  QC-to-QCO and QCO-to-QC conversion without introducing scratch memory.
- Update verification, parsing and printing, RegionBranch interfaces,
  tied-value helpers, builders, mapping, tensor traversal, module equivalence,
  and affected optimizations for the segmented result model.
- Keep unsupported Jeff lowering explicit without constraining the standard
  QC → QCO → QC → QIR pipeline.

## Motivation

This capability was identified while working on the OpenQASM integration in
#1910 and while reducing the structured-control-flow fixes that became #1935.
QCO conditionals previously represented only linear quantum results. Preserving
classical SCF results therefore required either rejecting otherwise valid
programs or lowering classical state through temporary memory.

The new representation keeps classical values in ordinary SSA form while
retaining QCO's explicit single-use quantum flow:

1. classical result prefix;
2. tied linear quantum result suffix.

Conditional region arguments remain linear-only. Classical inputs continue to
use normal SSA capture, while `qco.yield` returns the complete
classical-plus-linear result signature.

## Implementation notes

- Both conditional operations use `AttrSizedResultSegments` and expose explicit
  classical and linear result accessors.
- Custom parsers infer the linear suffix from the `args(...)` assignments and
  retain quantum-only textual compatibility.
- Parent-aware `qco.yield` verification preserves the stricter modifier
  contracts.
- QC↔QCO conversions retain classical def-use chains directly and replace only
  linear results with QC references on the QC side.
- Mapping handles every index-switch case and the default region, preserving
  classical yield operands while realigning only quantum values.
- Module equivalence compares classical yield prefixes positionally while
  keeping the linear suffix permutation-aware.

## Validation

All affected targets build successfully. The focused suites pass 933 tests in
total:

- 450 QCO IR tests
- 134 QC-to-QCO tests
- 132 QCO-to-QC tests
- 82 QCO utility tests
- 3 QTensor utility tests
- 15 mapping tests
- 117 Jeff round-trip tests

Repository lint, changed-source `clang-tidy`, and `git diff --check` pass. Two
independent read-only reviews were performed; the final review found no
remaining actionable issues.

### DIFF
--- a/.agent/plans/qco-if-classical-results.md
+++ b/.agent/plans/qco-if-classical-results.md
@@ -63,10 +63,10 @@ their consumers while QCO's explicit single-use quantum flow remains intact.
 - [x] (2026-07-24 22:47Z) Move the QC-to-QCO-to-QC composition regressions into
       a dedicated round-trip test target so each directional conversion test
       links only its own conversion library.
-- [ ] Rebuild the affected targets, run focused suites and repository lint,
-      independently verify the follow-up, and record the final evidence
-      (completed: 718 affected tests, changed-source clang-tidy, repository
-      lint, and `git diff --check`; remaining: independent verification).
+- [x] (2026-07-24 22:53Z) Rebuild the affected targets, pass 718 affected tests,
+      pass changed-source clang-tidy and repository lint, run
+      `git diff --check`, and complete an independent read-only verification of
+      the follow-up.
 
 ## Surprises & Discoveries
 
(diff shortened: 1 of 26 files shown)


burgholzer:
This feels very reasonable to me and it passes all tests. I am proceeding to merge.

@MatthiasReumann for awareness as this also touches the IndexSwitchOp and its mapping. Hope this does not tae away too much of the work you had already done to support it.

PR #1944, opened 2026-07-25
⚡️ Improve Python CI performance

burgholzer:
🤖 *AI text below* 🤖

## Summary

- Update all reusable workflow pins to the immutable v2.2.1 release commit.
- Share one scikit-build-core build directory across Python test sessions.
- Configure coverage.py's `sysmon` core and parallel data collection centrally.
- Require pytest-cov 7.1 for the coverage configuration.

The shared build directory lets the minimum- and regular-resolution nox sessions reuse the same CMake and Ninja state. On Windows, reusable workflows v2.2.1 initialize native MSVC and use Ninja so the existing job-local `sccache` installation is active.

## Evidence

The final test-only experiment passed on macOS and Windows in [run 30434315809](https://github.com/munich-quantum-toolkit/core/actions/runs/30434315809). The full Core CI run for the same source revision also passed on every supported runner in [run 30434310296](https://github.com/munich-quantum-toolkit/core/actions/runs/30434310296).

The strongest reuse was on Windows: after the minimum-resolution builds, all five regular-resolution sessions completed in 47 seconds and `sccache` reported a 35.93% hit rate in [run 30404816970](https://github.com/munich-quantum-toolkit/core/actions/runs/30404816970).

## Validation

- [x] `uv lock --check`
- [x] Full Core lint session
- [x] macOS and Windows performance experiment
- [x] Full Core CI matrix
- [x] No mutable workflow references or benchmark-only workflow remain


pyproject.toml
```diff
@@ -135,6 +135,7 @@ BUILD_MQT_CORE_BINDINGS = "ON"
 BUILD_MQT_CORE_TESTS = "OFF"
 BUILD_MQT_CORE_SHARED_LIBS = "ON"
 BUILD_MQT_CORE_MLIR = "ON"
+DISABLE_GIL = "0"
```
burgholzer: This should not be necessary and shall be removed.
```suggestion
```

Issue #1867, opened 2026-07-08
♻️ Layout Improvements: Injective Mapping and Non-Consecutive Qubit Indices


MatthiasReumann:
The `Layout` class currently has the following issues:
- The bidirectional mapping is implemented using `SmallVector`s. This assumes consecutive qubit indices. However, one can easily imagine a scenario where this isn't the case.
- The mapping enforces bijection. However, usually `hardware.size() >= program.size()`. The Layout class should somehow reflect this property.

PR #1951, opened 2026-07-27
✨ Implement "vote-and-restore" strategy for `qco.index_switch`

MatthiasReumann:
## Description

This pull request implements the "vote-and-restore" strategy for `qco.index_switch` operation. The final layout of each case region is considered as a "vote" of the respective case (the "voter") and the average (winning) layout is determined using the Borda Rank.

## Checklist


MatthiasReumann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1028,6 +1032,30 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+    const auto ncandidates = (*layouts.begin()).nqubits();
+
+    SmallVector<size_t> scores(ncandidates, 0);
+    for (const Layout& layout : layouts) {
+      for (const auto [rank, hw] : enumerate(layout.getProgramToHardware())) {
+        scores[hw] += ncandidates - rank - 1;
+      }
+    }
+
+    auto mapping = llvm::to_vector(llvm::seq(ncandidates));
```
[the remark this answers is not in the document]
MatthiasReumann: The layouts are assumed to be of equal size.

MatthiasReumann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1298,47 +1337,37 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
             OpOperand* const qubit = ifOp.getTiedQubit(res);
-            for (size_t i = 0; i < 2; ++i) {
-              Region& region = ifOp.getRegion(i);
-
-              const auto arg = region.getRegionNumber() == 0
-                                   ? ifOp.getTiedThenBlockArgument(qubit)
-                                   : ifOp.getTiedElseBlockArgument(qubit);
-              children[i].infos.insertOrUpdate(children[i].infos.size(), prog);
-              children[i].wires.emplace_back([&] -> Value {
+            for (size_t r = 0; r < 2; ++r) {
```
burgholzer: Would unrolling this loop (which only has two iterations) lead to cleaner code?

burgholzer (approved):
This looks really clean!
I just have one minor observation where I'd like your input. Otherwise this can go in!

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: I think a lambda might be the way to go here! Let me know what you think!

burgholzer:
🤖 *AI text below* 🤖

### Q-01: does the Borda vote have the intended semantics for `Layout`?

The current implementation treats `getProgramToHardware()` as a preference ranking: position 0 receives the highest weight, position 1 the next-highest weight, and so on. However, those positions are program-qubit indices, not stated preferences. In effect, the hardware location of program qubit 0 contributes more points than the location of program qubit 1, etc.

That makes the selected layout sensitive to an otherwise semantics-preserving renaming of logical qubits. The hardware-index tie-break can amplify this. The examples below use `[h0, h1, ...]` to mean “program qubit `i` is mapped to hardware qubit `hi`.” Restore costs are minimum adjacent-SWAP counts on a line topology.

#### Three-qubit example

The voters

- `[0, 1, 2]`
- `[0, 1, 2]`
- `[1, 2, 0]`

produce Borda scores `[4, 4, 1]`, so the hardware-index tie-break selects `[0, 1, 2]`.

After a cyclic renaming of program qubits, the equivalent voters are

- `[2, 0, 1]`
- `[2, 0, 1]`
- `[0, 1, 2]`

Equivariance under renaming would make the winner `[2, 0, 1]`, i.e., the correspondingly renamed original winner. The current Borda calculation instead selects `[0, 2, 1]`. On a three-node line, restoring all voters to the equivariant winner costs 2 adjacent SWAPs in total; restoring them to the Borda winner costs 7.

A smaller transposition example also changes the result: voters `[0, 2, 1]`, `[0, 2, 1]`, `[1, 0, 2]` should inherit winner `[0, 2, 1]` from the renamed original election, but the current tie-break selects `[0, 1, 2]` (line cost 2 versus 3).

#### Four-qubit example

The voters

- `[0, 1, 2, 3]`
- `[0, 1, 3, 2]`
- `[0, 2, 1, 3]`

select `[0, 1, 2, 3]`. After the cyclic program renaming, the equivalent voters are

- `[3, 0, 1, 2]`
- `[2, 0, 1, 3]`
- `[3, 0, 2, 1]`

The correspondingly renamed winner is `[3, 0, 1, 2]`, but the current vote selects `[0, 3, 2, 1]`. On a four-node line, the total minimum restore cost is 2 for the equivariant winner and 16 for the Borda winner.

So the central question is: **is prioritizing lower-numbered program qubits intentional?** If it is, the current rule should probably be documented as such rather than described as an “average” layout. If it is not, a topology-aware alternative would be to choose the branch-layout medoid: select the child layout minimizing the sum of `restore` costs from all child layouts, with an explicit deterministic tie-break. Benchmarking parent restoration, the current Borda rule, and such a medoid on representative mapped programs would make the trade-off measurable.

The added regression test deliberately does not decide this question: its routed case layouts are identical, so it only proves that the vote-and-restore path is actually used.

MatthiasReumann:
@burgholzer 
> Q-01: does the Borda vote have the intended semantics for `Layout`?

I am not really sure where this "cycling renaming" comes from, but here's how I see the problem:

Each of the `N` cases ends with a layout. Each layout is equivalent to a program-to-hardware mapping 
```
prog (0, 1, 2, 3, ..., M)
  hw (0, 1, 2, 3, ..., M) // Using the identity here.
```

The goal is to find a "average" "winning" layout - or in other words - a program-to-hardware mapping. Finding this mapping should be somewhat efficient / fast. 

The [Borda rank](https://en.wikipedia.org/wiki/Borda_count) assigns points based on the position of the candidate (here: hardware index) on its ballot (here: program-to-hardware mapping). The program-qubit with index `0` will always be at position `0` and program-qubit with index `M` will always be at position `M` in this voting system. Consequently, from a program-qubit perspective `0` will always win and `M` will always loose. 

Thus, only the order of hardware indices determines the points. For the example given by Cody 
- U: `(2, 0, 1)`
- V: `(2, 0, 1)`
- W: `(0, 1, 2)`

the Borda Ranking is

| Candidate | U Points | V Points | W points | Total |
|--------|--------|--------|--------|--------|
| 0 | 1 | 1 | 2 | 3 |
| 1 | 0 | 0 | 1 | 2 |
| 2 | 2 | 2 | 0 | 4 |

The winner is thus
```
prog (0, 1, 2)
  hw (2, 0, 1)
```

Given that the program indices are ordered as discussed above, this is the correct choice.

burgholzer:
🤖 *AI text below* 🤖

I put the entire reasoning through another round of scutinizing and took some time to go through the reasoning, see below. To me the argumentation here makes sense and the proposed restauration strategy feels more natural and performant. What do you think?

@MatthiasReumann Thanks for laying out your interpretation. I think there are two separate points here: one arithmetic issue in the example and one more fundamental question about what the aggregation should optimize.

First, for the three ballots

- U: `(2, 0, 1)`
- V: `(2, 0, 1)`
- W: `(0, 1, 2)`

I get the following totals:

| Hardware candidate | U | V | W | Total |
|---|---:|---:|---:|---:|
| 0 | 1 | 1 | 2 | 4 |
| 1 | 0 | 0 | 1 | 1 |
| 2 | 2 | 2 | 0 | 4 |

Thus, candidates 0 and 2 tie with four points. Since the implementation breaks Borda ties by the lower hardware index, it returns `(0, 2, 1)`, not `(2, 0, 1)`.

The more fundamental question is whether program-qubit indices should serve as preference ranks in the first place.

The implementation interprets a layout `(h0, h1, ..., hM)` as a ballot ranking hardware candidates according to the program index occupying them. Consequently, the hardware occupied by program qubit 0 always receives the maximum points, while the hardware occupied by the last program qubit receives none. Program qubits are not actually candidates in this election; their numeric indices define the preference order.

That is a valid Borda construction if ascending program index is intentionally an ordinal ranking. However, in the mapping pass, program indices are identities derived from extraction order. I do not see a reason why program qubit 0 should have greater importance than program qubit 1, or why the numeric distance between program indices should be meaningful for restoration.

This is where the renaming argument comes from. If we consistently rename the logical qubits throughout the program and every branch layout, we have only renamed the tokens being routed. The hardware topology and the required physical movement have not changed. The restoration cost between two layouts is invariant under such a consistent renaming. Therefore, a convergence heuristic intended to minimize restoration overhead should ideally produce the correspondingly renamed result.

This problem is not limited to Borda ties. For example, consider:

- A: `(0, 1, 2)`
- B: `(0, 2, 1)`
- C: `(1, 0, 2)`

The Borda scores are `(5, 3, 1)`, so the result is strictly `(0, 1, 2)`.

Now consistently swap program-qubit identities 1 and 2. The equivalent layouts are:

- A′: `(0, 2, 1)`
- B′: `(0, 1, 2)`
- C′: `(1, 2, 0)`

Their scores are `(4, 3, 2)`, so Borda again strictly returns `(0, 1, 2)`. The correspondingly renamed original winner would be `(0, 2, 1)`. There are no score ties in either calculation. On a three-qubit line, restoring to the renamed original winner requires two adjacent SWAPs in total, while restoring to the new Borda result requires three.

There are also counterexamples that do not rely on renaming at all. Consider three branch layouts on a four-qubit line:

- A: `(0, 2, 1, 3)`
- A again: `(0, 2, 1, 3)`
- B: `(1, 3, 0, 2)`

Despite two of the three branches agreeing exactly on A, Borda produces scores `(7, 5, 4, 2)` and selects the identity layout `(0, 1, 2, 3)`, which no branch proposed.

Restoring all branches to A requires two adjacent SWAPs in total. Restoring them to the Borda result requires five. This is tie-free and shows that the Borda objective can disagree substantially with the restoration cost we ultimately care about.

Given that minimizing inserted SWAPs and restoration overhead is the actual objective, would it make more sense to choose a topology-aware medoid among the branch layouts?

Concretely, for every child layout `Lj`, we could evaluate

`cost(Lj) = sum_i restore(Li, Lj).size()`

and select the child layout with the smallest total cost. The restoration sequences for the winning candidate could be cached and inserted directly, avoiding another round of computation.

This has several properties that seem attractive here:

- it directly minimizes the quantity we subsequently insert;
- it accounts for the actual device topology through `restore`;
- it is invariant under consistently renaming program qubits;
- it always selects an already existing, valid branch layout;
- it cannot manufacture a consensus layout that is farther from every relevant branch under the restoration metric.

The selection requires quadratically many `restore` evaluations in the number of regions. That seems acceptable to me for an `index_switch`, although we should benchmark it for switches with many regions. For equal-cost candidates, we could use stable region order as a deterministic tie-break rather than program- or hardware-index ordering.

What do you think about replacing the Borda consensus with this branch-layout medoid? If compile time turns out to be a concern, we could compare the two strategies on representative switches before deciding, but the medoid seems more closely aligned with the optimization goal of the pass.

burgholzer:
Side note: The PR itself is obviously functionally correct and passes CI. So I am also happy to continue to merge this after resolving the conflicts and defer to a follow up.

MatthiasReumann:
@burgholzer 

Many thanks! Besides me failing to add three numbers, I now know where this is coming from 👍🏻

I think the "topology-aware" argument is the strongest one here. I will try to find a solution for that.  

However, I am not sure if I am a fan of the medoid idea as this will add a significant overhead for many-case switches. Then, one could argue "just restore to the origin" is a better runtime / quality tradeoff. 

Otherwise, let's merge this for now to get #1867 going again.

burgholzer:
> @burgholzer 
> 
> 
> 
> Many thanks! Besides me failing to add three numbers, I now know where this is coming from 👍🏻
> 
> 
> 
> I think the "topology-aware" argument is the strongest one here. I will try to find a solution for that.  
> 
> 
> 
> However, I am not sure if I am a fan of the medoid idea as this will add a significant overhead for many-case switches. Then, one could argue "just restore to the origin" is a better runtime / quality tradeoff. 
> 
> 
> 
> Otherwise, let's merge this for now to get #1867 going again. 

I agree, let's merge for now and improve in a follow-up! 👍🏻

I am not so sure if the overhead would really be that bad. But, to be honest, all of this is hypothetical. I have not yet really seen a use of the index switch in an algorithm, which would allow us to judge what to prioritize for performance.

PR #1930, opened 2026-07-23
⚡️ Improve `walkProgramGraph` driver performance

MatthiasReumann:
## Description

I spent this morning investigating and profiling the mapping pass using the benchmark I recently created and noticed a few interesting bits. As a consequence I tried to optimize as much as possible - overall a ~30% reduction in runtime on the set of circuits I benchmarked on (See below for table). 

Moreover, I've also learned that `isa`, `cast`, and consequently `TypeSwitch` on interfaces is really expensive in MLIR as it always traverses a sorted array of interface libraries to find the correct context (See [here](https://github.com/llvm/llvm-project/blob/38b7237069f2e51ee8562cbce2014ae29e6459b5/mlir/include/mlir/Support/InterfaceSupport.h#L268) for the MLIR implementation). I would expect another maybe not so insignificant improvement in runtime by swapping the UnitaryInterface with `isa<XOp, YOp, ...>`. To put this in numbers: It requires roughly the same amount of CPU cycles to perform the cast to `UnitaryOpInterface` as to manage the DenseMap for the pending wires. 

Consequently, I tried to avoid redundant `isa` and `cast` operation for the interface in the driver function. Additionally, I moved the `isa<AllocOp, ...>` elsewhere to avoid the check for each iteration step (also noticed while profiling). 

## Benchmarks 

Using 12x10 qubit grid as architecture and
```cpp
const mlir::qco::MappingPassOptions options{.nlookahead = 20,
                                            .lambda = 0.5,
                                            .niterations = 1,
                                            .ntrials = 18,
                                            .seed = 39393939};
```
```
| circuit                                                | time_branch | time_main | time_branch / time_main |
|--------------------------------------------------------|-------------|-----------|-------------------------|
| bmw_quark_copula_nativegates_ibm_falcon_opt2_100       | 420         | 583       | 0.72                    |
| bmw_quark_copula_nativegates_ibm_falcon_opt2_30        | 21          | 27        | 0.78                    |
| bmw_quark_copula_nativegates_ibm_falcon_opt2_40        | 32          | 43        | 0.74                    |
| bmw_quark_copula_nativegates_ibm_falcon_opt2_50        | 68          | 88        | 0.77                    |
| bmw_quark_copula_nativegates_ibm_falcon_opt2_60        | 89          | 122       | 0.73                    |
| bmw_quark_copula_nativegates_ibm_falcon_opt2_70        | 162         | 217       | 0.75                    |
[...]

PR #1955, opened 2026-07-27
⚡️ Add MLIR mapping performance benchmark

burgholzer:
🤖 *AI text below* 🤖

## Summary

- add deterministic frontier-heavy and routing-heavy MLIR mapping workloads
- add a paired A/B runner with randomized within-pair order, raw samples, robust summaries, and bootstrap confidence intervals
- record source-worktree state, executable SHA-256 digests, and compiler/CMake metadata
- document the timing boundary and the separate correctness and route-quality validation requirements

## Motivation

Follow-up to #1930. The local harness used during that review was useful for separating genuine mapping-pass improvements from behavior changes. Publishing it provides a shared starting point for iterating on representative workloads, measurement methodology, and performance-regression evaluation.

The benchmark is opt-in through `BUILD_MQT_CORE_BENCHMARKS` and does not change production behavior.

## Validation

- configured and built `mqt-core-mlir-mapping-eval` in release mode with MLIR 22.1.3
- exercised both scenarios through the paired runner
- passed all applicable repository hooks for the changed files, including CMake formatting, clang-format, Ruff, ty, Markdown, policy, and lockfile checks
- independently reviewed the exact branch head

## Draft iteration points

- workload representativeness and additional mapping/routing scenarios
- route-quality metrics alongside timing
- cross-platform build and metadata handling
- suitable CI integration, if any, as a separate decision

### DIFF
--- a/eval/mlir_mapping.cpp
+++ b/eval/mlir_mapping.cpp
@@ -1,0 +1,189 @@
+/*
+ * Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
+ * Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#include "mlir/Dialect/QCO/Builder/QCOProgramBuilder.h"
+#include "mlir/Dialect/QCO/IR/QCODialect.h"
+#include "mlir/Dialect/QCO/Transforms/Mapping/Mapping.h"
+
+#include <llvm/ADT/DenseSet.h>
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/Support/CommandLine.h>
+#include <llvm/Support/LogicalResult.h>
+#include <mlir/Dialect/Arith/IR/Arith.h>
+#include <mlir/Dialect/Func/IR/FuncOps.h>
+#include <mlir/Dialect/SCF/IR/SCF.h>
+#include <mlir/IR/BuiltinOps.h>
+#include <mlir/IR/DialectRegistry.h>
+#include <mlir/IR/MLIRContext.h>
+#include <mlir/IR/Verifier.h>
+#include <mlir/Pass/PassManager.h>
+
+#include <algorithm>
+#include <chrono>
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <numeric>
+#include <random>
+#include <string>
+#include <tuple>
+#include <utility>
+
+using namespace mlir;
+using namespace mlir::qco;
+
+namespace {
+
+llvm::cl::opt<std::string>
+    scenario("scenario", llvm::cl::desc("Workload: frontier or routing"),
+             llvm::cl::value_desc("name"), llvm::cl::init("frontier"));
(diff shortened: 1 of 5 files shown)


burgholzer:
@MatthiasReumann FYI: this is what GPT-5.6-sol came up with when asked for a benchmarking suite for the mapping pass.
I haven't looked too deeply myself, but I would be interested in how that compares to what you have set up locally and whether this might be something to further extend.

eval/mlir_mapping.cpp
```diff
@@ -0,0 +1,189 @@
...
+  if (failed(result)) {
+    llvm::errs() << "mapping failed\n";
+    return 1;
+  }
+  if (failed(verify(module.get()))) {
+    llvm::errs() << "mapped module verification failed\n";
+    return 1;
+  }
+  std::cout << elapsed.count() << '\n';
+  return 0;
```
MatthiasReumann, line 188: Instead of wrapping `pm.run` we could also use MLIR's pass instrumentation. In my local benchmarks I implemented it as follows.

```cpp
struct PassExecutionStats : public mlir::PassInstrumentation {
  PassExecutionStats(Statistics &stats) : stats_(&stats) {}

  void runBeforePass([[maybe_unused]] mlir::Pass *pass,
                     mlir::Operation *op) override {
    nswapsPrev_ = countSwaps(op);
    t0_ = std::chrono::steady_clock::now();
  }

  void runAfterPass([[maybe_unused]] mlir::Pass *pass,
                    mlir::Operation *op) override {
    const auto t1 = std::chrono::steady_clock::now();
    const auto duration =
        std::chrono::duration_cast<std::chrono::milliseconds>(t1 - t0_).count();
    const auto nswapsAfter = countSwaps(op);

    stats_->emplace("time", duration);
    stats_->emplace("nswaps", nswapsAfter - nswapsPrev_);
  }

private:
  std::size_t countSwaps(mlir::Operation *op) {
    std::size_t cnt{0};
    op->walk([&cnt](mlir::qco::SWAPOp) { ++cnt; });
    return cnt;
  }

  Statistics *stats_;
  std::chrono::steady_clock::time_point t0_;
  std::size_t nswapsPrev_;
};
```

```cpp
int main() {
  ...
  pm.addInstrumentation(
          std::make_unique<support::PassExecutionStats>(stats));
}
```

Not sure if this brings any real benefit, but I think `mlir-timing` is also implemented this way.

eval/mlir_mapping.cpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+    if (i + width < count) {
+      couplingSet.insert({i, i + width});
+      couplingSet.insert({i + width, i});
+    }
+  }
+  return couplingSet;
+}
+
+OwningOpRef<ModuleOp> makeProgram(MLIRContext& context) {
```
MatthiasReumann, line 87: I really like this idea! Eventually it would be really awesome if we could simply use `mqt-bench` in Python to generate QASM circuits and run the mapping pass one those. Maybe this is even possible today - Codex could probably implement this with relative ease!

MatthiasReumann (commented):
Thanks @burgholzer! Looks really promising 👀 

I've left two comments - nothing dramatic - with some additional ideas.

eval/mlir_mapping.cpp
MatthiasReumann: If this was possible, we could also think about using this setup as a test-framework instead of the current unit-tests.

PR #1956, opened 2026-07-27
✨ Extend Layout API for partial injections

rturrado:
🤖 *AI text below* 🤖

## Description

Extend `Layout` to support `M <= N` program-to-hardware mappings where `N - M` hardware slots stay unmapped:

- Unmapped hardware entries in `hardwareToProgram_` carry an `UNMAPPED` sentinel.
- Split `nqubits()` into `nProgramQubits()` and `nHardwareQubits()`. `random()` takes both sizes.
- Add `hasProgramAt(hw)` to distinguish mapped from unmapped hardware slots.

A possible follow-up could exercise `nProg < nHw` and the new query API from inside the mapping pass. 

Part of #1867

## Checklist

Assisted-by: Claude Opus 4.7 via Claude Code


burgholzer:
@rturrado Thanks for another PR! 🙌

@MatthiasReumann care to take a first look here since this is closest to the stuff you have been working on? 😌

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1378,69 +1396,32 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
-              insertSWAPs<Mode>(snd, children[1], stats, rewriter);
-              return convergedLayout;
-            })
-            .template Case<IndexSwitchOp>([&](IndexSwitchOp) {
-              for (auto& child : children) {
-                const auto swaps = restore(child.layout, parent.layout);
-                insertSWAPs<Mode>(swaps, child, stats, rewriter);
-              }
-              return parent.layout;
-            });
```
MatthiasReumann: The converge strategy for `qco.if` and the restore strategy only of the "yielding" branch of the `scf.while` operation is definitely not a bug but a feature.

MatthiasReumann (commented):
@rturrado Thanks for the effort 🚀 Really appreciate it!  

I've left one comment regarding the strategies for mapping SCF which probably shouldn't be reverted to "restore-all". The essential idea is that the SCF operations (besides `scf.for`) act like permutation networks changing the layout (which needs to be propagated to the parent). If you have any questions, feel free to reach out!

rturrado:
> @rturrado Thanks for another PR! 🙌

Thanks! This one touches a really beautiful area.

> @MatthiasReumann care to take a first look here since this is closest to the stuff you have been working on? 😌

But I have to admit that it is also more complicated code, so human review here is essential.

rturrado:
> @rturrado Thanks for the effort 🚀 Really appreciate it!

Many thanks!

> I've left one comment regarding the strategies for mapping SCF which probably shouldn't be reverted to "restore-all". The essential idea is that the SCF operations (besides `scf.for`) act like permutation networks changing the layout (which needs to be propagated to the parent). If you have any questions, feel free to reach out!

Perfect, thanks! Yes, this was my main concern about this PR. That the new wire-index-equals-hardware-index invariant always has to restore to the parent layout and never converges. That `which needs to be propagated to the parent` constraint is key here.

MatthiasReumann:
@rturrado 
> That the new wire-index-equals-hardware-index invariant always has to restore to the parent layout and never converges.

I think the "wire-index-equals-hardware-index" is a pretty neat idea, which eventually I would have also looked into. In a previous version of the mapping pass (before SCF mapping), we implemented "wire-index-equals-program-index" which worked pretty nicely. 

Nonetheless, I think it would make sense to split this PR into two:
- The layout improvements + everything else regarding `getComputation`, etc.
-  The "wire-index-equals-hardware-index" logic.

Especially since it's very likely that #1951 is merged before this one.

rturrado:
🤖 *AI text below* 🤖

@MatthiasReumann Agreed on the split. Rough sketch:

PR 1 (safe): `Layout` API additions (`nProgramQubits`, `nHardwareQubits`, `hasProgramAt`, `swap` with an empty side), the `random(nProg, nHw, seed)` variant that only places `nProg` real programs (still with placeholder fill), and any `getComputation` cleanup that does not depend on the invariant. Keeps `WireInfos` and the current `dispatch` (`converge` for `qco.if`, restore-yielding-only for `scf.while`, and `vote-and-restore` for `qco.index_switch` from #1951).

PR 2 (invariant): `wire index == hardware index` everywhere, sentinel `WireIterator`, delete `WireInfos`, `random` drops the placeholder fill, rewrite `dispatch` for the invariant.

- Would you want PR 1 now, or should I wait for #1951 to land first?
- Happy to help with PR 2 in the future.

Assisted-by: Claude Opus 4.7 via Claude Code

burgholzer:
I think we'll try to get #1951 in asap (either tonight or tomorrow). Then PR1 can go on top 😌

rturrado:
> I think we'll try to get #1951 in asap (either tonight or tomorrow). Then PR1 can go on top 😌

Perfect. Don't rush it. I'll rebase when you're done with #1951.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -913,8 +915,20 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
     SmallVector<IndexPairType, 6> expansionSet;
 
+    const auto materializeKey = [](const Layout& layout) {
+      SmallVector<size_t> key(layout.nHardwareQubits());
+      for (size_t prog = 0; prog < layout.nHardwareQubits(); ++prog) {
+        key[prog] = layout.getHardwareIndex(prog);
+      }
+      return key;
+    };
+
```
MatthiasReumann: Couldn't we use `getProgramToHardware` instead as it now returns a `SmallVector<size_t>`? For the mapping pass it must always hold `nHardwareQubits() == nProgramQubits()`.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -913,8 +915,20 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
     frontier.emplace(root);
 
     DenseMap<ArrayRef<size_t>, size_t> bestDepth;
+    // Owns the materialized layout snapshots that `bestDepth` keys point into.
+    // std::deque never invalidates pointers to existing elements on push_back,
+    // so ArrayRefs already in `bestDepth` stay stable as we grow it.
+    std::deque<SmallVector<size_t>> keyStorage;
```
MatthiasReumann:
```suggestion
    std::deque<SmallVector<size_t>> keys;
```

Personal preference; not a must-have.

mlir/include/mlir/Dialect/QCO/Utils/Layout.h
```diff
@@ -61,28 +75,42 @@ class Layout {
...
   /// Maps a program qubit index to its hardware index.
-  SmallVector<size_t> programToHardware_;
+  DenseMap<size_t, size_t> programToHardware_;
   /// Maps a hardware qubit index to its program index.
-  SmallVector<size_t> hardwareToProgram_;
+  DenseMap<size_t, size_t> hardwareToProgram_;
+  /// Number of program qubits this layout was declared with.
+  size_t nProgramQubits_ = 0;
+  /// Number of hardware qubits this layout was declared with.
+  size_t nHardwareQubits_ = 0;
```
MatthiasReumann: Why are these required? Shouldn't `nProgramQubits` (`nHardwareQubits`) also rather return `programToHardware_.size()` (`hardwareToProgram_.size()`)?

mlir/lib/Dialect/QCO/Utils/Layout.cpp
```diff
@@ -38,44 +47,74 @@ Layout Layout::fromMapping(ArrayRef<size_t> mapping) {
...
+  hardwareToProgram_.erase(hwB);
+  if (hasA) {
+    hardwareToProgram_[hwB] = progA;
+    programToHardware_[progA] = hwB;
+  }
+  if (hasB) {
+    hardwareToProgram_[hwA] = progB;
+    programToHardware_[progB] = hwA;
+  }
 }
```
MatthiasReumann, line 110: I think this function should assume (assert) that both hardware qubits have program qubits assigned instead of inserting `0`s.

MatthiasReumann (changes requested):
Great work. Looking forward for merging this 🚀

**Some notes / ideas:**
We can now reasonably argue that the program and hardware qubits are a consecutive range of numbers (#1993). That is, a `Layout` holds `[M] → [N]`, where `M <= N`. Maybe this allows us to avoid `DenseMap`s and use `SmallVector`s again. The `Layout` is an essential data structure in the mapping pass, so we need to make sure the implementation is very efficient. 

Moreover, if we now already invest the time to improve this data structure, it could also make sense to add a generalized `InjectiveMap<K, V>`, where the `Layout` would then be a subclass of that (If that is even necessary). Using templates, we may need to keep `DenseMap`; or specialize on integer types using `SmallVector`. Just an idea - maybe a follow-up!

rturrado:
@MatthiasReumann Many thanks for the thorough review and the guidance. This PR was just the rebase onto `main` and doesn't yet take into account the new invariant from #1993: that program and hardware qubits are now consecutive range of numbers. I'll update the PR accordingly.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
rturrado: Fair enough!

mlir/lib/Dialect/QCO/Utils/Layout.cpp
rturrado: Done. `swap` now asserts both hardware slots have program qubits assigned.

burgholzer:
@MatthiasReumann @rturrado Except for the minor conflict with `main`, what is the status here? This looks like it would be pretty close to ready, right?

rturrado:
> @MatthiasReumann @rturrado Except for the minor conflict with `main`, what is the status here? This looks like it would be pretty close to ready, right?

@burgholzer Thanks for flagging this. I'll have a look into that conflict!

mlir/include/mlir/Dialect/QCO/Utils/Layout.h
```diff
@@ -20,23 +20,36 @@
...
+
+  /// Construct and return a random layout that maps every program qubit
+  /// index in `[0, nProgramQubits)` to a distinct hardware index drawn from
+  /// `[0, nHardwareQubits)`.
+  static Layout random(size_t nProgramQubits, size_t nHardwareQubits,
+                       size_t seed);
 
   /// Construct a layout from a program-to-hardware mapping,
   /// where mapping[prog] = hw.
+  /// Sets both `nProgramQubits` and `nHardwareQubits` to `mapping.size()`.
```
MatthiasReumann:
```suggestion
  /// Construct a layout from a bijective program-to-hardware mapping,
  /// where mapping[prog] = hw.
  /// Sets both `nProgramQubits` and `nHardwareQubits` to `mapping.size()`.
```

Another option would be to explicitly allow the caller to set some program qubits to `UNMAPPED`. This would require to change the semantics of mapping to `mapping[hw] = prog`.

mlir/unittests/Dialect/QCO/Utils/test_layout.cpp
```diff
@@ -11,31 +11,142 @@
...
+}
+
+TEST(LayoutTest, RandomLeavesExtraHardwareUnmapped) {
+  constexpr size_t nProg = 2;
+  constexpr size_t nHw = 5;
+  const auto layout = Layout::random(nProg, nHw, /*seed=*/0);
+
+  const auto mappedCount =
+      std::ranges::count_if(std::views::iota(size_t{0}, nHw),
+                            [&](size_t hw) { return layout.hasProgramAt(hw); });
```
MatthiasReumann: Nitpick: Could we do this with `llvm::` utilities? For example, `llvm::seq` should be able to replace `iota`.

mlir/unittests/Dialect/QCO/Utils/test_layout.cpp
```diff
@@ -11,31 +11,142 @@
...
+            ArrayRef<size_t>(mapping));
   EXPECT_EQ(layout.getProgramIndex(0), 1);
   EXPECT_EQ(layout.getProgramIndex(1), 2);
   EXPECT_EQ(layout.getProgramIndex(2), 0);
 }
 
 TEST(LayoutDeathTest, RejectDuplicateHardwareIndex) {
   constexpr std::array<size_t, 3> mapping{0, 0, 2};
-  EXPECT_DEATH((void)qco::Layout::fromMapping(mapping),
+  EXPECT_DEATH((void)Layout::fromMapping(mapping),
```
MatthiasReumann: Nitpick: Any way to avoid that C-style cast here?

MatthiasReumann (changes requested):
Thanks for your continued work on this! @rturrado  Much appreciated🙏

The pull request looks really good already! I've only left some very nitpicky comments, which should be fairly easy to resolve. Lastly, I think a changelog entry is missing also. Otherwise, let's get this PR merged ASAP 🛫

burgholzer:
> Lastly, I think a changelog entry is missing also. 

In that regard, please just fold this PR number into the existing changelog entry for the mapping pass (and add your name to it) 😌

mlir/include/mlir/Dialect/QCO/Utils/Layout.h
rturrado: Thanks, applied the `bijective` addition.

On `mapping[hw] = prog`: agreed as a natural next step when a caller needs partial injection through `fromMapping`. Today no caller does; `Mapping.cpp` only uses it with pure permutations. Leaving it for the follow-up that will start exercising `M < N` from the mapping pass.

mlir/unittests/Dialect/QCO/Utils/test_layout.cpp
rturrado: Beautiful. Done, replaced both `std::ranges::count_if` / `std::views::iota` with `llvm::count_if` / `llvm::seq`.

mlir/unittests/Dialect/QCO/Utils/test_layout.cpp
rturrado: Very well spotted. Done, dropped the cast. `fromMapping` is not `[[nodiscard]]`, so calling it as a plain statement compiles cleanly and the cast was defensive.

rturrado:
> In that regard, please just fold this PR number into the existing changelog entry for the mapping pass (and add your name to it) 😌

Done, folded `#1956` into the `place-and-route` entry and added my name.

MatthiasReumann (approved):
Looks great!  Many, _many_, thanks for your work on this @rturrado. 
Let's merge this 🛫

burgholzer (approved):
Nice! Thanks for all the work. Let's get this in! 🚀

rturrado:
Many thanks for the kind words, guys!

PR #1986, opened 2026-08-01
✨ Normalize compiler-wide global phases

### LINKED ISSUE #1641: ✨ Merge `GPhaseOp`s
We currently do not have a canonicalization pattern that merges multiple `GPhaseOp` operations. We should add such a pattern (or a corresponding fold) to `qco::GPhaseOp`.

In principle, two `GPhaseOp`s can be merged into one by adding their angles. Similarly, two `GPhaseOp`s with opposite angles cancel each other out.

However, implementing such a canonicalization pattern is not straightforward. Because the operation is, well, global, we cannot easily detect all `GPhaseOp`s present in a given module. The naive approach of looping over all operations in the module seems expensive.

For more details, see https://github.com/munich-quantum-toolkit/core/pull/1407#discussion_r3090092259.

burgholzer:
🤖 *AI text below* 🤖

## Summary

- add the shared `normalize-global-phases` pass for the QC and QCO dialects
- keep at most one direct global-phase operation per basic block while treating functions, CFG blocks, SCF regions, and unknown region-owning operations as hard boundaries
- expose normalization through the C++ and Python program APIs and integrate it with cleanup, synthesis, Jeff, and QIR pipelines
- repair branch-sensitive principal-power canonicalizations and reduce phase materialization in Euler, Weyl, native synthesis, rotation merging, and Hadamard lifting
- use MQT-specific transform target names and the `mlir::mqt` namespace throughout the shared implementation

## Quantum semantics

The modifier rewrites preserve complete unitaries, not merely equivalence modulo global phase:

- inverse extraction negates the phase
- power extraction is limited to finite, compile-time integral exponents
- fractional and dynamic powers retain their nested phase because MQT uses principal matrix powers
- a phase in a controlled body becomes a relative phase on the control register: `P` for one control and a smaller controlled `P` for multiple controls
- QCO control results are threaded through the relative-phase operation without changing target results

The tests compare complete matrices and repeat modifier rewrites under an additional outer control, where an incorrectly extracted global phase becomes observable.

## Follow-up refinements

- replace the defensive non-finite/overflow machinery with a simple verifier contract: direct constant phase angles must be finite and have magnitude at most `1e4`; dynamic angles have the same runtime precondition
- represent accumulated phases as a compact postfix expression and materialize the ordered arithmetic once at the stopping scope
- avoid rescanning dynamic backward slices while transporting phases through nested modifiers
- retain the existing tolerance-based canonicalization rules
- remove the redundant normalization pass after single-qubit rotation merging in the default pipeline
- document and enforce `moduleOp` naming instead of the C++20 keyword `module`

The postfix representation keeps normalization linear without reassociation or symbolic cancellation of dynamic floating-point expressions. A local nested-power benchmark improved from approximately 532 ms at depth 3,200 with the previous implementation to 0.75 ms at depth 1,024 with the revised implementation; increasing depth eightfold increased runtime by about 6.5x.

## Validation

- refreshed `origin/main` integrated before publication
- release build
- all 4,431 configured CTest cases passed; two environment-dependent QDMI job-ID cases were skipped
- 27 focused global-phase normalization tests
- 74 QCO optimization tests
- 206 compiler pipeline tests
- local clang-tidy 22.1.8 on every changed C++ source
- all repository hooks and `git diff --check`
- exact-unitary coverage for modifiers, Euler/Weyl decomposition, native synthesis, Jeff/QIR lowering, and end-to-end OpenQASM pipelines

GPT-5.6 via Codex materially assisted with implementation, semantic auditing, tests, performance evaluation, validation, and drafting this description.

Fixes #1641


burgholzer:
@munich-quantum-toolkit/mqt-cc for awareness.

This looks pretty clean to me and removes a lot of the global phase boilerplate we currently create as part of synthesis or gate fusion.
It's also a nice blueprint for passes shared between QC and QCO.

PR #1995, opened 2026-08-03
✨ Fold constant `arith` trees when normalizing global phases

simon1hofmann:
🤖 *AI text below* 🤖

## Summary

- Global-phase normalization previously treated `arith.mulf` / `addf` / `sitofp`-derived `gphase` angles as dynamic, then merged them into long `addf` chains.
- Later constant-folding could produce `|θ| > 1e4`, so `qco.gphase` verification failed during `decompose-multi-controlled`.
- Add `valueToConstantDouble` to fold pure constant `arith` trees (including `sitofp`/`uitofp`) and materialize a single `normalizeAngle`'d constant so merged phases stay within the practical angle limit.

## Checklist


burgholzer (changes requested):
This generally looks good to me! I just keep wondering: This is talking a lot about folding, but implementing quite a bit of custom code across the phase handling. Is there maybe a more builtin/MLIR-way of doing this by relying on its fold infrastructure?
Maybe pointing an LLM at https://mlir.llvm.org/docs/Canonicalization/#canonicalizing-with-the-fold-method and the MLIR sources would reveal some more potential for simplification and removal of custom code.
Could you try that?

simon1hofmann:
> This generally looks good to me! I just keep wondering: This is talking a lot about folding, but implementing quite a bit of custom code across the phase handling. Is there maybe a more builtin/MLIR-way of doing this by relying on its fold infrastructure? Maybe pointing an LLM at https://mlir.llvm.org/docs/Canonicalization/#canonicalizing-with-the-fold-method and the MLIR sources would reveal some more potential for simplification and removal of custom code. Could you try that?

I tried to do that and this was the summary:
> MLIR fold is one-level/ConstantLike-only, so a full canonicalize isn’t the right dependency here — but we can still reuse the fold hooks. valueToConstantDouble then recursively feeds operand attrs into Operation::fold (arith’s own folders) instead of a custom op switch. The remaining custom code is PhaseExpression’s deferred materialization through modifiers, which isn’t a constant-folding problem.

### Advantages

- Stays aligned with arith’s own fold semantics (APFloat, identities like addf x, -0, casts, etc.) instead of reimplementing +/-/* in C++.
- New foldable ops come for free — e.g. divf now folds without another dyn_cast arm.
- Less risk of drifting from what canonicalize would eventually compute.

### Not really advantages

- Same job for the original bug: recognize constant phase DAGs and normalize them.
- Not simpler overall — PhaseExpression is unchanged; the helper is just a different folding engine.
- Slightly more overhead (recursive attr collection + fold dispatch) for no practical win here.

So: better reuse of MLIR infrastructure, not a clearer or faster solution to the phase-limit problem.

Should we do that?

burgholzer (approved):
I see that it is not a clear win, but I think this feels better overall as the previous version. So I'd say, let's get this in 🙌
Something we might come back to in the future.

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

burgholzer:
Good catch by the rabbit. Looks good to merge to me 👍🏼

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

PR #2002, opened 2026-08-04
🐛 Fix NaN gphase and add host/SSA paths for merge-single-qubit-rotation-gates

### LINKED ISSUE #1640: ⚡️  Improve merging of rotation gates with static angles
The pass added in https://github.com/munich-quantum-toolkit/core/pull/1407 works for rotation gates with dynamic and static angles. However, the pass uses `arith` and `math` operations also when all angles are static. We should investigate whether there's a benefit to using STL functions for this case. If so, it might be worth duplicating parts of the code.

For more details, see https://github.com/munich-quantum-toolkit/core/pull/1407#discussion_r3047940003 and the replies thereof.

simon1hofmann:
🤖 *AI text below* 🤖

## Summary

- Rework `merge-single-qubit-rotation-gates` around a shared dual-backend `Val<T>` algorithm: compile-time-constant chains merge on host scalars (`Val<double>` / STL), otherwise through `arith`/`math` SSA (`Val<Value>`).
- Fix NaN `gphase` / invalid angles on pure-Z (and other XY-aligned) merges: MLIR’s `math.atan2` constant folder returns NaN for both-zero operands, which can poison Euler extraction under slight `beta` FP drift. Force the gimbal-lock path when `|x|` and `|y|` are near zero, and sanitize the `atan2(-x, y)` operand so the folder never sees `(0, 0)`.
- On the host path, omit near-zero phase corrections; on the SSA path always emit `gphase` and let global-phase normalization clean up foldable zeros. Normalize foldable constant corrections into the practical `gphase` range (same contract as #1995).
- Extend tests for pure-Z static merges, dynamic-angle SSA merges, and multi-`P` dynamic phase accumulation; tidy clang-tidy issues in the pass and tests.

Fixes #1640

## Checklist


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

mlir/lib/Dialect/QCO/Transforms/Optimizations/MergeSingleQubitRotationGates.cpp
```diff
@@ -539,15 +540,41 @@ struct MergeSingleQubitRotationGatesPattern final
     // safe2 = beta not within boundary eps around PI: |beta - PI| >= eps
     auto safe2 = arith::CmpFOp::create(rewriter, loc, arith::CmpFPredicate::OGE,
                                        absBetaMinusPi, constants.eps);
-    // is safe (not in gimbal lock) when both hold (safe1 AND safe2)
-    auto safe = arith::AndIOp::create(rewriter, loc, safe1, safe2);
+
+    // Pure Z (and other XY-axis-aligned) quaternions have x = y = 0.
+    // MLIR's math.atan2 constant folder returns NaN for both-zero operands
+    // (unlike IEEE libm, which returns ±0 / ±π). Floating-point drift can also
```
burgholzer: The mention of IEEE libm feels a bit strange here. I am pretty sure this is an iteration of artifact.
I also can't help but feel that the entire handling here has become more complex than it likely needs to be.
Would it make sense to take a step back, look at the problem with a fresh context and see whether there is an overall cleaner solution for handling this?

burgholzer (changes requested):
Generally, this looks fine. I am just wondering whether this could be a bit simpler overall.
If not, this is fine.

mlir/lib/Dialect/QCO/Transforms/Optimizations/MergeSingleQubitRotationGates.cpp
simon1hofmann: We could also compute ZYZ angles on the host (std::atan2 / std::acos) if w,x,y,z all fold via valueToConstantDouble and emit arith.constants.
Would be more code though but less complex overall.

burgholzer: Could you evaluate https://github.com/munich-quantum-toolkit/core/issues/1640#issuecomment-4256679054 in that regard? could be a promising direction (also to keep the overhead manageable)

simon1hofmann: Should be possible to come up with a solution that directly resolves that issue as well 👍

burgholzer (approved):
The current solution feels pretty good and comes at roughly the same complexity as the previous solution. Once CI is green, this can go in if you are happy

simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

PR #2004, opened 2026-08-04
🐛 Replay cold-preview SWAP plan during hot routing

simon1hofmann:
## Description

Reverts the `place` logic introduced in #1997. Instead of only creating `qco.static` ops for materialized program qubits, all available hardware qubits are initialized. A subsequent canonicalization pass run will then remove unused static ops. 

## Checklist


simon1hofmann:
@MatthiasReumann As already talked about previously, I created a test case that was failing and a proposed fix. Would appreciate your feedback 🙏

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -369,9 +373,16 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
     SmallVector<size_t> materializedPrograms(wires.size());
     std::iota(materializedPrograms.begin(), materializedPrograms.end(), 0);
     if (comp->hasTwoQubitOperations) {
+      // Cold preview records the A* SWAP plan that Hot must reproduce after
+      // sparse workspace materialization.
+      swapPlan.clear();
+      swapPlanCursor = 0;
+      swapPlanMode = SwapPlanMode::Record;
```
MatthiasReumann: I don't think this should be necessary.

MatthiasReumann (commented):
Many thanks @simon1hofmann 🚀

I don't think this is quite right, yet. It is actually by design that the final forward `Hot` routes the program again. By doing so we can directly insert SWAPs into the IR without any replay mechanism. 

The mapping pass works as follows 

```
generateLayout:
[0]: (forward, backward)^{niterations} of route<Cold> using ntrials many initial random layouts. 
[1]: Find the "best" final layout which has been produced by the final *backward* iteration.

place:
[0]: Use this "best" layout to initially place the circuit 

route<Hot>: Final forward iteration while inserting SWAPs
```

I've pushed a few commits which (I think) will fix the bug also. Essentially, what I think is happening is that in `insertSWAPs` a SWAP wants to be inserted on an unmaterialized program qubit for which the `infos` weren't initialized. 

@burgholzer FYI: I've removed the `materializedPrograms` for now. At some point, I want to get rid of the majority of the code in the `place` method anyway (inline `extend` the SCF ops while hot routing, for example) - so maybe let's revisit that idea again later and let the canonicalization pass remove unused static ops for now.

simon1hofmann:
> Many thanks @simon1hofmann 🚀
> 
> I don't think this is quite right, yet. It is actually by design that the final forward `Hot` routes the program again. By doing so we can directly insert SWAPs into the IR without any replay mechanism.
> 
> The mapping pass works as follows
> 
> ```
> generateLayout:
> [0]: (forward, backward)^{niterations} of route<Cold> using ntrials many initial random layouts. 
> [1]: Find the "best" final layout which has been produced by the final *backward* iteration.
> 
> place:
> [0]: Use this "best" layout to initially place the circuit 
> 
> route<Hot>: Final forward iteration while inserting SWAPs
> ```
> 
> I've pushed a few commits which (I think) will fix the bug also. Essentially, what I think is happening is that in `insertSWAPs` a SWAP wants to be inserted on an unmaterialized program qubit for which the `infos` weren't initialized.
> 
> @burgholzer FYI: I've removed the `materializedPrograms` for now. At some point, I want to get rid of the majority of the code in the `place` method anyway (inline `extend` the SCF ops while hot routing, for example) - so maybe let's revisit that idea again later and let the canonicalization pass remove unused static ops for now.

Thanks a lot for the quick fix @MatthiasReumann 🙏 

 I reran the benchpress tests and your solutions seems to fix them as well 👍

mlir/unittests/Dialect/QCO/Transforms/Mapping/test_mapping.cpp
```diff
@@ -1569,5 +1597,35 @@ TEST_P(MappingPassTest, MapIndexSwitchUsesVotedLayout) {
...
+      runPass(m.get(), target, MappingPassOptions{.ntrials = 1}).succeeded());
+  ASSERT_TRUE(succeeded(verify(*m)));
+  EXPECT_TRUE(isExecutable(getEntryPoint(m.get()), target));
+}
+
+INSTANTIATE_TEST_SUITE_P(ThreeByThreeSquareGrid, MappingPassTest,
+                         testing::Values(getSquareGridTarget(3)));
+INSTANTIATE_TEST_SUITE_P(FourByFourSquareGrid, MappingPassTest,
+                         testing::Values(getSquareGridTarget(4)));
+INSTANTIATE_TEST_SUITE_P(ThenByThenSquareGrid, MappingPassTest,
```
burgholzer:
```suggestion
INSTANTIATE_TEST_SUITE_P(TenByTenSquareGrid, MappingPassTest,
```
Suppose this is a typo.

burgholzer (approved):
Looks like a fine solution to me 👍🏻
Just spotted one potential typo. Then this can happily go in.

burgholzer:
Ah also one more comment: do we really want to run a full canonicalization here or simply apply the specific pattern for removing static-sink pairs? Could be more efficient.

mlir/unittests/Dialect/QCO/Transforms/Mapping/test_mapping.cpp
MatthiasReumann: Thanks for catching that. Sometimes my brain does weird things.

MatthiasReumann:
> Ah also one more comment: do we really want to run a full canonicalization here or simply apply the specific pattern for removing static-sink pairs? Could be more efficient.

Good idea! Done.

PR #2013, opened 2026-08-06
✨ Pull `GPhaseOp`s out of multi-operation `ctrl` modifiers

### LINKED ISSUE #1759: ✨ Pull `GPhaseOp`s out of multi-operation `ctrl` modifiers
#### Description

#1751 has dropped the condition that modifier bodies can only contain a single unitary operation followed by a `yield` operation.

We should add a canonicalization that pulls out `GPhaseOp`s from multi-operation `ctrl` modifiers. For example, 

```mlir
%const = arith.constant 0.123 : f64
qc.ctrl(%c0, %c1) targets(%t0, %t1) {
  qc.gphase(%const)
  qc.rxx(%const) %t0, %t1 : !qc.qubit, !qc.qubit
}
```

is unrolled to

```mlir
%const = arith.constant 0.123 : f64
qc.ctrl(%c0) targets(%c1) {
  qc.p(%const) %c1 : !qc.qubit
}
qc.ctrl(%c0, %c1) targets(%t0, %t1) {
  qc.rxx(%const) %t0, %t1 : !qc.qubit, !qc.qubit
}
```

Notice that we already have such a canonicalization for one-operation `ctrl` modifiers. It should be possible to adapt some of the existing code.

#### Acceptance criteria

- The described canonicalization is implemented in QC and QCO
- The [two](https://github.com/munich-quantum-toolkit/core/pull/1603#issuecomment-4230138934) [comments](https://github.com/munich-quantum-toolkit/core/pull/1603#issuecomment-4243599023) made in [#1603](https://github.com/munich-quantum-toolkit/core/pull/1603) are taken into account

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Add canonicalization patterns that pull `GPhaseOp`s out of multi-operation `ctrl` modifiers in the QC and QCO dialects.

A controlled global phase is materialized as:

- a `POp` on the control for a single-control modifier;
- a reduced `CtrlOp` applying `POp` to the final control for a multi-control modifier.

The existing single-operation and new multi-operation canonicalizations share the same controlled-phase construction helper. The QCO implementation also threads the updated control-qubit SSA values through the remaining modifier.

Global phases whose angle is defined inside the modifier body are conservatively left unchanged to avoid violating SSA dominance.

Fixes #1759

AI assistance was used to inspect the existing canonicalizations, implement and refactor the QC and QCO patterns, add regression tests, and run validation.

## Checklist

### DIFF
--- a/mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
@@ -38,17 +38,15 @@
 using namespace mlir;
 using namespace mlir::qco;
 
-namespace {
-
 /**
  * @brief Materialize a global phase controlled by @p controls.
  * @return The updated control qubits in their original order.
  */
-SmallVector<Value> createControlledPhase(PatternRewriter& rewriter,
-                                         const Location controlledLoc,
-                                         const Location phaseLoc,
-                                         const ValueRange controls,
-                                         const Value theta) {
+static SmallVector<Value> createControlledPhase(PatternRewriter& rewriter,
+                                                const Location controlledLoc,
+                                                const Location phaseLoc,
+                                                const ValueRange controls,
+                                                const Value theta) {
   assert(!controls.empty());
   if (controls.size() == 1) {
     return {POp::create(rewriter, controlledLoc, controls.front(), theta)
@@ -62,6 +60,8 @@ SmallVector<Value> createControlledPhase(PatternRewriter& rewriter,
       });
   return SmallVector<Value>(controlledPhase.getOutputQubits());
 }
+
+namespace {
 
 /**
  * @brief Merge nested control modifiers into a single one.
--- a/mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
@@ -41,6 +41,29 @@ using namespace mlir::qco;
 namespace {
 
 /**
+ * @brief Materialize a global phase controlled by @p controls.
+ * @return The updated control qubits in their original order.
+ */
+SmallVector<Value> createControlledPhase(PatternRewriter& rewriter,
+                                         const Location controlledLoc,
+                                         const Location phaseLoc,
+                                         const ValueRange controls,
+                                         const Value theta) {
+  assert(!controls.empty());
+  if (controls.size() == 1) {
+    return {POp::create(rewriter, controlledLoc, controls.front(), theta)
+                .getOutputQubit(0)};
+  }
+
+  auto controlledPhase = CtrlOp::create(
+      rewriter, controlledLoc, controls.drop_back(), controls.back(),
+      [&](Value target) -> Value {
+        return POp::create(rewriter, phaseLoc, target, theta).getOutputQubit(0);
+      });
+  return SmallVector<Value>(controlledPhase.getOutputQubits());
+}
+
+/**
  * @brief Merge nested control modifiers into a single one.
  */
 struct MergeNestedCtrl final : OpRewritePattern<CtrlOp> {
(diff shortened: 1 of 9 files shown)


simon1hofmann:
@coderabbitai full review

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
```diff
@@ -38,6 +38,29 @@
...
+                .getOutputQubit(0)};
+  }
+
+  auto controlledPhase = CtrlOp::create(
+      rewriter, controlledLoc, controls.drop_back(), controls.back(),
+      [&](Value target) -> Value {
+        return POp::create(rewriter, phaseLoc, target, theta).getOutputQubit(0);
+      });
+  return SmallVector<Value>(controlledPhase.getOutputQubits());
+}
```
burgholzer, line 58: This could potentially be optimized a little bit by splitting it into the one-control case (which could simply return a Value instead of a Vector), and the rest.
Also holds for the QC case.

mlir/unittests/programs/qc_programs.cpp
```diff
@@ -2196,6 +2196,44 @@ Value nestedCtrlTwo(QCProgramBuilder& b) {
...
+  b.cx(q[0], q[1]);
+  return measureAndReturn(b, q.qubits);
+}
+
+Value ctrlMultipleGPhases(QCProgramBuilder& b) {
+  auto q = b.allocQubitRegister(4);
+  b.ctrl({q[0], q[1]}, {q[2], q[3]}, [&](ValueRange targets) {
+    b.gphase(0.123);
+    b.x(targets[0]);
+    b.gphase(0.456);
```
burgholzer, line 2220: Slightly confused here: shouldn't these global phases be combined by the new logic and then the result would only be a single cp gate?

I suppose we do not yet support merging of gates across modifiers (such as the cp's here)..

burgholzer (commented):
Just two comments here while I have some time to kill

mlir/lib/Dialect/QCO/IR/Modifiers/CtrlOp.cpp
simon1hofmann: Optimized it for both cases.

mlir/unittests/programs/qc_programs.cpp
simon1hofmann: True, should work now.

simon1hofmann:
@coderabbitai review

burgholzer:
Hi @simon1hofmann, thanks again for putting this together. 
I had another close look after merging the current main locally. 
And I have to say that this PR kind of feels a bit obsolete by now.
#1986, together with #1995, has introduced the compiler-wide `normalize-global-phases` pass, which already covers this behavior for QC and QCO—including controlled phases, SSA handling, QCO result threading, and exact-unitary tests.

As written, this PR would therefore add a second, narrower implementation in the CtrlOp canonicalizations ahead of the shared pass. The new tests also run the full cleanup pipeline, so the existing normalization pass can make them pass even if the new patterns do nothing.

My inclination would be to consider #1759 covered by the existing pass and close this PR as superseded. The main reason to keep it would be if we specifically want the standalone canonicalizer to perform this transformation without normalize-global-phases. 

Do you think that standalone behavior is important, or would you agree that the existing pass covers the intended use case?

simon1hofmann:
> Hi @simon1hofmann, thanks again for putting this together. I had another close look after merging the current main locally. And I have to say that this PR kind of feels a bit obsolete by now. #1986, together with #1995, has introduced the compiler-wide `normalize-global-phases` pass, which already covers this behavior for QC and QCO—including controlled phases, SSA handling, QCO result threading, and exact-unitary tests.
> 
> As written, this PR would therefore add a second, narrower implementation in the CtrlOp canonicalizations ahead of the shared pass. The new tests also run the full cleanup pipeline, so the existing normalization pass can make them pass even if the new patterns do nothing.
> 
> My inclination would be to consider #1759 covered by the existing pass and close this PR as superseded. The main reason to keep it would be if we specifically want the standalone canonicalizer to perform this transformation without normalize-global-phases.
> 
> Do you think that standalone behavior is important, or would you agree that the existing pass covers the intended use case?

Fine for me to close this PR if the other passes already handle it 👍

Issue #1639, opened 2026-04-15
✨ Extend merge pass to non-rotation gates


denialhaag:
The merge pass added in https://github.com/munich-quantum-toolkit/core/pull/1407 only acts on rotation gates. We could extend the pass to non-rotation gates so that single-qubit gates are fused into a single `UOp` or other rotation gates.

For more details, see https://github.com/munich-quantum-toolkit/core/pull/1407#discussion_r3047637595 and the replies thereof.

J4MMlE:
I just dump the TODO i had for that here:
```
Gate  | Quaternion Q(w,x,y,z)                      | Global Phase 
------+--------------------------------------------+--------------
X     | Q(0, 1, 0, 0)                              |  π/2
Y     | Q(0, 0, 1, 0)                              |  π/2
Z     | Q(0, 0, 0, 1)                              |  π/2
S     | Q(cos(π/4), 0, 0,  sin(π/4))               |  π/4
Sdg   | Q(cos(π/4), 0, 0, -sin(π/4))               | -π/4
T     | Q(cos(π/8), 0, 0,  sin(π/8))               |  π/8
Tdg   | Q(cos(π/8), 0, 0, -sin(π/8))               | -π/8
H     | Q(0, 1/√2, 0, 1/√2)                        |  π/2
SX    | Q(cos(π/4),  sin(π/4), 0, 0)               |  π/4
SXdg  | Q(cos(π/4), -sin(π/4), 0, 0)               | -π/4
```

Special handling (no quaternion conversion):
`Id`: skip, don't break chain (contributes nothing)
`Barrier`: break chain
`GPhase`: skip rotation, accumulate phase only

burgholzer:
> I just dump the TODO i had for that here:
> ```
> Gate  | Quaternion Q(w,x,y,z)                      | Global Phase 
> ------+--------------------------------------------+--------------
> X     | Q(0, 1, 0, 0)                              |  π/2
> Y     | Q(0, 0, 1, 0)                              |  π/2
> Z     | Q(0, 0, 0, 1)                              |  π/2
> S     | Q(cos(π/4), 0, 0,  sin(π/4))               |  π/4
> Sdg   | Q(cos(π/4), 0, 0, -sin(π/4))               | -π/4
> T     | Q(cos(π/8), 0, 0,  sin(π/8))               |  π/8
> Tdg   | Q(cos(π/8), 0, 0, -sin(π/8))               | -π/8
[...]

PR #2015, opened 2026-08-06
✨ Add `unroll-modifiers` pass for unrolling multi-operation modifiers

### LINKED ISSUE #1758: ✨ Add pass for unrolling multi-operation modifiers
#### Description

#1751 has dropped the condition that modifier bodies can only contain a single unitary operation followed by a `yield` operation.

We should add a pass that unrolls such modifiers into a sequence of one-operation modifiers. For example, 

```mlir
%const = arith.constant 0.123 : f64
qc.ctrl(%c0, %c1) targets(%t0, %t1) {
  qc.x %t0 : !qc.qubit
  qc.rxx(%const) %t0, %t1 : !qc.qubit, !qc.qubit
}
```

is unrolled to

```mlir
%const = arith.constant 0.123 : f64
qc.ctrl(%c0, %c1) targets(%t0) {
  qc.x %t0 : !qc.qubit
}
qc.ctrl(%c0, %c1) targets(%t0, %t1) {
  qc.rxx(%const) %t0, %t1 : !qc.qubit, !qc.qubit
}
```

Notice that the block arguments (i.e., `targets`) should be adjusted as well. When unrolling `inv` modifiers, the order of the operations must be inverted as well.

Once implemented, this pass should be used to prepare the IR before the QCO-to-`jeff` and QC-to-QIR conversions. These cannot currently fully convert multi-operation modifiers. The QC-to-QIR conversion can furthermore be simplified by making `LoweringState::inCtrlOp` a `bool`.

#### Acceptance criteria

- The described pass is implemented for `ctrl`, `inv`, and `pow` modifiers
- The described pass is used in the QCO-to-`jeff` and QC-to-QIR conversions
- `LoweringState::inCtrlOp` is simplified

denialhaag:
## Description

🤖 *AI text below* 🤖

This PR adds an `unroll-modifiers` pass that splits `ctrl`, `inv`, and `pow` modifiers into a sequence of modifiers holding a single operation each. 

For `inv`, the order of the operations is reversed, since $(ab)^{-1} = b^{-1} a^{-1}$. `pow` is only unrolled if the body operations act on disjoint qubits, since $(ab)^r$ is generally not equal to $a^r b^r$.

Classical operations are moved in front of the modifier.

Following #1986, the pass is shared between QC and QCO and lives in `mlir::mqt`. It now runs before the QCO-to-`jeff` and QC-to-QIR conversions, which cannot fully handle multi-operation modifiers.

Fixes #1758

## AI Notice

The first version of the changes was assisted by Opus 5 via Claude Code. I subsequently iterated on the changes.

## Checklist


denialhaag:
@coderabbitai full review

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
```diff
@@ -0,0 +1,261 @@
...
+    auto controlsOut = ctrlOp.getControlsOut();
+    controls.assign(controlsOut.begin(), controlsOut.end());
+    qubits.map(unitary->getResults(), ctrlOp.getTargetsOut());
+  }
+
+  SmallVector<Value> results(controls);
+  for (auto yielded : body->getTerminator()->getOperands()) {
+    results.push_back(qubits.lookup(yielded));
+  }
+  rewriter.replaceOp(op, results);
```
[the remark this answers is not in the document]
denialhaag: https://github.com/munich-quantum-toolkit/core/pull/2015/changes/0cf9047ddfaf3925573f3a44a4078c425bb16469 addresses this, but not exactly as you suggest. Does it look good to you?

denialhaag:
@coderabbitai review

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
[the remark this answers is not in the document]
denialhaag: Does https://github.com/munich-quantum-toolkit/core/pull/2015/commits/7707cba50152fa6cef6dd3a4b114b43916a611b4 resolve the issue?

denialhaag:
@coderabbitai review

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+    // The walk visits nested modifiers before their parents, so unrolling the
+    // collected modifiers in order reaches a fixpoint in a single sweep.
+    IRRewriter rewriter(&getContext());
+    for (auto* modifier : modifiers) {
+      llvm::TypeSwitch<Operation*>(modifier)
+          .Case<qc::CtrlOp, qc::InvOp, qco::CtrlOp, qco::InvOp>([&](auto op) {
+            static_cast<void>(unrollModifier(op, rewriter));
+          });
+    }
```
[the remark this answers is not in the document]
denialhaag: This is fixed by https://github.com/munich-quantum-toolkit/core/pull/2015/changes/9ff2e9d58238556dfc13ede95aefe20216e05988, right?

mlir/include/mlir/Dialect/Utils/Transforms/Passes.td
```diff
@@ -34,4 +34,23 @@ def NormalizeGlobalPhases : Pass<"normalize-global-phases", "mlir::ModuleOp"> {
...
+    The targets of the new modifiers are narrowed to the qubits of the
+    respective operation. For `inv`, the order of the operations is reversed.
+    Classical operations of the body are moved in front of the new modifiers.
+
+    If a modifier cannot be unrolled, the modifier is left untouched. Such
+    modifiers are skipped silently; the pass never fails.
+
+    In addition, `pow` modifiers are left untouched because `pow(r) { a; b }`
+    is generally not equivalent to `pow(r) { a }; pow(r) { b }`. Modifiers
+    nested in a `pow` body are still unrolled.
```
burgholzer: This feels a bit too narrow. The cases where this simplification holds are well known. IIRC, this holds for (positive?) integer powers. That should be easy to verify though with an LLM.

mlir/unittests/Conversion/QCToQIR/QCToQIRBase/test_qc_to_qir_base.cpp
```diff
@@ -49,19 +50,19 @@ namespace {
...
   friend std::ostream& operator<<(std::ostream& os,
                                   const QCToQIRBaseTestCase& info);
 };
 
 // NOLINTNEXTLINE(llvm-prefer-static-over-anonymous-namespace)
 std::ostream& operator<<(std::ostream& os, const QCToQIRBaseTestCase& info) {
-  return os << "QCToQIRBase{" << info.name
-            << ", original=" << mqt::test::displayName(info.programBuilder.name)
+  return os << "QCToQIRBase{" << info.name << ", original="
+            << ::mqt::test::displayName(info.programBuilder.name)
```
burgholzer, line 65: Do all these changes in the various files adding `::` serve any particular purpose?
The diff here in the tests feels much larger than necessary.

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
burgholzer, line 1: I looked through this and my gut feeling is telling me that this is more complicated than it needs to be.
I think the entire handling of block arguments and their reduction could be handled more efficiently. I might be misremembering, but I think there already was a pass/canonicalization that drops unused block arguments. So I was thinking whether we could simplify quite a bit of the inlining logic here and rely on that pass/canon.
And just to question that: Does MLIR provide some more convenient inlining functionality that we could use here?

burgholzer (changes requested):
Thanks @denialhaag 🙏🏼 I looked through this and, on a quick glance, I have the feeling there is some room for improvement here. I hope these comments make sense.

mlir/include/mlir/Dialect/Utils/Transforms/Passes.td
denialhaag: Unless I'm missing something fundamental right now, `pow(r) { a; b } = pow(r) { a }; pow(r) { b }` requires _at least_ that `a` and `b` commute, no? This should already be a simple enough counterexample: $(XY)^2 = XYXY \neq XXYY = X^2Y^2$, where $X$ and $Y$ are the respective Pauli matrices.

denialhaag: I guess we can always expand `pow(r) { a; b }` if `a` and `b` are acting on different qubits, as they definitely commute then. 🤔 Should be easy enough to check.

burgholzer: ah. I was definitely looking at this too quickly. You are right, and I would not overcomplicate the current implementation with Commutation Analysis.

simon1hofmann: I encountered a related problem while extending `MergeSingleQubitRotationGates` to fixed gates in #1639.

For the OpenQASM program

```qasm
gate hs q {
  h q;
  s q;
}

pow(2) @ hs q;
```

the merge pass turns the `h; s` body into a `u` operation plus a separate `gphase`. This is unitary-equivalent outside a modifier, but inside `pow` it changes the body from

```text
pow(2) { h; s }
```

to roughly

```text
pow(2) { gphase(φ); u(...) }
```

That currently breaks the standard QC-to-QIR pipeline because the resulting multi-operation `pow` cannot be legalized.

For an integer exponent, the phase could theoretically be handled with

```text
pow(n) { gphase(φ); U } = gphase(nφ); pow(n) { U }
```

but that would require an explicitly Pow-aware global-phase rewrite, as it is not the same as distributing the power over arbitrary body operations.

Is that something that should be taken care of in this PR?

burgholzer: Isn't the gphase normalization handled by the respective pass that was recently added?

simon1hofmann: My thinking is as follows:

Rotation merge pass:
- Calculates the phase contributions of the input gates.
- Accounts for changes of SU(2) representative during Euler normalization.
- Emits the required global phase.

Global-phase pass:
- Combines emitted global phases.
- Cancels or moves them where valid.
- Normalizes the resulting angle.

We could leave the emitted correction unnormalized and rely on the later phase pass for normalization. However, the correction itself must still be calculated by the rotation merge pass. Normalizing it locally keeps its output within the global phase angle limit and makes the pass independently valid.

Or am I overseeing something?

burgholzer: The rotation merge pass must still compute the correction and emit the phase instruction. However, the phase normalization pass can be run directly after the pass in the pipeline (similar to the various other places where it has currently been added).
But maybe I am the one overseeing something 😌

simon1hofmann: No you're right, I just realized that this is part of `MergeSingleQubitRotationGates`:
```
    if (failed(applyPatternsGreedily(op, std::move(patterns))) ||
        failed(mlir::mqt::normalizeGlobalPhases(op))) {
      signalPassFailure();
    }
```

The bug I encountered in #1639 was due to some other problem, sorry for the confusion 😄

simon1hofmann: The problem was that pow(2) { u } is not unrolled and cannot be lowered to QIR then 🤔

burgholzer: Yeah, we have no canonicalization for `pow @ u`. Feels like an oversight though. At least for (positive) integer powers, we can surely use our synthesis methods to compute the updated `u` parameters.
Potentially, there even is an analytic formula for integer powers.

denialhaag: https://github.com/munich-quantum-toolkit/core/pull/2015/commits/3f9ce38c4620ed1ceec083c8dc0e4de04e889764 adds a pattern that unrolls a `PowOp` if all body operations are disjoint.

mlir/unittests/Conversion/QCToQIR/QCToQIRBase/test_qc_to_qir_base.cpp
denialhaag: This was necessary because the `mlir::mqt` and `mqt::test` namespaces began to conflict. https://github.com/munich-quantum-toolkit/core/pull/2015/commits/ac7b14eb4ea4a6a84f2ab8a2e60eedc81edffe21 cleans this up, but also doesn't really reduce the diff. 😬

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
denialhaag: I tried quite a few things, but I didn't find a way to use any built-in pass to drop unused block arguments. The closest match would be in [`ControlFlowInterfaces.h`](https://mlir.llvm.org/doxygen/ControlFlowInterfaces_8h.html). In particular, `scf::ForOp` uses [`populateRegionBranchOpInterfaceCanonicalizationPatterns`](https://mlir.llvm.org/doxygen/namespacemlir.html#ad2c5d55e2b30e192dd378eb3a8ee0929). We certainly cannot use that in QC because the patterns expect all block arguments to be yielded (which we do not do). In QCO, I ran into several problems with our linearity checks, but I will try again later.

For now, I have added custom canonicalization patterns in https://github.com/munich-quantum-toolkit/core/pull/2015/commits/216524b0077b219c953733e07c39e20d0149f537. While they did not make the implementation in `UnrollModifiers.cpp` _much_ simpler (e.g., the QCO values still need to be threaded), I think the patterns are good to have regardless. Again, I'll have another look to see whether there are built-in passes that cover at least some functionality, but I'm not too hopeful at this point.

mlir/include/mlir/Dialect/Utils/Transforms/Passes.td
simon1hofmann: This probably supersedes #2041, could you check if that is similar to what is covered by your pattern?

denialhaag: The implementation here only splits `pow(2) { gphase(..); u(...) }` into `pow(2) { gphase(..) }; pow(2) { u(...) }`, but it doesn't then inline `pow(2) { u(...) }`. If I'm not missing something right now, I still see the value in #2041. 🤔

simon1hofmann: Okay great, then I will keep that open 👍

denialhaag:
@coderabbitai review

mlir/include/mlir/Dialect/Utils/Transforms/Passes.td
burgholzer: 🤖 *AI text below* 🤖

Implemented the bounded version at `93c8cdabd`: `unroll-modifiers` splits only body operations whose qubit sets are disjoint. Overlapping operations and non-integral powers remain untouched; no commutation analysis or general `pow @ u` synthesis was added. Structural tests assert both the split and preservation cases before cleanup.

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
burgholzer: 🤖 *AI text below* 🤖

The unroller now constructs each split `ctrl`, `inv`, or `pow` with exactly the qubits used by that body unitary, so it no longer relies on a cleanup canonicalization for correctness. QCO linear SSA threading remains explicit. The retained drop-unused patterns are short adapters over shared helpers and serve as general cleanup when a later rewrite turns a previously used block argument into pass-through. Structural tests inspect the narrowed modifiers before cleanup, and an independent pass found no remaining simplification blocker.

mlir/unittests/Conversion/QCToQIR/QCToQIRBase/test_qc_to_qir_base.cpp
burgholzer: 🤖 *AI text below* 🤖

Resolved by restoring the original `mqt::test` namespace usage and reversing the namespace-only churn from `ac7b14eb4`. The effective PR diff no longer carries the broad `::` qualification changes; only qualifications directly required by the implementation remain.

mlir/lib/Dialect/QCO/IR/Modifiers/PowOp.cpp
```diff
@@ -357,159 +375,220 @@ struct FoldPowIntoGate final : OpRewritePattern<PowOp> {
...
               rewriter.replaceOpWithNewOp<decltype(gate)>(
-                  op, op.getInputTarget(0), op.getInputTarget(1), newParam);
+                  op,
+                  utils::getValueFromBlockArgument(gate.getInputTarget(0),
+                                                   op.getQubitsIn()),
+                  utils::getValueFromBlockArgument(gate.getInputTarget(1),
+                                                   op.getQubitsIn()),
+                  newParam);
               return success();
             })
```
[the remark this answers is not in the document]
burgholzer: 🤖 *AI text below* 🤖

Resolved in `cc2169a60` and completed in `93c8cdabd`. Power folds now create the replacement first and map the modifier terminator yields through the body unitary to the replacement outputs, preserving logical wire association for permuted operands and yields. This covers the parametric two-qubit gates and iSWAP, plus Barrier and even Hermitian identity folds. Reordered RZX, Barrier, and even-SWAP regressions pass.

mlir/lib/Dialect/QCO/IR/Modifiers/ModifierUtils.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+LogicalResult verifyModifierBody(Operation* modifierOp, Block& body) {
+  if (llvm::any_of(body, [](Operation& operation) {
+        return isa<AllocOp, SinkOp, StaticOp, MeasureOp, ResetOp,
+                   qtensor::ExtractOp, qtensor::InsertOp>(operation);
+      })) {
+    return modifierOp->emitOpError(
+        "body must not contain non-unitary quantum operations or modify a "
+        "quantum register");
+  }
```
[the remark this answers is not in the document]
burgholzer: 🤖 *AI text below* 🤖

Resolved in `290111305`. The shared QCO modifier verifier now recursively walks nested regions and rejects forbidden non-unitary operations there. The regression covers `ctrl`, `inv`, and `pow` with a nested measurement that uses only a modifier block argument. External qubit captures are also categorically rejected, while classical captures remain allowed. The complete QCO IR suite passes 476/476.

mlir/unittests/Dialect/QCO/Transforms/NativeSynthesis/test_target_synthesis.cpp
```diff
@@ -45,39 +45,29 @@
...
-using mlir::qco::SWAPOp;
-using mlir::qco::XOp;
-using mlir::qco::ZOp;
-
-[[nodiscard]] static mlir::func::FuncOp mainFunction(ModuleOp module) {
-  return *module.getBody()->getOps<mlir::func::FuncOp>().begin();
+using namespace mlir::qco;
+
+[[nodiscard]] static func::FuncOp mainFunction(ModuleOp module) {
+  return *module.getBody()->getOps<func::FuncOp>().begin();
```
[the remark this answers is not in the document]
burgholzer: 🤖 *AI text below* 🤖

The unrelated target-synthesis test churn was removed from the effective PR diff when the broad namespace changes were reversed. The remaining changed QC-to-QIR helper consistently uses `moduleOp`. Repository-wide lint and changed-file `clang-tidy` pass at `93c8cdabd`.

mlir/unittests/ExactUnitaryTest.h
```diff
@@ -32,19 +32,19 @@ namespace mqt::test {
  * becomes observable here and when the rewritten function is put under an
  * additional control.
  */
-inline void expectFullUnitaryEqual(mlir::ModuleOp expectedModule,
-                                   mlir::ModuleOp actualModule,
+inline void expectFullUnitaryEqual(ModuleOp expectedModule,
+                                   ModuleOp actualModule,
                                    const std::size_t numQubits,
                                    const double tolerance = 1e-12) {
```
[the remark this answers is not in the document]
burgholzer: 🤖 *AI text below* 🤖

`ExactUnitaryTest.h` is no longer changed by this PR after the unrelated namespace/style churn was reversed, so this pre-existing line is outside the effective diff. New and modified code in the remediation uses unqualified `size_t` with direct `<cstddef>` includes. Repository-wide lint passes at `93c8cdabd`.

mlir/unittests/Dialect/QC/IR/test_qc_ir.cpp
```diff
@@ -1509,3 +1529,140 @@ INSTANTIATE_TEST_SUITE_P(
...
+  expectUnrollsTo(context.get(), powTwoDisjoint, powTwoDisjointUnrolled);
+}
+
+TEST_F(QCTest, UnrollModifiersLeavesOverlappingPowUntouched) {
+  expectUnrollsTo(context.get(), powTwo, powTwo);
+}
+
+TEST_F(QCTest, UnrollModifiersLeavesNonIntegerPowUntouched) {
+  expectUnrollsTo(context.get(), powHalfDisjoint, powHalfDisjoint);
+}
```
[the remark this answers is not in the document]
burgholzer: 🤖 *AI text below* 🤖

Resolved in `889041bd7`. The QC and QCO test helpers now run structural assertions immediately after `unroll-modifiers` and before cleanup/equivalence checking. They verify split counts, narrowed footprints, inverse order, preserved overlapping/non-integral powers, and QCO consumer/result SSA ordering. The complete QC and QCO suites pass 326/326 and 476/476.

burgholzer:
🤖 *AI text below* 🤖

Remediation is pushed at `93c8cdabd`.

In addition to the inline-thread fixes, the QC-to-QIR conversion now arms control state only for a sole non-Barrier/non-Id unitary, so empty and no-op control bodies cannot leak control state to a following gate. Direct conversion regressions cover both cases.

The branch includes current `main` (`a9105f908`) and is mergeable. Local validation passed a complete debug build, 1,441 affected unit tests, the new mqt-cc QIR-output ctest, repository-wide lint, and changed-file clang-tidy. A fresh independent exact-head review found no remaining correctness, simplification, performance, documentation, scope, or test-coupling issue.

I replied to the open review threads but left their resolution state unchanged.

burgholzer:
Had some Codex quota left before the reset, so I figured I may as well use it on this PR.

mlir/lib/Dialect/Utils/Transforms/UnrollModifiers.cpp
denialhaag: Makes sense. Removing the superfluous block arguments already during the pass also felt cleaner to me. With https://github.com/munich-quantum-toolkit/core/pull/2015/commits/1f278299baa9c262403ecf2c8a68c3643df8a1ec, we are pretty much exactly back to where we were before, which felt pretty compact to me. 🤔

denialhaag:
@burgholzer, from my side, this is ready to go now. Feel free to merge whenever.

mlir/unittests/Conversion/QCToQIR/QCToQIRBase/test_qc_to_qir_base.cpp
denialhaag: I think https://github.com/munich-quantum-toolkit/core/pull/2015/commits/8f3484342bee0a139903626bd264d75e0576fee6 pretty much exactly reverted https://github.com/munich-quantum-toolkit/core/commit/ac7b14eb4ea4a6a84f2ab8a2e60eedc81edffe21, restoring the state at the time of the initial review. Either option is fine with me.

burgholzer (approved):
Just pushed one commit that should hopefully fix the docs build and pull out a couple more common functionality.
Tagging this for auto-merge now.

PR #2039, opened 2026-08-10
✨ Enhance classical control transformations in QCO dialect

### LINKED ISSUE #1928: Investigate classical-control replacement for RZ and RZZ gates
🤖 *AI text below* 🤖

## Context

PR #1756 replaces quantum controls whose input qubits have already been measured with classical `qco.if` conditions. Its target/control swap is intentionally limited to the supported phase gates `Z`, `S`, `T`, `P`, `Sdg`, `Tdg`, and `Id`, whose matrix form is `diag(1, …, e^{iφ})`.

`RZ` and `RZZ` are diagonal gates, but they do not satisfy that same phase-gate form, so the current direct target/control swap is not generally valid. This follow-up should determine whether the optimization can be extended with compensating phases or another semantics-preserving construction.

Related PR: https://github.com/munich-quantum-toolkit/core/pull/1756

## Scope

- Derive the exact conditions and identities under which measured controls or targets of `RZ` and `RZZ` operations can be replaced by classical control flow.
- Decide whether the additional operations and compiler complexity are justified by practical optimization value.
- If beneficial, implement the transformation with semantic-equivalence tests for representative angles, control positions, and multi-qubit cases.
- Otherwise, document why these gates remain outside the supported phase-gate subset.

simon1hofmann:
🤖 *AI text below* 🤖

## Description

This PR extends the QCO qubit-reuse optimizations to handle controlled `RZ` and
`RZZ` gates when controls or targets have already been measured. The rewrites
replace quantum dependencies on measured qubits with classical data flow while
preserving phase kickback on every surviving coherent control.

Fixes #1928.

## Changes

- Rewrite measured quantum controls into one `qco.if`; unmeasured controls
  remain quantum controls inside it.
- Replace a controlled `RZ` with a measured target by an outcome-selected phase
  on the remaining controls. Remove the operation directly when its control and
  target qubits have all been measured.
- Replace a controlled `RZZ` with one measured target by an outcome-selected
  controlled `RZ` on the other target. With two measured targets, retain only
  the outcome-selected phase on coherent controls. Remove the operation when
  every participating qubit has been measured.
- Preserve the mapping from modifier-body `qco.yield` operands to `RZZ` results,
  including bodies that yield transformed targets in a different order.
- Remove a single-qubit diagonal gate immediately before computational-basis
  measurement. Multi-qubit diagonal gates are commuted instead of deleted so
  phase kickback on other qubits remains observable.
- Add semantic regressions for both measured-target positions, multiple and
  measured controls, direct `RZ`/`RZZ` removal, reversed yields, observable
  phase kickback, and representative control widths and angles.

The implementation selects rotation angles directly from measurement outcomes.
This avoids the earlier nested phase-correction construction while keeping the
same exact quantum semantics.

## AI assistance

Codex materially assisted with reviewing the original implementation,
simplifying the measured-phase identities, preparing tests and documentation,
and running local validation. The contributor remains responsible for reviewing
and accepting these changes.

## Validation

- Focused measurement and classical-control tests: 38/38 passed.
- Full QCO optimization unit-test binary: 105/105 passed.
- Repository lint passed.
- LLVM 22.1.8 clang-tidy reported no source diagnostics in the four changed C++
  files.
- Generated MLIR documentation and the full Sphinx/Doxygen documentation build
  passed.
- Independent exact-revision verification found no actionable issues.

## Checklist

- [x] The PR is focused and relevant to the requested change.
- [x] Automated tests cover the changed behavior.
- [x] Pass documentation matches the implementation.
- [x] The changelog credits the PR author.
- [x] The changes follow the project's style guidelines.
- [x] AI assistance is disclosed above.
- [x] The contributor accepts responsibility for reviewing the assisted work.


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

burgholzer:
Had some Codex quota left before the reset, so I figured I may as well use it on this PR.

burgholzer:
Okay, I definitely want to simplify this a little bit further. On it.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -244,39 +244,17 @@ def HadamardLifting : Pass<"hadamard-lifting", "mlir::ModuleOp"> {
...
-    ┤ X ├┤ Meas ├ => ┤ Meas* ├┤ X ├
-    └───┘└──────┘    └───────┘└───┘
-    ```
-    Here, Meas* is a measurement after which the outcome is classically negated. X can also be replaced by Y, which
-    is likewise anti-diagonal.
+    This pass moves computational-basis measurements toward the start of a
+    circuit. A measurement of a control can move before the controlled
+    operation. A single-qubit diagonal operation immediately before a
+    measurement is removed because it cannot affect either the outcome or the
+    post-measurement qubit. Multi-qubit diagonal operations are only commuted:
```
DRovara, line 253: I guess doing this potentially drops a global phase. What's the `mqt-cc` policy on that? Do we care about global phases?

burgholzer: We generally care about global phases; however, I of not believe that this drops a global phase. If at all, it is a relative phase that is dropped, but that relative phase is eliminated by the measurement.

One concrete example, where the previous change was actually not correct:
Imagine `Z -- measure --`, which previously would be changed to `-- measure -- Z`.
This is not faithful/correct. Assume the measurement results in `|1>`. Then in the first circuit the state at the end of the circuit is `|1>`. In the second circuit, the result is `-|1>`.
So I would even be tempted to say that the previous replacement was not correct.

DRovara: Wait, are you sure about that? The formula for amplitude update after measurements is $\frac{\ket{1}\bra{1}\ket{\psi}}{\sqrt{\bra{\psi}\ket{1}\bra{1}\ket{\psi}}}$

If our state is -|1>, the denominator simplifies to 1, and the numerator remains -|1>, so the outcome is -|1>, right?

Actually, I just double-checked with Claude, who also agrees the correct outcome should be -|1> :thinking:

burgholzer: You are right! My bad. The "correct" outcome is -|1>.
Deleting the (global) phase is still safe though in my opinion as this phase cannot change any measurable outcome.

In my opinion, we should always try to maintain global phase equivalence for unitary synthesis and replacement/optimization. In cases like qubit reuse, it's much less about preserving a unitary (the measurement itself makes the whole process non-unitary), but more so about preserving the (non-unitary) quantum process.

Does that sound convincing?

DRovara: I have no problem with getting rid of global phase in qubit reuse, so I don't mind. Though I do feel like the line for when it is okay and when not is still a bit vague. After all, qubit reuse is also an optimisation. Maybe worth considering in more detail and then clarifying in the docs somewhere.

burgholzer: The thing here is, that we are not getting rid of a global phase; we are getting rid of a conditional-global-phase. When the measurement result is |0>, the Z gate does not add a phase, when the result is |1>, it adds a phase.

I do not really know, where one would point this out in the docs.
I'd currently be fine to just proceed with this now until we notice that we definitely need that thorough treatment for any use case (which I do not foresee at the moment).

burgholzer: Just FYI: I will merge this PR now, because it is not in conflict with main at the moment and ready except for this discussion, which also feels semi-resolved. If necessary, Let's take this to a potential follow-up issue/discussion instead of stalling this PR 😌

burgholzer: Just FYI: I will merge this PR now, because it is not in conflict with main at the moment and ready except for this discussion, which also feels semi-resolved. If necessary, Let's take this to a potential follow-up issue/discussion instead of stalling this PR 😌

PR #2043, opened 2026-08-10
🧪 Test QDMI device admission with Slurm

burgholzer:
🤖 *AI text below* 🤖

## Summary

- add a real Slurm 25.11+ cluster fixture with one controller and two compute nodes on Ubuntu 26.04
- use systemd, Munge, cgroup v2, `select/cons_tres`, two CPUs per node, two DDSIM licenses, and one SC license
- build one Python 3.14 wheel on the runner, install it in the disposable image with pinned uv, and run without `PYTHONPATH`
- rely on the packaged Slurm unit's `Delegate=yes` and assert that property at runtime
- reject live non-unit, AND, and OR QDMI license requests through the mechanism-specific selector
- run two DDSIM Bell jobs concurrently on separate nodes and hold both allocations after execution
- require a third DDSIM job to remain pending for `Licenses` while CPUs remain free
- run an SC job on a free CPU, then release one DDSIM job and require the pending job to execute

Each valid job requests one license and uses `mqt.core.qdmi.slurm.open_device_from_license()`. Each DDSIM result contains 256 samples and only Bell outcomes `00` and `11`.

## Resource-management boundary

The tutorial separates four controls:

1. Slurm licenses provide cluster-wide admission and accounting.
2. The mutable license environment selects a registered QDMI device.
3. The provider reports device availability and queue state.
4. Provider credentials or operating-system isolation authorize access.

The example is suitable for admission testing on a cooperative cluster. Strict isolation of a local device requires a GRES `File=` entry with `ConstrainDevices=yes`. Remote IQM and Amazon Braket devices require provider-scoped credentials.

The QDMI Python namespace from #2074 is now part of `main`. This PR remains the direct child of #2025 and the top layer of the #2025 → #2043 native stack.

## Validation

- Python and shell syntax
- Docker Compose configuration
- changed-file hooks and complete repository lint
- complete FoMaC C++, QDMI driver, and DDSIM suites
- QDMI Python suite on Python 3.10 and 3.14
- recursive nanobind stub regeneration with no source diff
- Sphinx documentation with warnings as errors
- Python 3.14 wheel build and wheel-content inspection
- tutorial length: 1,177 words
- independent exact-head adversarial audit: no source blocker
- exact-head privileged Slurm workflow: passed, including the installed wheel without `PYTHONPATH`, packaged `Delegate=yes`, non-unit/AND/OR rejection, concurrent admission, SC independence, DDSIM readmission, and Bell results

Replacement standard CI for the signed post-merge revision is in progress.

### DIFF
--- a/.github/workflows/slurm.yml
+++ b/.github/workflows/slurm.yml
@@ -1,0 +1,69 @@
+name: Slurm integration
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/slurm.yml"
+      - "bindings/qdmi/**"
+      - "cmake/**"
+      - "include/mqt-core/fomac/**"
+      - "include/mqt-core/qdmi/**"
+      - "pyproject.toml"
+      - "python/mqt/core/qdmi/**"
+      - "src/fomac/**"
+      - "src/qdmi/**"
+      - "test/slurm/**"
+      - "uv.lock"
+  pull_request:
+    paths:
+      - ".github/workflows/slurm.yml"
+      - "bindings/qdmi/**"
+      - "cmake/**"
+      - "include/mqt-core/fomac/**"
+      - "include/mqt-core/qdmi/**"
+      - "pyproject.toml"
+      - "python/mqt/core/qdmi/**"
+      - "src/fomac/**"
+      - "src/qdmi/**"
+      - "test/slurm/**"
+      - "uv.lock"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  slurm:
+    name: Slurm 25.11 and QDMI devices
+    runs-on: ubuntu-26.04
+    timeout-minutes: 45
+    env:
+      UV_NO_PROGRESS: "1"
+    steps:
+      - name: Check out MQT Core
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+
+      - name: Set up uv and Python
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+        with:
+          enable-cache: true
+          python-version: "3.14"
+
+      - name: Set up MLIR
+        uses: munich-quantum-software/setup-mlir@fc08baf0a172e98008e6528d465de3b5dd99bc68 # v1.4.1
+        with:
+          llvm-version: 22.1.7
+
+      - name: Build the MQT Core wheel
+        run: uv build --wheel --out-dir test/slurm/dist
+
+      - name: Test Slurm admission and QDMI execution
+        run: uv run --no-project --python 3.14 test/slurm/run_integration.py
--- a/docs/qdmi/slurm.md
+++ b/docs/qdmi/slurm.md
@@ -1,0 +1,218 @@
+# Use QDMI devices with Slurm
+
+This example uses Slurm 25.11 or newer on Ubuntu 26.04. It has one controller,
+two compute nodes, and two CPUs on each compute node.
+
+A Slurm license controls admission to a cluster-wide resource. In this setup,
+the license name is a stable QDMI device ID. A license does not show provider
+availability. It does not show the device queue. The QDMI provider supplies that
+information when its interface supports it.
+
+## Understand the control boundaries
+
+This example is suitable for admission and accounting tests on a cooperative
+cluster. It does not make a Slurm license an access-control credential. The
+controls are independent:
+
+- Slurm admits jobs and accounts for the configured license count.
+- The MQT Core adapter uses the license environment to select a registered QDMI
+  device.
+- The QDMI provider reports device availability and queue data.
+- The provider or the operating system authorizes access to the device.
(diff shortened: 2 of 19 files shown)


.github/workflows/slurm.yml
```diff
@@ -0,0 +1,75 @@
...
+    paths:
+      - ".github/workflows/slurm.yml"
+      - "bindings/qdmi/**"
+      - "cmake/**"
+      - "include/mqt-core/fomac/**"
+      - "pyproject.toml"
+      - "src/fomac/**"
+      - "src/qdmi/**"
+      - "test/slurm/**"
+      - "uv.lock"
```
burgholzer, line 31: Are all of these paths still up to date given the recent FoMaC to QDMI rename in MQT Core as part of this PR stack?

.github/workflows/slurm.yml
```diff
@@ -0,0 +1,75 @@
...
+    steps:
+      - name: Check out MQT Core
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.14"
```
burgholzer: there is no need to set up python here. this can all be folded into the setup-uv action.

.github/workflows/slurm.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: |
+          uv build --wheel --out-dir test/slurm/runtime/dist
+          uv pip install \
+            --python "$pythonLocation/bin/python" \
+            --target test/slurm/runtime/site-packages \
+            --no-deps \
+            test/slurm/runtime/dist/*.whl
+
+      - name: Test Slurm admission and QDMI execution
+        run: python test/slurm/run_integration.py
```
burgholzer: This feels like a very odd way of building MQT Core for the tests. There has to be a more elegant/native/modern way to do this with `uv`, Docker and GitHub Actions.

docs/qdmi/slurm.md
burgholzer, line 1: uses outdated FoMaC wording that needs to be replaced and rewritten throughout the entire PR.

test/slurm/slurmd-override.conf
burgholzer, line 1: Is this necessary? What is this for?

burgholzer (commented):
Just some comments I stumbled over while reviewing.

.github/workflows/slurm.yml
burgholzer: 🤖 *AI text below* 🤖

Addressed in `38daee37ec8dfce2c6329a4537a24ee70ddccc77`. The runner builds one wheel after the existing MLIR setup and places it in the Docker build context. The image copies the pinned official uv binary and installs that wheel with `uv pip install --system --no-deps`. The staged `site-packages` directory and every `PYTHONPATH` override are gone. The runtime mount now contains only the Munge key, job output, and result files, and the harness asserts that `mqt.core` was not imported from `/workspace` or `/runtime`. I am leaving the thread unresolved for re-review.

.github/workflows/slurm.yml
burgholzer: 🤖 *AI text below* 🤖

Addressed in `38daee37ec8dfce2c6329a4537a24ee70ddccc77`. The workflow now includes `include/mqt-core/qdmi/**` and `python/mqt/core/qdmi/**`. It retains the C++ FoMaC headers and sources because `fomac::slurm` and the `MQT::CoreFoMaC` target remain the current C++ API. Bindings, device/driver sources, build configuration, documentation, and the complete Slurm fixture remain in scope. I am leaving the thread unresolved for re-review.

test/slurm/slurmd-override.conf
burgholzer: 🤖 *AI text below* 🤖

Addressed in `38daee37ec8dfce2c6329a4537a24ee70ddccc77`. The custom `slurmd-override.conf` and its Docker copy are deleted. The fixture now relies on the modern packaged unit and explicitly checks `systemctl show slurmd --property=Delegate --value` before running the cgroup scenarios. A package regression therefore fails with a direct diagnostic instead of being masked by the test image. I am leaving the thread unresolved for re-review.

.github/workflows/slurm.yml
burgholzer: 🤖 *AI text below* 🤖

Addressed in `38daee37ec8dfce2c6329a4537a24ee70ddccc77`. The workflow no longer uses `actions/setup-python`. `astral-sh/setup-uv` now provisions Python 3.14, and the host integration harness runs through uv. I am leaving the thread unresolved for re-review.

docs/qdmi/slurm.md
burgholzer: 🤖 *AI text below* 🤖

Addressed in `38daee37ec8dfce2c6329a4537a24ee70ddccc77`. The tutorial now uses QDMI terminology and keeps FoMaC only where it names the current C++ namespace or target. It separately describes Slurm admission/accounting, mutable QDMI device selection, provider availability/queues, and provider or operating-system authorization. It also identifies the cooperative-cluster scope, GRES `File=` with `ConstrainDevices=yes` for strict local isolation, and provider-scoped credentials for IQM or Braket. I am leaving the thread unresolved for re-review.

burgholzer:
This also looks like a solid start for now. Once CI is green. This stack can be merged.

PR #2046, opened 2026-08-10
🔧 Add a lower bound for sphinx-llm

burgholzer:
🤖 *AI text below* 🤖

## Summary

- Require sphinx-llm 0.4.1 or newer, the latest available release when this direct lower bound was established.
- Follow the dependency-group policy of using the latest available version as the floor for direct requirements, keeping uv lowest-direct resolution intentional and warning-free.
- Synchronize the matching direct-requirement metadata in uv.lock without changing the normally resolved sphinx-llm version.

## Validation

- Python 3.10 uv sync dry run with lowest-direct resolution: resolved 177 packages without the unpinned sphinx-llm warning
- uv lock --check: passed
- Repository lint session: passed
- git diff --check: passed
- Commit signature: verified locally

A full documentation build was not repeated because this metadata-only change does not alter the locked documentation environment, which already resolves sphinx-llm 0.4.1.

AI assistance: GPT-5.6 via Codex prepared the original dependency metadata change, preserved the maintainer-selected version floor, regenerated uv.lock, and ran the reported validation.


pyproject.toml
```diff
@@ -334,7 +334,7 @@ docs = [
   "sphinx-autoapi>=3.6.1",
   "sphinx-copybutton>=0.5.2",
   "sphinx-design>=0.6.1",
-  "sphinx-llm",
+  "sphinx-llm>=0.1.2",
```
```suggestion
  "sphinx-llm>=0.4.1",
```
No need to go lower than necessary here.

burgholzer:
@denialhaag for awareness

denialhaag:
> @denialhaag for awareness

Thanks! I'll add this in the next round!

PR #2047, opened 2026-08-10
⚡ Speed up native documentation builds

burgholzer:
🤖 *AI text below* 🤖

## Summary

- build the documentation extension in `MinSizeRel` with unity builds enabled
- make MQT's native and MLIR targets unity-build compatible instead of excluding individual source files
- keep only the incompatible `capnp` target out of unity builds, colocated with its Jeff dependency setup
- use CMake's default unity batch size rather than maintaining an RtD-specific override
- omit the NA and SC QDMI provider shared libraries from the docs environment while retaining all native libraries, bindings, MLIR reference generation, and the DDSIM provider used by the notebooks

## Design

The earlier implementation used `SKIP_UNITY_BUILD_INCLUSION` for source files that happened to collide in four-source batches. Those exclusions were sensitive to source ordering and batch size. This revision removes every first-party unity exclusion and resolves the underlying translation-unit-local conflicts instead:

- common QC/QCO modifier-region construction and QCO linear-type detection use shared helpers
- implementation-local rewrite patterns, constants, and formatting helpers have descriptive, non-conflicting names
- default function arguments remain on declarations rather than being repeated on definitions

The global unity setting therefore applies across eligible MQT and Jeff targets. Cap'n Proto's concrete `capnp` target remains the sole explicit opt-out because its upstream translation units contain conflicting internal helpers; that target property is set directly after Jeff makes the transitive dependency available.

No explicit `CMAKE_UNITY_BUILD_BATCH_SIZE` is set. A fresh configure consequently uses [CMake's default batch size of eight](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD_BATCH_SIZE.html), avoiding a project-specific value that had not been measured on RtD and that hid additional source-order conflicts.

## Motivation

A recent successful `main` Read the Docs build took 1,661 seconds overall, with 1,578 seconds spent in the nox docs command. The earlier MinSizeRel/provider-only experiment still took 1,547 seconds overall and 1,454 seconds in nox, so build type and provider pruning alone were not sufficient. The logs show that native compilation dominates; dependency preparation and installation take only seconds.

Unity builds reduce repeated parsing of LLVM-heavy headers. This PR does not set `UV_LINK_MODE=copy`: that would suppress the hardlink fallback warning without materially reducing the dominant native compilation time.

## Validation

- clean native editable build with the exact docs CMake configuration and a fresh CMake cache
- confirmed default eight-source unity units and no first-party `SKIP_UNITY_BUILD_INCLUSION` exclusions
- 44 generated unity units across the MQT MLIR and Jeff build trees
- warnings-as-errors Sphinx build including Doxygen, MLIR reference pages, and `llms.txt`
- all seven executable MyST notebooks completed successfully after removing the Sphinx/Jupyter cache
- full `nox -s lint`

The final uncached Sphinx/notebook phase completed locally in 51 seconds with a warm native/compiler cache. The independent Read the Docs build remains the meaningful constrained-worker measurement and should be compared against the timings above.

AI assistance was used to analyze the Read the Docs logs, investigate unity-build boundaries, refactor the native sources for broader unity compatibility, validate fresh-cache behavior, and prepare the implementation and PR text. The complete native build, executable documentation, and lint suite were run before publication.

### DIFF
--- a/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
@@ -8,7 +8,6 @@
  * Licensed under the MIT License
  */
 
-#include "mlir/Dialect/QCO/IR/QCODialect.h"
 #include "mlir/Dialect/QCO/IR/QCOOps.h"
 #include "mlir/Dialect/QCO/QCOUtils.h"
 
--- a/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
@@ -10,6 +10,7 @@
 
 #include "mlir/Dialect/QCO/IR/QCODialect.h"
 #include "mlir/Dialect/QCO/IR/QCOOps.h"
+#include "mlir/Dialect/QCO/QCOUtils.h"
 
 #include <llvm/ADT/BitVector.h>
 #include <llvm/ADT/STLExtras.h>
@@ -35,15 +36,6 @@
 
 using namespace mlir;
 using namespace mlir::qco;
-
-static bool isLinearType(Type type) {
-  if (isa<QubitType>(type)) {
-    return true;
-  }
-  const auto tensorType = dyn_cast<RankedTensorType>(type);
-  return tensorType && tensorType.getRank() == 1 &&
-         isa<QubitType>(tensorType.getElementType());
-}
 
 void IfOp::build(OpBuilder& odsBuilder, OperationState& odsState,
                  Value condition, ValueRange qubits,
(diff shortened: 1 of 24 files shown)


mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include "mlir/Dialect/QCO/IR/QCODialect.h"
```
burgholzer:
```suggestion
```

mlir/lib/Dialect/QCO/IR/SCF/IndexSwitchOp.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include "mlir/Dialect/QCO/IR/QCODialect.h"
```
burgholzer:
```suggestion
```

PR #2048, opened 2026-08-11
🛠️ Support RTTI-free MLIR toolchains

burgholzer:
🤖 *AI text below* 🤖

## Description

Allow MQT Core to consume LLVM and MLIR built without C++ runtime type information.

- Set `LLVM_ENABLE_RTTI=OFF` and let LLVM's standard target policy configure ordinary MLIR targets.
- Apply one narrow target helper to exception-owning MLIR targets and direct LLVM consumers that otherwise escape or override that policy, including QIR JIT, the DDSIM QDMI device, the QIR runner, and both shipped and candidate Qiskit C-API adapter objects.
- Keep exception handling independent of RTTI: the explicit exception compatibility boundaries and test-only exception targets still finish with RTTI disabled.
- Keep RTTI only for the MLIR Python extension because nanobind's public headers directly instantiate `typeid`; every linked MQT MLIR library remains RTTI-free.
- Configure the unit-test program fixture libraries through the same central MLIR target policy.
- Update CI, Read the Docs, the development container, and the deployment image to the published portable LLVM/MLIR 22.1.8 no-RTTI release.
- Record the work under the generic QC/QCO compiler-infrastructure changelog entry.

Part of native GitHub stack #2064: #2054 → #2051 → #2048.

The producer-side change in [portable-mlir-toolchain#74](https://github.com/munich-quantum-software/portable-mlir-toolchain/pull/74) has merged, and [release 2026.08.11](https://github.com/munich-quantum-software/portable-mlir-toolchain/releases/tag/2026.08.11) publishes LLVM/MLIR 22.1.8 without RTTI for Linux, macOS, and Windows.

Fixes #1589.

## Validation

- A fresh Release cache against the published arm64 macOS LLVM/MLIR 22.1.8 archive builds all 598 native targets.
- All 4,547 discovered CTests pass; two live job-ID tests are skipped by design.
- All 118 direct Qiskit bridge tests pass against matching IR, FoMaC, and MLIR extensions.
- Compile-command audit confirms native LLVM/MLIR consumers use no RTTI, `mqt-cc` uses neither RTTI nor exceptions, and only the nanobind extension sources retain RTTI.
- `mqt-cc` lists all configured QDMI devices and compiles the Bell program for `mqt.ddsim.default`.
- Repository lint, workflow/config validation, and `git diff --check` pass. Replacement exact-head CI is tracked separately while checks are pending.

GPT-5.6 via Codex materially assisted with implementation, review remediation, testing, rebasing, and publication under maintainer direction.

PR #2051, opened 2026-08-11
🔧 Build native MQT MLIR targets without C++ exceptions

### LINKED ISSUE #1590: 🚸 MLIR - Enable building without exceptions enabled
Due to the integration with the `qc::QuantumComputation` class we currently need to build the MLIR code with exceptions enabled even though we do not need exceptions in the MLIR part of the code base.
Since exception handling adds extra code that may hinder performance, we want to look into disabling exceptions again for the MLIR code base.

This is intended as a tracking issue for removing the requirement to build with exceptions enabled.

burgholzer:
🤖 *AI text below* 🤖

## Description

Build the complete MQT MLIR compiler collection against an LLVM/MLIR toolchain built with `LLVM_ENABLE_EH=OFF`.

- Use LLVM's normal `llvm_update_compile_flags` policy for ordinary MLIR targets; `mqt-cc` and the rest of the compiler graph remain exception-disabled.
- Keep the complete compiler graph instead of introducing a feature option, strict preset, sentinel, or conditional omission of FoMaC, DD, QIR, or CLI targets.
- Limit explicit production exception overrides to compatibility boundaries that still instantiate throwing APIs: the FoMaC adapter, QCO DD functionality, and the standalone QIR runner. The Python extension retains nanobind's normal exception model outside that MLIR target policy.
- Retain exceptions for the small set of test executables whose translation units directly instantiate throwing Core IR or DD header code.
- Express the test-only policy through the existing target helper; there is no directory-wide generator-expression framework or custom exception target property.
- Record the work under the generic QC/QCO compiler-infrastructure changelog entry.

Part of native GitHub stack #2064: #2054 → #2051 → #2048.

Fixes #1590.

## Validation

- The #2051 layer builds the complete graph against LLVM/MLIR 22.1.3 with exceptions disabled by default; all 990 MLIR tests pass.
- Compile-command inspection confirms `mqt-cc` is exception-disabled and only the documented production and test boundaries retain exceptions.
- Repository lint, clang-tidy remediation, and `git diff --check` pass.

GPT-5.6 via Codex materially assisted with implementation, review remediation, testing, rebasing, and publication under maintainer direction.


.github/workflows/ci.yml
```diff
@@ -32,6 +32,9 @@ jobs:
           - runs-on: ubuntu-24.04
             compiler: gcc
             preset: debug
+          - runs-on: ubuntu-24.04
+            compiler: gcc
+            preset: no-exceptions
```
burgholzer: This should not be necessary and just adds extra CI time.

```suggestion
```

cmake/CompilerOptions.cmake
```diff
@@ -25,8 +25,10 @@ function(enable_project_options target_name)
     # always include debug symbols (avoids common problems with LTO)
     target_compile_options(${target_name} INTERFACE -g)
 
-    # ensure that exceptions are enabled
-    target_compile_options(${target_name} INTERFACE -fexceptions)
+    # ensure that exceptions are enabled unless a consuming target explicitly opts out
+    target_compile_options(
+      ${target_name}
+      INTERFACE "$<$<NOT:$<BOOL:$<TARGET_PROPERTY:MQT_DISABLE_EXCEPTIONS>>>:-fexceptions>")
```
burgholzer: I believe instead of this way around, I'd rather disable exceptions by default and enable them explicitly for targets that need them. That seems to be the better default in my opinion. Pivot accordingly.

CHANGELOG.md
```diff
@@ -106,6 +106,8 @@ releases may include breaking changes.
 
 ### Changed
 
+- 🔧 Allow building the native MQT MLIR target graph without C++ exception
+  support ([#2051]) ([**@burgholzer**])
```
burgholzer: This can be folded into the general entry  for the compiler collection.

CMakePresets.json
burgholzer, line 1: These should not be necessary in general and only served for validation purposes.

docs/installation.md
burgholzer, line 1: This is a templated file and must not be modified.

mlir/CMakeLists.txt
```diff
@@ -10,9 +10,23 @@ if(MSVC)
...
 
+option(MQT_MLIR_ENABLE_EXCEPTIONS "Build MQT MLIR targets with C++ exception support" ON)
+if(NOT MQT_MLIR_ENABLE_EXCEPTIONS)
+  if(NOT CMAKE_CXX_COMPILER_ID MATCHES "GNU|.*Clang")
+    message(
+      FATAL_ERROR "MQT_MLIR_ENABLE_EXCEPTIONS=OFF requires a GCC- or Clang-compatible compiler")
+  endif()
+  message(STATUS "Building MQT MLIR libraries without C++ exception support")
+  add_compile_options(-fno-exceptions)
+endif()
```
burgholzer: I do not want to make this conditional. The compiler collection should compile without exceptions enabled by default except for the parts that still need them.

mlir/tools/CMakeLists.txt
```diff
@@ -6,4 +6,6 @@
 #
 # Licensed under the MIT License
 
-add_subdirectory(mqt-cc)
+if(MQT_MLIR_ENABLE_EXCEPTIONS)
+  add_subdirectory(mqt-cc)
+endif()
```
burgholzer: mqt-cc shall not require exceptions to be enabled.

burgholzer (commented):
Just some comments with requests for changes that came up while reviewing the PR.

burgholzer:
🤖 *AI text below* 🤖

Superseded by #2125, which combines exception-free and RTTI-free support for stock LLVM and MLIR builds on the current `main` branch.

PR #2040, opened 2026-08-10
✨ Add OpenQASM-native fixed-width angles

burgholzer:
🤖 *AI text below* 🤖

## Description

Import OpenQASM 3.1 fixed-width `angle[N]` values into typed QC MLIR without adding a custom angle type or using the Quant dialect.

- Retain angle kind and width in the OpenQASM semantic model while lowering storage, arithmetic, and structured state to builtin signless `iN` and unsigned MLIR operations.
- Implement exact modulo-2π, ties-to-even constant and runtime conversion, width promotion, resizing, casts, comparisons, bit operations, scalar bit access, and gate-parameter semantics for widths 1 through 64.
- Add shared canonical angle-bit/radian builders and matchers for later boundary layers.
- Preserve scalar interface kind and name metadata while deriving width from the builtin MLIR type.
- Keep QC/QCO gate definitions, the default compiler pipeline, target schema, legacy circuit IR, and QIR ABI unchanged in this layer.

Part of native GitHub stack: #2040 → #2061 → #2062.

Part of #1128.

## Validation

- The angle-conversion utility and typed OpenQASM frontend/target suites pass locally.
- Tests cover unsized and explicit widths, exact rounding, modular arithmetic, mixed-width promotion, casts, comparisons, bit-register bridges, scalar bit access, gate parameters, and structured control flow.
- The final aggregate stack builds successfully and passes all 4,547 discovered CTests; two live QDMI job-ID tests are skipped by design.
- Repository lint and `git diff --check` pass.

GPT-5.6 via Codex materially assisted with implementation, independent review remediation, testing, restacking, and publication under maintainer direction.

PR #2061, opened 2026-08-12
♻️ Preserve fixed-width angles in OpenQASM export

burgholzer:
🤖 *AI text below* 🤖

## Description

Preserve fixed-width angle semantics when exporting QC MLIR back to OpenQASM 3.

- Reconstruct canonical integer-to-angle and float-to-angle conversions as official `angle[N]` casts.
- Preserve angle-valued inputs, outputs, arithmetic, comparisons, divisions, casts, custom-gate captures, and structured loop-carried state.
- Emit legal bit-register intermediates where exact angle bit patterns must be reconstructed.
- Keep the contract semantic: exported OpenQASM reparses to the same angle bit patterns and gate parameters.

Depends on #2040.

## Validation

- OpenQASM target and QC translation suites pass locally.
- Round-trip coverage includes constants, dynamic values, bit-pattern casts, composite helpers, scalar widths 1 through 64, and structured control flow.

GPT-5.6 via Codex materially assisted with implementation, independent review remediation, testing, restacking, and publication under maintainer direction.

### DIFF
--- a/docs/mlir/OpenQASM.md
+++ b/docs/mlir/OpenQASM.md
@@ -35,26 +35,57 @@ mqt-cc --input-format=qasm program.txt
 
 ### Input support
 
-| OpenQASM concept           | Support and restrictions                                                                                                                                                                        |
-| -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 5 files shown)


burgholzer:
🤖 *AI text below* 🤖

Superseded by #2169, which handles the compile-time input semantics and emits valid canonical OpenQASM 3 output without exporter-side angle reconstruction or format-specific MLIR attributes. Closing this PR in favor of the replacement.

PR #2062, opened 2026-08-12
✨ Add terminal QCO gate-angle quantization

burgholzer:
🤖 *AI text below* 🤖

## Description

Add an explicit, opt-in QCO pass for quantizing actual gate parameters to OpenQASM fixed-width angle precision.

- Add `quantize-gate-angles{precision-bits=N}` with required precision in `[1,64]`.
- Expose `QCOProgram::quantizeGateAngles` and `QCOProgram.quantize_gate_angles`.
- Quantize constants and runtime parameters with exact modulo-2π, ties-to-even semantics, including nested control, inverse, and power regions without changing power exponents.
- Make quantization terminal so ordinary cleanup runs before it and cannot erase its canonical angle conversions afterward.
- Preserve same-width idempotence and exact different-width angle resizing.

Depends on #2061.

## Validation

- QCO optimization and compiler suites pass locally.
- Tests cover constant and dynamic values, negative and multi-turn inputs, widths 1/8/53/64, modifiers, global phases, repeated passes, cleanup boundaries, C++ and Python APIs, and OpenQASM export.

GPT-5.6 via Codex materially assisted with implementation, independent review remediation, testing, restacking, and publication under maintainer direction.

### DIFF
--- a/bindings/mlir/register_mlir.cpp
+++ b/bindings/mlir/register_mlir.cpp
@@ -638,6 +638,10 @@ operations.)pb");
            &BooleanMemberAdapter<
                &mlir::QCOProgram::mergeSingleQubitRotationGates>::call,
            "Merge compatible consecutive single-qubit rotation gates.")
+      .def("quantize_gate_angles",
+           &BooleanMemberAdapter<&mlir::QCOProgram::quantizeGateAngles>::call,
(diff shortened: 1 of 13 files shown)


burgholzer:
Closing this PR as all its predecessors have been closed as well. We'll revisit this at a later point in time.

PR #2054, opened 2026-08-11
💥 Remove the legacy circuit-to-MLIR translation

burgholzer:
🤖 *AI text below* 🤖

## Description

Remove the one-way translator from the legacy `qc::QuantumComputation` circuit representation to the compiler's QC dialect.

- Delete the translator, its public C++ and Python factories, its fixtures, and its translation tests.
- Remove `QuantumComputation` from the generic Python compiler input types.
- Remove the compiler and Python binding dependencies on `MQT::CoreIR` that existed only for this translation.
- Keep OpenQASM, Qiskit circuits, textual MLIR, and typed MLIR programs as compiler inputs.
- Keep direct Qiskit circuit import and export from #2031.
- Keep direct QCO decision-diagram construction and simulation. Tests can still use `QuantumComputation` as an independent result oracle.
- Document the breaking change and the OpenQASM 3 migration path. The legacy `mqt.core.load` and Qiskit converter APIs do not change.

Part of #1590.

## Validation

- Configured and built `mqt-core-mlir-unittests-compiler` with LLVM/MLIR 22.1.3.
- 131 compiler tests pass.
- Repository lint and `git diff --check` pass.

GPT-5.6 via Codex materially assisted with the history rewrite, conflict resolution, implementation review, documentation, testing, and publication under maintainer direction.

PR #2050, opened 2026-08-11
♻️ Propagate OpenQASM semantic errors explicitly

burgholzer:
🤖 *AI text below* 🤖

## Description

Replace the OpenQASM semantic analyzer's private exception unwinding with explicit `mlir::LogicalResult` and `mlir::FailureOr<T>` propagation. The analyzer retains the first diagnostic and returns a typed program only after every semantic phase succeeds.

The conversion covers constant evaluation, casts, declarations, assignments, structured control flow, qubit and bit resolution, recursive-gate validation, and output initialization while preserving the public frontend result. Review remediation keeps gate-parameter expressions such as `theta + 1` valid through contextual constant promotion without allowing ordinary runtime integer/angle mixing. The small assignment helper now moves the successful value rather than the `FailureOr` wrapper and carries no clang-tidy suppression outside its definition.

## Validation

- All 98 `OpenQASMFrontendTest` cases pass locally, including recursive-gate, source-location, and gate-angle regressions.
- No throw, catch, private semantic exception, or standard exception header remains in `OpenQASMSemantics.cpp`.
- The final aggregate stack builds successfully and passes all 4,547 discovered CTests; two live QDMI job-ID tests are skipped by design.
- Repository lint, targeted clang-tidy remediation, and `git diff --check` pass.

GPT-5.6 via Codex materially assisted with implementation, review remediation, testing, rebasing, and publication under maintainer direction.

PR #2063, opened 2026-08-12
🛡️ Enforce frozen QIR 2.1 profile boundaries

### LINKED ISSUE #1128: 📐 MLIR - Represent OpenQASM 3 angle as fixed-point (quantized) type
## Background
OQ3 angles correspond to binary angular measure (fixed precision). Some hardware supports limited precision (e.g., a few bits).

## Problem
Floats are convenient but may misrepresent precision/semantics; bit-arrays are clunky.

## Proposal:
- Evaluate MLIR’s quant/fixed-point facilities to model angles; define a mapping and conversions to/from `arith` when needed.
- Decide where angle typing lives (Ref vs Opt vs frontend-only) and document trade-offs.

This partially depends on #30 if we want to build the support for the angle type into the regular MQT Core IR.
Alternatively, one could think about adapting the existing OQ3 parser to act as a direct translator to MLIR. There is no current tracking issue for that.

## Acceptance criteria:
- Design note with chosen representation and examples; parser/lowering hooks planned or implemented.

burgholzer:
🤖 *AI text below* 🤖

## Description

Enforce the official frozen QIR 2.1 entry-point and profile boundaries for fixed-width angle programs.

- Normalize generated QIR entry points to parameterless functions returning `void` and keep QIS calls on their prescribed `void` ABI.
- Emit the frozen profile module-flag tuple and validate source interfaces before lowering.
- Allow constant angle conversions that fold before QIR emission.
- Reject dynamic angle bridges, modular wrapping, lossy narrowing, unsupported intrinsics, assertion runtime calls, and scalar inputs/outputs when the selected Base or Adaptive profile cannot represent their semantics.
- Document the boundary against the frozen QIR 2.1 publication rather than later additions on the specification main branch.

Depends on #2062.

Fixes #1128.

## Validation

- Base and Adaptive QIR suites pass locally, including source-facing rejection diagnostics and module/entry-point ABI checks.

GPT-5.6 via Codex materially assisted with specification validation, production precedent research, implementation, independent review remediation, testing, restacking, and publication under maintainer direction.

### DIFF
--- a/docs/qir/index.md
+++ b/docs/qir/index.md
@@ -98,6 +98,36 @@ QIR entry points take no arguments and return an `i64` exit code. Runtime and
 QIS declarations are checked before JIT compilation; a mismatched or unsupported
 declaration is reported with its actual and accepted LLVM function types.
 
+QC-to-QIR conversion therefore rejects unresolved OpenQASM `input` declarations
+instead of exposing them as non-profile entry-point parameters. Bind or
+specialize source inputs before QIR lowering. Source scalar outputs are not QIR
+function results either: current conversion records measurement results and
+rejects other scalar outputs until they have an explicit output-record lowering.
+This is especially important for `uint[64]` and `angle[64]`, whose complete
+unsigned domains have no single standard QIR 2.1 output primitive.
+
+Adaptive Profile emission records the classical integer and floating-point
+widths used by runtime computations in the required module flags. Base Profile
+angle constants are folded to constant `f64` QIS parameters, so their internal
+integer representation does not introduce local classical computation. Dynamic
+fixed-width angle conversion is rejected for both profiles because the required
+`bitcast` and unsigned-integer-to-floating-point instructions are not part of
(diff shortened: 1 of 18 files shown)


burgholzer:
🤖 *AI text below* 🤖

Closing this PR without rebasing. The independently useful QIR 2.1 metadata correction was extracted and merged in #2066. The remaining changes depend on the unmerged fixed-width angle infrastructure in #2040, #2061, and #2062 and implement only a partial, OpenQASM-specific source validation pass. Issue #1128 concerns fixed-width angle representation and quantization, so it remains open and continues through those three PRs.

A comprehensive QIR-profile validator should be proposed independently when there is a concrete backend capability contract. It should validate the final LLVM-dialect representation through profile-aware MLIR target legality.

Commit `0bcbd5357` remains available for comparison. It is also still present in the ancestry of the open #2050, #2054, #2051, and #2048 branches and must be omitted during their separately authorized rebases.

PR #2077, opened 2026-08-13
✨ Add QCO DD sampling and classical registers

simon1hofmann:
🤖 *AI text below* 🤖

## Summary

Extend QCO decision-diagram execution and expose it through Python.

- Add `build_functionality`, `simulate`, and zero-state `sample` bindings.
- Execute measurement/reset, concrete classical SSA, CBit registers, `qco.if`, `qco.index_switch`, bounded `scf.for`, and non-recursive single-block calls.
- Sample correlated returned CBit registers, with final-basis sampling as the fallback.
- Reuse one DD evolution for deferrable terminal measurements.

## Scope

- Functionality construction remains limited to compile-time-known unitaries.
- Sampling starts from zero; simulation accepts a live input DD from the supplied package.
- Qtensors, `scf.while`, recursion, and multi-block functions remain unsupported. Nested execution has a shared 10,000-step budget.
- Direct C++ APIs require verified, linear QCO.
- The implementation relies on the static-qubit ownership and uniqueness guarantees now provided by `main`.
- Cleanup remains explicit because canonicalization can remove terminal wires and change fallback sample width.

## Validation

- Release C++ suite: 4,042/4,042 passed, with one documented skip.
- Focused QCO DD suite: 47/47 C++ and 6/6 Python tests passed.
- Python Nox matrix: 3,143 tests passed across Python 3.11–3.14, with documented skips.
- Stub generation, repository-wide lint, C++ lint, the full release build, and `git diff --check` passed.

GPT-5.6 via Codex materially assisted with implementation, testing, review remediation, and restructuring. I reviewed the resulting changes and take responsibility for them.

## Checklist


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp
```diff
@@ -736,28 +811,121 @@ static LogicalResult applyOp(Operation& op, WalkState& walk, StateDD& state) {
...
+        if (failed(walkBlock(callee.getBody().front(), walk, state))) {
+          return failure();
+        }
+
+        // Map callee return operands onto call results via the return op.
+        auto returnOp =
+            cast<func::ReturnOp>(callee.getBody().front().getTerminator());
+        return bindValuePairs(returnOp.getOperands(), call.getResults(), walk,
+                              call);
+      })
```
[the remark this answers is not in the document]
simon1hofmann: should be fixed

burgholzer (approved):
Hey @simon1hofmann 👋🏼 
I pushed a couple of changes here that simplify the implementation quite a bit. I also looked through all the resulting changes and I believe this looks good now.
Maybe you could have a look as well and decide whether it also looks good for you. If yes, feel free to merge once CI is green!

simon1hofmann:
> Hey @simon1hofmann 👋🏼 I pushed a couple of changes here that simplify the implementation quite a bit. I also looked through all the resulting changes and I believe this looks good now. Maybe you could have a look as well and decide whether it also looks good for you. If yes, feel free to merge once CI is green!

I pushed four more commits, do they make sense?

- **`scf.for` carried values were rebound sequentially**, corrupting qubit, scalar, CBit-register, and induction-variable permutations. Fixed by snapshotting source bindings before updating destinations. ([`[61c6fedca](https://github.com/munich-quantum-toolkit/core/commit/61c6fedca)`](https://github.com/munich-quantum-toolkit/core/commit/61c6fedca))

- **Duplicate `qco.static` indices created independent SSA roots for the same physical wire**, allowing deferred measurements to cross later operations on that wire. Such aliases are now rejected at the shared execution boundary. ([`[bf005cf24](https://github.com/munich-quantum-toolkit/core/commit/bf005cf24)`](https://github.com/munich-quantum-toolkit/core/commit/bf005cf24))

- **`simulate` accepted an initial DD narrower than the program**, producing an output width dependent on which wires the program touched. Inputs must now span every program wire; wider states remain supported. ([`[dacf5f246](https://github.com/munich-quantum-toolkit/core/commit/dacf5f246)`](https://github.com/munich-quantum-toolkit/core/commit/dacf5f246))

- **The `seed=None` documentation was inaccurate**: the implementation rejects every measurement and reset without an RNG, not only mid-circuit operations. The C++ and Python documentation and generated stub now state this correctly. ([`[2e98e48ea](https://github.com/munich-quantum-toolkit/core/commit/2e98e48ea)`](https://github.com/munich-quantum-toolkit/core/commit/2e98e48ea))

burgholzer:
> I pushed four more commits, do they make sense?
> 
> * **`scf.for` carried values were rebound sequentially**, corrupting qubit, scalar, CBit-register, and induction-variable permutations. Fixed by snapshotting source bindings before updating destinations. ([`[61c6fedca](https://github.com/munich-quantum-toolkit/core/commit/61c6fedca)`](https://github.com/munich-quantum-toolkit/core/commit/61c6fedca))

I believe that makes sense!

> * **Duplicate `qco.static` indices created independent SSA roots for the same physical wire**, allowing deferred measurements to cross later operations on that wire. Such aliases are now rejected at the shared execution boundary. ([`[bf005cf24](https://github.com/munich-quantum-toolkit/core/commit/bf005cf24)`](https://github.com/munich-quantum-toolkit/core/commit/bf005cf24))

Hm. The fix itself kind of makes sense. Although I am was expecting that we have canonicalizations that eliminate such duplicate static qubit statements. Which makes me question whether we should potentially run canonicalization before simulation to have stricter guardrails of how the progams to be simulated look like. Could be a fairly useful way to simplify parts of the implementation here (including the changes in this commit, which would not be necessary under the assumption that canonicalization can remove duplicate static qubit operations).

> * **`simulate` accepted an initial DD narrower than the program**, producing an output width dependent on which wires the program touched. Inputs must now span every program wire; wider states remain supported. ([`[dacf5f246](https://github.com/munich-quantum-toolkit/core/commit/dacf5f246)`](https://github.com/munich-quantum-toolkit/core/commit/dacf5f246))

Makes a lot of sense 👍🏼 

> * **The `seed=None` documentation was inaccurate**: the implementation rejects every measurement and reset without an RNG, not only mid-circuit operations. The C++ and Python documentation and generated stub now state this correctly. ([`[2e98e48ea](https://github.com/munich-quantum-toolkit/core/commit/2e98e48ea)`](https://github.com/munich-quantum-toolkit/core/commit/2e98e48ea))

Here I am also not too sure. Does the tertiary nature of the seed parameter really make too much sense (especially given that the default is `None`)?
Wouldn't it be better to always supply a `seed` and simply default to `0` (the non-deterministic case)?

simon1hofmann:
> > * **Duplicate `qco.static` indices created independent SSA roots for the same physical wire**, allowing deferred measurements to cross later operations on that wire. Such aliases are now rejected at the shared execution boundary. ([`[bf005cf24](https://github.com/munich-quantum-toolkit/core/commit/bf005cf24)`](https://github.com/munich-quantum-toolkit/core/commit/bf005cf24))
> 
> Hm. The fix itself kind of makes sense. Although I am was expecting that we have canonicalizations that eliminate such duplicate static qubit statements. Which makes me question whether we should potentially run canonicalization before simulation to have stricter guardrails of how the progams to be simulated look like. Could be a fairly useful way to simplify parts of the implementation here (including the changes in this commit, which would not be necessary under the assumption that canonicalization can remove duplicate static qubit operations).

I'm not really sure that works, I traced it back using Codex and this is the summary:

Because CSE removes the exact evidence that `bf005cf24` checks.

Before cleanup:

```mlir
%a = qco.static 0
%b = qco.static 0
qco.measure %a
qco.x %b
```

`qco.static` is marked `Pure` and is completely identified by its index ([[QCOOps.td:77-92](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/include/mlir/Dialect/QCO/IR/QCOOps.td#L77-L92)](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/include/mlir/Dialect/QCO/IR/QCOOps.td#L77-L92)). Cleanup runs CSE ([[Passes.cpp:127-133](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Support/Passes.cpp#L127-L133)](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Support/Passes.cpp#L127-L133)), producing:

```mlir
%q = qco.static 0
qco.measure %q
qco.x %q
```

Now the duplicate-index loop sees only one `qco.static`, so it passes ([[DDFunctionality.cpp:960-969](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp#L960-L969)](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp#L960-L969)). However, `%q` has two consumers, violating QCO’s linear qubit semantics. The normal MLIR verifier does not currently enforce that invariant.

The interpreter’s qubit lookup is non-consuming, so both operations resolve `%q` to physical qubit 0 ([[DDFunctionality.cpp:70-94](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp#L70-L94)](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp#L70-L94)). Sampling also considers the measurement deferrable because it only examines the measurement’s output uses ([[DDFunctionality.cpp:1076-1086](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp#L1076-L1086)](https://github.com/munich-quantum-toolkit/core/blob/2e98e48ea07caaa26db50d62dc1084103f40b66a/mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp#L1076-L1086)). Consequently, `x` executes first and the deferred measurement observes `1`.

Both checks are therefore necessary:

- Before CSE: two static definitions, each single-use → duplicate-index guard catches it.
- After CSE: one static definition with two uses → linear-use guard catches it.

The proper long-term home is QCO verification; a `prepare` check is the minimal protection for #2077.


> > * **The `seed=None` documentation was inaccurate**: the implementation rejects every measurement and reset without an RNG, not only mid-circuit operations. The C++ and Python documentation and generated stub now state this correctly. ([`[2e98e48ea](https://github.com/munich-quantum-toolkit/core/commit/2e98e48ea)`](https://github.com/munich-quantum-toolkit/core/commit/2e98e48ea))
> 
> Here I am also not too sure. Does the tertiary nature of the seed parameter really make too much sense (especially given that the default is `None`)? Wouldn't it be better to always supply a `seed` and simply default to `0` (the non-deterministic case)?

Makes sense, pushed another commit to align that.

burgholzer:
> I'm not really sure that works...
>
> Before cleanup:
> 
> ```mlir
> %a = qco.static 0
> %b = qco.static 0
> qco.measure %a
> qco.x %b
> ```
> ... producing ...
> ```mlir
> %q = qco.static 0
> qco.measure %q
> qco.x %q
> ```


Okay, first things first: the input program is clearly illegal and should be detected as such since it violates linear typing. (there must not be two static `qco.static` operations with the same index as part of a single program)
I also very much doubt that there currently is a way to produce that program from QC, Qiskit, OpenQASM, or even from `jeff`. QCO is not an input dialect. So we can be much more assertive on "correct" programs.
I would like to avoid working around such hypotheticals and avoid having to add lots of guardrails for practically irrelevant programs.
MLIR's guide says that passes (and I'd consider the simulation also just some kind of pass) may assume to get valid IR.
Let's fix this at its root here if the verifier does not do so yet. Then let's strip all the unnecessary checking code from the PR that only reassures that the IR is valid.
Lastly, let's consider adding the canonicalization to the simulation pipeline, which allows us to make even more assumptions about the programs and their shape; and likely allows us to delete more code.

simon1hofmann:
> > I'm not really sure that works...
> > Before cleanup:
> > ```mlir
> > %a = qco.static 0
> > %b = qco.static 0
> > qco.measure %a
> > qco.x %b
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ... producing ...
> > ```mlir
> > %q = qco.static 0
> > qco.measure %q
> > qco.x %q
> > ```
> 
> Okay, first things first: the input program is clearly illegal and should be detected as such since it violates linear typing. (there must not be two static `qco.static` operations with the same index as part of a single program) I also very much doubt that there currently is a way to produce that program from QC, Qiskit, OpenQASM, or even from `jeff`. QCO is not an input dialect. So we can be much more assertive on "correct" programs. I would like to avoid working around such hypotheticals and avoid having to add lots of guardrails for practically irrelevant programs. MLIR's guide says that passes (and I'd consider the simulation also just some kind of pass) may assume to get valid IR. Let's fix this at its root here if the verifier does not do so yet. Then let's strip all the unnecessary checking code from the PR that only reassures that the IR is valid. Lastly, let's consider adding the canonicalization to the simulation pipeline, which allows us to make even more assumptions about the programs and their shape; and likely allows us to delete more code.

That makes a lot of sense. I still have to get my head around the fact that not every corner case is possible. 
Would probably be good to merge #2220 first so `qco::verifyLinearity` can be used.

burgholzer:
Yeah, let me pull out #2220 real quick to not block this for too long 👍🏼

burgholzer:
#2220 just merged 🚀

simon1hofmann:
🤖 *AI text below* 🤖

Follow-up after merging #2220:

`verifyLinearity` previously checked only that every linear qubit SSA value had exactly one use. That does not catch two distinct values such as two `qco.static 0` operations: each value can have one use while both still alias the same physical qubit.

Changed in `fb25224f5`:

- Track `qco.static` indices per nearest `IsIsolatedFromAbove` scope, normally each `func.func`.
- Reject duplicate static indices within the same scope while allowing independent functions to reuse an index.
- Keep the existing one-use verification unchanged.

This puts the invariant at the shared checked boundary used by `QCOProgram::fromModule`, so the DD executor can assume valid linear IR and its duplicate validity checks can be removed.

`33f66d57a` changes the tracking set to `uint64_t`, matching `StaticOp::getIndex()` and resolving the `bugprone-narrowing-conversions` warning. The replacement GitHub C++ lint passed with zero annotations; all 488 focused QCO IR tests also pass.

simon1hofmann:
🤖 *AI text below* 🤖

Follow-up to the static-qubit discussion: the earlier per-function interpretation was too weak, and there is a valid producer for repeated static references.

- `OPENQASM 3.1; h $0; x $0;` produced two `qc.static 0` operations before lowering. This is valid source input, so rejecting it only after it becomes QCO would reject a real frontend path.
- [`c242b1436`](https://github.com/munich-quantum-toolkit/core/commit/c242b14365b2869147cc8d5de2cf3a3262970207) now coalesces repeated `qc.static` references at the start of QC-to-QCO conversion, before capture analysis. Repeated references in structured control flow therefore lower to one linear QCO value.
- `qco::verifyLinearity` now treats static indices as module-wide physical identities, matching the requirement that one program must not contain two `qco.static` roots for the same site. Helper functions must receive qubits through arguments instead of declaring another root.
- The simulator still assumes verified QCO and does not duplicate these validity checks. Cleanup remains explicit: the general cleanup pipeline can erase terminal wires and thereby change the width of fallback sampling, so it is not safe to insert unconditionally here.

Regression coverage includes duplicate references across `scf.if` regions, two duplicated indices, repeated OpenQASM hardware-qubit use, and rejection of cross-function aliases. The full release suite passes 4,036/4,036 tests with one documented skip; repository-wide lint passes. The standalone local C++ lint session requires clang-tidy 22, which is unavailable on this machine, so the hosted C++ lint check remains authoritative for that tool.

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
```diff
@@ -42,7 +46,37 @@ namespace mlir::qco {
...
+  func::FuncOp entryPoint;
+  if (auto moduleOp = dyn_cast<ModuleOp>(root)) {
+    for (auto funcOp : moduleOp.getOps<func::FuncOp>()) {
+      if (funcOp->hasAttr(
+              mqt::MQTDialect::EntryPointAttrHelper::getNameStr())) {
+        entryPoint = funcOp;
+        break;
+      }
+    }
+  }
```
burgholzer, line 52: Didn't we have a utility function for this somewhere already?

python/mqt/core/mlir.pyi
```diff
@@ -541,6 +542,60 @@ class QIRProgram(Program):
...
+            value produces reproducible results.
+
+    Returns:
+        Histogram of returned CBit registers in return order, each MSB first. If
+        no CBit result exists, final ``measureAll`` bitstrings instead.
+
+    Raises:
+        ValueError: When the program is unsupported for sampling.
+    """
+
```
burgholzer: It feels like these could either be methods of the `QCOProgram` class, or they could be moved to an `mlir.dd` submodule. Might help to avoid this file growing larger and larger.

mlir/include/mlir/Dialect/QCO/Utils/DDFunctionality.h
burgholzer: Does the distinction between these two really make too much sense? We do not distinguish them in the Python interface anymore.
I believe all simulation methods should get an RNG. The distinction about how the simulation is handled should be an implementation detail.

burgholzer (changes requested):
Hey @simon1hofmann 👋🏼 
I pushed a couple of simplification to the PR that streamline the `static` handling a little bit. Hope these make sense. I have a couple of comments left to address. But this should be fairly close to mergeable! 😌

mlir/lib/Dialect/QCO/IR/QCOUtils.cpp
simon1hofmann: Seems like this has been resolved already.

python/mqt/core/mlir.pyi
simon1hofmann: 🤖 *AI text below* 🤖

Addressed in `0db7f5c1a`. The DD bindings are now `QCOProgram.build_functionality`, `QCOProgram.simulate`, and `QCOProgram.sample`. The tests and generated stub use the methods, and the module-level exports are gone.

mlir/include/mlir/Dialect/QCO/Utils/DDFunctionality.h
simon1hofmann: 🤖 *AI text below* 🤖

Addressed in `0db7f5c1a`. The public `mlir::qco::simulate` API now always takes `std::mt19937_64&`. Nullable RNG handling remains private to the implementation for optimized deferred sampling, and all C++ callers and tests now pass an RNG.

burgholzer (approved):
LGTM 👍🏻 let's get this in!

mlir/lib/Dialect/QCO/Utils/DDFunctionality.cpp
```diff
@@ -869,110 +972,215 @@ FailureOr<dd::MatrixDD> buildFunctionality(func::FuncOp func, dd::Package& dd) {
...
+           }
+           return isa<func::ReturnOp>(use.getOwner());
+         });
 }
 
-[[nodiscard]] static bool requiresDynamicSampling(func::FuncOp func) {
-  bool dynamic = false;
-  bool measured = false;
+static bool isDeferrableMeasurement(MeasureOp measure, Block* entry,
+                                    const ArrayRef<Value> outputs) {
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Addressed in `919bc8c19`. I removed top-level `const` from the listed parameters and the remaining sibling by-value parameters and `ArrayRef` view in `DDFunctionality.cpp`, so the file follows the MLIR rule consistently. Full lint and all 126 QCO utility tests pass.

simon1hofmann:
🤖 *AI text below* 🤖

Addressed the outside-diff `arith.extui` finding from the latest CodeRabbit review in `01e8bb8aa`. Both dispatch layers now recognize `arith::ExtUIOp`, the evaluator reuses the existing unsigned-cast implementation, the public support list is updated, and `SimulateExtUI` makes the zero extension observable through `qco.if`.

One correction to the suggested test: `arith.extui i1 -> index` is invalid MLIR because `arith.extui` only accepts fixed-width integer-like types. The regression therefore covers the valid `i1 -> i8` form. Full lint and all 127 QCO utility tests pass; the standalone `cpp-lint` session cannot start on this host because the project requires clang-tidy 22.

PR #2083, opened 2026-08-13
⚡ Reduce build and CI times

burgholzer:
🤖 *AI text below* 🤖

## Summary

- enable unity builds for first-party CMake presets and Python package builds
- keep lint non-unity and verify interface headers only in the lint configuration
- reduce the routine C++ matrix from nine jobs to six and remove dormant extensive-CI paths

## Motivation

Native compilation dominates MQT Core CI. This is especially costly on macOS, where the organization-wide five-runner limit causes long queue times. Cold local AppleClang measurements reduced a representative native and MLIR build from 111.2 seconds to 73.2 seconds with unity compilation, about 34 percent.

The final matrix keeps one Debug and one Release C++ job on each operating system. The unchanged Python platform jobs retain Release coverage for each omitted C++ architecture. This reduces the macOS matrix from five jobs to four and leaves one organization runner available.

## Implementation

Normal Debug, Release, Coverage, Windows, and scikit-build-core configurations enable unity builds. CMake supplies its default batch size of eight. The lint preset explicitly disables unity and enables interface-header verification so clang-tidy can diagnose missing includes and source-order dependencies.

Jeff and its transitive Cap'n Proto dependencies compile without unity through a function-scoped `CMAKE_UNITY_BUILD` setting. This avoids target discovery and keeps the setting out of first-party targets and embedded consumers. The test-only collisions exposed by unity builds use distinct helpers, global namespace qualification, and contained using directives. No first-party target is excluded from unity compilation.

The CI cleanup removes the x64 Release C++ row on Linux and Windows, the ARM Release row on macOS, and all extensive C++ and Python jobs and label conditions. Python tests, minimums, coverage, packaging, and wheel matrices are unchanged. Interprocedural optimization keeps its existing behavior and remains disabled by CI deployment configuration.

## Validation

- fresh Release unity build: 4,488 native tests passed, with two expected skips
- fresh Debug unity build: 4,488 native tests passed, with two expected skips
- generated first-party unity units contain at most eight sources; Jeff and Cap'n Proto compile from separate source files
- non-unity lint build completed; its compile database contains 1,112 interface-header verification entries and no unity source
- `tests-3.14`: 598 passed, 3 expected Qiskit skips
- `minimums-3.14`: 601 passed
- repository lint and `git diff --check` passed

Alternative linkers, identical code folding, and persistent compiler caching remain separate measured experiments. Linux CI already uses mold. The portable MLIR toolchain can support later shared-workflow experiments with its bundled mold on Linux and LLD on Windows.

AI assistance implemented and validated the scoped build and CI changes. A human remains responsible for reviewing and merging the contribution.


cmake/ExternalDependencies.cmake
```diff
@@ -32,12 +32,10 @@ FetchContent_Declare(
 # leak into this third-party dependency.
 function(_mqt_core_make_jeff_available)
   set(BUILD_TESTING OFF)
+  # Jeff's transitive Cap'n Proto dependency contains source files that cannot share a unity
```
denialhaag:
```suggestion
  # jeff's transitive Cap'n Proto dependency contains source files that cannot share a unity
```

denialhaag: Maybe something we can add in AGENTS.md. 🤔

noxfile.py
```diff
@@ -159,16 +159,10 @@ def docs(session: nox.Session) -> None:
...
+        # The docs exercise only the DDSIM provider. The common Python package
+        # configuration enables unity builds.
         "SKBUILD_CMAKE_ARGS": (
-            "-DBUILD_MQT_CORE_DOCUMENTATION=ON;"
-            "-DBUILD_MQT_CORE_QDMI_NA_DEVICE=OFF;"
-            "-DBUILD_MQT_CORE_QDMI_SC_DEVICE=OFF;"
-            "-DCMAKE_UNITY_BUILD=ON;"
-            "-DCMAKE_VERIFY_INTERFACE_HEADER_SETS=OFF;"
-            "-DENABLE_IPO=OFF"
+            "-DBUILD_MQT_CORE_DOCUMENTATION=ON;-DBUILD_MQT_CORE_QDMI_NA_DEVICE=OFF;-DBUILD_MQT_CORE_QDMI_SC_DEVICE=OFF"
```
denialhaag, line 165: Maybe it's nicer to have these on separate lines (unless `ruff` formats this like this).

noxfile.py
```diff
@@ -210,13 +204,7 @@ def stubs(session: nox.Session) -> None:
...
         # documentation build, which lets both sessions share a build tree.
         "SKBUILD_CMAKE_BUILD_TYPE": "MinSizeRel",
-        "SKBUILD_CMAKE_ARGS": (
-            "-DBUILD_MQT_CORE_QDMI_NA_DEVICE=OFF;"
-            "-DBUILD_MQT_CORE_QDMI_SC_DEVICE=OFF;"
-            "-DCMAKE_UNITY_BUILD=ON;"
-            "-DCMAKE_VERIFY_INTERFACE_HEADER_SETS=OFF;"
-            "-DENABLE_IPO=OFF"
-        ),
+        "SKBUILD_CMAKE_ARGS": ("-DBUILD_MQT_CORE_QDMI_NA_DEVICE=OFF;-DBUILD_MQT_CORE_QDMI_SC_DEVICE=OFF"),
```
denialhaag, line 207: Same comment here.

denialhaag (approved):
This LGTM! I just have three very minor comments. 😌

denialhaag:
Do we also delete the extensive labels then? 🤔

cmake/ExternalDependencies.cmake
burgholzer: Yeah, feels like this belongs in the AGENTS guidelines.

burgholzer:
> Do we also delete the extensive labels then? 🤔

We could. At least here. I see no real downside.

noxfile.py
burgholzer: This is ruff's doing, unfortunately. It reformats as soon as everything fits in one line.

noxfile.py
burgholzer: Same as above 😌

burgholzer:
@denialhaag In the interest of speeding up our CI here as soon as possible, I'll leave this one to you for merging. If you feel like everything is addressed, please resolve the conversations and hit auto-merge 😌

denialhaag:
> @denialhaag In the interest of speeding up our CI here as soon as possible, I'll leave this one to you for merging. If you feel like everything is addressed, please resolve the conversations and hit auto-merge 😌

Let's get this in! :shipit:

PR #2105, opened 2026-08-14
🐛 Fix conversions of nested control flow between `jeff` and QCO

denialhaag:
## Description

🤖 *AI text below* 🤖

This fixes two bugs that prevented nested control flow from surviving conversion between QCO and `jeff`.

An `scf.for` loop nested directly inside another `scf.for` loop failed to convert to `jeff`, and the outer loop was reported as not legalizable. `moveRegion` clones the operations of the source region, so a nested control flow operation reaches the conversion driver as a new operation of the same kind as the one just matched. The driver then rejects it with "pattern was already applied", and legalization of the outer operation fails with it. The patterns that move a region into a `jeff` operation now declare bounded rewrite recursion. The recursion terminates because each application moves one nesting level of the original program into its `jeff` counterpart. Directly nested `scf.while` loops were broken by the same mechanism and are fixed as well.

In the other direction, converting a `jeff.switch` nested in a `jeff.while` back to QCO crashed on an out-of-bounds access. `ConvertJeffSwitchOpToQCO` decided which in-values are qubits from the types of the operands, but by that point the operands may already carry converted types, which `isLinearType` does not recognize. It then read the results of the new `qco.if` using a count derived from the original `jeff` result types, and indexed past the end. Both loops now consult the result types, which still carry `jeff` types and correspond to the in-values positionally. `ConvertJeffWhileOpToQCO` had the same defect and is fixed as well.

This adds a `NestedForLoopForOp` round-trip test case, which covers the only nesting the `SCFForOpTest` suite was missing, and a `NestedWhileOpIfOp` case for the second fix.

## AI notice

This PR and its contents were created with the assistance of Opus 5 via Claude Code.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,7 +92,7 @@ releases may include breaking changes.
   correction ([#1407], [#1674], [#2002], [#2038]) ([**@J4MMlE**],
   [**@denialhaag**], [**@MatthiasReumann**], [**@simon1hofmann**])
 - ✨ Add conversions between `jeff` and QCO ([#1479], [#1548], [#1565], [#1637],
-  [#1676], [#1706], [#1776], [#1836], [#1934], [#2000], [#2018])
+  [#1676], [#1706], [#1776], [#1836], [#1934], [#2000], [#2018], [#2105])
   ([**@denialhaag**], [**@burgholzer**])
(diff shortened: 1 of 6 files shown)


burgholzer (approved):
Looks nice and clean 👍🏼

Issue #2085, opened 2026-08-13
🧹 MQT Core v4 cleanup and ownership tracking


burgholzer:
🤖 *AI text below* 🤖

## Goal

Use the MQT Core v4 boundary to reduce legacy surface, remove unused APIs, and move specialized functionality to the repositories that actually own it. The aim is a smaller and clearer Core package with fewer public targets, transitive dependencies, compatibility layers, and domain-specific concepts.

The sub-issues are grouped into four tracks:

1. straightforward removals and API pruning;
2. dependency and packaging simplification;
3. ownership transfers to QMAP, QCEC, DDSIM, and QDMI-on-IQM; and
4. bounded architectural cleanup of CoreIR, DD, QDMI, QIR, and OpenQASM support.

Each sub-issue should remain independently reviewable. Cross-repository transfers should prepare the downstream owner before removing the corresponding Core API.

## Scope principles

- Backward compatibility is not a reason to retain deprecated or unused machinery in v4.
- Preserve useful user-facing behavior while simplifying its implementation and ownership.
- Do not replace one maintenance burden with a generic abstraction that has no demonstrated consumer.
- Record removed targets, headers, namespaces, Python APIs, options, and migration paths in `CHANGELOG.md` and `UPGRADING.md`.
- Validate installed shared and static CMake packages, Python wheels, downstream source builds, documentation, and dependency inspection where applicable.

## Explicit non-goals

- The compiler collection remains in MQT Core.
- The generic QDMI layer remains in MQT Core for now.
- The Qiskit and PennyLane adapters are not being assigned a new owner before v4.
- The DDSIM and SC QDMI devices remain supported.
- Jeff remains supported.
- This tracker does not pursue CMake templating merely to remove repeated declarations.

## Audit basis

The initial inventory was performed against MQT Core `main` at `2953866e3f0b424c434cd4c4f882795e3a18e85e`. Individual issues should refresh usage and downstream evidence before implementation.

Issue #1348, opened 2025-12-03
✨ Support IQM JSON, QIR, and QPY in the QDMI Qiskit backend


marcelwa:
### Problem Statement

The QDMI Qiskit backend introduced in #1243 currently supports QASM2 and QASM3 for job submission exclusively. 

### Proposed Solution

We should also support IQM JSON, QIR, and QPY, as they're supported by QDMI 1.2.

burgholzer:
I am closing this one as stale for now. We have recently introduced serializers, which can also be vendor provided. QDMI v1.4 will also bring a couple of related changes that go beyond the scope identified in this issue.

PR #2114, opened 2026-08-14
✨ Add program serializers to the Qiskit backend and hand IQM JSON to QDMI-on-IQM

### LINKED ISSUE #2094: 🚚 Move IQM conversion support to QDMI-on-IQM
🤖 *AI text below* 🤖

## Goal

Move the provider-specific IQM JSON converter and related IQM operation handling from generic MQT Core code to QDMI-on-IQM.

## Scope

- Move `qiskit_to_iqm_json` and its tests to QDMI-on-IQM.
- Move or register provider-specific `MoveGate` conversion behavior without making the generic Qiskit adapter depend on IQM.
- Introduce the smallest codec/program-format registration seam needed for a provider to advertise and implement IQM JSON.
- Reconcile this ownership change with the behavior currently discussed in #1348.
- Do not decide the ownership of the generic Qiskit adapter in this issue.

## Acceptance criteria

- Core contains no IQM-specific JSON conversion logic.
- QDMI-on-IQM owns and tests IQM program conversion.
- Generic QDMI and Qiskit code selects provider codecs without hard-coded IQM branches.
- Existing IQM execution behavior remains covered by cross-repository tests.

marcelwa:
🤖 *AI text below* 🤖

## Description

This adds a registration seam that lets any package supply the conversion from a Qiskit circuit into a program format, and moves MQT Core's own vendor-specific conversion out through it.

The Qiskit backend used to decide in its own code how to turn a circuit into a program, with one branch per format. It now takes every format from a registered **program serializer**, and MQT Core registers its OpenQASM 2 and OpenQASM 3 serializers the same way as everyone else. `_serialize_circuit` is one ordered walk with no format-specific branch left.

That makes the vendor carve-out possible. MQT Core carried `qiskit_to_iqm_json` and an IQM `MoveGate` inside its vendor-neutral adapter; nothing in Core used either, and the sole consumer is the external `iqm-qdmi` package. Removing them alone would have broken IQM execution, because the IQM QDMI device advertises only `QIR_BASE_STRING` and `IQM_JSON`, with no OpenQASM to fall back to. Registered through the seam, IQM JSON keeps working from outside Core.

### The seam

A new `mqt.core.plugins.qiskit.serializers` module holds a registry keyed by `ProgramFormat`. A **program serializer** turns one circuit into one program in one format.

A format fixes the kind of payload it carries, so there are two signatures:

```python
def serialize(circuit: QuantumCircuit, backend: QDMIBackend) -> str: ...   # text format
def serialize(circuit: QuantumCircuit, backend: QDMIBackend) -> bytes: ...  # binary format
```

The serializer takes the backend, so it can reach the device through the new `QDMIBackend.device` property and the supported operations through `QDMIBackend.target`. The API is `register_program_serializer(fmt, fn, *, replace=False)`, `unregister_program_serializer(fmt)`, `program_serializer(fmt)`, and `preferred_program_formats(formats)`.

A package advertises a serializer through an entry point group:

```toml
[project.entry-points."mqt.core.qiskit.program_serializers"]
IQM_JSON = "iqm.qdmi.serializers:qiskit_to_iqm_json"
```

Entry points rather than only a runtime call, because `QDMIProvider` builds a plain `QDMIBackend` for every registered device: a user can reach a registered IQM device without ever importing `iqm.qdmi`, and an import-time registration would miss that path. An entry point naming an unknown format, naming a format with no program payload, or failing to import warns and is skipped, so one broken package cannot hide the others.

Core registers its own OpenQASM 2 and OpenQASM 3 serializers through the same registry, so `_serialize_circuit` is one ordered walk with no format-specific branch, and a provider can replace the built-ins with `replace=True`.

### Which format wins

`PROGRAM_FORMAT_PREFERENCE` states the order explicitly, most preferred first:

```text
IQM_JSON, CUSTOM1 ... CUSTOM5,
QIR_ADAPTIVE_MODULE, QIR_ADAPTIVE_STRING,
QPY, QASM3,
QIR_BASE_MODULE, QIR_BASE_STRING,
QASM2
```

A device-native format comes first, because a package that registers a serializer for its own device's format wants that format used, and that is the precedence IQM JSON had. The standardized formats follow in order of what a circuit may contain: the QIR adaptive profile allows classical control, QPY carries a Qiskit circuit without loss, and OpenQASM 3 expresses control flow, while the QIR base profile forbids classical feedback and OpenQASM 2 has no control flow at all. Encoding only breaks a tie within one profile. `CALIBRATION` and `BATCH_JOB` are absent because a serialized circuit is not what they carry. The list decides, not the order the device reports — `test_backend_respects_format_preference` covers that.

### Text and binary payloads

Core's C++ client already knew which formats are binary and which carry no program at all: `src/qdmi/Client.cpp` had `isBinaryProgramFormat` and `hasNoGenericProgramPayload` in an anonymous namespace, and the text `submitJob` overload rejects a binary format. Rather than restate that in Python, both move to `include/mqt-core/qdmi/Client.hpp` and are bound as `mqt.core.qdmi.is_binary_program_format` and `mqt.core.qdmi.has_program_payload`. The backend checks a serializer's return type against its format and raises `TranslationError` on a mismatch. `Device.submit_job` already overloads on `str | bytes`, so a binary payload reaches the device unchanged.

### The gate

`MoveGate` moves to `iqm.qdmi.gates`. `QDMIBackend` gains an empty `_EXTRA_GATES` class variable and an `__init_subclass__` that rebuilds the gate maps per subclass, so a backend for a device with a native gate outside Qiskit's standard library supplies it without global state. `_map_operation_to_gate` and `_map_qiskit_gate_to_operation_names` become classmethods so a subclass reads its own maps. Every Qiskit provider with a non-standard native gate does the same: `GPIGate` and friends in qiskit-ionq, `HamiltonianGate` in qiskit-pasqal-provider, `MoveGate` in qiskit-on-iqm.

### Deliberately left alone

- The `"r": {"prx"}` alias stays in `_GATE_ALIASES`. `prx` is IQM terminology, but Core's own SC device model uses it and #2085 keeps that device supported.
- `json/sc/iqm-garnet.json` and `json/sc/iqm-emerald.json` stay: device models, not conversion logic.
- `TranslationError` keeps its name. It belongs to a hierarchy shared with the PennyLane plugin (`PennyLaneTranslationError` and friends), so renaming it is a separate cross-plugin change.

### Related

The downstream half is iqm-finland/QDMI-on-IQM#189, which adds the serializer, the `MoveGate`, and the entry point. It pins this branch and stays open until MQT Core v4 ships. It must follow this PR's renames: `iqm.qdmi.serializers`, the `mqt.core.qiskit.program_serializers` group, and the `(circuit, backend)` signature.

This also reframes #1348: QIR and QPY support becomes a serializer registered by whoever owns the format rather than another branch in the backend.

Design notes, including the Qiskit ecosystem cross-check and the decision log, are in `.agent/plans/2094-iqm-conversion-carveout.md`.

Fixes #2094

## Validation

- `uv run --no-sync pytest test/python` — 619 passed, 4 skipped
- `uv run --no-sync pytest test/python/plugins/qiskit` and `test/python/qdmi` — 356 passed
- `./build/release/test/qdmi/mqt-core-qdmi-test` — 274 passed, no compiler warnings
- `uvx nox -s stubs` — the stub diff contains only the two new functions
- `uvx nox --non-interactive -s docs` — passes with `-W -n`
- `uvx nox -s lint` — passes, full `prek` hook set

Built against LLVM/MLIR 22.1.0. Not verified locally: the cross-repository check with `iqm-qdmi` installed alongside, which needs QDMI-on-IQM#189.

## Checklist

AI assistance: written by Claude Opus 5 via Claude Code, under my direction. The last checkbox is left for me to tick after my own review.


.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+[#2085](https://github.com/munich-quantum-toolkit/core/issues/2085).
+
+## Terms
+
+- **QDMI** — the Quantum Device Management Interface, the C API through which
+  Core talks to a device. `mqt.core.qdmi` is its Python binding.
+- **Program format** — the syntax a device accepts a program in, named by a
+  member of the `ProgramFormat` enum (`QASM2`, `QASM3`, `IQM_JSON`, `QPY`, …).
+- **Program codec** — a function that converts one Qiskit circuit into one
+  program format. Signature: `(circuit, device) -> str`.
```
burgholzer: Two aspects here:
- The word "codec" is not really established anywhere in quantum and I haven't seen that being used anywhere. I am also not sure the literal meaning of the word matches that well to what it is supposed to represent. Terms used in the community include "convert/conversion" or "translate/translation", which would both be a better fit overall in my opinion.
- We need at least two signatures for these conversions, one for string based formats (QASM, QIR string, IQM JSON) and one for binary based formats (QIR binary, QPY). The latter likely returns `bytes`. This may require a bit of a redesign for how conversions are registered in the QDMI qiskit backend.

.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+- `backend.py::_build_gate_mappings_for_backend` hard-coded `"move": MoveGate()`
+  into the map from device operation names to Qiskit gates.
+- `__init__.py` exported both symbols publicly.
+- `test/python/plugins/qiskit/test_mock_backend.py` held twelve
+  `test_qiskit_to_iqm_json_*` tests and three tests about `move`.
+
+The IQM device advertises only `QIR_BASE_STRING` and `IQM_JSON`, never any
+OpenQASM. So deleting the branch without a replacement would leave every IQM
+circuit with nothing to convert to. The seam is not optional polish; it is what
+keeps IQM execution working.
```
burgholzer: Besides the fact that I keep rolling my eyes on LLM-sentences like the last one here (which I suppose will go away over time), I'd also like to include the following in the general reasoning:
Mid-term (likely in v4+ of MQT Core), we will likely make use of `mqt-cc` to compile from Qiskit Circuits to 
- any version of QIR advertised by a backend (binary over string representations, adaptive over base profile)
- OpenQASM 3 (I believe our exporter may be more capable than Qiskit's QASM exporter at the moment; we cannot and probably won't generally emit OpenQASM 2; except if we also want to make a distinction similar to the QIR base and adaptive profiles)
- QPY (likely possible through the qiskit C-API at some point)

This may change some of the design decisions here.

.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+- `program_codec(fmt)` returns the codec or `None`.
+
+The first of those calls loads the entry point group
+`mqt.core.qiskit.program_codecs` once and caches the result. The entry point
+name is the `ProgramFormat` member name; the value points at the codec:
+
+```toml
+[project.entry-points."mqt.core.qiskit.program_codecs"]
+IQM_JSON = "iqm.qdmi.converters:qiskit_to_iqm_json"
+```
```
burgholzer: Should MQT Core register its QASM conversion as a "codec" so that the backend has a uniform way of handling conversions? Would potentially streamline things quite a bit.

.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+discovered by Core, not the other way round: a user can enumerate devices with
+`QDMIProvider` and get a plain `QDMIBackend` over an IQM device without ever
+importing `iqm.qdmi`. An import-time registration would miss that path.
+
+An entry point naming an unknown format, or one that fails to import, produces a
+warning and is skipped. One broken package must not make every other codec
+unreachable.
+
+`_convert_circuit` walks the device's supported formats in the order the device
+reports them and uses the first one with a registered codec. Only then does it
```
burgholzer: I believe we should encode the priority/order that the program formats are tried and not leave that up to some `map` ordering. See the above comment for a potential starting point.
I am not fully sure yet, whether we want
- "explicit registrations (for custom formats) take precedence over standard formats", or
- "always try standard(ized) formats first.

We definitely want binary over strings, and more advanced capabilities over simpler programs.

.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+### The gate
+
+`MoveGate` moves to `iqm.qdmi.gates`. Core's `_build_gate_mappings_for_backend`
+takes a second argument, `extra_gates`, and `QDMIBackend` gains a `_EXTRA_GATES`
+class variable, empty by default. `__init_subclass__` rebuilds
+`_QISKIT_TO_QDMI_GATE_MAP` and `_OPERATION_TO_GATE_MAP` for each subclass from
+its own `_GATE_ALIASES` and `_EXTRA_GATES`, so a subclass adds a device-native
+gate without touching global state. `_map_operation_to_gate` and
+`_map_qiskit_gate_to_operation_names` become classmethods so a subclass reads
+its own maps rather than the base class's.
```
burgholzer: Just to cross-check: Is this close to what Qiskit does for other external backends? For example
- https://github.com/qiskit-community/qiskit-ionq
- https://github.com/qiskit-community/qiskit-aqt-provider
- https://github.com/qiskit-community/qiskit-pasqal-provider
- https://github.com/qiskit-community/qiskit-quantinuum-provider

There is quite a bit of a precedent for how Qiskit has been extended in the past for various providers. It may make sense to analyze these integrations and extract potential commonalities that apparently worked for the community.

Maybe also worth checking Qiskit documentation pages like https://qiskit.qotlabs.org/docs/guides/custom-backend

.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Deliberately not in scope
+
+- The alias `"r": {"prx"}` stays in Core's `_GATE_ALIASES`. `prx` is IQM
+  terminology, but Core's own superconducting device model uses it, and #2085
+  keeps that device supported.
+- `json/sc/iqm-garnet.json` and `json/sc/iqm-emerald.json` stay. They are device
+  models for the generic SC device, not conversion logic.
+
+Both are worth revisiting; neither belongs in this change.
```
burgholzer: I agree that both of these are here to stay (long term).

.agent/plans/2094-iqm-conversion-carveout.md
```diff
@@ -0,0 +1,181 @@
...
+separate pull request in a separate repository.
+
+## How to see it working
+
+Build and install the package, then run the adapter tests:
+
+```console
+./.agent/run.sh uv sync --inexact --only-group build --only-group test
+MLIR_DIR=<llvm>/lib/cmake/mlir ./.agent/run.sh uv sync --inexact --no-build-isolation-package mqt-core
+./.agent/run.sh uv run --no-sync pytest test/python/plugins/qiskit -q
```
burgholzer: This uses "old" (😅) syntax that is no longer present in AGENTS.md
May also be worth to update.

burgholzer (changes requested):
Hey @marcelwa 👋🏼 
Thanks a lot! I generally did not think too much about the replacement of the custom converters in MQT Core when the initial issue was created and thought it could simply be removed. But I now fully see and agree that it needs some kind of conversion registration that external packages can provide.

I purposely only reviewed the ExecPlan here and added requests for changes to it because I think this most effectively steers the implementation in the right direction.
I hope the comments themselves make sense.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Both addressed, and the term is now **serializer**.

On the word: I went with serialization rather than conversion or translation because Qiskit already uses it for exactly this operation — `qasm3.dumps` is documented as "Serialize a `QuantumCircuit` object in an OpenQASM 3 string", the class behind it is `qasm3.exporter.Exporter`, and `qpy.dump` writes the binary form. IQM uses the same word (`IQMBackend.serialize_circuit`, `serialize_instructions`). I deliberately avoided "translation" because `qiskit.transpiler.translation` and `BasisTranslator` mean basis translation, which this same adapter relies on, so the word would carry two meanings a few lines apart. I avoided "conversion" because MQT Core already uses it for in-memory conversion between its IR and Qiskit (`qiskit_to_mqt`, `mqt_to_qiskit`).

On the two signatures: `TextProgramSerializer` returns `str`, `BinaryProgramSerializer` returns `bytes`. The redesign turned out smaller than expected, because the program format already determines the payload kind — and MQT Core already knew that. `src/qdmi/Client.cpp` had `isBinaryProgramFormat` (QIR modules, QPY) and `hasNoGenericProgramPayload` (calibration, batch jobs) in an anonymous namespace, and the text `submitJob` overload already rejects a binary format. Rather than restate that in Python, I promoted both to `include/mqt-core/qdmi/Client.hpp` and bound them as `mqt.core.qdmi.is_binary_program_format` and `has_program_payload`. So one registry keyed by format still suffices; the backend checks the returned type against the format and raises on a mismatch, and `Device.submit_job` already overloads on `str | bytes`, so the bytes path needed no further change. Registering a serializer for `CALIBRATION` or `BATCH_JOB` is a `ValueError`.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Fair on both counts. That sentence and its relatives are gone.

The plan now has a "Where this is going" section that records the `mqt-cc` outlook: any version of QIR a backend advertises, preferring a binary module over a string and adaptive over base; OpenQASM 3 through MQT Core's own exporter rather than Qiskit's; QPY through the Qiskit C API once that exists; and OpenQASM 2 not generally emitted.

It did change a design decision. It is the main reason the format preference is now one explicit, ordered tuple in one module rather than control flow: when `mqt-cc` can emit QIR for a device that also has a vendor format, choosing between them is a one-line edit to that tuple. And those future serializers will replace MQT Core's registered built-ins through the same registry instead of adding branches to the backend.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Yes, and it does streamline things. MQT Core now registers `_serialize_to_qasm3` and `_serialize_to_qasm2` through the same registry at import of `backend.py`, so `_serialize_circuit` is one ordered walk with no format-specific branch left:

    for fmt in preferred_program_formats(formats):
        serializer = program_serializer(fmt)
        if serializer is None:
            continue
        ...

A side effect worth naming: the built-ins are now replaceable like anything else, so a provider whose device needs a different OpenQASM 3 export can pass `replace=True` instead of subclassing.

This is what settled the signature question. The OpenQASM 3 serializer needs `backend.target.operation_names` to decide which `stdlib.inc` definitions to suppress, so a `(circuit, device)` signature would have forced the built-ins to stay hard-coded. A serializer now takes `(circuit, backend)`, which is also what qiskit-ionq does (`qiskit_to_ionq(circuit, backend, ...)`), and `QDMIBackend` gained a public `device` property.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Encoded. `mqt.core.plugins.qiskit.serializers.PROGRAM_FORMAT_PREFERENCE` is one tuple, most preferred first:

    IQM_JSON, CUSTOM1 ... CUSTOM5,
    QIR_ADAPTIVE_MODULE, QIR_ADAPTIVE_STRING,
    QIR_BASE_MODULE, QIR_BASE_STRING,
    QPY, QASM3, QASM2

`preferred_program_formats()` applies it to whatever a device reports, and `CALIBRATION` and `BATCH_JOB` are absent because they carry no program payload. `test_backend_respects_format_preference` has the device report QASM2 first and asserts QASM3 is submitted, so the list decides and not the device.

On your open question I picked **explicit registrations take precedence**, on the grounds that a package registers a serializer for its own device's native format because it wants that format used, and that is the precedence IQM JSON has today. Reversing it is a one-line reorder of that tuple, and the reasoning is in the Decision Log, so say the word if you would rather draw it the other way — I think it becomes a live question once `mqt-cc` emits QIR.

Within the standardized formats I made capability the primary key and encoding the tiebreaker (adaptive module, adaptive string, base module, base string), because the profile decides what a circuit may contain while the encoding only decides how the program travels. Your comment lists binary first, so if you meant binary as the primary key that is also a one-line change.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Good call — I checked all four, plus qiskit-ibm-runtime, qiskit-on-iqm, Qiskit itself, and qBraid. Findings are recorded in the plan under "What the Qiskit ecosystem does". The four that mattered:

**There is no Qiskit entry point group for circuit-to-payload conversion.** Qiskit defines exactly eight groups and all are transpiler stages (`qiskit.unitary_synthesis`, `qiskit.synthesis`, and the six `qiskit.transpiler.*`). `qiskit.qasm3` and `qiskit.qpy` are closed modules with no plugin lookup. So there is no existing interface to conform to here.

**Providers do not need a seam because they own both halves.** `qiskit_ionq/helpers.py::qiskit_to_ionq(circuit, backend, ...)`, `qiskit_aqt_provider/circuit_to_aqt.py::qiskit_to_aqt_circuit(circuit)`, and `qiskit_pasqal_provider`'s `gen_seq(register, device, circuit)` are all plain functions called from their own `run`. That works when the package owns the backend and the format. MQT Core owns the backend but not the vendor formats, which is the asymmetry that makes registration necessary. qiskit-quantinuum-provider is archived, BackendV1-era, and just calls `circuit.qasm()`.

**The Qiskit-native pattern for this shape is entry point plus backend hook.** A backend returns a plugin *name* from `get_translation_stage_plugin()`, resolved through `qiskit.transpiler.translation` — AQT, IQM, and qiskit-ibm-runtime all do it. It confirms entry points as the discovery channel, and it is a second reason not to reuse "translation" for anything else in this adapter.

**The closest structural precedent is outside Qiskit: qBraid.** It registers program formats under a `qbraid.programs` entry point group as `ProgramSpec(program_type, alias, serialize=..., validate=...)` objects, and a device holds a list of them with `set_target_program_type()` to override the target. Entry point group, callable per format, explicit selection policy — the same three pieces as here, which I take as reassurance the shape is right.

On multiple formats per backend, no provider negotiates the format with the device at run time. IonQ picks between its flat JSON and OpenQASM 3 with a predicate on the circuit (`circuit_requires_qasm3`) and tags the payload `ionq.qasm3.v1` / `ionq.circuit.v1`; qBraid uses an explicit list; qiskit-ibm-runtime negotiates only a QPY *version*, `min(SERVICE_MAX_SUPPORTED_QPY_VERSION, QISKIT_QPY_VERSION)`. An explicit ordered list is the conventional answer, which is what the preference tuple now is.

On the gate seam, every provider with a non-standard native gate defines a `Gate` subclass in its own package and injects it into its `Target`: `GPIGate`/`GPI2Gate`/`MSGate`/`ZZGate` in `qiskit_ionq/ionq_gates.py`, `HamiltonianGate` in qiskit-pasqal-provider, `MoveGate` in qiskit-on-iqm. That is what `_EXTRA_GATES` lets a `QDMIBackend` subclass do, so I left that part as it was.

The IBM custom-backend guide says nothing about payload conversion — it covers only `target`, `max_circuits`, `_default_options`, and `run`.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Agreed, both stay. The plan's out-of-scope section keeps them and no longer hedges about revisiting.

.agent/plans/2094-iqm-conversion-carveout.md
marcelwa: 🤖 *AI text below* 🤖

Updated. The plan now uses the current commands (`uv sync --locked --only-group dev`, `uv sync --inexact --no-dev --no-build-isolation-package mqt-core`, `uv run --no-sync pytest`, `uvx nox -s stubs`, `uvx nox -s lint`) with no `.agent/run.sh`.

While there I noticed the file did not follow `.agent/PLANS.md` at all — `Progress`, `Surprises & Discoveries`, `Decision Log`, and `Outcomes & Retrospective` are mandatory and all four were missing — so I restructured it to the skeleton and filled them in. It also no longer nests code fences, per the formatting rule.

bindings/qdmi/qdmi.cpp
```diff
@@ -223,6 +223,35 @@ when the custom slot is unsupported.)pb");
...
+Returns:
+    True if the format requires exact-byte submission.)pb");
+
+  qdmiModule.def(
+      "has_program_payload", &qdmi::hasProgramPayload, "program_format"_a,
+      R"pb(Returns whether a program format carries a program payload.
+
+``CALIBRATION`` and ``BATCH_JOB`` name a device-side routine instead of holding
+a program, so :meth:`Device.submit_job` rejects them. Every other format
+carries a program.
```
burgholzer: I am with you on this for the batch jobs, but I am not so sure about the Calibration side. At least in case of IQM, a calibration job also carries a payload (string-based in case of IQM).
I would generally expect that the payload of a calibration job is some form of configuration for the job.
Maybe this changes a few aspects in this PR.

docs/qdmi/qdmi_backend.md
```diff
@@ -427,19 +427,86 @@ except UnsupportedFormatError as e:
...
+A device usually accepts several formats. The backend walks them in the order of
+{py:data}`~mqt.core.plugins.qiskit.serializers.PROGRAM_FORMAT_PREFERENCE` and
+uses the first one that has a serializer, so the order of the list decides and
+not the order the device reports:
+
+```text
+IQM_JSON, CUSTOM1 ... CUSTOM5,
+QIR_ADAPTIVE_MODULE, QIR_ADAPTIVE_STRING,
+QIR_BASE_MODULE, QIR_BASE_STRING,
+QPY, QASM3, QASM2
```
burgholzer: I would put `QPY` and `QASM3` over `QIR_BASE_*` because they are more capable.

python/mqt/core/plugins/qiskit/backend.py
```diff
@@ -110,6 +112,102 @@ def _build_gate_mappings_for_backend(
...
+
+    Args:
+        circuit: The circuit to serialize.
+        backend: The backend that runs the circuit. Qiskit's OpenQASM 2 exporter
+            takes no information from it.
+
+    Returns:
+        The OpenQASM 2 program.
+    """
+    return qasm2.dumps(circuit)
```
burgholzer, line 189: Is there a convincing reason that qasm3 does way more handling than qasm2 here?
Likely because it was that way before, right?
Fine with me to keep it that way, but may be worth to explore whether there is room for improvement.

python/mqt/core/plugins/qiskit/serializers.py
```diff
@@ -0,0 +1,269 @@
...
+def _load_entry_points() -> None:
+    """Load the serializers advertised through :data:`ENTRY_POINT_GROUP` once.
+
+    An entry point that names an unknown program format, names a format without
+    a program payload, or fails to load produces a warning and is skipped, so
+    one broken package cannot make every other serializer unreachable.
+    """
+    global _ENTRY_POINTS_LOADED  # ruff:ignore[global-statement] Guards a one-time import side effect
+    if _ENTRY_POINTS_LOADED:
+        return
```
burgholzer: Does this really need to be `global`? Is there no other convenient way to handle this?

CHANGELOG.md
burgholzer, line 1: The changelog entries added here are extremely wordy compared to the other changelog entries. I believe it would be good to align the style. Most changelog entries are a single sentence. The longer, more elaborate description can go into the upgrade guide.

UPGRADING.md
burgholzer, line 1: I believe it may be worth to split this into separate entries, one for the removal of IQM and the other one for the addition of the serializer concept.

burgholzer (changes requested):
Thanks @marcelwa 🙏🏼 This is looking great already!
I have a couple of requests for changes inline; some of them are related to pre-existing code that may not be fully ideal.
More generally, I believe this PR's title and description should be updated to better match what the PR adds.

docs/qdmi/qdmi_backend.md
marcelwa: Makes sense. Will gladly reorder.

python/mqt/core/plugins/qiskit/backend.py
marcelwa: Exactly, it was already like that before. Probably because Qiskit's `qasm3.dumps()` and `qasm2.dumps()` have differing APIs. I could add a comment to explain the asymmetry here and file a follow-up issue to further streamline things. TBD how that would look.

CHANGELOG.md
marcelwa: Fair point. Will adjust those.

UPGRADING.md
marcelwa: Also fully on board.

python/mqt/core/plugins/qiskit/backend.py
burgholzer: Let's keep it simple here and just preserve the status quo. OpenQASM2 shouldn't be too important for now.

python/mqt/core/plugins/qiskit/serializers.py
marcelwa: I thought the `@cache` decorator on `_load_entry_points()` might solve the problem here, but it doesn't, because the cache is only populated after the function terminates. 

The alternative to `global` would be some kind of local object that holds the state. It would also improve testability by getting rid of the current monkey-patching. But honestly, that would just be the `global` variable with a fancy hat on. 

If you prefer that, I can pitch a proposal.

bindings/qdmi/qdmi.cpp
marcelwa: Okay, this one took me a bit of back-and-forth with Claude and a dive into the QDMI spec to grasp what is actually happening here. 

`QDMI_DEVICE_JOB_PARAMETER_PROGRAM` says "This parameter is required". `CALIBRATION` is then the explicit exception to that, stating `@brief void* A calibration program.` and the details say triggering a calibration run "does not require a program to be set". It seems to me that the payload is optional there, which makes sense because QDMI-on-IQM actually does send one.

However, MQT Core currently unconditionally rejects a format that doesn't carry a payload in `submitJob`:
```c++
if (!hasProgramPayload(format)) {
  throw std::invalid_argument(
      "Calibration and batch jobs do not use a generic program payload");
}
```
That doesn't seem correct to me. Do we want to fix that here or in a separate PR?

Also, there is something going on with `BATCH_JOB` that I haven't fully understood yet. According to the QDMI spec, it requires a "`QDMI_Job` list" as `QDMI_Job*` or `QDMI_Device_Job*`. But MQT Core's `submitJob` accepts `std::span<const std::byte>`. I don't know of a meaningful way to convert one into the other. 

Could you help shed some light on this situation?

.agent/plans/2094-iqm-conversion-carveout.md
burgholzer: "Serializer" makes a lot more sense 👍🏼

python/mqt/core/plugins/qiskit/serializers.py
burgholzer: Had a bit of a discussion with gpt-5.6-sol on this. Below is the summary it spit out after a bit of iteration.

🤖 *AI text below* 🤖

I took a closer look at this. The `@cache` point is correct: it stores the result only after the call returns, so it cannot guard against re-entry while `EntryPoint.load()` imports third-party code. Some process-wide state is therefore necessary.

I still think a private registry object would be more than a cosmetic wrapper if it owns both the state and the behavior. It could:

- represent `NOT_STARTED`, `LOADING`, and `LOADED` explicitly, instead of using a flag that means both “loading” and “loaded”;
- collect discovered serializers separately and publish them only after discovery completes, while `setdefault` preserves the precedence of explicit registrations;
- reset to `NOT_STARTED` if discovery aborts, rather than leaving the registry permanently marked as loaded;
- accept the entry-point discovery function as a constructor dependency, so tests can create isolated registries instead of monkey-patching `_SERIALIZERS`, `_ENTRY_POINTS_LOADED`, and `entry_points`.

The existing public module functions can remain unchanged and delegate to one private singleton. A state holder introduced only to avoid the `global` keyword would indeed be “the global with a fancy hat”; a registry that owns these invariants and provides the test seam would be a useful abstraction.

I would prefer that direction. I would also add regression tests for re-entrant discovery and retry after an interrupted discovery.

.agent/plans/2094-iqm-conversion-carveout.md
burgholzer: Thanks for the cross-check! 👍🏼

python/mqt/core/plugins/qiskit/serializers.py
burgholzer: Here's what it proposed code wise
```python
class _LoadState(Enum):
    NOT_STARTED = auto()
    LOADING = auto()
    LOADED = auto()


class _ProgramSerializerRegistry:
    def __init__(
        self,
        discover: Callable[[], Iterable[EntryPoint]],
    ) -> None:
        self._discover = discover
        self._serializers: dict[ProgramFormat, ProgramSerializer] = {}
        self._load_state = _LoadState.NOT_STARTED

    def get(self, fmt: ProgramFormat) -> ProgramSerializer | None:
        self._load_entry_points()
        return self._serializers.get(fmt)

    def _load_entry_points(self) -> None:
        if self._load_state is not _LoadState.NOT_STARTED:
            return

        self._load_state = _LoadState.LOADING
        discovered: dict[ProgramFormat, ProgramSerializer] = {}

        try:
            for entry_point in self._discover():
                loaded = self._load_entry_point(entry_point)
                if loaded is not None:
                    fmt, serializer = loaded
                    discovered.setdefault(fmt, serializer)
        except BaseException:
            self._load_state = _LoadState.NOT_STARTED
            raise

        # Explicit registrations made before or during discovery still win.
        for fmt, serializer in discovered.items():
            self._serializers.setdefault(fmt, serializer)

        self._load_state = _LoadState.LOADED
```

bindings/qdmi/qdmi.cpp
burgholzer: Let me try:
Very likely that something is "broken" here in the sense that it is not implemented in a spec compliant way (or that the spec is not really great).
Both are likely artifacts of the fact that the two concepts have merely been added for practical reasons to facilitate the fulfillment of some requirement, but have not been fully thought through and were not part of QDMI initially.

For calibration, the payload is `void*` because it is up to the device to specify how it wants its calibration payload to look. Although I suppose that in most cases it will be some string describing configuration (like in the IQM case).
Overall, I would say that MQT Core should not fail if a payload is set for a calibration job; but it should also not fail when no payload is set, because it is optional.

Now batch jobs are a little more complicated and I would generally argue that MQT Core has no proper support for these yet. But I am also not aware of any device that currently implements that feature. IIRC, this was added for the multi-chip neutral atom demonstrator, which is 10*100 qubits, and the idea was that one bundles up a couple of job handles as the payload and then submits them as a batch job.
Given the uncertainty around batch jobs, I would be willing to simply drop the support for them explicitly in MQT Core and potentially add it back (properly) when it turns out to be needed.

Does that make sense?

marcelwa: Yes, I think it does. What I get from that is that we leave the current behavior as is in this PR. And I'll file a follow-up with the breaking changes that I'll stack on top of this one.

python/mqt/core/plugins/qiskit/serializers.py
marcelwa: Thanks for the example code. I wasn't considering the "loading" state and instead assumed it was binary. We have to explicitly state that neither solution is thread-safe, though.

burgholzer: Yeah. But I'd assume that thread-safety is not really a big thing here for now, given how Qiskt does not support free-threading Python, so this code can never be executed in a multi-thread setting.

burgholzer (approved):
This feels like it is ready to go in 🚀 
I'll wait for the stacked PR to also turn green and merge this as one stack afterwards! Then, I'll likely kick off a backport after the PennyLane PR is merged so that this makes it onto the v3.x branch and into the 3.9 release.

PR #2118, opened 2026-08-14
♻️ Make dead-gate removal an explicit QCO pass

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Make dead-gate removal an explicit QCO transformation instead of an implicit
canonicalization effect.

- Add the `remove-dead-gates` pass, preserving the existing backward removal
  from `qco.sink` and `qco.reset`.
- Remove dead-gate elimination from `SinkOp` and `ResetOp` canonicalization.
- Run the new pass explicitly in the `mqt-qubit-reuse` pipeline.
- Preserve unobserved quantum operations during ordinary QCO cleanup and target
  compilation, including measurement-free programs and reusable quantum
  building blocks.
- Move the existing dead-gate tests to the optimization-pass suite and add
  coverage for cleanup and target-compilation preservation.
- Document the explicit behavior and add the pass to the changelog.

No target-compilation option or Python API change is introduced.

This PR was implemented and submitted with OpenAI Codex assistance after the
author explicitly authorized the branch and PR work. Human maintainer review is
still required before merge.

## Validation

- 121/121 QCO optimization tests passed.
- 480/480 QCO IR tests passed.
- 134/134 compiler tests passed.
- The complete repository lint passed.
- The warning-as-error Sphinx documentation build passed.
- The branch was rebased and revalidated against `main` at `d9e7e1cd3`.

## Checklist

- [x] I have added migration instructions to the upgrade guide (not needed; no public API is changed).


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

mlir/include/mlir/Dialect/QCO/QCOUtils.h
```diff
@@ -26,6 +27,12 @@
 
 namespace mlir::qco {
 
+/// Stable canonicalizer label for patterns that erase unobserved quantum
+/// operations.
+inline constexpr llvm::StringLiteral
+    UNOBSERVED_QUANTUM_OPERATION_ELIMINATION_PATTERN_LABEL =
+        "qco-unobserved-quantum-operation-elimination";
```
burgholzer: This feels like an iteration artifact. We are not adding such labels for any other pass so far.

bindings/mlir/register_mlir.cpp
```diff
@@ -774,11 +775,24 @@ operations.)pb");
...
+            const mlir::TargetCompilationOptions options{
+                .preserveUnobservedQuantumOperations =
+                    preserveUnobservedQuantumOperations};
+            requireSuccess(program.compileForTarget(
+                target, options, enableTiming, enableStatistics));
+          },
+          "target"_a, nb::kw_only(),
+          "preserve_unobserved_quantum_operations"_a = false,
+          "enable_timing"_a = false, "enable_statistics"_a = false,
+          R"pb(Compile this QCO program for the target in place.
```
burgholzer: This opens a bit of a bigger box; how we want to configure our compiler.
This also affects `mqt-cc`.
I am not quite sure this is the best solution here. At least I am a bit hesitant to commit to this now based on this small flag.

If we were to go this way, we would likely directly expose the `TargetCompilationOptions` to Python as well to make the bindings and the usage easier once the number of options grows.

docs/mlir/target_compilation.md
```diff
@@ -40,6 +40,25 @@ compilation to an existing QCO program. For pass-level benchmarking, the C++ API
 exposes separate factories for pre-routing optimization, mapping, native
 synthesis, and conformance verification.
 
+By default, target compilation removes quantum operations whose final qubit
+values are not observed by a measurement or returned from the program. Callers
+that intentionally compile state-preparation circuits without observations can
```
burgholzer: This is narrower than it needs to be. State-preparation is not the only use case for when one would potentially not include measurements. For example, when someone simply wants to optimize a larger building block for a target native instruction set. When working through the other requests for changes, this should also be accounted for.

docs/mlir/target_compilation.md
```diff
@@ -40,6 +40,25 @@ compilation to an existing QCO program. For pass-level benchmarking, the C++ API
...
+    output=OutputFormat.QCO,
+)
+qco.compile_for_target(
+    target,
+    preserve_unobserved_quantum_operations=True,
+)
+```
+
+Preservation is intended for compiler analysis and benchmarking. The resulting
+program still has no classical observation of those qubit values.
```
burgholzer: Feels like a bit of an odd addition. As pointed out by the comment above, there may be valid use-cases beyond benchmarking for feeding measurement-free programs into a compiler.

mlir/include/mlir/Support/Passes.h
burgholzer, line 1: I did not really find a place where to put this, but I think that the current solution is not fully ideal.
Given how we noticed that it may not always be desired to run the dead gate removal, I think this is evidence enough that it should not be a canonicalization, but rather a dedicated pass that is called when it is explicitly desired to run.
The main reason for its introduction was @DRovara's work on qubit reuse, where the elimination is explicitly wanted.
So I am starting to think that we should simply turn this from a canonicalization to a dedicated pass and insert that pass in the pass pipeline for qubit re-use.
While it is a bit sad, I think this also means that we could reduce quite a bit of the test programs again if we wanted because they need not return something anymore if the removal pass is not executed. However, we can also keep the "complete" programs. It's just that we gain a couple more degrees of freedom back.

Let me know if you think this feels reasonable.

mlir/include/mlir/Support/Passes.h
```diff
@@ -56,10 +56,15 @@ runPassPipeline(mlir::ModuleOp module, mlir::StringRef pipeline,
...
+ * @brief Populate a configurable QCO-oriented cleanup pipeline.
+ * @param eliminateUnobservedQuantumOperations Whether to remove unobserved
+ * quantum operations. This parameter defaults to true. When false, only the
+ * labeled unobserved-operation elimination patterns are disabled.
+ * Canonicalization, phase normalization, CSE, QTensor shrinking, and dead-value
+ * removal remain enabled.
  */
-void populateQCOCleanupPipeline(mlir::OpPassManager& pm);
+void populateQCOCleanupPipeline(
+    mlir::OpPassManager& pm, bool eliminateUnobservedQuantumOperations = true);
```
burgholzer: Very random place to put this, but I am fairly unhappy with the naming suggestion for this change. The pass this concerns is called "Dead Gate Removal". So I would argue the most natural thing is to not invent a completely different synonymous naming, but refer to this as something like "runDeadGateRemoval".
Given the other review comments, I do not think that the option itself will survive, but for future additions this is something that I would value. I also already tried to institutionalize this in the AGENTS.md guidance, but maybe this is not effective enough yet.

burgholzer (changes requested):
Thanks @simon1hofmann 🙏🏼 
I believe this will require a bit of work to be complete and clean. As you will probably see from the comments, I am not fully happy with the direction this PR is taking.
However, I hope that the proposal in one of the comments actually makes sense and makes this much cleaner.

bindings/mlir/register_mlir.cpp
simon1hofmann: Makes sense, I changed the PR to simply making dead-gate removal an explicit pass and therefore don't have change the current configuration of our compiler.

mlir/include/mlir/Support/Passes.h
simon1hofmann: I changed this PR to making dead gate removal an optional pass but did not change the test programs yet.

burgholzer (approved):
So much cleaner 👍🏼 Great work!

PR #2124, opened 2026-08-15
✨ Add SpecAudits for finding and removing spec debt

marcelwa:
Sorry for the wall of text. I believe it is important to understand this addition and its application extensively. What follows is that extensive description plus a guided usage example. 

🤖 *AI text below* 🤖

## Description

This adds **SpecAudits**: a documented method, a small probe script, and one worked example, for finding tests that assert behaviour nobody ever specified — and for spending the freedom their removal buys.

No production code changes. No test is deleted or modified in this PR. It is additive: one process document, one script, a nine-line section in `AGENTS.md`, and one audit ledger.

### Why this is worth adding

`.agent/PLANS.md` is the repository's forward pass. It instructs an agent to "resolve ambiguities autonomously" and to "describe comprehensive tests for new capabilities". That is the right instruction, and it works. But it has a predictable side effect, and nothing in the repository currently checks for it.

An agent receives a request that does not state every detail. It fills the gaps and writes code. In the same commit it writes tests for that code. The tests pass, so they look correct. They are also a specification that nobody wrote on purpose. The next agent reads those tests as the contract and builds on them. A third agent tries to simplify the code, finds the change "breaks tests", and reverts. The accidental choice is now permanent.

The cost lands twice: the suite grows without covering more intent, and the code underneath freezes, because every simplification looks like a regression.

The repository already forbids the first half of this. `AGENTS.md` says:

> Add tests that protect intended behavior or reproduce a concrete regression. Do not test provisional implementation choices that are not part of the supported contract.

And `PLANS.md` says the same thing from the other side:

> Err on the side of over-explaining user-visible effects and under-specifying incidental implementation details.

Both rules are good. Neither is checked after the work lands. A SpecAudit is that check — the backward pass that `PLANS.md` implies but does not provide.

Two things make this worth doing here specifically rather than in the abstract. Roughly 500 commits carry `Assisted-by:` trailers, and across those commits `mlir/lib` sees 347 file touches against `mlir/unittests`' 298 — implementation and its tests written in lockstep, by the same agent, in the same commit, at scale. And Codecov enforces patch targets of 90% (C++) and 95% (Python) with essentially nothing excluded, so unreachable defensive branches create real pressure to write tests that manufacture impossible inputs. `test/fomac/test_fomac.cpp` casts `0` to an enumeration behind a `NOLINT` for `EnumCastOutOfRange` purely to reach a `detail::` throw; that test now pins the defensive branch in place.

### What this adds

| Path | What it is |
| --- | --- |
| `.agent/AUDITS.md` | The method, self-contained, as a peer of `PLANS.md` |
| `.agent/audits/<scope>.md` | One audit ledger per audited scope, as a peer of `.agent/plans/` |
| `.agent/audit-probe.sh` | POSIX `sh`; runs the evidence loop so every candidate gets identical treatment |
| `AGENTS.md` | A nine-line `## Spec Audits` section beside `## ExecPlans` |

It lives under `.agent/` because that is the tracked, vendor-neutral agent surface here — `.claude/` and `CLAUDE.md` are git-ignored, and the disclosed trailers show Codex, Claude Code, Cursor and Copilot all in use on this repository. `AGENTS.md` is the only discovery entry point, and it is repo-owned (`templating.yml` passes `synchronize-agents-md: false`), so extending it does not fight the templating bot. No template-managed file is touched.

The method in brief:

- **A spec ladder.** Every assertion is asked which promise it defends and who made it. External and machine-checked sources rank first (the pinned QDMI headers, the MLIR TableGen files, the OpenQASM grammar, the QIR profiles), then published promises (`CHANGELOG.md`, `UPGRADING.md`, `docs/`, Doxygen on exported headers), then the issue or review comment that requested the work, then an ExecPlan — and *the code* and *a test* are explicitly **not** promises. An assertion whose only support is "the code does this" is a mirror; it cannot fail unless someone changes the code, which is exactly when you want the freedom to change it.
- **Five verdict classes** — anchored, over-specified, redundant, contract-free, and coverage-driven — with the standing rule **narrow before merge, merge before delete**.
- **A tiered evidence protocol.** Coverage delta, then fault injection, then scoped mutation, escalating only as far as a candidate's value justifies, because a C++ reconfigure-and-build here costs several minutes. The rule that matters: **no verdict without an executed experiment.**
- **Unlock analysis** — for every assertion narrowed or deleted, what does this now permit? Including the class that usually goes unnoticed: work the code does *only* to satisfy an over-specified assertion.
- **Role isolation.** Spec cartographers are forbidden from reading the tests, because a cartographer that reads them rediscovers their assumptions as promises, which silently turns the whole audit into a rubber stamp. An independent advocate then argues to save every accused assertion, without sight of the case against it.

The method is deliberately vendor-neutral: roles are described as charters with inputs, forbidden inputs and an output schema, and a degraded mode is specified for agents with no subagent primitive — run the stages as separate fresh sessions, because the isolation is the method and the parallelism is only a speed optimisation.

### How to use it

Point whichever agent you use at `.agent/AUDITS.md` and give it a bounded scope — one plugin, one dialect, one device, one pass. Whole-repository audits are refused by the method; at roughly 2,250 tests the evidence protocol would mean nothing at that size.

```
Audit test/python/plugins/qdmi_pennylane against python/mqt/core/plugins/pennylane/
following .agent/AUDITS.md. Stop at the ledger.
```

The audit writes `.agent/audits/<scope-slug>.md` and stops. It never applies its own findings — a human picks which verdicts are worth acting on, and each one lands as its own pull request. That is what `docs/ai_usage.md` requires anyway: a human stays accountable, and an unreviewed bulk deletion of tests is precisely the "extractive contribution" that document warns about.

The probe script runs the two cheap evidence tiers directly:

```sh
# T2 — break the code an assertion claims to guard, see which tests notice
./.agent/audit-probe.sh t2 --lang python \
  --tests test/python/plugins/qdmi_pennylane \
  --inject python/mqt/core/plugins/pennylane/converter.py:228 \
  --with '    return repr(_finite_parameter(parameter, operation_name))'
```

It refuses to run on a dirty working tree, restores every file it edits, uses the build and test entry points `AGENTS.md` documents, and prints an evidence block to paste into the ledger. `--lang cpp` drives the coverage preset and a `ctest` filter instead. Running the loop by hand is fine; running it *differently* for different candidates is not, which is the only reason the script exists.

Reading a T2 result:

- another test fails → the candidate is **redundant**;
- nothing fails and no published promise makes the fault visible → the behaviour is **contract-free**, and the code you just broke is now the suspect;
- nothing fails but a promise *does* make it visible → the assertion is the sole anchor. **Keep it**, and record the fault it caught so the next auditor does not repeat the experiment.

### What the first audit found

Worked example: `.agent/audits/pennylane-plugin.md`, over `python/mqt/core/plugins/pennylane/` (915 lines) and `test/python/plugins/qdmi_pennylane/` (834 lines, 26 tests, 33 collected cases). Chosen because it needs no C++ rebuild, so fault injection costs seconds. Baseline `18b619af5`; every `file:line` was read at that commit; **seven fault-injection probes were executed** and the failing-test lists in the ledger are copied from the probe output.

To be clear about what this is not: the plugin is good work, and none of the below is a defect report. The point is that nothing checked the *test* side afterwards, which is exactly the gap this PR proposes to close.

**1. Seventeen-digit float literals block readable OpenQASM output.** `converter.py:228` renders parameters with `format(value, ".17g")`, so the emitted program contains `rxx(0.10000000000000001)`. Three assertions in `test_converter.py:131-133` pin those literals. Replacing `.17g` with `repr` — which gives the shortest representation that still round-trips exactly, preserving the docstring's stated intent — fails exactly one test:

```
failing tests   : 1
  - test_converter.py::test_qasm3_resolves_ddsim_aliases_and_inverse_gates
```

I verified the unlock end to end on a throwaway branch: with the assertion narrowed to parse-and-compare, the full suite passes 33/33 and the emitted program becomes `rxx(0.1)` at identical precision. Nothing published says anything about parameter rendering, and `_format_parameter` is private.

**2. One rule implemented twice, with each wording pinned.** `device.py:187-202` and `converter.py:132-147` are the same QASM3-then-QASM2-then-raise rule, differing only in the message: `"advertises neither OpenQASM 3 nor OpenQASM 2."` versus `"supports none of the required program formats: OpenQASM 3 or OpenQASM 2."`. `test_device.py:218` pins the first, `test_converter.py:212` the second, so unifying them breaks a test either way. Rewriting one message to match the other fails exactly one test. Neither wording is published. `device.py:163` already caches the selected format and uses it at `device.py:225-228`; `device.py:346` simply does not pass it, so `converter.py:413` re-derives it once per tape.

**3. A sample ordering that is asserted but never promised.** Two probes on `device.py:274` together say something precise: removing `sorted(...)` fails **nothing**, while reversing it fails `test_device.py::test_histogram_only_device_reconstructs_samples`. So the assertion pins *an* order — blocking any future change to the expansion strategy — but does not protect the `sorted` call, because the stub's `Counter` insertion order already equals sorted order. The load-bearing part of that test is multiplicity, not order, and no other test exercises `expose_shots=False`.

**4. The most useful result was a failure of the method's own prosecution.** The advocate argued that `test_device.py:180` is the only guard against a double diagonalizing rotation — drop `rotations=False` at `converter.py:384` and a QASM2-only device would silently return wrong expectation values. It was the strongest argument offered for any assertion in the scope. Flipping the flag fails **zero** tests. By the time a tape reaches `_convert_qasm2` through the device, `measurements_from_samples` has already replaced the observable with a `SampleMP`, so there are no diagonalizing gates left to add; the flag is inert on the only production path. The argument was careful, cited the plan and the pipeline, and was wrong — and only running it showed that. This is the clearest case for why the method forbids verdicts settled by argument.

**5. An anchor confirmed, and which tests actually held it.** `device.py:300` reverses each QDMI bit string; the endianness convention is stated nowhere except the code comment above that line. Removing the reversal fails three tests — all three in `test_ddsim.py`, including the differential comparison against `default.qubit`. Every stub-based test in `test_device.py` stayed green, because the stub's Bell histogram is palindromic. The end-to-end tests are the only defence for the most consequential unwritten convention in the plugin.

**6. The largest unlock, and the one I would not take unilaterally.** `convert_program` and `ConvertedProgram` are exported from the package `__all__`, but the only callers anywhere are `device.py:346` and `test_converter.py`, and neither name appears in the documentation. That three-argument free-function signature is what prevents the device from handing down state it has already computed: `supports_operation` therefore rebuilds the advertised-operation table on every call, and it is PennyLane's per-operation `stopping_condition`. For a 100-gate circuit on an 18-operation device, one preprocessing pass makes 101 `device.operations()` round trips and 1818 `Operation.name()` calls where 1 and 18 would do. Unblocking it means rewriting a 236-line test file against `QDMIDevice` plus `StubDevice.submissions` — a technique `test_device.py:125-126` already uses. It is ranked last in the ledger despite the payoff, because it is a maintainer decision.

Two things the ledger is explicit about, and which I would rather flag than bury. `test_converter.py:171-181` and the `nan`/`inf` matrix are defended but **unprobed**, and labelled unverified rather than cleared — severity and confidence are tracked separately. And four cleanups found along the way (an unreachable duplicate shots guard at `device.py:247-249`, a discarded duplicate lambda, a two-function parameter wrapper, a duplicated `ConvertedProgram` construction) are recorded in a separate section stating plainly that **no assertion blocks them**, so the audit claims no credit for them. A method that takes credit for everything it noticed cannot be evaluated.

The ledger also records how the scope came to look this way, because it is the mechanism rather than a criticism: `git log --diff-filter=A` shows `.agent/plans/qdmi-pennylane-device.md` was added in the same commit as the code and tests it describes, `077b73f80`, already carrying completed checkboxes and a retrospective. The method therefore treats an ExecPlan's stated purpose as requested intent, and details it invented on the way there as not a promise.

### What I am asking for

Mainly whether the method is worth having and whether `.agent/` is the right home for it. The verdicts in the ledger are deliberately not acted on here; if the approach is welcome, finding 1 is the smallest useful first patch and I am happy to open it separately.

## Checklist

- [ ] I have added appropriate tests that cover the new/changed functionality. — *n/a: no production code changes. The script is exercised by the seven probes recorded in the ledger, and `shellcheck` reports only two `SC1007` warnings, both raised by the deliberate `CDPATH= cd --` idiom used to resolve the script's own location.*
- [x] I have updated the documentation to reflect these changes. — *`AGENTS.md` gains the `## Spec Audits` section; the method document is itself the documentation.*
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals. — *added under `Added`. Noting for the record that the two closest precedents, `.agent/PLANS.md` ([#1907]) and `AGENTS.md` ([#1518]), both shipped without one; happy to drop the entry if you would rather keep agent-facing process docs out of the changelog.*

[#1907]: https://github.com/munich-quantum-toolkit/core/pull/1907
[#1518]: https://github.com/munich-quantum-toolkit/core/pull/1518
- [ ] I have added migration instructions to the upgrade guide (if needed). — *n/a.*
- [x] The changes follow the project's style guidelines and introduce no new warnings. — *`uvx nox -s lint` passes on all files.*

*AI assistance disclosure: the method document, the probe script, and the audit ledger were drafted with Claude Opus 5 via Claude Code, under the commit trailer `Assisted-by: Claude Opus 5 via Claude Code`.*


.agent/AUDITS.md
```diff
@@ -0,0 +1,537 @@
...
+./.agent/run.sh uv run --no-sync pytest --cov-config=pyproject.toml
+```
+
+For C++, use the coverage preset:
+
+```sh
+./.agent/run.sh cmake --preset coverage
+./.agent/run.sh cmake --build --preset coverage
+./.agent/run.sh ctest --preset coverage
+```
```
burgholzer: One of the last PRs actually dropped the run.sh script. I believe the respective parts here should be updated accordingly.

.agent/audit-probe.sh
burgholzer, line 1: I believe the run_wrapper can be dropped here based on the other comment

burgholzer (changes requested):
I _really_ like this. Thanks for the addition @marcelwa 🙏 that was a long but interesting read.

I only have one small observation about some recently outdated wording.

I'd also appreciate the follow-up to see this in action!

marcelwa:
I'm glad you like it. Hope it's helpful. Both of your comments are addressed :slightly_smiling_face:

burgholzer (approved):
Nice. Let's get this in and see it in action 🚀

PR #2125, opened 2026-08-15
🔧 Support stock LLVM builds without exceptions or RTTI

### LINKED ISSUE #1589: 🚸 MLIR - Enable building without RTTI
Due to the integration with the `qc::QuantumComputation` class and its type hierarchy, we currently need to compile MLIR with RTTI enabled.
This is not the default because of performance reasons and we actually do not require RTTI in the MLIR part of the code base except for the translation from `qc::QuantumComputation` to MLIR.

This is intended as a tracking issue for removing the requirement to build with RTTI enabled.
As soon as this is enabled, the MLIR toolchain builds we host may be updated to drop the respective flag.

### LINKED ISSUE #1590: 🚸 MLIR - Enable building without exceptions enabled

burgholzer:
🤖 *AI text below* 🤖

## Summary

- inherit the exception and RTTI policy from the imported LLVM and MLIR 22.1.8 configuration
- compile ordinary MLIR targets and `mqt-cc` without exceptions or RTTI when using a stock LLVM build
- retain exceptions only at the QDMI, QCO DD, QIR, DDSIM, test, and nanobind compatibility boundaries that require them
- convert QDMI discovery and target-construction failures to `llvm::Error` before they reach `mqt-cc`
- update every repository LLVM and MLIR pin to the portable 22.1.8 release
- pin CI to the exact reusable-workflows commit that uses setup-mlir v1.4.2 and can install newly released toolchains
- replace a fragile OpenQASM member-template with ordinary overloads so Apple Clang cannot leave an undefined symbol in the nanobind extension

## Motivation

MQT Core forced exceptions and RTTI on for MLIR builds. This conflicted with stock LLVM and MLIR installations, which disable both by default. It also allowed the compiler driver to call throwing QDMI C++ interfaces directly.

The central target helpers now apply LLVM's imported compile policy to MLIR libraries, their object libraries, tools, and test fixtures. Small compatibility targets retain exceptions while using LLVM's platform-specific no-RTTI flag when required. `mqt-cc` uses error-returning QDMI adapters and no longer needs a top-level exception handler.

This is a breaking build-policy change for downstream code that relied on MQT Core forcing exceptions or RTTI on. The complete compiler remains enabled by default; this change adds no build option, reduced configuration, or CI matrix.

## Validation

- fresh Release build against `/home/nvidia/.local/opt/mqt-llvm-mlir/22.1.8/lib/cmake/mlir`
- complete CTest suite: 4,382 passed, with 2 expected skips
- supported Python 3.10 through 3.14 suites: 3,369 passed
- focused post-rebase LLVM 22.1.8 rebuild and 10 QDMI adapter and driver tests
- MLIR documentation target
- full lint suite
- `git diff --check`
- compile database audit: 55 ordinary MLIR and `mqt-cc` entries use `-fno-exceptions -fno-rtti`; 9 documented MLIR compatibility entries retain exceptions with `-fno-rtti`; only the nanobind extension retains compiler-default exceptions and RTTI

The full Sphinx session could not run locally because the host has no Doxygen executable. The MLIR documentation target passed.

Fixes #1589.
Fixes #1590.

Supersedes #2051 and #2048.

AI assistance: Codex implemented and validated this change under human direction. A human must review and merge it.


docs/mlir/target_compilation.md
```diff
@@ -96,6 +94,12 @@ if (!qco || !qco->compileForTarget(*target)) {
 }
 ```
 
+The ID factory and {code}`mlir::registeredQDMIDeviceIds` contain exceptions from
+the QDMI C++ client and return them as {code}`llvm::Error` values. This path is
+safe for an exception-free caller such as {code}`mqt-cc`. The lower-level
+{code}`mlir::compilerTargetFromDevice` overload provides the same error boundary
+for callers that already own a QDMI device.
+
```
burgholzer:
```suggestion
```
Feels like this is not totally needed as explanation

mlir/tools/mqt-cc/CMakeLists.txt
```diff
@@ -6,8 +6,9 @@
 #
 # Licensed under the MIT License
 
-# Build the compiler driver executable
-add_mlir_tool(mqt-cc mqt-cc.cpp DEPENDS MQTCompilerPipeline MQTCompilerQDMIAdapter SUPPORT_PLUGINS)
+# Keep statically linked LLVM and MLIR symbols private. Exporting them lets a loaded QDMI device
+# interpose LLVM's process-wide registries.
```
burgholzer: I am wondering: could we enable plugins here, but tweak the configuration of the DDSIM QDMI device in a similar minimally invasive way so that the underlying conflict is resolved?

pyproject.toml
```diff
@@ -299,10 +299,10 @@ test-skip = [
 [tool.cibuildwheel.linux]
 before-all = """
 uv tool install sccache>=0.10.0
-curl -LsSf https://github.com/munich-quantum-software/setup-mlir/releases/download/v1.4.1/setup-mlir.sh | bash -s -- -v 22.1.7 -p /opt/llvm-22.1.7
+curl -LsSf https://github.com/munich-quantum-software/setup-mlir/releases/download/v1.4.1/setup-mlir.sh | bash -s -- -v 22.1.8 -p /opt/llvm-22.1.8
```
```suggestion
curl -LsSf https://github.com/munich-quantum-software/setup-mlir/releases/download/v1.4.2/setup-mlir.sh | bash -s -- -v 22.1.8 -p /opt/llvm-22.1.8
```
This may not work otherwise.

test/qir/mqt-cc/CMakeLists.txt
```diff
@@ -13,3 +13,11 @@ add_test(
...
     "-DOUTPUT_DIR=${CMAKE_CURRENT_BINARY_DIR}/output-$<CONFIG>" -P
     "${CMAKE_CURRENT_SOURCE_DIR}/verify_qir_output.cmake")
+
+add_test(
+  NAME mqt-core-mqt-cc-qdmi-test
+  COMMAND
+    ${CMAKE_COMMAND} "-DMQT_CC=$<TARGET_FILE:mqt-cc>"
+    "-DINPUT_FILE=${CMAKE_CURRENT_SOURCE_DIR}/bell.qasm"
+    "-DOUTPUT_DIR=${CMAKE_CURRENT_BINARY_DIR}/qdmi-output-$<CONFIG>" -P
+    "${CMAKE_CURRENT_SOURCE_DIR}/verify_qdmi.cmake")
```
burgholzer: Doesn't really feel like this adds value
```suggestion
```

test/qir/mqt-cc/verify_qdmi.cmake
burgholzer, line 1: This can be removed.

burgholzer (commented):
Some request for changes.

burgholzer:
This should be ready now and also unblocks us from using the no-RTTI LLVM 22.1.8 release.

@munich-quantum-toolkit/mqt-cc for awareness, I pinned the workflows to a particular commit from `main` here for now to get the `setup-mlir` update without going through the churn of having to prepare and release a new workflow version now. Before the next release, at the latest, we need a new release of the workflows out.

PR #2135, opened 2026-08-17
✨ Add typed structured quantum benchmarks

### LINKED ISSUE #1115: 🧪 MLIR - Add QPE and Iterative Phase Estimation examples
## Background
We have manual MLIR sketches using `scf`/`arith` for classical control and `memref<?xi1>` for classical registers. See the first two comments below.

## Problem
We need runnable, reviewed examples that exercise classical-quantum interplay and passes.

## Proposal:
- Commit `.mlir` examples for QPE and IPE in MQTRef; include tests for parse/roundtrip and simple analysis checks; optional simulator hook to validate expected classical results.

## Acceptance criteria:
- Examples and tests present; comments explain design choices.

This is one of the first steps in elaborating how to represent hybrid quantum-classical workflows.

denialhaag:
## Description

🤖 *AI text below* 🤖

Add a typed foundation for structured quantum benchmarks.

The new MLIR-free `MQT::CoreBench` library owns each benchmark's parameters, validation, logical output, analytic reference, and evaluation. This vertical slice contains:

- Bernstein--Vazirani with an explicit hidden bitstring and static or dynamic method;
- GHZ with qubit count, topology, and measurement basis;
- Grover search with an explicit marked bitstring and automatic or explicit iteration count;
- QFT with qubit count, period exponent, and standard or semiclassical method;
- QPE with precision, an exact rational phase in turns, and standard or iterative method.

Strict, versioned JSON instances materialize all defaults. Self-checking manifests record the resolved parameters, output convention, reference model, and a deterministic case ID. The C++, Python, and `mqt-core-bench` APIs expose the same contract. The command-line tool lists and describes benchmarks, generates structured QC or `jeff`, and evaluates counts against a manifest.

The MLIR layer contains compact, structured QC emitters for the five benchmark families. It consumes typed instances and keeps its private dispatch table aligned with the semantic registry instead of duplicating defaults. The Python `mqt.core.bench` module delegates program generation to the existing MLIR extension, which keeps one MLIR runtime per process.

Distribution-level execution tests against the DD simulator remain a follow-up on the stack that starts at #2077.

Fixes #1115

## AI notice

This PR and its contents were created with the assistance of Opus 5 via Claude Code and GPT-5.6 Sol via Codex.

## Checklist


mlir/benchmark/programs/QFT.cpp
```diff
@@ -0,0 +1,74 @@
+/*
```
MatthiasReumann: 👋🏻 Sorry for coming in early! 

I really love these files! Any chance we could make these programs the "single-source of truth" (not just for JEFF) and move them somewhere for further reuse?

denialhaag: No worries! I guess we could just promote the programs to `mlir/benchmark/programs` and keep the `jeff` specifics in `mlir/benchmark/jeff`. 🤔

MatthiasReumann: sounds good! An associated header file in the `include/` folder and I will be very happy :)

MatthiasReumann: @burgholzer Maybe this is also relevant for the existing "programs" used for unit-testing as these are rather snippets than programs. Following the discussion in #2118, maybe it does make more sense that these snippets don't require measurements.

burgholzer: Yeah. What you are suggestion @MatthiasReumann feels very reasonable. This is similar to how in MQT Bench, we define the benchmark generation scripts as the single source of truth and then build infrastructure for conversion and compilation around it.

I'd also be in favor of keeping the unittest snippets as close to "unit" as possible and would be fine if we make them smaller again.

denialhaag: https://github.com/munich-quantum-toolkit/core/pull/2135/changes/c0882d70fe99ed49c5ba62a9ac77a8549b8297ed promotes the programs as discussed. Let me know if you have any feedback on the directory structure! 🙂

denialhaag:
Modulo some more testing, we should be able to mark this PR as fixing #1115.

denialhaag:
🤖 *AI text below* 🤖

## Benchmark coverage

This branch now covers **26 of the 28 programs** listed in the [`jeff` structured benchmark README](https://github.com/unitaryfoundation/jeff/blob/main/benchmarks/structured/README.md), and **25 of the 26 rows** of Table IV of the challenge paper. The two counts differ because the README splits GHZ into linear and star, and lists one program that Table IV does not.

Two programs are missing. Both are blocked for the same reason: the README marks each of them with two ❓ entries, which its own legend defines as "requires some further research". The unresolved columns are *dynamic classical values* and *conditionals on originally classical values*.

| Program | Reference | Why it is missing |
| --- | --- | --- |
| X-Ray Absorption Spectroscopy | [arXiv:2405.11015](https://arxiv.org/abs/2405.11015), [PennyLane demo](https://pennylane.ai/qml/demos/tutorial_xas) | Fixed size, so there is no structural parameter to write against. What makes it this benchmark is a specific molecular Hamiltonian and its Trotterization, which the reference has to supply. |
| Quantum Metropolis Sampling | [arXiv:0911.3635](https://arxiv.org/abs/0911.3635) | Fixed size. Also absent from Table IV of the paper, so the README is the only source that asks for it. |

Everything else from the README is implemented, including all five entries of the Error Correction & Fault Tolerance section.

denialhaag:
🤖 *AI text below* 🤖

## `jeff` limitations found while writing the benchmarks

`jeff-mlir` converts a fixed set of `arith` operations (`lib/Conversion/NativeToJeff/NativeToJeff.cpp`). Two gaps stopped a faithful program from converting. `arith.cmpi` is supported, so comparing classical values is not the problem.

| Missing | Effect |
| --- | --- |
| `arith.select` | A classical value cannot be chosen by a measurement outcome. Measurement-based quantum computation needs this to pick the angle of the next measurement, so it branches on the measured bit and applies a constant rotation in each arm instead. |
| `arith.extui` / `extsi` / `trunci`, `sitofp` / `uitofp` | A measured bit cannot be widened to an integer or a float. VQE counts an Ising energy from the measured register to decide whether to run another round, which needs exactly this. |

denialhaag:
@coderabbitai full review

denialhaag:
@coderabbitai review

mlir/benchmark/programs/ControlledMultiplyModN.cpp
```diff
@@ -0,0 +1,102 @@
...
+SmallVector<Value> controlledMultiplyModN(qc::QCProgramBuilder& b,
+                                          const uint64_t n) {
+  const auto total = static_cast<int64_t>(n) - 2;
+  const auto factor = total / 2;
+  const auto size = total - factor;
+  auto ctrl = b.allocQubit();
+  auto x = b.allocQubitRegister(factor, "x");
+  auto acc = b.allocQubitRegister(size, "acc");
+  auto anc = b.allocQubit();
+  auto c = b.allocClassicalBitRegister(size, "c");
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/FanOut.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+  // The first qubit holds the value that is copied. The fan-out spreads it
+  // over the register so that the following layer acts on every qubit at once.
+  b.h(q[0]);
+  b.scfFor(1, size, 1, [&](Value iv) { b.cx(q[0], b.loadQubit(q.value, iv)); });
+
+  b.scfFor(1, size, 1, [&](Value iv) { b.t(b.loadQubit(q.value, iv)); });
+
+  // The copies are uncomputed so that only the first qubit stays entangled.
+  b.scfFor(1, size, 1, [&](Value iv) { b.cx(q[0], b.loadQubit(q.value, iv)); });
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/MeasurementBasedComputation.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+SmallVector<Value> measurementBasedComputation(qc::QCProgramBuilder& b,
+                                               const uint64_t /*n*/) {
+  auto q = b.allocQubitRegister(MBQC_CLUSTER, "q");
+  auto c = b.allocClassicalBitRegister(MBQC_CLUSTER, "c");
+
+  resetRegister(b, q.value, MBQC_CLUSTER);
+
+  // The cluster state is a chain of qubits in the plus state that CZ joins.
+  b.scfFor(0, MBQC_CLUSTER, 1, [&](Value i) { b.h(b.loadQubit(q.value, i)); });
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/QFTAdderClassical.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+  // Adding a classical constant in the Fourier basis needs no controls. Each
+  // qubit takes one phase that depends only on the constant, so the whole
+  // addition is a single layer of phase gates between the two transforms.
+  fourierTransform(b, q.value, size, 1.0);
+
+  for (int64_t i = 0; i < size; ++i) {
+    const auto angle = std::numbers::pi * static_cast<double>(ADDEND) /
+                       static_cast<double>(int64_t{1} << i);
+    b.p(angle, q[size - 1 - i]);
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/SyndromeMeasurement.cpp
```diff
@@ -0,0 +1,72 @@
...
+          b.cx(b.loadQubit(q.value, next), anc);
+          b.measure(anc, syndrome, i);
+        });
+
+        b.scfFor(0, size - 1, 1, [&](Value i) {
+          auto next = arith::AddIOp::create(b, i, one);
+          b.scfIf(syndrome, i, [&] { b.x(b.loadQubit(q.value, next)); });
+        });
+
+        b.scfCondition(syndrome, 0);
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/ToffoliHeavy.cpp
```diff
@@ -0,0 +1,50 @@
...
+SmallVector<Value> toffoliHeavy(qc::QCProgramBuilder& b, const uint64_t n) {
+  const auto size = static_cast<int64_t>(n);
+  auto q = b.allocQubitRegister(size, "q");
+  auto c = b.allocClassicalBitRegister(size, "c");
+
+  resetRegister(b, q.value, size);
+  b.scfFor(0, size, 1, [&](Value iv) { b.h(b.loadQubit(q.value, iv)); });
+
+  // A chain of Toffoli gates. Each gate takes its two controls from the
+  // preceding qubits, so the chain carries a running AND along the register.
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/VQE.cpp
```diff
@@ -0,0 +1,77 @@
...
+            auto next = arith::AddIOp::create(b, i, one);
+            b.cx(b.loadQubit(q.value, i), b.loadQubit(q.value, next));
+          });
+        });
+        auto improved = b.measure(q[0]);
+        scf::ConditionOp::create(b, improved, ValueRange{angle});
+      },
+      [&](OpBuilder&, Location, ValueRange args) {
+        auto next = arith::MulFOp::create(b, args[0], decay);
+        scf::YieldOp::create(b, ValueRange{next});
```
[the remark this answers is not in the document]
denialhaag: I believe this benchmark has moved to #2204.

mlir/benchmark/programs/Multiplexer.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <cstdint>
+#include <numbers>
+
+namespace mqt::benchmark {
+
+using namespace mlir;
+
+SmallVector<Value> multiplexer(qc::QCProgramBuilder& b, const uint64_t n) {
+  const auto numControls = static_cast<int64_t>(n) - 1;
+  const auto numStates = int64_t{1} << numControls;
```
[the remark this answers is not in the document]
denialhaag: Is this fixed by the `maximumSize` field added in https://github.com/munich-quantum-toolkit/core/pull/2135/commits/bcaec8dea552c9d04991c98abe190552b277401c?

mlir/benchmark/Compile.cpp
```diff
@@ -0,0 +1,65 @@
...
+using namespace mlir;
+
+std::optional<QCProgram> buildQCProgram(const Benchmark& benchmark,
+                                        const uint64_t n) {
+  if (n < benchmark.minimumSize) {
+    return std::nullopt;
+  }
+  if (benchmark.maximumSize != 0 && n > benchmark.maximumSize) {
+    return std::nullopt;
+  }
```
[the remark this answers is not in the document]
denialhaag: Is this fixed by https://github.com/munich-quantum-toolkit/core/pull/2135/commits/67ef253d73d07544ef4f503737283d5937b77c71?

mlir/unittests/benchmark/test_benchmark_jeff.cpp
```diff
@@ -0,0 +1,70 @@
...
+      << assembly;
+}
+
+TEST_P(JeffBenchmarkTest, RejectsSizesBelowTheMinimum) {
+  const auto& benchmark = GetParam();
+  if (benchmark.minimumSize == 0) {
+    GTEST_SKIP() << "the benchmark accepts every size";
+  }
+  EXPECT_FALSE(buildQCProgram(benchmark, benchmark.minimumSize - 1));
+}
```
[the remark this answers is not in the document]
denialhaag: I think a smoke test might just be good enough for now. Once we have test infrastructure that can actually check the results, we should definitely add such tests, though.

mlir/benchmark/programs/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+# Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+# The benchmark programs, one file per program. The programs build QC dialect modules and carry no
+# format-specific code, so any backend can reuse them.
+file(GLOB PROGRAM_SOURCES *.cpp)
```
denialhaag:
```suggestion
file(GLOB PROGRAM_SOURCES *.cpp)
```

mlir/benchmark/programs/BenchmarkUtils.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+void scfForWithAngle(qc::QCProgramBuilder& b, Value lower, Value upper,
+                     const std::variant<double, Value>& start,
+                     const double factor,
+                     const function_ref<void(Value, Value)>& body) {
+  auto one = b.indexConstant(1);
+  auto first = std::holds_alternative<Value>(start)
+                   ? std::get<Value>(start)
+                   : b.floatConstant(std::get<double>(start));
```
denialhaag: Let's use `variantToValue` here.

mlir/benchmark/programs/BenchmarkUtils.cpp
```diff
@@ -0,0 +1,58 @@
...
+  b.scfFor(0, size, 1, [&](Value i) { b.reset(b.loadQubit(reg, i)); });
+}
+
+void measureRegister(qc::QCProgramBuilder& b, Value reg, const int64_t size,
+                     Value bits) {
+  b.scfFor(0, size, 1,
+           [&](Value i) { b.measure(b.loadQubit(reg, i), bits, i); });
+}
+
+void scfForWithAngle(qc::QCProgramBuilder& b, Value lower, Value upper,
```
denialhaag: Maybe this needs a more descriptive name, given how specific it is.

mlir/benchmark/Generate.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace mqt::benchmark {
+
+using namespace mlir;
+
+std::optional<QCProgram> buildQCProgram(const Benchmark& benchmark,
+                                        const uint64_t n) {
+  if (n < benchmark.minimumSize) {
+    return std::nullopt;
+  }
```
denialhaag: Does it make sense to emit an error via `llvm::errs` here and below?

mlir/benchmark/jeff/MQTJeffBenchmarks.cpp
```diff
@@ -0,0 +1,115 @@
...
+                    llvm::cl::value_desc("directory"), llvm::cl::init("."));
+
+static llvm::cl::opt<std::string> programFilter(
+    "program",
+    llvm::cl::desc("Generate only the named program instead of every one"),
+    llvm::cl::value_desc("name"), llvm::cl::init(""));
+
+/// Builds one benchmark and writes it as a `jeff` file.
+static bool generate(const mqt::benchmark::Benchmark& benchmark,
+                     const uint64_t n, const std::filesystem::path& directory) {
```
denialhaag: Could we reuse `buildQCProgram` here?

mlir/include/mlir/Benchmark/BenchmarkUtils.h
denialhaag, line 1: Maybe it does make sense to move this back to `mlir/benchmark/jeff/programs`. I don't think we want to use these functions outside of that location.

mlir/include/mlir/Benchmark/Generate.h
denialhaag, line 1: Would something like `Generate.h` be a better fit? Also, is there a more fitting name for `buildQCProgram`?

mlir/unittests/benchmark/BenchmarkTestUtils.h
denialhaag, line 1: Let's get rid of this, even if it means duplicating some code.

mlir/unittests/benchmark/test_benchmark_jeff.cpp
```diff
@@ -0,0 +1,86 @@
...
+  EXPECT_FALSE(buildQCProgram(benchmark, std::numeric_limits<uint64_t>::max()));
+}
+
+TEST_P(JeffBenchmarkTest, RejectsSizesAboveTheMaximum) {
+  const auto& benchmark = GetParam();
+  if (benchmark.maximumSize == 0) {
+    GTEST_SKIP() << "the benchmark has no upper size limit";
+  }
+  EXPECT_FALSE(buildQCProgram(benchmark, benchmark.maximumSize + 1));
+}
```
denialhaag: If we reuse the `buildQCProgram` function, these tests and the one above would be redundant.

mlir/benchmark/programs/GHZLinear.cpp
```diff
@@ -0,0 +1,43 @@
...
+SmallVector<Value> ghzLinear(qc::QCProgramBuilder& b, const uint64_t n) {
+  const auto size = static_cast<int64_t>(n);
+  auto q = b.allocQubitRegister(size, "q");
+  auto c = b.allocClassicalBitRegister(size, "c");
+
+  resetRegister(b, q.value, size);
+
+  b.h(q[0]);
+  b.scfFor(1, size, 1, [&](Value iv) {
+    auto previous = arith::SubIOp::create(b, iv, b.indexConstant(1));
```
denialhaag: Can we define `auto one =  b.indexConstant(1)` outside of the loop?

mlir/benchmark/programs/IQFT.cpp
```diff
@@ -0,0 +1,58 @@
...
+                      auto index = arith::AddIOp::create(b, offset, step);
+                      b.scfIf(res, index, [&] { b.p(angle, q); });
+                    });
+
+    b.h(q);
+    auto last = b.indexConstant(bits - 1);
+    auto target = arith::SubIOp::create(b, last, i);
+    b.measure(q, res, target);
+    b.reset(q);
+  });
```
denialhaag: Also here: Can we define the constants outside of the loop?

mlir/benchmark/programs/Multiplexer.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    b.mcry(angle, controls.qubits, target);
+
+    for (int64_t bit = 0; bit < numControls; ++bit) {
+      if (((state >> bit) & 1) == 0) {
+        b.x(controls[bit]);
+      }
+    }
+  }
```
denialhaag: Is there no way to write this with `scf::ForOp`?

mlir/benchmark/programs/QPE.cpp
```diff
@@ -0,0 +1,72 @@
...
+  b.scfFor(0, counting, 1, [&](Value i) {
+    auto lower = arith::AddIOp::create(b, i, b.indexConstant(1));
+    auto upper = b.indexConstant(counting);
+    scfForWithAngle(b, lower, upper, -std::numbers::pi / 2.0, 0.5,
+                    [&](Value angle, Value j) {
+                      b.cp(angle, b.loadQubit(q.value, j),
+                           b.loadQubit(q.value, i));
+                    });
+    b.h(b.loadQubit(q.value, i));
+  });
```
denialhaag: Also here: Let's pull the constants out.

denialhaag (commented):
Some comments:

mlir/benchmark/programs/CMakeLists.txt
denialhaag: 🤖 *AI text below* 🤖

Addressed in 25785ccea.

mlir/benchmark/programs/BenchmarkUtils.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in 2299580e6.

mlir/benchmark/programs/BenchmarkUtils.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in 9d218c6bc.

mlir/benchmark/Generate.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a4457437b.

mlir/benchmark/jeff/MQTJeffBenchmarks.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a4457437b.

mlir/include/mlir/Benchmark/BenchmarkUtils.h
denialhaag: 🤖 *AI text below* 🤖

Addressed in 6310621bb.

mlir/include/mlir/Benchmark/Generate.h
denialhaag: 🤖 *AI text below* 🤖

Addressed in f5375d4f4.

mlir/unittests/benchmark/BenchmarkTestUtils.h
denialhaag: 🤖 *AI text below* 🤖

Addressed in b0f6c6005.

mlir/unittests/benchmark/test_benchmark_jeff.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in dc7a81dc1.

mlir/benchmark/programs/GHZLinear.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in 2e32e8e43.

mlir/benchmark/programs/IQFT.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in 2e32e8e43.

mlir/benchmark/programs/QPE.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in 2e32e8e43.

mlir/benchmark/programs/Multiplexer.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in 78186417e.

denialhaag:
@coderabbitai review

mlir/benchmark/programs/Multiplexer.cpp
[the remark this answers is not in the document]
denialhaag: Can we resolve this?

mlir/benchmark/Compile.cpp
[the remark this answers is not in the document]
denialhaag: The test has moved in https://github.com/munich-quantum-toolkit/core/pull/2135/commits/dc7a81dc1b670cecf89566733bf8a330c44cc032, but the fix is in. Can you confirm?

mlir/benchmark/programs/QPE.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  b.scfFor(0, counting, 1, [&](Value i) {
+    auto lower = arith::AddIOp::create(b, i, one);
+    phaseRotationLoop(b, lower, upper, -std::numbers::pi / 2.0, 0.5,
+                      [&](Value angle, Value j) {
+                        b.cp(angle, b.loadQubit(q.value, j),
+                             b.loadQubit(q.value, i));
+                      });
+    b.h(b.loadQubit(q.value, i));
```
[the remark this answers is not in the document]
denialhaag: iQFT stands for iterative QFT, _not_ inverse QFT. This should be fine as is.

mlir/benchmark/programs/IQFT.cpp
```diff
@@ -0,0 +1,59 @@
...
+  auto last = b.indexConstant(bits - 1);
+
+  b.scfFor(0, bits, 1, [&](Value i) {
+    auto offset = arith::SubIOp::create(b, total, i);
+
+    phaseRotationLoop(b, zero, i, std::numbers::pi / 2.0, 0.5,
+                      [&](Value angle, Value step) {
+                        auto index = arith::AddIOp::create(b, offset, step);
+                        b.scfIf(res, index, [&] { b.p(angle, q); });
+                      });
```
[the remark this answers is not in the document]
denialhaag: iQFT stands for iterative QFT, not inverse QFT. This should be fine as is.

mlir/benchmark/programs/Registry.cpp
```diff
@@ -0,0 +1,36 @@
...
+      {.name = "ghz-star", .build = &ghzStar, .minimumSize = 1},
+      {.name = "grover", .build = &grover, .minimumSize = 2},
+      {.name = "iqft", .build = &iqft, .minimumSize = 1},
+      {.name = "iqpe", .build = &iqpe, .minimumSize = 1},
+      // The loop runs over the 2^(n-1) control states, so a larger size cannot
+      // hold that count in a signed 64-bit integer.
+      {.name = "multiplexer",
+       .build = &multiplexer,
+       .minimumSize = 2,
+       .maximumSize = 63},
```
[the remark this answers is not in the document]
denialhaag: Fixed in https://github.com/munich-quantum-toolkit/core/pull/2135/commits/c8ea84fb7b43dd06a5fa0c3126b6ca287a207da6. Please confirm.

mlir/benchmark/programs/BenchmarkUtils.cpp
```diff
@@ -0,0 +1,84 @@
...
+  auto one = b.indexConstant(1);
+
+  auto loop = scf::ForOp::create(b, lower, upper, one, ValueRange{first});
+  OpBuilder::InsertionGuard guard(b);
+  b.setInsertionPointToStart(loop.getBody());
+  auto angle = loop.getRegionIterArg(0);
+  body(angle, loop.getInductionVar());
+  scf::YieldOp::create(b, ValueRange{advance(angle)});
+}
+} // namespace
```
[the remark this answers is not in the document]
denialhaag: Fixed in https://github.com/munich-quantum-toolkit/core/pull/2135/commits/2b888dd364b314fd22ddde8c8e56498e7b68ddf9. Please confirm.

mlir/benchmark/programs/QPE.cpp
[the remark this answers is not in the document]
denialhaag: Sorry, I responded to the wrong comment. This should be fixed by https://github.com/munich-quantum-toolkit/core/pull/2135/commits/16f576368a3bc2b4454beecc0674b98a2efe99c7. Please confirm.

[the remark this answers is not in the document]
denialhaag: That's not true. During the first iteration, the inner loop is trivial, and only `h` is applied.

Also, I won't add a regression test for something we are adding in this PR.

denialhaag:
This should finally be ready for review now. Note that I decided to split the PR; more benchmarks will be added in #2204. All programs added here [already have an OpenQASM counterpart in `jeff`](https://github.com/unitaryfoundation/jeff/tree/benchmarks-dot-jeff/benchmarks/structured). Note that the OpenQASM files differ from those on `main`; all changes are detailed in the description of https://github.com/unitaryfoundation/jeff/pull/96. I hope this split makes this PR relatively easy to review. 😌

burgholzer:
I am taking a look at this now. From a first glance, this may need a bit of work to be ready for merging. I have a couple of ideas how we could turn this into something that is more useful long-term (and may even, at some point, evolve into a successor for MQT Bench; or at least feed it with benchmarks).
Still have to find the best way for how to write and/or code all of that up.

docs/benchmarks.md
```diff
@@ -0,0 +1,218 @@
...
+  number_source_lines: true
+---
+
+# Structured quantum benchmarks
+
+MQT Core defines a benchmark as validated, benchmark-specific parameters plus an
+analytic reference. One instance produces a structured QC program, a resolved
+manifest, and a stable case ID. Each benchmark returns one classical register
+named `result`. Outcome strings are big-endian: the highest-index result bit is
+the leftmost character.
```
DRovara: I feel like this is a bit difficult to understand, especially considering the loaded term "benchmark".

If I understand it correctly, how about a rewording like:

```
Benchmarks in MQT Core are defines as a tuple of:
- benchmark-specific parameters
- an analytic reference
They allow the construction of:
- A structured QC program
- a resolved manifest
- a stable case ID

Resulting programs return one classical register named [...]
```

For instance, from the current description, I still wouldn't understand that we are talking about specific programs here. And since your first sentence defines a benchmark as parameters + analytic reference, the sentence "each benchmark returns [...]" feels a bit weird - how can parameters + analytic reference even return anything?

docs/benchmarks.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+catalog = run_bench("list")
+[entry["id"] for entry in catalog["benchmarks"]]
+```
+
+```{code-cell} ipython3
+qft_schema = run_bench("describe", "qft")
+qft_schema["properties"]["parameters"]
+```
```
DRovara, line 28: Isn't it a bit weird to illustrate example usages of a CLI tool with python code and `subprocess`? This makes it seem like subprocess is the exected way to run it...

include/mqt-core/bench/QFT.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "bench/mqt_core_bench_export.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <string_view>
+
+namespace mqt::bench {
+
+/// Circuit method used for the quantum Fourier transform.
+enum class QFTMethod : uint8_t { Standard, Semiclassical };
```
DRovara: I just realised that (I think) there is no documentation anywhere on what these options mean. Here, I didn't quite get what "Semiclassical" means. Is it the same iterative approach as the iQPE?

mlir/bench/programs/BV.cpp
```diff
@@ -0,0 +1,89 @@
...
+                     .getResult();
+      builder.scfIf(bit, [&] { builder.cz(query, flag); });
+      builder.h(query);
+      builder.measure(query, result, index);
+      builder.reset(query);
+    });
+    return {result};
+  }
+
+  auto query = builder.allocQubitRegisterStorage(width, "query");
```
DRovara, line 69: This is just a quick random thought but would it make sense/how much effort would it be to also provide an option that can be used for all benchmark programs that replaces any and all qubit registers with multiple individually allocated single qubits?

DRovara (commented):
This is really cool! I didn't expect to find an entire benchmark suite with actual evaluations here when I opened the PR but this is pretty nice.

In fact, looking ahead, I can really imagine adding further custom benchmarks to this to e.g. create a reproduction environment for paper results, but that's beyond the scope for now.

I guess the one disadvantage I see here is that it seems like adding another benchmark program requires quite some effort, but I guess there could be worse things.

The programs themselves look correct, I didn't notice any issues there.

docs/benchmarks.md
burgholzer: Yeah. I like the direction of your suggestion! I'll take that into the next iteration!

docs/benchmarks.md
burgholzer: Yeah. It certainly is; especially since there is Python code for doing this.
I suppose this was mostly about showing how the CLI can be used, but there are better ways to do that in Jupyter Notebooks. I'll revisit this.

include/mqt-core/bench/QFT.hpp
burgholzer: No, these should probably be documented inline with Doxygen comments.

Same idea as iQPE. `semiclassical` is just the established name from the literature.

mlir/bench/programs/BV.cpp
burgholzer: Hm. I see the idea. Feels like this would be a bit complicated though. Especially for the algorithms using loops because they currently index into registers in the loops, which is not directly possible if the program just has loose qubits.
At the top of my head, I don't see a very easy solution. But maybe I am also overlooking things.

mlir/bench/programs/Grover.cpp
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+    b.mcz(controls, target);
+    for (size_t i = 0; i < benchmark.qubits(); ++i) {
+      if (options.markedBitstring[benchmark.qubits() - 1 - i] == '0') {
+        b.x(q[i]);
+      }
+    }
+
+    b.scfFor(0, search, 1, [&](Value iv) { b.h(b.loadQubit(q.value, iv)); });
```
MatthiasReumann, line 51: Seeing this, I think it could make sense to have an abstraction that "applies" single qubit gates to registers.

burgholzer: One could even think about allowing registers as arguments of the gates and defining broadcast rules similar to OpenQASM. But I am not quite sure we want this, because it adds a whole bunch of complexity.

What should be easy is adding convenience methods in the builders for easing the construction of the construct here.

MatthiasReumann: I think a modifier along the lines 
```mlir 
%out:9 = qco.broadcast %in:9 (%target) { qco.h(%target) } : (!qco.qubit, ...., !qco.qubit)
%tensor_out = qco.broadcast %tensor_in (%target) { qco.x(%target) } : (!tensor.tensor<!qco.qubit>)
```

could be kind of nice as this doesn't add overhead to the individual gates. But definitely not a must have right now.

burgholzer: There may be a precedent in MLIR for how to handle this. This is kind of like vectorization; and there is a `vector` dialect in MLIR. Might be worth to brainstorm these ideas with an LLM and collect the thoughts in an issue.
This si nothing to be addressed in this PR.

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+        period_exponent=1,
+        method=QFTMethod.SEMICLASSICAL,
+    )
+)
+benchmark.options.method, benchmark.output.name, benchmark.output.width
+```
+
+Python also accepts {py:class}`fractions.Fraction` for QPE phases. It reduces
+the phase modulo one turn before it enters the C++ API. The reduced denominator
+must fit in an unsigned 64-bit integer.
```
denialhaag: This feels a bit out of place as written.

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+import json
+
+
+request = json.loads(benchmark.request_json)
+manifest = json.loads(benchmark.manifest_json)
+request, {
+    "case_id": manifest["case_id"],
+    "outputs": manifest["outputs"],
+    "reference": manifest["reference"],
+}
```
denialhaag: Is there a better way to show this sort of output in the documentation? Right now, this is rendered as a tuple of dictionaries. It's not really straightforward to discern what the `request` looks like. Are `print` statements maybe better?

This applies to the entire page.

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+{
+    "total_variation_distance": evaluation.total_variation_distance,
+    "squared_hellinger_fidelity": evaluation.squared_hellinger_fidelity,
+    "success_probability": evaluation.success_probability,
+}
+```
+
+Total variation distance zero and squared Hellinger fidelity one identify an
+exact distribution. Bernstein--Vazirani and Grover also report the observed
+success probability.
```
denialhaag: The second sentence feels out of place. I would hope that the other benchmarks work as well; I don't think we need to mention this in the documentation.

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+
+Generation returns the same `QCProgram` type as the MLIR compiler bindings. The
+program can enter the normal compiler pipeline.
+
+```{code-cell} ipython3
+program = benchmark.generate()
+assert program.is_valid
+structured_ir = program.ir
+assert program.copy().to_qco().is_valid
+structured_ir
```
denialhaag: Here, we definitely need a `print`. Right now, this is rendered in a single line.

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+Total variation distance zero and squared Hellinger fidelity one identify an
+exact distribution. Bernstein--Vazirani and Grover also report the observed
+success probability.
+
+## Generate structured IR
+
+Generation returns the same `QCProgram` type as the MLIR compiler bindings. The
+program can enter the normal compiler pipeline.
```
denialhaag: Can we maybe link to our MLIR docs here? Maybe also a link to the API docs for `QCProgram`?

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+root = Path(temporary.name)
+request_path = root / "request.json"
+counts_path = root / "counts.json"
+output_directory = root / "generated"
+request_path.write_text(benchmark.request_json, encoding="utf-8")
+counts_path.write_text(
+    json.dumps({"schema_version": 1, "counts": {"000": 5, "100": 5}}),
+    encoding="utf-8",
+)
+```
```
denialhaag, line 153: Can the output cell be hidden?

docs/benchmarks.md
```diff
@@ -0,0 +1,197 @@
...
+program_path.name, manifest_path.name
+```
+
+```{code-cell} ipython3
+!mqt-core-bench evaluate --manifest {manifest_path} --counts {counts_path}
+```
+
+```{code-cell} ipython3
+temporary.cleanup()
+```
```
denialhaag, line 174: Can this cell be hidden altogether? Can the setup of the temporary directory be hidden as well?

include/mqt-core/bench/QFT.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "bench/Evaluation.hpp"
+#include "bench/mqt_core_bench_export.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <string_view>
+
+namespace mqt::bench {
+
+/// Circuit method used for the quantum Fourier transform.
```
denialhaag: Let's keep the docstring consistent and use the shorthand. This also applies to other instances in this file (and potentially more in other files).

```suggestion
/// Circuit method used for the QFT.
```

include/mqt-core/bench/QPE.hpp
```diff
@@ -0,0 +1,78 @@
...
+  [[nodiscard]] uint64_t denominator() const noexcept;
+
+  friend bool operator==(const Phase&, const Phase&) = default;
+
+private:
+  uint64_t numerator_;
+  uint64_t denominator_;
+};
+
+/// Circuit method used for quantum phase estimation.
```
denialhaag: See my comment above.

```suggestion
/// Circuit method used for QPE.
```

mlir/benchmark/programs/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
+# Copyright (c) 2025 - 2026 Munich Quantum Software Company GmbH
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+add_library(MQTBenchmarkPrograms STATIC BV.cpp Fourier.cpp GHZ.cpp Grover.cpp)
```
denialhaag: Would a glob pattern make sense here?

mlir/benchmark/programs/Fourier.cpp
```diff
@@ -0,0 +1,241 @@
...
+phaseRotationLoop(qc::QCProgramBuilder& builder, Value lower, Value upper,
+                  const double start, const double factor,
+                  const function_ref<void(Value angle, Value index)>& body) {
+  auto one = arith::ConstantIndexOp::create(builder, 1).getResult();
+  auto first =
+      arith::ConstantOp::create(builder, builder.getF64FloatAttr(start))
+          .getResult();
+  auto scale =
+      arith::ConstantOp::create(builder, builder.getF64FloatAttr(factor))
+          .getResult();
```
denialhaag: Should we add helper methods to the `QCProgramBuilder`? It already has methods for some constants. It could make things easier to read.

CC @burgholzer

mlir/benchmark/programs/Fourier.cpp
denialhaag, line 1: Hmm. I get the idea behind combining the implementations in this file, but I feel it would also be nice to have one `.cpp` file per benchmark. Opinions?

CC @burgholzer

mlir/benchmark/Generate.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Licensed under the MIT License
+ */
+
+#include "mlir/Bench/Generate.h"
+
+#include "bench/JSON.hpp"
```
denialhaag: A general comment: We've now added the following paths:

- `bindings/bench`
- `include/mqt-core/bench`
- `mlir/benchmark`
- `mlir/include/mlir/Bench`

Can this be streamlined a bit, or are these sufficiently diverse to warrant three different versions of the term "benchmark"? Please keep in mind the added `mqt::bench` namespace as well. Also, check how MLIR benchmarks are handled upstream in the LLVM project repository.

mlir/bench/Generate.cpp
```diff
@@ -0,0 +1,132 @@
...
+std::optional<QCProgram> generate(const BV& benchmark) {
+  return buildProgram("bv", [&](qc::QCProgramBuilder& builder) {
+    return bv(builder, benchmark);
+  });
+}
+
+std::optional<QCProgram> generate(const GHZ& benchmark) {
+  return buildProgram(
+      "ghz", [&](qc::QCProgramBuilder& b) { return ghz(b, benchmark); });
+}
```
denialhaag, line 58: Let's consistently use `b` instead of mixing it with `builder`.

mlir/bench/Generate.cpp
```diff
@@ -0,0 +1,132 @@
...
+template <class Benchmark>
+[[nodiscard]] static std::optional<GeneratedBenchmark>
+generateRequest(const std::string_view id, const Benchmark& benchmark) {
+  auto program = generate(benchmark);
+  if (!program) {
+    return std::nullopt;
+  }
+  return GeneratedBenchmark{std::string(id), caseId(benchmark),
+                            toManifestJSON(benchmark), std::move(*program)};
+}
```
denialhaag, line 84: Maybe this is just me, but I find the term "request" confusing. What is this requesting? It cannot be the benchmark, since it's already being generated here. Is it a request for an eventual simulation? 🤔 

This is likely relevant for the entire PR.

CC @burgholzer

mlir/bench/MQTCoreBench.cpp
```diff
@@ -0,0 +1,323 @@
...
+    llvm::cl::cat(benchmarkOptions), llvm::cl::sub(generateCommand));
+
+static llvm::cl::opt<std::string> manifestInputPath(
+    "manifest", llvm::cl::desc("Benchmark manifest JSON file"),
+    llvm::cl::value_desc("file"), llvm::cl::Required,
+    llvm::cl::cat(benchmarkOptions), llvm::cl::sub(evaluateCommand));
+static llvm::cl::opt<std::string> countsInputPath(
+    "counts", llvm::cl::desc("Counts JSON file, or '-' for standard input"),
+    llvm::cl::value_desc("file|-"), llvm::cl::Required,
+    llvm::cl::cat(benchmarkOptions), llvm::cl::sub(evaluateCommand));
```
denialhaag, line 79: I don't see a pattern behind the spacing between the options. Can you streamline this?

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
denialhaag: If we are already editing it, can we align this to the new docstring style specified in `docs/development.md`? Please do not update any additional docstrings.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -216,9 +217,9 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
    * @return The memref value representing the qubit register
    *
-   * @details The register is tracked for automatic deallocation and remains
-   * intact until an element is loaded. Use `loadQubit` to obtain references at
-   * their points of use.
+   * @details Every allocated qubit is initialized to |0⟩. The register is
+   * tracked for automatic deallocation and remains intact until an element is
+   * loaded. Use `loadQubit` to obtain references at their points of use.
    */
   Value allocQubitRegisterStorage(int64_t size, StringRef name = {});
```
denialhaag: Same here.

mlir/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ endfunction()
 add_subdirectory(include)
 add_subdirectory(lib)
 add_subdirectory(tools)
+add_subdirectory(benchmark)
```
denialhaag: I think these were generally sorted alphabetically before. If the only change is the position of benchmarking-related things, please do adjust it. This applies to the entire PR.

src/bench/EvaluationUtils.hpp
```diff
@@ -0,0 +1,90 @@
...
+    if (successOutcome && outcome == *successOutcome) {
+      successShots = count;
+    }
+  }
+  if (totalShots == 0) {
+    throw std::invalid_argument("total shot count must be positive");
+  }
+
+  /// Extended precision prevents avoidable loss while summing distributions.
+  /// NOLINTBEGIN(google-runtime-float)
```
denialhaag: Shouldn't this be `//` instead of `///`?

src/bench/EvaluationUtils.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  const auto missingIdealMass = std::max(0.L, 1.L - observedIdealMass);
+  const auto totalVariation =
+      std::clamp((observedDistance + missingIdealMass) / 2.L, 0.L, 1.L);
+  const auto fidelity = std::clamp(coefficient * coefficient, 0.L, 1.L);
+  const auto success =
+      successOutcome ? std::optional<double>{static_cast<double>(successShots) /
+                                             static_cast<double>(totalShots)}
+                     : std::nullopt;
+  /// NOLINTEND(google-runtime-float)
```
denialhaag: Same here.

src/bench/GHZ.cpp
```diff
@@ -0,0 +1,71 @@
...
+  if (options_.qubits == 0 || options_.qubits > GHZOptions::MAX_QUBITS) {
+    throw std::invalid_argument("GHZ qubits must be between 1 and 1000000");
+  }
+  if (options_.topology != GHZTopology::Linear &&
+      options_.topology != GHZTopology::Star) {
+    throw std::invalid_argument("unknown GHZ topology");
+  }
+  if (options_.basis != GHZBasis::Z && options_.basis != GHZBasis::X) {
+    throw std::invalid_argument("unknown GHZ measurement basis");
+  }
```
denialhaag, line 34: Not super important, but isn't this guaranteed by construction? Please also check the other benchmarks if you do agree.

src/bench/Grover.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <numbers>
+#include <stdexcept>
+#include <string_view>
+#include <utility>
+
+namespace mqt::bench {
+namespace {
+
+/// Extended precision keeps the analytic reference stable near Grover peaks.
+/// NOLINTBEGIN(google-runtime-float)
```
denialhaag: Here and across the file, shouldn't this be `//` instead of `///`? Please check for other occurrences in this PR.

src/bench/SHA256.hpp
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+#pragma once
+
+#include <string>
+#include <string_view>
+
+namespace mqt::bench::detail {
+
+[[nodiscard]] std::string sha256Hex(std::string_view input);
```
denialhaag, line 19: Maybe this could use a docstring to explain why we need it. 🤔

denialhaag (commented):
Thanks a lot for all the improvements, @burgholzer! I really like this a lot! 🙂 


I went through the PR pretty carefully now and left a comment wherever I stumbled. I'll feed these comments into Codex in a bit. I'm not sure about some of my comments, which is why I sometimes explicitly tagged you.

docs/benchmarks.md
burgholzer: Certainly can. Myst-NB has all the means to do so 🙌

mlir/benchmark/programs/Fourier.cpp
burgholzer: Yeah. If it helps to reduce boilerplate code that we have to write often it feels right to add a few utilities to the builder 👍🏻

mlir/benchmark/programs/Fourier.cpp
burgholzer: Yeah. I'd argue it's code duplication worth living with 👍🏻 let's keep benchmarks self contained

mlir/benchmark/Generate.cpp
burgholzer: Strongly in favor of consistency.
Maybe lower case `bench` everywhere would be most fitting

mlir/bench/Generate.cpp
burgholzer: I see the point. Also not fully happy. Is it a "specification"? An "instance"/"instantiation"? Open for better suggestions.

burgholzer:
> Thanks a lot for all the improvements, @burgholzer! I really like this a lot! 🙂 
> 
> 
> I went through the PR pretty carefully now and left a comment wherever I stumbled. I'll feed these comments in a bit. I'm not sure about some of my comments, which is why I explicitly tagged you sometimes. 

Nice! I think I commented on every tag. Let me know if I missed something!

mlir/bench/Generate.cpp
denialhaag: Funnily enough, GPT-5.6 suggested "instance specification," which is maybe a bit too much. I'm leaning toward "instance" right now and would go with that. We can leave this thread open so that we can revisit the topic once everything has been renamed. 😌

mlir/benchmark/Generate.cpp
denialhaag: Let's go with lowercase `bench` then! 😎

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The paragraph now presents `Fraction` as an example of QPE's family-specific typed validation, so it fits the surrounding narrative.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The notebook now uses labeled `print` calls and indented JSON throughout instead of raw tuple and dictionary representations.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The text now states precisely that only Bernstein–Vazirani and Grover report a success probability because those families define a distinguished success outcome.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The generated IR is now passed to `print`, which preserves its multiline presentation.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The section now links both the `QCProgram` API reference and the MQT Core MLIR compiler guide.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The file-writing cell now uses the MyST-NB `remove-output` tag.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Addressed in f1507b2a6. The temporary-directory setup and cleanup cells now use `remove-cell`, while the useful CLI workflow remains visible.

include/mqt-core/bench/QFT.hpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. The enum documentation now uses the QFT shorthand consistently.

include/mqt-core/bench/QPE.hpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. The corresponding enum documentation now uses the QPE shorthand.

mlir/benchmark/programs/CMakeLists.txt
denialhaag: 🤖 *AI text below* 🤖

Reviewed while updating this source list in a35fe2fd2, but intentionally not changed to a glob. The explicit list makes source additions visible and avoids relying on glob-driven CMake reconfiguration.

mlir/benchmark/programs/Fourier.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. `QCProgramBuilder` now provides helpers for recurring floating-point and index constants, and the benchmark emitters use them.

mlir/benchmark/programs/Fourier.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. QFT and QPE now have separate, self-contained `.cpp` files, accepting the small amount of duplication discussed here.

mlir/benchmark/Generate.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. The MLIR source, include, and unit-test paths now consistently use lowercase `bench`; the existing `mqt::bench` namespace and public `bench` paths already match.

mlir/bench/Generate.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. The generator now uses `b` consistently instead of mixing `b` and `builder`.

mlir/bench/Generate.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed throughout the PR in cab9140f8. The resolved benchmark configuration is now called an `instance`; it is not a request for a later simulation. The API, CLI, schemas, tests, stubs, documentation, and glossary use the new term.

mlir/bench/MQTCoreBench.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. The command-line option declarations now use consistent spacing.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. This edited declaration now follows the current documentation style, without broadening the change to unrelated docstrings.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. This edited declaration now follows the current documentation style as well; unrelated docstrings remain untouched.

mlir/CMakeLists.txt
denialhaag: 🤖 *AI text below* 🤖

Addressed across the affected CMake files in a35fe2fd2. The new benchmarking entries now appear in their established alphabetical positions.

src/bench/EvaluationUtils.hpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. This implementation annotation now uses `//` instead of the Doxygen marker `///`.

src/bench/EvaluationUtils.hpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. This annotation and the other applicable implementation comments now use `//`.

src/bench/GHZ.cpp
denialhaag: 🤖 *AI text below* 🤖

Reviewed during a35fe2fd2, but intentionally left in place. Although normal construction supplies a valid enumerator, an invalid scoped-enum value can still enter through a cast, deserialization, or a binding boundary. The check and its test protect the public constructor boundary.

src/bench/Grover.cpp
denialhaag: 🤖 *AI text below* 🤖

Addressed across the relevant implementation files in a35fe2fd2. These annotations now use ordinary `//` comments rather than Doxygen comments.

src/bench/SHA256.hpp
denialhaag: 🤖 *AI text below* 🤖

Addressed in a35fe2fd2. The declaration now explains that this private SHA-256 helper provides stable benchmark case identifiers without adding an external cryptographic dependency.

test/python/test_cli.py
```diff
@@ -52,6 +54,34 @@ def test_cli_version(script_runner: ScriptRunner) -> None:
...
+        patch.object(benchmark_cli.sys, "argv", ["mqt-core-bench", "list"]),
+        patch.object(benchmark_cli, "distribution") as distribution_mock,
+        patch.object(benchmark_cli.os, "execv") as execv_mock,
+    ):
+        distribution_mock.return_value.locate_file.return_value = executable
+        benchmark_cli.main()
+
+    distribution_mock.assert_called_once_with("mqt-core")
+    distribution_mock.return_value.locate_file.assert_called_once_with(f"mqt/core/bin/mqt-core-bench{suffix}")
+    execv_mock.assert_called_once_with(executable, [str(executable), "list"])
```
denialhaag: These two tests separate `mqt-core-cli` tests. Can we move them to the end?

include/mqt-core/bench/QFT.hpp
denialhaag: This comment was meant to apply to the entire file. Please check again.

include/mqt-core/bench/QPE.hpp
denialhaag: This comment was meant to apply to the entire file. Please check again.

mlir/benchmark/programs/Fourier.cpp
denialhaag: Maybe it's a good opportunity to align the docstrings of the four `ConstOp` helpers in `QCProgramBuilder.h`.

docs/benchmarks.md
denialhaag: This still feels overly specific for a generic page, but I shall be fine.

docs/benchmarks.md
denialhaag: Can you make the statement a bit less strong? It's very specific and subject to change whenever we add an applicable benchmark. I don't want this of all things to become a maintenance burden.

denialhaag (commented):
Another set of comments:

test/python/test_cli.py
denialhaag: 🤖 *AI text below* 🤖

Moved both benchmark CLI tests to the end of the file so they no longer split the mqt-core-cli tests. Addressed in a641c1b4e.

include/mqt-core/bench/QFT.hpp
denialhaag: 🤖 *AI text below* 🤖

Checked the complete QFT header. The remaining expanded algorithm name now uses QFT, and the other algorithm references already use the shorthand. Addressed in a641c1b4e.

include/mqt-core/bench/QPE.hpp
denialhaag: 🤖 *AI text below* 🤖

Checked the complete QPE header. The remaining expanded algorithm name now uses QPE, and the other algorithm references already use the shorthand. Addressed in a641c1b4e.

mlir/benchmark/programs/Fourier.cpp
denialhaag: 🤖 *AI text below* 🤖

Aligned the docstrings of all four constant helpers in QCProgramBuilder.h. They now use concise /// comments and state each result type. Addressed in a641c1b4e.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Agreed. Removed the QPE-specific phase and denominator details. The generic page now only states that each family validates its options when it creates an instance. Addressed in a641c1b4e.

docs/benchmarks.md
denialhaag: 🤖 *AI text below* 🤖

Made this future-proof: the text now says that some benchmark families report a success probability for a distinguished outcome, without listing the current families. Addressed in a641c1b4e.

denialhaag:
> I think I commented on every tag. Let me know if I missed something!

Thanks a lot for the quick input, @burgholzer! 🙂 

I might have gotten a bit lost in the details here, but I think I'm happy with this now. I'm kind of indifferent about adding more benchmarks now or in one or more follow-ups. The latter would certainly make reviewing easier, but only if we leave the infrastructure (largely) alone. Either way, we should add most (if not all) structured benchmarks mentioned in the `jeff` repository sooner rather than later. 🤔

mlir/bench/Generate.cpp
burgholzer: Haven't looked through the code in detail yet, but "instance" is definitely already a step in the right direction. I don't really think `InstanceSpecification` is too bad; it's at least expressive. If one wouldn't care about length (maybe we should not), `BenchmarkInstanceSpecification` would be fairly appropriate in my opinion. It very much says what this is describing.

burgholzer:
I'll try to get a review in for this tomorrow, so that we can get this merged and not hold it up for too long. Follow-ups should come quite naturally.

mlir/bench/Generate.cpp
denialhaag: You have a point! Unfortunately, I don't think I'll manage to iterate on this further today, so I'll leave the PR to you again. I'm happy to jump in again tomorrow evening if ther should more to do! Either way, we can go with `BenchmarkInstanceSpecification` (and "instance specification" in prose) from my side. 😌

denialhaag: I briefly did the renaming in https://github.com/munich-quantum-toolkit/core/pull/2135/changes/5ddae1baad906e56c43af8983df925bbe762d8f3.

python/mqt/core/bench.pyi
burgholzer, line 1: Would it make sense to separate the individual benchmarks into individual submodules so that this is not as cluttered? With 20+ benchmarks, this will not look pretty.
Also mirrors a little closer how we handle this in MQT Bench.
The main module could hold the generic helpers and utility functions while the individual submodules hold the code for a particular benchmark.
Something like `mqt.core.bench.benchmarks.<...>` could make sense. Maybe that is one level too deep though.
I am just looking for a little bit more structure here. Throwing this into Codex with Plan mode is likely gonna yield a satisfactory result.

burgholzer (approved):
Thanks @denialhaag 🙏🏼 I think we almost have it now.
I just have one comment left besides the obvious conflict with `main` at the moment.
I am already approving here as I am sure any resolution to the last open comment will be fine. Feel free to proceed and subsequently merge 🚀

python/mqt/core/bench.pyi
denialhaag: Very good point! Done in https://github.com/munich-quantum-toolkit/core/pull/2135/commits/63dffbff1a429647858ff6055fc5d9b2e3832ac9. I went with `mqt.core.bench.qft` instead of `mqt.core.bench.benchmarks.qft`.

denialhaag:
> I just have one comment left besides the obvious conflict with `main` at the moment.

Thanks a lot for the additional feedback! Both of the points are addressed now, and I also updated the changelog and upgrade guide in https://github.com/munich-quantum-toolkit/core/pull/2135/commits/b3b982845bad51548fb8a97693a73b28c04261d7.

> I am already approving here as I am sure any resolution to the last open comment will be fine. Feel free to proceed and subsequently merge 🚀 

I already set this to auto-merge. I'm not a huge fan of one of the Python tests, but I'll remove that in one of the follow-up PRs that add the additional benchmarks. 😌

denialhaag:
@coderabbitai approve

PR #2147, opened 2026-08-18
🧪 Apply the PennyLane plugin SpecAudit

marcelwa:
🤖 *AI text below* 🤖

## Description

This applies the repository's first SpecAudit, [`.agent/audits/pennylane-plugin.md`](https://github.com/munich-quantum-toolkit/core/blob/main/.agent/audits/pennylane-plugin.md), which landed with the method itself in `#2124` and has not been acted on since. The audit is the ledger; this is what acting on it looks like.

All six verdicts are applied, along with the four cleanups the audit found along the way. Every `file:line` the audit cites was re-read at `3354fdaab` and still resolved to the line it names, so the drift since its `18b619af5` baseline invalidated nothing.

### Verdicts

| #   | Verdict                                            | Outcome  | Result                                    |
| :-- | :------------------------------------------------- | :------- | :---------------------------------------- |
| 1   | Parameter rendering pinned to seventeen digits     | Applied  | Readable OpenQASM at identical precision  |
| 2   | The format rule implemented twice                  | Applied  | One implementation, one wording           |
| 3   | Histogram expansion order asserted, never promised | Applied  | Multiplicity asserted, order freed        |
| 4   | A defence the experiment refuted                   | Narrowed | Assertion narrowed; no code change earned |
| 5   | An assertion that cannot fail for its own purpose  | Applied  | Accumulated value asserted exactly        |
| 6   | A public function that exists for its tests        | Applied  | Advertised gate set read once per session |

### What changed, in order

Each verdict is two commits, per guardrail 7 of [`.agent/AUDITS.md`](https://github.com/munich-quantum-toolkit/core/blob/main/.agent/AUDITS.md): the assertion change first, then the code change it unblocked, so a bisection can tell which one broke.

**Verdict 1.** Three assertions pinned `rxx(0.10000000000000001)` and its siblings. Nothing published states how a parameter is rendered; the stated intent is that no precision is lost. The tests now parse the literals and compare them as exact floats, and `_format_parameter`'s `.17g` becomes `repr`, which gives the shortest literal that reads back as the same double. Every OpenQASM program this plugin sends to every device becomes readable.

**Verdict 4.** `count("ry(-1.5707963267948966) q[0];") == 1` was believed to be the only guard against a double diagonalizing rotation. Re-running the audit's probe at this commit confirms its finding: `rotations=True` fails no test, because `measurements_from_samples` has already replaced the observable by the time a tape reaches the QASM2 serializer. The assertion is narrowed to the claim it can still make. `rotations=False` stays and is protected by nothing; the reconciled audit records that rather than papering over it.

**Verdict 5.** `execution_time >= 0.0` could not fail for its own purpose: delete the accumulation and the attribute stays `0.0`. A new test freezes the clock over a three-job shot vector and asserts the exact total. Replacing the accumulation with `+= 0.0` now fails it.

**Verdicts 2 and 6**, the largest change and the reason for the diff size. The three-argument `convert_program(tape, device, wires)` was public only because `test_converter.py` called it, and that signature was what stopped the device from handing down state it had already computed. `supports_operation` is PennyLane's per-operation `stopping_condition`, so it re-read the whole advertised gate set on every gate.

Measured on a 100-gate circuit over an 18-operation device, instrumenting the QDMI boundary:

|                                               | `Device.operations()` | `Operation.name()` |
| :-------------------------------------------- | --------------------: | -----------------: |
| Before, per preprocessing and conversion pass |                   101 |               1818 |
| After, once per opened device session         |                     1 |                 18 |

`test_converter.py` is rewritten against `QDMIDevice` and `StubDevice.submissions`, so every test now asserts what a device actually receives. `ProgramConverter` then binds the conversion to one opened session, `converter._preferred_format` (a second implementation of the format rule, with a second wording of its error) is deleted, and `convert_program` and `ConvertedProgram` leave the public surface. `UPGRADING.md` covers the removal.

### Coverage

Guardrail 9 asks for this plainly. Two branches lost their old callers when the tests moved to the device boundary: the OpenQASM 3 and OpenQASM 2 "operation not advertised" refusals, which `decompose` now reaches first. Both tests regained them by also executing a tape through `Device.execute` directly, which is the path where conversion, not preprocessing, has to do the refusing.

Codecov then rejected the patch at 92.8% of the diff, which exposed something the audit had not looked for: `_validate_qdmi_contract` reads advertised sites, site pairs, and the device coupling map, and no test exercised any of it beyond one site-pair rejection. Those tests now exist. Plugin coverage goes from 85% to 90%, the converter from 82% to 94%, and every source line this change adds is covered. The one branch that could not be covered, a third program format in `supports`, was unreachable and is gone.

### Reconciliation

`.agent/audits/pennylane-plugin.md` is updated in place, per the Reconciling section of the method: a status column, an applied note under each verdict, the coverage consequence, and dated progress entries. The one open item, probing the exact QASM2 payload assertion against a bumped PennyLane, stays open.

Two facts the reconciliation records rather than hides. Verdict 4 is narrowed, not applied. And verdict 6 removed round trips, not lines: the source lost 37 lines while the test tree gained 57, because driving conversion through the device costs QNode boilerplate that a direct converter call did not.

### A note on granularity

`AGENTS.md` asks for one pull request per verdict. This one bundles all six deliberately, as the first end-to-end demonstration of the method; the per-verdict commit pairs preserve the bisection property the separate pull requests would have given. Verdicts 2 and 6 could not have been split anyway: verdict 2 alone changes `convert_program`'s public signature, so it would have shipped a breaking change to an API that verdict 6 removes.

## Checklist

- [x] Every agent-authored or agent-edited public text body begins with the visible disclosure `🤖 *AI text below* 🤖`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,11 +137,9 @@ releases may include breaking changes.
 
 ### Changed
 
-- 🧪 Apply the first SpecAudit to the PennyLane plugin. Tests that pinned
-  unspecified behavior are narrowed or replaced, the emitted OpenQASM parameter
-  literals become readable at identical precision, and program conversion reads
-  the advertised QDMI gate set once per device session instead of once per gate
-  ([#2147]) ([**@marcelwa**])
+- 🧪 Apply the first SpecAudit to the PennyLane plugin, narrowing tests that
+  pinned unspecified behavior and reading the advertised QDMI gate set once per
+  device session ([#2147]) ([**@marcelwa**])
 - 💥 Prune dead and misleading CoreIR APIs, including renaming the non-garbage
   logical output count to `getNoutputQubits()` and `num_output_qubits` ([#2112])
   ([**@simon1hofmann**])
@@ -180,8 +178,8 @@ releases may include breaking changes.
   `dd/Approximation.hpp` header, `dd::ApproximationMetadata`, and
   `dd::approximate`. No replacement is provided ([#2154]) ([**@burgholzer**])
 - 💥 Remove `convert_program` and `ConvertedProgram` from
-  `mqt.core.plugins.pennylane`. The program-conversion surface behind
-  `QDMIDevice` is now private ([#2147]) ([**@marcelwa**])
+  `mqt.core.plugins.pennylane`, making program conversion private behind
+  `QDMIDevice` ([#2147]) ([**@marcelwa**])
 - 💥 Remove `nlohmann_json` from the public package contract. MQT Core no longer
   installs or exports the library, no installed header exposes a `nlohmann`
   type, and the decision-diagram statistics report through strings and streams
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,8 @@ releases may include breaking changes.
   and Python FoMaC APIs, and expose optional device queue length and job queue
   position ([#2008], [#2010]) ([**@burgholzer**])
 - 🐍 Start building CPython 3.15 wheels ([#2011]) ([**@denialhaag**])
-- ✨ Add PennyLane support for gate-based QDMI devices ([#2005], [#2147]) ([**@burgholzer**],
-  [**@marcelwa**])
+- ✨ Add PennyLane support for gate-based QDMI devices ([#2005], [#2147])
+  ([**@burgholzer**], [**@marcelwa**])
 - ✨ Integrate QDMI devices as MLIR compiler targets across C++, Python, and
   `mqt-cc` ([#1687]) ([**@MatthiasReumann**], [**@burgholzer**])
 - ✨ Add structured OpenQASM emission from the QC dialect to the C++ and Python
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -34,8 +34,9 @@ releases may include breaking changes.
   and Python FoMaC APIs, and expose optional device queue length and job queue
   position ([#2008], [#2010]) ([**@burgholzer**])
 - 🐍 Start building CPython 3.15 wheels ([#2011]) ([**@denialhaag**])
-- ✨ Add PennyLane support for gate-based QDMI devices ([#2005])
-  ([**@burgholzer**])
+- ✨ Add PennyLane support for gate-based QDMI devices and read each device's
+  advertised gate set once per session ([#2005], [#2147]) ([**@burgholzer**],
```
```suggestion
- ✨ Add PennyLane support for gate-based QDMI devices ([#2005], [#2147]) ([**@burgholzer**],
```

burgholzer (approved):
Thanks @marcelwa 🙏🏼 This is great. I iterated a bit on the code. But this can happily go in now.

A takeaway for the future: We should get better at instructing our clankers to properly manage the changelog and upgrade guide.

PR #2137, opened 2026-08-17
💥 Remove the neutral-atom stack

denialhaag:
## Description

🤖 *AI text below* 🤖

MQT Core no longer contains any neutral-atom functionality. [MQT QMAP](https://github.com/munich-quantum-toolkit/qmap) is now its sole owner and absorbs the transferred code in munich-quantum-toolkit/qmap#1111.

This removes the neutral-atom computation model (`MQT::CoreNA`), the neutral-atom FoMaC device session, the neutral-atom QDMI device together with its runtime configuration parser and device schema, the `mqt.core.na` Python module, and `AodOperation`. It also drops the `Move`, `Bridge`, `AodActivate`, `AodDeactivate`, and `AodMove` operation kinds from the closed `qc::OpType` enum, along with the OpenQASM names they dumped and parsed, `QuantumComputation::bridge`, and the `BUILD_MQT_CORE_QDMI_NA_DEVICE` option. The bundled `mqt.na.default` QDMI device is no longer registered. `UPGRADING.md` lists every removed name.

Several tests used the neutral-atom device merely as a stand-in for some device rather than for anything neutral-atom-specific. The QDMI driver, device-ownership, and Python QDMI tests now use the superconducting device instead. Because the superconducting device is a circuit-model device, the ownership test that asserted a site is a zone now asserts the opposite.

Three tests covered the rejection of zone-model devices at the circuit-target boundary: `CompilerQDMIAdapterTest.RejectsNeutralAtomZoneModels`, `test_compiler_target_rejects_qdmi_zone_model`, and `test_zoned_operation_rejected_at_backend_init`. The rejection logic stays and still guards external zone-model devices, but MQT Core no longer bundles a device that exercises it, so these tests are removed. MQT QMAP is the natural home for them once it owns the device.

Fixes #2092

## AI notice

This PR and its contents were created with the assistance of Opus 5 via Claude Code.

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

[...]

PR #2148, opened 2026-08-18
🐛 Let the QDMI client trigger a calibration run

marcelwa:
🤖 *AI text below* 🤖

> **Stacked on #2114.** The base of this PR is `agent/2094-move-iqm-conversion`, so the diff shown here is only this change. Merge #2114 first.

## Description

Follow-up to the discussion in #2114, where reviewing the program-format classification turned up a bug behind it.

`Device::submitJob` rejected `CALIBRATION` and `BATCH_JOB` together, under one predicate that read as "carries no program payload":

```cpp
if (hasNoGenericProgramPayload(format)) {
  throw std::invalid_argument(
      "Calibration and batch jobs do not use a generic program payload");
}
```

They are not the same case, and the calibration half was wrong.

### Calibration

QDMI declares the format as `` `void*` A calibration program `` and says of it:

> Triggering a calibration run **does not require** a program to be set via `QDMI_DEVICE_JOB_PARAMETER_PROGRAM`.

That is the explicit exception to `QDMI_DEVICE_JOB_PARAMETER_PROGRAM` being "required" — the payload is *optional*, not absent, and QDMI-on-IQM sends one. MQT Core rejected the format before any device call, so a calibration run could not be started at all, with or without a payload. Meanwhile `QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION` is exposed as `Device::getNeedsCalibration()` and `needs_calibration()`: we told a caller a device needs calibration and gave them no way to act on it, even though the spec says exactly how.

This adds a dedicated entry point:

```python
device.submit_calibration_job()                 # no payload
device.submit_calibration_job("configuration")  # text payload
device.submit_calibration_job(b"\x01\x02")      # exact bytes
```

`Device::submitCalibrationJob` in C++, with `std::optional<std::span<const std::byte>>` and a `std::string` convenience overload. It sets no shot count, because a calibration run executes no circuit. Its own entry point rather than a special case in `submitJob`, since the payload is optional and the shot count does not apply — `submitJob` now points there instead of failing silently on intent.

### Batch jobs

A different problem. The spec requires the program for a batch job, but types it as a `QDMI_Job*` list — an array of live job handles. `submitJob` takes `std::span<const std::byte>`, so it structurally cannot express one, whatever the check says. MQT Core has no support for batch jobs and no device currently implements the feature; per the discussion, it was added for the multi-chip neutral-atom demonstrator, whose stack left MQT Core in #2137.

So this states the limitation instead of describing it as a missing payload, and leaves the door open:

```text
MQT Core does not support batch jobs. A batch job's program is a list of job
handles, which this API cannot express
```

## Validation

- `./build/release/test/qdmi/mqt-core-qdmi-test` — 277 passed
- `uv run --no-sync pytest test/python` — 626 passed, 4 skipped
- `uvx nox -s stubs` — the stub diff contains only `submit_calibration_job`
- `uvx nox -s lint` — passes, full `prek` hook set

The bundled DD simulator and SC device both report `NEEDSCALIBRATION = 0` and decline the format themselves, so the new tests assert what is actually verifiable here: the client no longer refuses before asking, and the failure arrives as a device error rather than an argument error. A device that supports calibration is needed to exercise the happy path.

## Checklist

AI assistance: written by Claude Opus 5 via Claude Code, under my direction. The last checkbox is left for me to tick after my own review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ releases may include breaking changes.
   ([**@marcelwa**])
 - ✨ Add `mqt.core.qdmi.is_binary_program_format`, which states whether a
   program format requires exact-byte submission ([#2114]) ([**@marcelwa**])
+- ✨ Add `Device::submitCalibrationJob` and
+  `mqt.core.qdmi.Device.submit_calibration_job` for triggering a calibration run
+  ([#2140]) ([**@marcelwa**])
 - ✨ Add SpecAudits, a method and probe script for auditing tests that pin
   behavior the project never specified ([#2124]) ([**@marcelwa**])
 - ✨ Extract the existing QCO dead-gate elimination patterns from
@@ -164,6 +167,8 @@ releases may include breaking changes.
 
 ### Removed
 
+- 💥 Remove batch job submission from the QDMI client. `Device::submitJob` now
+  states that MQT Core does not support batch jobs ([#2140]) ([**@marcelwa**])
 - 💥 Remove the IQM JSON converter `qiskit_to_iqm_json` and the `MoveGate` from
   the Qiskit plugin, which [QDMI-on-IQM] now owns ([#2114]) ([**@marcelwa**])
 - 💥 Remove the unused decision-diagram approximation algorithm, including the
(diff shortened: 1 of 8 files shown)


burgholzer (approved):
This also feels ready. Thanks for the fix @marcelwa 🙏🏼 
Let's get this in 🚀

PR #2149, opened 2026-08-18
✨ Add inspection methods to `QCProgram`

denialhaag:
## Description

🤖 *AI text below* 🤖

This PR adds inspection methods to `QCProgram` that are all exposed to Python.

We need the functionality of counting two-qubit gates to benchmark `mqt-cc` against other compilers in [`ucc-bench`](https://github.com/unitaryfoundation/ucc-bench).

The methods count static gate occurrences in the entry-point IR. They count gates in each structured control-flow region once, regardless of how often the region executes. Operations within gate modifiers are not counted recursively, and barriers are skipped.

## AI notice

This PR and its contents were created with the assistance of Opus 5 via Claude Code and GPT-5.6 Sol via Codex.

## Checklist


denialhaag:
While working on this, I noticed that we have `QCProgram.from_qasm_str()`, `QCProgram.from_qasm_file()`, and `QCProgram.to_openqasm3()`. We should probably align those a bit.

mlir/lib/Compiler/Programs.cpp
```diff
@@ -370,6 +372,30 @@ std::optional<QIRProgram> QCProgram::intoQIR(const QIRProfile profile) && {
...
+      if (auto ctrl = dyn_cast<qc::CtrlOp>(parent)) {
+        numQubits += ctrl.getNumControls();
+      } else if (!isa<qc::InvOp, qc::PowOp>(parent)) {
+        break;
+      }
+    }
+    if (numQubits == 2) {
+      ++count;
+    }
+  });
```
burgholzer: Maybe I am too naive here, but couldn't this entirely rely on `op.getNumQubits()` and simply traverse the UnitaryOpInterface operations in the module once (without recursing into modifiers)?
Feels quite a bit simpler. 
This would count modifiers with more than one containing operation as a single gate, but I feel that is generally fair. And towards the hardware side, this is irrelevant anyway as the hardware will not have the concept of a modifier.

I was also wondering whether we should be walking the entry point(s) instead of the entire module.

(I'd still skip Barrier ops)

burgholzer (commented):
Just one fairly general observation while looking at this.

burgholzer:
Follow-up thought from above: we likely also want further convenience functions for the programs
- single-qubit gates
- gates
- depth (?)
- gate counts (like Qiskit's operation dictionary that carries the frequency of particular operations)

For straight-line programs, these are easy to compute. As soon as SCF is involved, it becomes much harder. Especially with loops. Should still be possible though.
The gate dictionary is a bit tough, because of how one handles/tracks modifiers.

mlir/lib/Compiler/Programs.cpp
denialhaag: No, I think that's fair. https://github.com/munich-quantum-toolkit/core/pull/2149/changes/b959f56aba9a9d85df91d1b81e0bb95c17319614 should address all of the points you're raising. 😌

denialhaag: The handling of SCF operations is still outstanding, but I don't think that should be impossibly difficult. 🤔

denialhaag:
> Follow-up thought from above: we likely also want further convenience functions for the programs
> 
> * single-qubit gates
> * gates
> * depth (?)
> * gate counts (like Qiskit's operation dictionary that carries the frequency of particular operations)
> 
> For straight-line programs, these are easy to compute. As soon as SCF is involved, it becomes much harder. Especially with loops. Should still be possible though.
> The gate dictionary is a bit tough, because of how one handles/tracks modifiers.

Good points! Progress:

- [x] Handle SCF operations
- [x] Number of gates
- [x] Number of single-qubit gates
- [x] ~~Depth~~ (postponed to [#2260](https://github.com/munich-quantum-toolkit/core/pull/2260))
- [x] ~~Gate counts~~ (postponed to [#2260](https://github.com/munich-quantum-toolkit/core/pull/2260))

denialhaag:
🤖 *AI text below* 🤖

The gate-counting methods report static gate counts for the entry-point IR. They count gates in every SCF region once: both branches of `scf.if`, every `scf.index_switch` case, and all regions of `scf.for` and `scf.while`, regardless of the runtime path or iteration count. This matches static instruction counting in classical compilers. A dynamic gate count depends on runtime branches and loop iterations, so it would be a separate resource-estimation metric based on known bounds, runtime inputs, or profile data.

mlir/lib/Compiler/Programs.cpp
denialhaag: > The handling of SCF operations is still outstanding, but I don't think that should be impossibly difficult. 🤔 

Done in https://github.com/munich-quantum-toolkit/core/pull/2149/changes/ca64202db3cb5035eef881b802ca57ea1ad5fa28 and documented in https://github.com/munich-quantum-toolkit/core/pull/2149/commits/6bdbeea8647712545f9e4a10fc04c7e77f4ac5ec. See also https://github.com/munich-quantum-toolkit/core/pull/2149#issuecomment-5418636964.

denialhaag:
@burgholzer, apart from the missing changelog entry, this should be ready now. I postponed the other two methods to a follow-up PR in case they require some more iteration.

To save CI resources, I'm intentionally leaving the PR as a draft until we actually want to merge it.

burgholzer (approved):
Pushed two simplification commits. Otherwise I think this can go in!
A thought that came to mind, but is not important for this PR is that we likely want to also introduce such functionality for QCO to inform some passes with quality metrics. That's a future topic though.

PR #2158, opened 2026-08-18
♻️ Replace implicit classical-register memrefs with CBit IR

### LINKED ISSUE #2155: ✨ Add a first-class classical-bit register dialect
🤖 *AI text below* 🤖

## Problem Statement

Issue #1123 chose a mixed representation for classical results: measurement operations return SSA `i1` values, while source-level classical registers use `memref<?xi1>` storage. Issue #1726 implemented that choice in the QC and QCO builders, conversions, and QIR lowering. This avoided a new IR concept, but a generic memref does not identify a classical-bit register, its initialization semantics, or whether it is a public result.

The recent Qiskit and OpenQASM integrations expose the cost of reconstructing those semantics:

- #2140 makes Qiskit export recognize a leading sequence of false stores as implicit OpenQASM 2 register initialization. It also ignores unused `ub.poison` values emitted as OpenQASM 3 placeholders.
- QIR lowering independently scans generic memrefs and reconstructs classical-result slots and initialization rules.
- OpenQASM import keeps parallel SSA bit state and output memrefs, including per-element poison placeholders.
- Qiskit import previously omitted Qiskit's required zero initialization. A condition that read a clbit before its first measurement therefore did not preserve Qiskit semantics. #2140 adds the missing initialization through the builder-wide memref policy.

These are not isolated frontend defects. They show that the IR lacks the information needed by its producers and consumers. The compatibility logic merged in #2140 is intentionally temporary. It keeps the unreleased frontends usable while this issue replaces the implicit convention.

## Proposed Solution

Add a neutral `cbit` dialect that QC and QCO share unchanged.

The first version contains:

- `!cbit.reg<N>` for a static, positive-width, non-aliasing classical-bit register.
- `#cbit.init<zero>` and `#cbit.init<undefined>`.
- `cbit.alloc` with an initialization attribute and an optional source name.
- `cbit.load` and `cbit.store`, with `index` indices and `i1` values.
- Explicit allocation, read, and write memory effects.

An undefined element has undefined behavior until a store defines it. Verification rejects constant out-of-bounds indices. The first version has no casts, subviews, aliases, deallocation, or reverse recognition of arbitrary `memref<Nxi1>` values.

Only `!cbit.reg` values returned by the entry function are public result registers. Internal allocations do not become Qiskit or OpenQASM outputs. Program builders accept initialization per allocation and expose register-based load and store operations. Measurement helpers continue to return `i1` and store the result through `cbit.store`.

QC-to-QCO and QCO-to-QC conversions preserve CBit operations, register identity, names, initialization, and returned-output semantics. A dedicated `convert-cbit-to-memref` pass lowers CBit late when a generic memory representation is required:

- Zero initialization becomes an allocation followed by false stores.
- Undefined initialization becomes an allocation only.
- Loads, stores, function signatures, and returns are converted.

### Acceptance criteria

- Qiskit import allocates each circuit classical register with zero initialization, uses CBit loads and stores, and returns every Qiskit-visible register.
- Qiskit export discovers only returned CBit registers. It supports direct static measurement stores, zero-initialized unwritten bits, and fully overwritten undefined registers. It gives precise errors for dynamic destinations, classical loads or control flow, non-measurement stores, duplicate destinations, and returned undefined bits.
- OpenQASM 2 import uses zero initialization. OpenQASM 3 import uses undefined initialization. Both use CBit directly for assignments, structured control flow, and dynamic indices without parallel output memrefs or `ub.poison` placeholders.
- OpenQASM export emits declarations and assignments from CBit operations and returned registers without inspecting generic memrefs.
- QC-to-QCO and QCO-to-QC conversion preserve CBit unchanged.
- QIR lowering handles CBit directly, removes leading-zero-store recognition, maps direct measurement stores to result slots, preserves Base-profile static-index restrictions and Adaptive-profile dynamic-index behavior, and rejects unsupported output stores with precise diagnostics.
- QCO DD execution tracks CBit initialization and register state for indexed loads, stores, conditions, and measurement updates.
- QCO-to-jeff and jeff-to-QCO convert CBit directly and preserve supported register behavior.
- `convert-cbit-to-memref` has focused coverage for initialization, operations, signatures, and returned registers.
- The dialect is registered in compiler programs, tools, bindings, and test contexts. Generated dialect documentation, the MLIR index, conversion documentation, and OpenQASM documentation describe the supported contract.
- Tests cover parsing, printing, verification, memory effects, bounds, invalid types, builder output, frontend round trips, public-output selection, conversion, QIR profiles, DD execution, and jeff.
- Qiskit and QIR contain no leading-zero memref recognizer. Qiskit export has no UB-dialect dependency. OpenQASM emits no classical-output poison placeholder. No in-tree consumer depends on the generic classical-register-name memref convention.

burgholzer:
🤖 *AI text below* 🤖

## Description

This pull request replaces implicit `memref<Nxi1>` classical registers with a neutral CBit dialect shared by QC and QCO. The new IR makes register width, initialization, source name, access, and public return identity explicit.

The implementation adds `!cbit.reg<N>`, `#cbit.init<zero>`, `#cbit.init<undefined>`, `cbit.alloc`, `cbit.load`, and `cbit.store`. QC and QCO builders select initialization per allocation, with unnamed zero-initialized registers as the concise default. QC↔QCO preserves CBit identity, while `convert-cbit-to-memref` provides an explicit late lowering.

Qiskit and OpenQASM import and export, QIR Base and Adaptive lowering, QCO DD execution, and QCO↔jeff conversion now consume CBit directly. The old leading-zero memref recognizers, Qiskit UB dependency, parallel OpenQASM bit state, classical-output poison placeholders, and generic classical-register-name memref convention are removed.

Validation:

- The release build passes.
- `ctest --preset release` passes all 4,290 tests; one device test is skipped by its fixture.
- The focused CBit, QC↔QCO, QIR, QCO utility, jeff, compiler, QC translation, and OpenQASM binaries pass.
- The rebuilt binding passes 167 focused Qiskit and MLIR Python tests.
- The supported and minimum Python dependency matrices pass on Python 3.10 through 3.14.
- Stub generation, the MLIR documentation target, the documentation build, and lint pass.
- All fourteen pull-request commits pass `git verify-commit`.
- Pull request #2165 fixes the pre-existing external DOI, Qiskit policy, and old Read the Docs link-check failures and is queued for automatic merge.

Fixes #2155

## Checklist


burgholzer:
@denialhaag @simon1hofmann Could you please take a closer look at this?
I will push one more commit that simplifies the builder syntax and the diff noise in the tests, but otherwise I have looked through the entire diff here and this feels extremely natural and complete. It unifies a lot of the handling between the various import and export formats when it comes to dealing with classical registers.
One might even think about whether we'd want a similar abstraction for quantum registers; but let's not go there (yet).

bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -129,6 +130,8 @@ struct ExportState {
   llvm::DenseMap<mlir::Value, uint32_t> quantumSizes;
   llvm::DenseMap<mlir::Value, uint32_t> classicalBases;
   llvm::DenseMap<mlir::Value, uint32_t> classicalSizes;
+  llvm::DenseMap<mlir::Value, mlir::cbit::Initialization>
+      classicalInitializations;
```
denialhaag: Is there an argument to be made for consolidating these three into a struct? 🤔

docs/mlir/CBit.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Memory lowering
+
+Use `convert-cbit-to-memref` when a later pipeline requires generic memory. The
+pass converts `!cbit.reg<N>` to `memref<Nxi1>` in operations, function
+signatures, calls, branches, and returns. It lowers zero initialization to an
+allocation followed by false stores. It lowers undefined initialization to an
+allocation only.
+
+This conversion is one-way. MQT Core does not infer CBit semantics from an
```
denialhaag:
```suggestion
This conversion is one-way. `mqt-cc` does not infer CBit semantics from an
```

docs/mlir/CBit.md
```diff
@@ -0,0 +1,42 @@
...
+indices remain valid and consumers decide which dynamic operations they can
+support.
+
+An optional source name on `cbit.alloc` preserves an input-language register
+name. A source name does not make a register public. Only CBit registers
+returned by the entry function are public result registers. This rule keeps
+internal registers out of Qiskit and OpenQASM output.
+
+The CBit dialect is shared by QC and QCO. Their conversions preserve register
+identity, initialization, names, loads, stores, and returned results.
```
denialhaag: Not sure how we want to handle these dialect pages in general, but this feels like largely redundant information, given the descriptions in the TableGen files. 🤔

mlir/include/mlir/Conversion/QCOToJeff/QCOToJeff.td
```diff
@@ -21,5 +21,6 @@ def QCOToJeff : Pass<"qco-to-jeff"> {
     As the index is not preserved in `jeff`, it is not possible to round-tripping static qubits.
   }];
 
-  let dependentDialects = ["mlir::jeff::JeffDialect"];
+  let dependentDialects = ["mlir::cbit::CBitDialect",
+                           "mlir::jeff::JeffDialect"];
```
denialhaag: I don't think the conversion pass is _creating_ any `cbit` operations, right? 🤔 

```suggestion
  let dependentDialects = ["mlir::jeff::JeffDialect"];
```

mlir/include/mlir/Dialect/CBit/IR/CBitOps.td
denialhaag, line 1: We generally don't have the dialect prefix (i.e., `CBit_`) in front of the operations of the other dialects. 🤔

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
```diff
@@ -257,24 +242,33 @@ class QCProgramBuilder final : public ImplicitLocOpBuilder {
...
-   * %c = memref.alloc() : memref<3xi1>
+   * %c = cbit.alloc(#cbit.init<zero>) source_name = "c" : !cbit.reg<3>
    * ```
    */
-  Value allocClassicalBitRegister(int64_t size, StringRef name = {});
+  Value allocClassicalBitRegister(
+      int64_t size, StringRef name = {},
+      cbit::Initialization initialization = cbit::Initialization::Zero);
+
+  /** Load one value from a classical-bit register. */
```
denialhaag: Applies in more cases.

```suggestion
  /// Load one value from a classical-bit register.
```

mlir/include/mlir/Conversion/CBitToMemRef/CBitToMemRef.td
denialhaag, line 1: Is there an argument for a similar `CBitToTensor` conversion that could be used by the QCO-to-`jeff` conversion? 🤔

mlir/lib/Conversion/CBitToMemRef/CBitToMemRef.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+struct ConvertStoreOp final : OpConversionPattern<cbit::StoreOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(cbit::StoreOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter& rewriter) const override {
+    memref::StoreOp::create(rewriter, op.getLoc(), adaptor.getValue(),
+                            adaptor.getReg(), adaptor.getIndex());
+    rewriter.eraseOp(op);
```
denialhaag: This could probably also use `rewriter.replaceOpWithNewOp`.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
```diff
@@ -73,11 +77,10 @@ static Value resolveRegisterMeasurement(LoweringState& state, Operation* op,
...
 
 /**
- * @brief Converts classical-bit-register `memref.alloc` to `llvm.alloca`
+ * @brief Converts `cbit.alloc` to `llvm.alloca`
  */
-static LogicalResult convertClassicalBitMemRefAllocOp(
-    memref::AllocOp op, memref::AllocOp::Adaptor adaptor, LoweringState& state,
-    ConversionPatternRewriter& rewriter) {
+static LogicalResult convertCBitAllocOp(cbit::AllocOp op, LoweringState& state,
+                                        ConversionPatternRewriter& rewriter) {
```
denialhaag: At this point, we can probably inline the function again. 🤔

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
```diff
@@ -63,11 +65,10 @@ using namespace qco;
...
 
 /**
- * @brief Returns whether @p value is a classical-bit-register memref.
+ * @brief Returns whether @p value is a CBit register.
  */
 [[nodiscard]] static bool isClassicalRegister(Value value) {
-  auto memrefType = dyn_cast<MemRefType>(value.getType());
-  return memrefType && memrefType.getElementType().isInteger(1);
+  return isa<cbit::RegisterType>(value.getType());
 }
```
denialhaag: Maybe worth inlining this function.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -202,11 +220,9 @@ Value QCProgramBuilder::measure(Value qubit) {
 Value QCProgramBuilder::measure(Value qubit, Value reg,
                                 const std::variant<int64_t, Value>& index) {
   checkFinalized();
-  validateMemRefIndex(reg, index);
```
denialhaag, line 205: It looks like the `validateMemRefIndex` function is now dead and can be removed.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
```diff
@@ -40,19 +43,28 @@
...
+  }
+  if (const auto* constant = std::get_if<int64_t>(&index)) {
+    if (*constant < 0) {
+      llvm::reportFatalUsageError("Register index must be non-negative");
+    }
+    if (*constant >= type.getWidth()) {
+      llvm::reportFatalUsageError("Register index is out of bounds");
+    }
+  }
+}
```
denialhaag: This could moved into `Utils.h` and be shared with QCO.

mlir/unittests/Conversion/QCToQIR/QCToQIRBase/test_qc_to_qir_base.cpp
```diff
@@ -221,9 +222,9 @@ TEST(QCToQIRBaseNativeTest, RecordsReturnedRegisterMeasurement) {
   builder.initialize();
   const auto q = builder.allocQubit();
   const auto c = builder.allocClassicalBitRegister(1, "named_result");
```
denialhaag: This applies to a bunch of places and might be out of scope, but we can drop the `const` now. 

```suggestion
  auto c = builder.allocClassicalBitRegister(1, "named_result");
```

denialhaag (changes requested):
Thanks a lot, @burgholzer! 🙂 I really like how the new dialect makes a bunch of things a lot cleaner! We should really look into whether we want to do something similar for the qubit registers.

Below are some points I stumbled across while going through the changes; most are very minor. 😌

simon1hofmann (changes requested):
I also skimmed over the changes and it looks like a much better solution to the previous issue. 
Tested the changes in my benchpress integration and found the following problem, does that make sense?

🤖 AI text below 🤖
cbit.alloc, cbit.load, and cbit.store are not rejected inside inv, ctrl, or pow bodies. The QC and QCO modifier verifiers, as well as the QC→QCO preflight validation, still only account for the previous memref representation. I reproduced a cbit.store inside qc.inv successfully verifying and lowering to a qco.inv that retains the store.
Could we reject CBit operations in all three validation paths and add focused QC/QCO verifier and conversion regressions? Classical side effects inside these modifiers violate their unitary contract.

docs/mlir/CBit.md
burgholzer: Yeah, good point. I am, in general, not the biggest fan of the sheer amount of documentation that LLMs generate; especially if it is redundant.
I'd probably vote for just removing this here and potentially doing a bigger iteration of doc cleanup once a large majority of the features we want to get into v4 are in.

bindings/mlir/qiskit/QiskitExport.cpp
burgholzer: Yeah, feels like a reasonable suggestion 👍🏼

mlir/include/mlir/Conversion/QCOToJeff/QCOToJeff.td
burgholzer: That is true. I can't quite remember whether dialects only need to be listed here when new operations are created or when existing operations are manipulated as well. Probably the former.

mlir/include/mlir/Dialect/CBit/IR/CBitOps.td
burgholzer: Agreed. Could be good to consolidate. As long as the operations are in their own C++ namespace, it should be all good.

mlir/include/mlir/Conversion/CBitToMemRef/CBitToMemRef.td
burgholzer: I believe there is and there should be quite some synergies that can be exploited between the two passes. Let's add that 👍🏼

docs/mlir/CBit.md
denialhaag: > I am, in general, not the biggest fan of the sheer amount of documentation that LLMs generate; especially if it is redundant.

Yeah, same... Tangentially related, but I'm also trying to keep the PR descriptions reasonably long with my local configuration. A one-line change shouldn't exactly come with three paragraphs. Maybe something we can add to AGENTS.md. We shouldn't generate so much text that a human doesn't want to read it anymore.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
burgholzer: Yeah. I'd generally let the agent look for opportunities of inlining one-liners, where they do not really bring a meaningful benefit (like here).

burgholzer:
> I also skimmed over the changes and it looks like a much better solution to the previous issue. Tested the changes in my benchpress integration and found the following problem, does that make sense?
> 
> 🤖 AI text below 🤖 cbit.alloc, cbit.load, and cbit.store are not rejected inside inv, ctrl, or pow bodies. The QC and QCO modifier verifiers, as well as the QC→QCO preflight validation, still only account for the previous memref representation. I reproduced a cbit.store inside qc.inv successfully verifying and lowering to a qco.inv that retains the store. Could we reject CBit operations in all three validation paths and add focused QC/QCO verifier and conversion regressions? Classical side effects inside these modifiers violate their unitary contract.

Yeah, that kind of makes sense. I think none of our Importers would ever create a program that contains cbit operations as part of modifiers, but it is likely still a good idea to verify that assumption explicitly. I'll include that in the next iteration!

bindings/mlir/qiskit/QiskitExport.cpp
burgholzer: 🤖 *AI text below* 🤖

Consolidated the three maps into one `ClassicalRegisterInfo` map.

docs/mlir/CBit.md
burgholzer: 🤖 *AI text below* 🤖

Updated the wording to name `mqt-cc`.

docs/mlir/CBit.md
burgholzer: 🤖 *AI text below* 🤖

Removed the prose that duplicated the generated TableGen reference; the lowering guidance remains.

mlir/include/mlir/Conversion/QCOToJeff/QCOToJeff.td
burgholzer: 🤖 *AI text below* 🤖

Removed CBit from `dependentDialects`; the pass now lists only the Jeff dialect it creates.

mlir/include/mlir/Dialect/CBit/IR/CBitOps.td
burgholzer: 🤖 *AI text below* 🤖

Removed the redundant `CBit_` prefix from the operation definitions.

mlir/include/mlir/Dialect/QC/Builder/QCProgramBuilder.h
burgholzer: 🤖 *AI text below* 🤖

Converted the one-line builder comments to `///` in both QC and QCO.

mlir/include/mlir/Conversion/CBitToMemRef/CBitToMemRef.td
burgholzer: 🤖 *AI text below* 🤖

Added reusable CBit-to-tensor state, type conversion, and operation patterns. QCO-to-jeff shares the neutral state and type layer while retaining Jeff-specific materialization.

mlir/lib/Conversion/CBitToMemRef/CBitToMemRef.cpp
burgholzer: 🤖 *AI text below* 🤖

Replaced the create-and-erase pair with `replaceOpWithNewOp`.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
burgholzer: 🤖 *AI text below* 🤖

Inlined the single-use CBit allocation lowering into the conversion pattern.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
burgholzer: 🤖 *AI text below* 🤖

Removed the one-line helper and folded its type check into the shared CBit conversion state.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
burgholzer: 🤖 *AI text below* 🤖

Removed the unused memref-index validator.

mlir/lib/Dialect/QC/Builder/QCProgramBuilder.cpp
burgholzer: 🤖 *AI text below* 🤖

Moved static register-index validation to `Utils.h` and now share it between the QC and QCO builders.

mlir/unittests/Conversion/QCToQIR/QCToQIRBase/test_qc_to_qir_base.cpp
burgholzer: 🤖 *AI text below* 🤖

Dropped the redundant `const` qualifiers from the affected CBit allocation locals.

burgholzer:
🤖 *AI text below* 🤖

Added recursive CBit rejection to QC and QCO modifier verification and QC→QCO preflight, with alloc/load/store regressions across `inv`, `ctrl`, and `pow`.

simon1hofmann (approved):

mlir/include/mlir/Conversion/CBitToTensor/CBitToTensor.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace mlir {
+class Operation;
+class Region;
+class RewritePatternSet;
+class TypeConverter;
+
+namespace cbit {
+
+/** Tracks the current tensor value for each CBit register during conversion. */
```
denialhaag:
```suggestion
/// Tracks the current tensor value for each CBit register during conversion.
```

burgholzer:
I'll fix the clang-tidy warnings. Otherwise, this should have addressed both of your concerns.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
denialhaag, line 1: In my head, this simplified further. Could we use `populateCBitToTensorConversionPatterns` and do a run before doing a run with `populateNativeToJeffConversionPatterns`? Otherwise, the CBit-to-`tensor` conversion might not be that useful. Maybe I'm missing something, though. 🤔

mlir/include/mlir/Dialect/Utils/Utils.h
```diff
@@ -137,22 +133,32 @@ variantToValue(OpBuilder& builder, Location loc,
...
+template <typename RegisterType>
+inline void
+validateStaticRegisterIndex(const Value reg,
+                            const std::variant<int64_t, Value>& index) {
+  const auto type = dyn_cast<RegisterType>(reg.getType());
+  if (!type) {
+    llvm::reportFatalUsageError("Expected a static register");
+  }
+  validateStaticRegisterIndex(type.getWidth(), index);
+}
```
denialhaag: Do we need this template function? At the moment, all instances call it with `cbit::RegisterType` anyway. 🤔

denialhaag (changes requested):

denialhaag:
I just have three more points. Apart from that, this LGTM now! 😌

burgholzer:
@denialhaag good points! I think there should be room for a couple more simplifications.

mlir/include/mlir/Conversion/CBitToTensor/CBitToTensor.h
burgholzer: 🤖 *AI text below* 🤖

Applied `///` and audited every single-line comment added by this PR. I also recorded the rule in the global agent guidance.

mlir/lib/Conversion/QCOToJeff/QCOToJeff.cpp
burgholzer: 🤖 *AI text below* 🤖

Refactored this into two stages: shared CBit-to-tensor lowering runs first, then native/QCO-to-jeff lowering. Loads, stores, function types, and returned register state now use the shared path; only allocation remains jeff-specific so the reverse conversion retains CBit identity.

mlir/include/mlir/Dialect/Utils/Utils.h
burgholzer: 🤖 *AI text below* 🤖

Removed the template. Static-index validation now belongs to CBit, and both builders call the CBit-specific function directly.

denialhaag (approved):
:shipit:

burgholzer:
> :shipit:

I simplified a couple more aspects triggered by your comment. This feels rather clean to me now. In the process, I removed the cbit-to-tensor logic again because jeff conversion was the only consumer and it did not really fit too well after all.
I'll tag this for auto-merge now so that some of the PRs that are influenced by this can progress.

denialhaag:
> In the process, I removed the cbit-to-tensor logic again because jeff conversion was the only consumer and it did not really fit too well after all.

Yeah, after seeing it in action, I agree that it didn't add any real value.

Issue #2131, opened 2026-08-17
✨ Declare target classical-control capabilities and reject unsupported programs


simon1hofmann:
🤖 *AI text below* 🤖

Lead: @simon1hofmann. Cross-repository coordination: @burgholzer.

## Problem

`CompilerTarget` describes sites, topology, and native quantum operations, but it cannot declare support for runtime classical control. The target-compilation pipeline consequently has no fail-closed capability validation for measurement-conditioned branches, loops, switches, or dynamic indexing before mapping and synthesis begin.

Current implementation:

- [`CompilerTarget` capability model](https://github.com/munich-quantum-toolkit/core/blob/53325922bfebe14f17fbb8af7d07fd68bb9f9704/mlir/include/mlir/Compiler/Target.h#L30-L171)
- [Target-compilation pipeline](https://github.com/munich-quantum-toolkit/core/blob/53325922bfebe14f17fbb8af7d07fd68bb9f9704/mlir/lib/Compiler/TargetCompilation.cpp#L23-L34)

Structured-control mapping exists for selected forms, including [`scf.for`](https://github.com/munich-quantum-toolkit/core/pull/1805), [`scf.while`](https://github.com/munich-quantum-toolkit/core/pull/1911), and [`qco.index_switch`](https://github.com/munich-quantum-toolkit/core/pull/1951), but target support is not part of the compiler-target contract.

## Goal

Add a frontend-neutral, conservative classical-control capability description to `CompilerTarget`. At minimum, targets must distinguish no runtime classical control from explicitly supported constructs. Target compilation must determine the program's required capabilities before target-mutating passes and reject unsupported programs with a stable diagnostic instead of continuing into an invalid or unsupported pipeline.

## Minimal reproducer

Compile a QCO program containing a measurement-conditioned `scf.if` for a target that declares no runtime control. Compilation must fail before mapping with a diagnostic naming the unsupported capability. A target declaring support for that conditional form must either compile it through final conformance or reject a more specific unsupported subfeature.

## Acceptance criteria

- The C++ and Python target APIs expose a documented, conservative classical-control capability model.
- An initial capability check identifies required control features and fails closed before mapping or synthesis.
- Diagnostics identify the unsupported construct or capability; no assertion, crash, or silently flattened control flow occurs.
- Declared supported forms remain valid through mapping, native synthesis, and final conformance.
[...]

PR #2162, opened 2026-08-19
✨ Legalize control flow for selected payloads

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Legalize structured control flow against the selected program capability model using MLIR normalization, bounded static-loop unrolling and dialect conversion.

Compiler-only workstream: **main → #2219 → #2162**. This PR no longer depends on the QDMI runtime or adapter stack. It targets **Core 4.0**, subject to Simon's compiler contract review and current CI. The compiler-only scope is tracked in #2131; it does not depend on QDMI 1.4 or its unresolved metadata contract. #2365 coordinates the separate Core 4.1/QDMI 1.4 adaptation.

Retain the existing prototype's safety limits, explicit linear quantum state and fail-closed unsupported control flow. Mechanical adaptations preserve current placement/decomposition and the canonical pipeline. The current LLVM handles the full-width loop range correctly; the regression checks that the loop and exact trip count survive.

Simon Hofmann's human co-authorship and the existing review history are preserved.

### Local validation

- Independent release build and CTest: 3,889 passed, one existing optional-device skip.
- MLIR documentation and repository lint passed.
- Full changed-file C++ lint passed on the final revision.
- No new Python API is introduced beyond the separately validated #2219 foundation.
- Hosted CI remains a separate final-head gate.

AI-assisted restructuring; human design review remains required.

### Outstanding hosted validation

The current [macOS upstream-Qiskit job](https://github.com/munich-quantum-toolkit/core/actions/runs/33866243571/job/101001658830) fails against Qiskit 2.6.0.dev0 (095d4bd26cee0bfc0bd362deb8b272c111c63ee2): deferred instruction export reports a one-qubit argument for a zero-qubit instruction. This is not the pinned local test configuration and needs separate triage; no claim of green hosted CI is made. The cancelled earlier aggregate CI check has a replacement run queued.

## Compiler acceptance

Preserve control flow supported by the selected target, lower or unroll unsupported forms when feasible within the established safety limits, and reject remaining unsupported programs clearly. Validate representative benchmark programs and ensure the final target-conformance check still holds. No QDMI 1.4 API is required.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,10 @@ releases may include breaking changes.
   [#1807], [#1808], [#1815], [#1824], [#1869], [#1872], [#1914], [#1925],
   [#1927], [#1935], [#1936], [#1938], [#1975], [#1976], [#2006], [#2014],
   [#2015], [#2017], [#2026], [#2028], [#2054], [#2058], [#2125], [#2136],
-  [#2149], [#2150], [#2158], [#2194], [#2210], [#2211], [#2215], [#2218],
-  [#2220], [#2323], [#2336], [#2344]) ([**@burgholzer**], [**@denialhaag**],
-  [**@taminob**], [**@DRovara**], [**@li-mingbao**], [**@Ectras**],
-  [**@MatthiasReumann**], [**@simon1hofmann**], [**@J4MMlE**])
+  [#2149], [#2150], [#2158], [#2162], [#2194], [#2210], [#2211], [#2215],
+  [#2218], [#2220], [#2323], [#2336], [#2344]) ([**@burgholzer**],
+  [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**],
+  [**@Ectras**], [**@MatthiasReumann**], [**@simon1hofmann**], [**@J4MMlE**])
 - ✨ Add a library for typed structured quantum benchmarks with versioned
   instance specifications, analytic references, deterministic manifests, and
   C++, Python, and command-line interfaces ([#2135], [#2299], [#2315], [#2324],
@@ -992,6 +992,7 @@ for previous changelogs._
 [#2175]: https://github.com/munich-quantum-toolkit/core/pull/2175
 [#2169]: https://github.com/munich-quantum-toolkit/core/pull/2169
 [#2168]: https://github.com/munich-quantum-toolkit/core/pull/2168
+[#2162]: https://github.com/munich-quantum-toolkit/core/pull/2162
 [#2158]: https://github.com/munich-quantum-toolkit/core/pull/2158
 [#2157]: https://github.com/munich-quantum-toolkit/core/pull/2157
 [#2156]: https://github.com/munich-quantum-toolkit/core/pull/2156
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,10 @@ releases may include breaking changes.
   [#1807], [#1808], [#1815], [#1824], [#1869], [#1872], [#1914], [#1925],
   [#1927], [#1935], [#1936], [#1938], [#1975], [#1976], [#2006], [#2014],
   [#2015], [#2017], [#2026], [#2028], [#2054], [#2058], [#2125], [#2136],
-  [#2149], [#2150], [#2158], [#2162], [#2194], [#2210], [#2211], [#2215],
-  [#2218], [#2220], [#2323], [#2336], [#2344]) ([**@burgholzer**],
-  [**@denialhaag**], [**@taminob**], [**@DRovara**], [**@li-mingbao**],
-  [**@Ectras**], [**@MatthiasReumann**], [**@simon1hofmann**], [**@J4MMlE**])
+  [#2149], [#2150], [#2158], [#2194], [#2210], [#2211], [#2215], [#2218],
+  [#2220], [#2323], [#2336], [#2344]) ([**@burgholzer**], [**@denialhaag**],
+  [**@taminob**], [**@DRovara**], [**@li-mingbao**], [**@Ectras**],
+  [**@MatthiasReumann**], [**@simon1hofmann**], [**@J4MMlE**])
 - ✨ Add a library for typed structured quantum benchmarks with versioned
   instance specifications, analytic references, deterministic manifests, and
   C++, Python, and command-line interfaces ([#2135], [#2299], [#2315], [#2324],
@@ -38,11 +38,12 @@ releases may include breaking changes.
   direct lowering and dense-array helpers for supported compiler inputs
   ([#1915], [#1973], [#2077], [#2078], [#2079], [#2334]) ([**@simon1hofmann**],
   [**@burgholzer**])
-- ✨ Add immutable MLIR compiler targets, selected payload specifications, QDMI
-  device integration, ordered operation applicability, directional native
-  synthesis, and target compilation through C++, Python, and `mqt-cc` ([#2285],
-  [#2219], [#2049], [#1999], [#1993], [#1687]) ([**@MatthiasReumann**],
-  [**@simon1hofmann**], [**@burgholzer**])
+- ✨ Add immutable MLIR compiler targets, selected payload specifications,
+  payload-aware control-flow legalization, QDMI device integration, ordered
+  operation applicability, directional native synthesis, and target compilation
+  through C++, Python, and `mqt-cc` ([#2285], [#2219], [#2162], [#2049],
+  [#1999], [#1993], [#1687]) ([**@MatthiasReumann**], [**@simon1hofmann**],
+  [**@burgholzer**])
 
 #### Import and export
 
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -82,7 +82,9 @@ Target compilation requires structured QCO/SCF input. Producers of raw CFG
 branches must normalize them before target compilation; runtime assertions are
 allowed. The pipeline removes unused symbols, propagates constants, unrolls
 unsupported static loops, and then runs the standard QCO cleanup pipeline. It
-applies these structural capabilities to the remaining control flow:
+uses `unroll-loops-for-payload` before cleanup and `legalize-control-flow` after
+cleanup, so unrolling can expose constant branches before legality checks. The
+latter pass applies these structural capabilities to the remaining control flow:
 
 | Capability           | Residual operations                                 |
 | -------------------- | --------------------------------------------------- |
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -78,11 +78,11 @@ typed `#mqt.payload_spec` attribute.
 
 ### Payload control flow
 
-Target compilation removes unused symbols, lifts reducible ControlFlow dialect
-graphs to SCF, and propagates constants. It checks constant loop ranges with
-widened arithmetic before generic canonicalization, unrolls unsupported static
-loops, and then runs the standard QCO cleanup pipeline. It applies these
-structural capabilities to the remaining control flow:
+Target compilation requires structured QCO/SCF input. Producers of raw CFG
+branches must normalize them before target compilation; runtime assertions are
+allowed. The pipeline removes unused symbols, propagates constants, unrolls
+unsupported static loops, and then runs the standard QCO cleanup pipeline. It
+applies these structural capabilities to the remaining control flow:
 
 | Capability           | Residual operations                                 |
 | -------------------- | --------------------------------------------------- |
@@ -102,13 +102,15 @@ arguments instead of capturing it.
 
 The supported constraints are `max-control-flow-nesting-depth` on all four
 capabilities, `max-iteration-count` on both iteration capabilities, and
-`max-case-count` on multiway branching. Limits are inclusive. The compiler must
+`max-case-count` on multiway branching, counting explicit cases without the
+default region. One explicit case plus a default is a supported index switch and
+does not require forward branching. Limits are inclusive. The compiler must
 prove a constrained loop's trip count. It currently proves constant `scf.for`
 bounds and rejects a constrained `scf.while` because no general termination
-bound is available. The compiler rejects a constant range when MLIR's native
-trip-count result disagrees with widened arithmetic. A zero, unknown, or
-misapplied constraint makes that capability group unusable. Missing or
-incomplete optional metadata never implies support.
+bound is available. MLIR computes static trip counts; full unrolling
+additionally requires bounds and scaled steps that fit its signed arithmetic. A
+zero, unknown, or misapplied constraint makes that capability group unusable.
+Missing or incomplete optional metadata never implies support.
 
 This stage checks structural control flow only. Later lowering stages remain
 responsible for scalar types and operations, measurement provenance, function
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -96,13 +96,19 @@ latter pass applies these structural capabilities to the remaining control flow:
 A finite `scf.for` that exceeds the selected counted-iteration contract is fully
 unrolled when this clones at most 65,536 body operations. Cleanup runs again
 because unrolling can make nested bounds and conditions constant. An unsupported
-index switch is lowered to nested forward branches when that form fits the
-selected contract. Generic SCF branches cannot capture or return QCO qubits or
-quantum tensors; use the corresponding QCO branch operation for linear quantum
-state. SCF loops must carry linear quantum state through their iteration
-arguments instead of capturing it. Both control-flow passes validate this loop
-input restriction before transforming loops or lowering switches. It is separate
-from QCO's exactly-one-SSA-use check.
+index switch is lowered to a linear chain of nested forward branches when that
+form fits the selected contract. Before expansion, the compiler checks the
+selected forward-branching nesting limit and a compiler safety limit of 256 total
+control-flow levels, including enclosing control flow. This compiler limit is
+not a QDMI requirement and does not apply to switches retained under multiway
+branching.
+
+Generic SCF branches cannot capture or return QCO qubits or quantum tensors; use
+the corresponding QCO branch operation for linear quantum state. SCF loops must
+carry linear quantum state through their iteration arguments instead of
+capturing it. Both control-flow passes validate this loop input restriction
+before transforming loops or lowering switches. It is separate from QCO's
+exactly-one-SSA-use check.
 
 The supported constraints are `max-control-flow-nesting-depth` on all four
 capabilities, `max-iteration-count` on both iteration capabilities, and
@@ -111,10 +117,12 @@ default region. One explicit case plus a default is a supported index switch and
 does not require forward branching. Limits are inclusive. The compiler must
 prove a constrained loop's trip count. It currently proves constant `scf.for`
 bounds and rejects a constrained `scf.while` because no general termination
-bound is available. MLIR computes static trip counts; full unrolling
-additionally requires bounds and scaled steps that fit its signed arithmetic. A
-zero, unknown, or misapplied constraint makes that capability group unusable.
-Missing or incomplete optional metadata never implies support.
+bound is available. The proof requires literal loop bounds and a literal step;
+it does not infer a trip count from symbolic bounds. MLIR computes static trip
+counts; full unrolling additionally requires bounds and scaled steps that fit
+its signed arithmetic. The scaled step must also fit the loop induction-variable
+type. A zero, unknown, or misapplied constraint makes that capability group
+unusable. Missing or incomplete optional metadata never implies support.
 
 This stage checks structural control flow only. Later lowering stages remain
 responsible for scalar types and operations, measurement provenance, function
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -98,9 +98,9 @@ unrolled when this clones at most 65,536 body operations. Cleanup runs again
 because unrolling can make nested bounds and conditions constant. An unsupported
 index switch is lowered to a linear chain of nested forward branches when that
 form fits the selected contract. Before expansion, the compiler checks the
-selected forward-branching nesting limit and a compiler safety limit of 256 total
-control-flow levels, including enclosing control flow. This compiler limit is
-not a QDMI requirement and does not apply to switches retained under multiway
+selected forward-branching nesting limit and a compiler safety limit of 256
+total control-flow levels, including enclosing control flow. This compiler limit
+is not a QDMI requirement and does not apply to switches retained under multiway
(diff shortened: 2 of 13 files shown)


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai full review

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ releases may include breaking changes.
 - ✨ Add immutable MLIR compiler targets, QDMI device integration, and target
   compilation through C++, Python, and `mqt-cc` ([#1687], [#1993], [#1999],
   [#2049]) ([**@MatthiasReumann**], [**@simon1hofmann**], [**@burgholzer**])
+- ✨ Add explicit runtime classical-control capabilities to compiler targets and
+  reject undeclared structured control before target compilation ([#2162])
+  ([**@simon1hofmann**])
```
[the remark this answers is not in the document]
simon1hofmann: Changed the wording.

simon1hofmann:
@coderabbitai review

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ releases may include breaking changes.
 - ✨ Add immutable MLIR compiler targets, QDMI device integration, and target
   compilation through C++, Python, and `mqt-cc` ([#1687], [#1993], [#1999],
   [#2049]) ([**@MatthiasReumann**], [**@simon1hofmann**], [**@burgholzer**])
+- ✨ Add explicit runtime classical-control capabilities to compiler targets and
+  reject undeclared structured control at the start of target compilation,
+  before target-mutating passes ([#2162]) ([**@simon1hofmann**])
```
burgholzer: Should likely be folded into another entry.

UPGRADING.md
burgholzer: Not necessary. Maybe also keep that in mind for further PRs in this and other stacks.

It seems likely by now that we should be adjusting our AGENTS.md guidance to avoid the constant back and fourth on the changelog and upgrade guide.

burgholzer:
I am starting to take a look at this. This is probably gonna become a little bit of a bigger discussion I fear. Finding the right way to describe the classical capabilities of a target and then matching that with the concepts used as part of a program isn't particularly trivial.
I'll try my best to come up with a reasonable review/plan here.

CHANGELOG.md
burgholzer: 🤖 *AI text below* 🤖

Addressed. The reduced PR folds #2162 into the existing MQT Compiler Collection changelog entry instead of adding a separate entry.

UPGRADING.md
burgholzer: 🤖 *AI text below* 🤖

Addressed by the split. The reduced control-flow PR no longer changes `UPGRADING.md`.

docs/mlir/target_compilation.md
```diff
@@ -37,6 +37,44 @@ MQT Core derives the compiler output from this payload specification and uses
...
+
+The supported constraints are `max-control-flow-nesting-depth` on all four
+capabilities, `max-iteration-count` on both iteration capabilities, and
+`max-case-count` on multiway branching. Limits are inclusive. The compiler must
+prove a constrained loop's trip count. It currently proves constant `scf.for`
+bounds and rejects a constrained `scf.while` because no general termination
+bound is available. The compiler rejects a constant range when MLIR's native
+trip-count result disagrees with widened arithmetic. A zero, unknown, or
+misapplied constraint makes that capability group unusable. Missing or
+incomplete optional metadata never implies support.
```
burgholzer, line 79: It feels like some of these descriptions could/should be upstreamed and included in the QDMI PR that introduces the program capabilities.
Some of these statements feel quite normative and helpful for making sure people export capabilities in a standardized fashion.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -108,6 +108,36 @@ def QuantumLoopUnroll
                         "set it to -1, and it will fully unroll.">];
 }
 
+//===----------------------------------------------------------------------===//
+// Payload legalization passes
+//===----------------------------------------------------------------------===//
+
+def UnrollUnsupportedPayloadLoops
+    : Pass<"unroll-unsupported-payload-loops", "mlir::ModuleOp"> {
```
burgholzer, line 118: The naming of this is a bit odd. Can we make this a little more natural? same holds for the other pass. Maybe the passes could even be combined. I like the "legalize-control-flow" part of this. Maybe a combination of both makes sense.

mlir/lib/Compiler/TargetCompilation.cpp
```diff
@@ -13,12 +13,20 @@
...
 
+#include <mlir/Conversion/ControlFlowToSCF/ControlFlowToSCF.h>
 #include <mlir/Pass/PassManager.h>
 #include <mlir/Transforms/Passes.h>
 
 namespace mlir {
 
 void populateTargetCompilationPipeline(OpPassManager& pm) {
+  pm.addPass(createSymbolDCEPass());
+  pm.addPass(createLiftControlFlowToSCFPass());
```
burgholzer, line 27: I might be largely mistaken, but none of the frontends we currently support will produce control flow dialect operations. Is this handling really necessary or could we simplify around that?

mlir/unittests/Compiler/test_compiler_pipeline.cpp
burgholzer, line 1: I'd critically question whether the programs being described here could actually reasonably be constructed by one of the frontends we currently consider.
If not, I'd rather make the assumptions a little more explicit and simplify around these new stricter assumptions.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+
+constexpr llvm::StringLiteral FORWARD_BRANCHING = "forward-branching";
+constexpr llvm::StringLiteral COUNTED_ITERATION = "counted-iteration";
+constexpr llvm::StringLiteral CONDITIONAL_LOOP = "conditional-loop";
+constexpr llvm::StringLiteral MULTIWAY_BRANCHING = "multiway-branching";
+
+constexpr llvm::StringLiteral MAX_NESTING_DEPTH =
+    "max-control-flow-nesting-depth";
+constexpr llvm::StringLiteral MAX_ITERATION_COUNT = "max-iteration-count";
+constexpr llvm::StringLiteral MAX_CASE_COUNT = "max-case-count";
```
burgholzer, line 62: This should likely be standardized in a more central place.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+  });
+}
+
+[[nodiscard]] static bool hasLinearBranchState(Operation* operation) {
+  return llvm::any_of(operation->getResultTypes(), isLinearQubitType) ||
+         hasLinearCapture(operation);
+}
+
+/// MLIR 22 computes index differences at their original bit width. Widen the
+/// constants first so an overflowing range cannot appear to have zero trips.
```
burgholzer: This feels like a case where a more explicit assumption on the input could really help to reduce complexity. This feels like a corner case that is merely hypothetical for practical programs.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
burgholzer, line 1: this feels way more complicated than it likely needs to be for practical use cases.

burgholzer (commented):
This one feels like it will still require a bit of work for simplification and making some assumptions more explicit.
The general direction looks very good though.

burgholzer:
@denialhaag maybe you have some comments on top of my existing ones.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+[[nodiscard]] static std::optional<llvm::APInt>
+getExactConstantTripCount(scf::ForOp loop) {
+  const auto constant = [](const Value value) -> std::optional<llvm::APInt> {
```
denialhaag:
```suggestion
  const auto constant = [](Value value) -> std::optional<llvm::APInt> {
```

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+[[nodiscard]] static bool isLegal(scf::IndexSwitchOp operation,
+                                  const PayloadControlSupport& support) {
+  const uint64_t cases = operation.getNumCases();
+  return !hasLinearBranchState(operation) && cases > 1U &&
+         support.coversMultiwayBranching(operation, cases);
+}
+
+[[nodiscard]] static bool canUseFullUnroll(scf::ForOp loop,
+                                           const uint64_t iterations) {
+  const auto constant = [&](const Value value) -> std::optional<int64_t> {
```
denialhaag:
```suggestion
  const auto constant = [&](Value value) -> std::optional<int64_t> {
```

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+        return ifOp;
+      }
+
+      Block& elseBlock = ifOp.getElseRegion().emplaceBlock();
+      const OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToEnd(&elseBlock);
+      scf::IfOp nested = self(self, index + 1U);
+      scf::YieldOp::create(rewriter, operation.getLoc(), nested.getResults());
+      return ifOp;
+    };
```
denialhaag, line 458: Not sure if this is really worth it, but maybe this function and its counterpart in `LowerQCOIndexSwitch` can become a shared helper. 🤔

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+    while (true) {
+      SmallVector<std::pair<scf::ForOp, llvm::APInt>> loops;
+      const WalkResult result =
+          getOperation().walk<WalkOrder::PreOrder>([&](scf::ForOp loop) {
+            if (hasLinearCapture(loop)) {
+              loop.emitError(
+                  "SCF loop captures QCO linear values; pass them as "
+                  "iteration arguments");
+              return WalkResult::interrupt();
+            }
```
denialhaag, line 498: This feels like one of the cases that could be excluded by construction. If we believe a linearity check is necessary regardless, maybe we can run the `verifyLinearity` function added in https://github.com/munich-quantum-toolkit/core/pull/2220 once at the beginning. That should pretty much guarantee that this check (and similar ones) cannot ever fail.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+        if (tripCount.isOne()) {
+          if (failed(loop.promoteIfSingleIteration(rewriter))) {
+            loop.emitError("failed to promote a single-iteration loop");
+            signalPassFailure();
+            return;
+          }
+          continue;
+        }
+        if (llvm::hasSingleElement(loop.getBody()->getOperations())) {
+          const ValueRange yielded = loop.getYieldedValues();
```
denialhaag:
```suggestion
          auto yielded = loop.getYieldedValues();
```

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+          if (llvm::all_of(yielded, [&](const Value value) {
```
denialhaag:
```suggestion
          if (llvm::all_of(yielded, [&](Value value) {
```

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
```diff
@@ -0,0 +1,636 @@
...
+    target.addDynamicallyLegalOp<IndexSwitchOp>(
+        [&](IndexSwitchOp operation) { return isLegal(operation, *support); });
+    target.addDynamicallyLegalOp<scf::IndexSwitchOp>(
+        [&](scf::IndexSwitchOp operation) {
+          return isLegal(operation, *support);
+        });
+    target.markUnknownOpDynamicallyLegal([](Operation* operation) {
+      if (isa<scf::ExecuteRegionOp, CtrlOp, InvOp, PowOp>(operation)) {
+        return true;
+      }
```
denialhaag, line 621: I might be missing something, but is there a reason we need to explicitly name these? I don't think any of our modifiers implement the `RegionBranchOpInterface`. 🤔

denialhaag (commented):
Thanks for getting this started, @simon1hofmann! 😎

Some more comments from my side:

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
burgholzer: We really need to institutionalize this in AGENTS.md. Maybe we should even feed it https://mlir.llvm.org/getting_started/DeveloperGuide/ and its subpages once and distill the essence out of those guidelines. I can try that in a separate PR.

simon1hofmann:
Thanks a lot for the review @burgholzer and @denialhaag, will work on them asap.

burgholzer:
> Thanks a lot for the review @burgholzer and @denialhaag, will work on them asap.

Before you start working on these, let's wait a bit. I planned to kick off a large-scale run on all the PR feedback in the stack, which should hopefully produce coherent edits across the board. I am almost done with the entire PR stack with feedback, so I should be able to kick that off later today.

simon1hofmann:
> > Thanks a lot for the review @burgholzer and @denialhaag, will work on them asap.
> 
> Before you start working on these, let's wait a bit. I planned to kick off a large-scale run on all the PR feedback in the stack, which should hopefully produce coherent edits across the board. I am almost done with the entire PR stack with feedback, so I should be able to kick that off later today.

Sounds good, then let me know when it's ready 👍

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
MatthiasReumann: I just want to add to that there is also the `QuantumLoopUnroll` pass defined right above that. I think these two could potentially be one.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
burgholzer: I am trying to tackle this one systematically first; before kicking off a large scale review and update of this PR stack. Ideally, the entire review process already benefits from the updated guidelines then. I might tag you, @denialhaag and @simon1hofmann, in a review for the updated guidelines once I am done with iterating them out.

simon1hofmann:
🤖 *AI text below* 🤖

I checked how much #2162 actually depends on the PRs below it. The result is substantially narrower than the current ancestry suggests:

- #2219 is the one hard functional dependency. It provides the selected payload specification, TargetEnvironment, its cached analysis, and the payload-aware compilation entry point consumed here.
- #2226 and #2227 are not functional dependencies of #2162. Their overlap with this PR is only editorial: the changelog for #2226, and the changelog plus target-compilation documentation for #2227.
- #2218, #2323, and #2215 are already merged into main and are only indirect foundations now.

I verified this by cherry-picking #2162's sole commit directly onto #2219 with #2226 and #2227 absent. The reduced tree configured and built successfully, all 10 PayloadControl tests passed, and the complete compiler suite passed 147/147.

The cleaner restack would therefore be main → #2219 → #2162, while #2226 → #2227 can remain a separate continuation above #2219. Since #2219 is currently behind and conflicting with main, it should be rebased first; #2162 should then be rebased or cherry-picked directly onto it rather than only changing its GitHub base.

This PR would also be valuable for the Benchpress integration. It gives Benchpress a target/payload-aware path for control flow: supported branches and loops can be preserved, unsupported static constructs can be legalized, and unsupported residual control flow produces deterministic diagnostics. That avoids baking one uniform control-flow capability model into the Benchpress side of the integration.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Removed the explicit `CtrlOp`, `InvOp`, and `PowOp` exemptions in [1ae3006c9](https://github.com/munich-quantum-toolkit/core/commit/1ae3006c9b64d44546e606eb1fc721a2d126eed3). They implement neither branch interface, so the existing fallback already accepts them. The `scf.execute_region` exception remains.

This is a one-line code change; the existing loop/switch capability snapshot and its safety limits are preserved. The branch is also restacked on the rebased #2219, retaining main's reusable-function inliner before control-flow legalization. All 179 compiler tests, including the ten payload-control tests, pass locally.

burgholzer:
I pushed a couple of simplifications here. Let's maybe merge 2219 first, then get this one in shape. Moved it out of draft for now.
@simon1hofmann can you take it from here again?

docs/mlir/target_compilation.md
simon1hofmann: 🤖 *AI text below* 🤖

The proposed definitions are already present in the open draft [QDMI #508](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/508): [control-flow capabilities](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/91955f3e878890e8038005929bd75024e8216ba4/include/qdmi/constants.h#L1148-L1169), [constraints](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/91955f3e878890e8038005929bd75024e8216ba4/include/qdmi/constants.h#L1211-L1231), and [grouping rules](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/91955f3e878890e8038005929bd75024e8216ba4/include/qdmi/constants.h#L1086-L1106).

These remain draft provider contracts, tracked by [QDMI #523](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/523) and [Core #2365](https://github.com/munich-quantum-toolkit/core/issues/2365). This PR keeps its compiler capability snapshot independent of that proposed API; it does not claim that the provider contract is settled.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
simon1hofmann: 🤖 *AI text below* 🤖

Renamed the passes to `unroll-loops-for-payload` and `legalize-control-flow` in 6214a2bd5 and updated their callers and descriptions.

The two stages stay separate because SCCP and QCO cleanup between them fold control flow made constant by unrolling. `QuantumLoopUnroll` also stays separate: it selects quantum loop-carried state and supports partial unroll factors, whereas payload legalization must handle classical loops too, preserve supported loops, and bound total cloning. Both already reuse MLIR's full-unroll utility.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Default-region inlining is already shared. I have kept the recursive builders separate: QCO explicitly threads linear targets through branch arguments and distinguishes classical and linear results, while SCF has different region and yield handling. Sharing that part would add callbacks or template machinery around those differences without simplifying the state flow.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Centralized the supported loop-input check at both pass entry points in 505b60215, and removed the repeated capture checks from the unroll traversal and loop-legality callbacks. The unroll pass still folds static branches before this validation.

The existing `verifyLinearity` checks exactly one SSA use, which does not exclude loop captures. Its general contract is unchanged; the shared check enforces this target pipeline's explicit iteration-argument requirement. The regression covers both standalone passes, including an inner capture beneath a dynamic outer loop, and checks that these failures leave the IR unchanged.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
simon1hofmann: 🤖 *AI text below* 🤖

The current implementation removes CFG lifting, uses MLIR's static trip-count calculation, and removes redundant cleanup. The follow-up commits centralize the loop-input checks and clarify the two stages as `unroll-loops-for-payload` and `legalize-control-flow`.

SCCP and QCO cleanup remain between those stages so unrolling can expose constant branches before legality checks. The remaining unroll limits and arithmetic checks protect the bounded full-unroll path.

mlir/unittests/Compiler/test_compiler_pipeline.cpp
simon1hofmann: 🤖 *AI text below* 🤖

The input contract is now restricted to structured QCO/SCF control flow. Raw CFG branches are rejected before cleanup, and SCF loops must carry quantum state through iteration arguments. The loop restriction is now checked at both pass boundaries in 505b60215.

These tests exercise the public typed-MLIR compilation boundary, including rejection cases; they do not assume that every fixture is produced by a current frontend.

mlir/lib/Dialect/QCO/Transforms/LegalizePayloadControlFlow.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Moved the seven identifiers to the existing `ProgramCapability` and `ProgramConstraint` types in `mlir/Compiler/TargetEnvironment.h` in 1c5d4cc66. Producers and legalization can now use the same constants; the spellings, extensible string model, and compiler behavior are unchanged. This does not introduce a dependency on the draft QDMI capability API.

burgholzer:
Is this one ready for review/finalization?

simon1hofmann:
> Is this one ready for review/finalization?

Yes from my side, addressed all open threads.

burgholzer (approved):
Let's get this in so that we are closer to a feature complete pipeline.

PR #2175, opened 2026-08-20
✨ Import captured Qiskit classical expressions

simon1hofmann:
🤖 *AI text below* 🤖

## Description

- Import Qiskit classical-expression bit and register leaves into first-class
  CBit loads and packed unsigned compiler expressions.
- Resolve condition-only, explicit-body, switch-target, and nested captures
  through the containing circuit and each enclosing native capture map.
- Avoid Qiskit 2.5's unsafe expression-valued switch accessor by normalizing
  that target through the public, structurally typed Python API.
- Reject malformed, duplicated, out-of-range, or unresolvable captures during
  import preflight.

This prerequisite is stacked on #2150. It keeps the import identity model
separate from Qiskit control-flow construction.

## Stack

1. #2150 — symbolic scalar parameters
2. #2175 — captured classical expressions (this PR)
3. #2176 — structured-control export for #2071
4. #2177 — measurement-store export
5. #2178 — parameter-vector provenance

## Testing

- Release Qiskit binding build: passed.
- `test/python/test_mlir_qiskit_translation.py`: 167 passed.
- Clang format, Ruff, Rumdl, Prettier, `ty`, `git diff --check`, and
  `uvx nox -s lint`: passed.

AI assistance: Codex assisted with implementation, review, testing, stack
construction, and this description.

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [x] Any agent that created, edited, or submitted GitHub content was explicitly authorized for that scope, as required by our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).
- [x] Every agent-authored or agent-edited public text body begins with the visible disclosure `🤖 *AI text below* 🤖` (titles are exempt).
- [x] I have disclosed AI assistance in the PR description.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

PR #2150, opened 2026-08-18
✨ Support symbolic Qiskit parameters

simon1hofmann:
🤖 *AI text below* 🤖

## Description

- Represent free Qiskit parameters as named `f64` compiler inputs and lower a
  bounded, frontend-neutral real expression tree to Arith/Math SSA.
- Preserve symbol identity and sharing, including equal display names with
  distinct identities, partially bound expressions, global phase, custom
  definitions, and lexical loop bindings.
- Reconstruct one shared Qiskit `Parameter` per compiler input and reject
  malformed, unsupported, non-finite, excessive, or unused named inputs during
  preflight without mutating either source object.
- Keep exact `ParameterVector` provenance out of this PR. The final stack item
  adds that metadata after structured-control export is in place.

The implementation uses the first-class CBit model introduced by #2158.

Closes #2067.

## Stack

1. #2150 — symbolic scalar parameters (this PR)
2. #2175 — captured classical expressions
3. #2176 — structured-control export for #2071
4. #2177 — measurement-store export
5. #2178 — parameter-vector provenance

## Testing

- Release Qiskit binding build: passed.
- `test/python/test_mlir_qiskit_translation.py`: 158 passed.
- Focused Clang format, Ruff, Rumdl, and committed-diff checks: passed.
- `uvx nox -s lint`: passed.

AI assistance: Codex assisted with implementation, review, testing, stack
construction, and this description.

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [x] Any agent that created, edited, or submitted GitHub content was explicitly authorized for that scope, as required by our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).
- [x] Every agent-authored or agent-edited public text body begins with the visible disclosure `🤖 *AI text below* 🤖` (titles are exempt).
- [x] I have disclosed AI assistance in the PR description.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

Issue #2071, opened 2026-08-12
✨ Export structured control flow to Qiskit


burgholzer:
🤖 *AI text below* 🤖

## Goal

Export compiler programs with structured control flow and runtime classical expressions to Qiskit. #2031 imports these constructs but keeps export flat because the Qiskit 2.5 C API provides inspection without the corresponding construction API.

## Implementation direction

- Extend export preflight and construction only after the supported Qiskit C API exposes constructors for the required control-flow operations and classical expressions.
- Map compiler regions and block arguments to Qiskit `if`/`else`, `for`, `while`, and `switch` constructs while preserving lexical scope, loop induction values, and resource capture.
- Reconstruct representative Bool, Uint up to 64 bits, and Float expressions with explicit type and width checks.
- Keep all version-specific construction calls in the version-specific translation unit and expose frontend-neutral writer operations to the generic exporter.
- Do not add a Python-object fallback to the native translation path. Unsupported Qiskit versions must continue to fail before allocating an output circuit.

## Edge cases

- Nested regions, empty branches, fallthrough behavior, switch defaults, loop step direction, zero iterations, captured bits, and shadowed induction variables.
- Expression width changes, signedness assumptions, non-finite floating-point constants, and operations without a Qiskit construction equivalent.
- Partially constructible programs must be rejected by preflight so no incomplete output circuit is allocated.

## Acceptance tests

- Nested round trips for `if`/`else`, `for`, `while`, and `switch` with representative Bool, Uint, and Float expressions.
- Tests for lexical binding and capture across nested regions.
- Version-dispatch tests proving that construction is enabled only where the required C API is available.
- Preflight tests for unsupported region or expression forms and checks that source programs remain unchanged.

Part of #917.

PR #2178, opened 2026-08-20
✨ Preserve Qiskit parameter-vector provenance

simon1hofmann:
🤖 *AI text below* 🤖

## Description

- Preserve genuine Qiskit `ParameterVectorElement` grouping, vector name,
  declared size, numeric index, and parameter order across compiler translation.
- Store provenance in one verified `mqt.input_group` dictionary while keeping
  similarly named standalone parameters scalar.
- Reuse the existing root/nested symbol table and perform one in-place
  root-level replacement, so structured-control blocks share each restored
  vector without recursive collection or an extra circuit copy.
- Keep vector-element loop variables lexical and bound individual plus aggregate
  declared vector sizes before allocation.

This is the optional leaf of the Qiskit translation stack. The refreshed
implementation is based directly on #2176 and removes the obsolete recursive
restoration machinery from the earlier version.

## Stack

1. #2150 — symbolic scalar parameters (merged)
2. #2175 — captured classical expressions (merged)
3. #2176 — structured-control and forwarded measurement-result export for #2071
4. #2178 — parameter-vector provenance (this PR)

## Testing

- Qiskit translation tests: 229 passed.
- Native MQT IR tests: 10 passed.
- Python stub regeneration: passed with no generated diff.
- Full repository lint: passed.
- `git diff --check`: passed.

AI assistance: Codex assisted with implementation, simplification, review,
testing, stack updates, and this description.

## Checklist


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
```diff
@@ -82,6 +84,43 @@ verifyEntryPoint(Operation* operation, const NamedAttribute attribute) {
...
+    return operation->emitError()
+           << "parameter input-group index and size must be nonnegative i64 "
+              "integers";
+  }
+  const auto expectedName =
+      groupName.str() + "[" + std::to_string(groupIndex.getInt()) + "]";
+  if (inputName.getValue() != expectedName) {
+    return operation->emitError()
+           << "parameter input name must match its group name and index";
+  }
```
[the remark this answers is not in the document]
simon1hofmann: Is this fixed?

bindings/mlir/qiskit/Qiskit2_5.cpp
```diff
@@ -1897,7 +1927,17 @@ class PythonClassicalBuilder final {
...
+  NativeSymbol(const std::string_view name,
+               std::optional<ParameterGroup> sourceGroup)
+      : group(std::move(sourceGroup)), parameter(name) {}
+
+  std::optional<ParameterGroup> group;
+  OwnedParameter parameter;
+};
+
+using NativeSymbolTable = std::unordered_map<std::string, NativeSymbol>;
+using PythonParameterGroups = std::unordered_map<std::string, nb::object>;
```
burgholzer: Should these potentially be llvm stringmaps? I would imagine that those are better optimized data structures. May apply in more cases than just these two.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
```diff
@@ -21,6 +21,10 @@ def MQTDialect : Dialect {
     across quantum dialect conversions. It defines no operations or types.
 
     `mqt.input_name` records the source-level name of a function input.
+    `mqt.input_group` optionally preserves its source-level vector identity,
+    name, element index, and size.
+    `mqt.loop_parameter_group` preserves the same metadata for a lexically
+    bound `scf.for` parameter.
```
burgholzer: What bothers me a little bit about these additions is that they are now Qiskit specific. OpenQASM import does not make use of these attributes.
Aren't parameter vectors just another kind of register? (With float values)
And couldn't we treat them as such?
Does that line of thought buy us anything?
I am a bit worried that we are too lenient in adding attributes here that then effectively define how QC IR looks like.

burgholzer (changes requested):
Didn't take a very detailed look here yet, but wanted to get first feedback out as early as possible.
As you will probably get from one of the comments, I am a bit worried whether this is the right way to handle this.
While it does allow for roundtrips, it does not generally add meaningful value for other input or output formats.
I didn't kick off my review agent for this yet, but maybe you can already think about whether you could also see a different solution here that does not introduce two new attributes.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
simon1hofmann: 🤖 *AI text below* 🤖

Thanks, I agree we should be conservative about growing the MQT/QC IR surface. These attributes are discardable source-provenance metadata, not runtime storage or new QC semantics. A circuit may use only a sparse subset of a `ParameterVector`; an element may have an index outside the current vector size; and an element can be a lexically bound loop parameter. A float register would impose storage, shape, bounds, and ABI semantics that do not describe those cases.

OpenQASM import does not attach equivalent metadata today because our frontend currently supports neither `input` nor `array` declarations, and QC-to-OpenQASM rejects function arguments. The dictionary itself contains only frontend-neutral group identity, name, element index, and size; no Qiskit-specific type or behavior. Its purpose is to preserve optional source grouping across QC/QCO conversions while leaving each executable parameter as an `f64` scalar.

If the remaining concern is specifically the two attribute names, I can consolidate them into one generic `mqt.parameter_group` attribute valid on function arguments and `scf.for`. I would avoid modeling them as float registers because that changes the IR semantics and still does not cover lexical or out-of-size elements.

burgholzer: Hm. I see the point. And thinking about it with a clearer mind, we will likely make use of the same attribute(s) when we support the respective inputs for OpenQASM import.
I believe I am convinced that this is generally a sound idea. I like the idea of a consolidated `mqt.parameter_group` attribute. Could you try to set this up?

simon1hofmann: Set that up in the two newest commits and also think that this is the most practical way to move forward.

burgholzer (approved):
Yeah, this looks much better now! 👍🏼 
Let's get this in!

PR #2179, opened 2026-08-20
✨ Preserve classical control through target mapping

simon1hofmann:
🤖 *AI text below* 🤖

## Description

This PR makes target mapping preserve structured classical control without
defining target capability policy:

- Scalarize supported static rank-one qubit tensors carried through `qco.if`.
- Keep unary structured operations as routing boundaries and select the qubit
  values that actually cross those boundaries.
- Restore SSA dominance with deterministic dependency ordering that preserves
  observable CBit and memref access order.
- Fail mapping without a partial reorder if the combined SSA and
  classical-memory dependencies contain a cycle.

This is the focused MQT Core 4.0 scope. It intentionally does not add
`CompilerTarget::ClassicalControl`, target-capability legality validation,
complete-topology identity placement, zero-SWAP preview optimizations, or
payload/QDMI 1.4 work. Target acceptance policy can therefore evolve separately
from the mapper's semantic-correctness guarantees.

## Preserved history

Before rewriting the PR branch, the full prior stack was preserved exactly:

- Original base: [`25b7eb8ef50db57abe0de94044821a160f384b89`](https://github.com/munich-quantum-toolkit/core/commit/25b7eb8ef50db57abe0de94044821a160f384b89)
- Original head: [`9f7a0eb08bb410f67312c802f212b544e0f2a547`](https://github.com/munich-quantum-toolkit/core/commit/9f7a0eb08bb410f67312c802f212b544e0f2a547)
- Archive ref: [`archive/pr-2179-full-classical-control-2026-09-01`](https://github.com/munich-quantum-toolkit/core/tree/archive/pr-2179-full-classical-control-2026-09-01), verified at the original head above

The range-diff drops the target-capability, complete-topology, and target-check
alignment commits and rewrites only the structured-control preservation commit
onto current `main`. The archive retains those deferred changes for future
extraction without keeping them in the Core 4.0 critical path.

## Testing

- QCO IR: 494 tests passed.
- QCO utilities: 138 tests passed.
- Mapping: 89 tests passed.
- Compiler: 138 tests passed.
- Changed-source C++ lint: 0 clang-format and 0 clang-tidy findings.
- `uvx nox -s lint`: passed.
- `git diff --check`: passed.

AI assistance: Codex assisted with the rescope, implementation review, testing,
and this description. The original human contribution and commit authorship
remain attributed to Simon Hofmann.

## Checklist

### DIFF
--- a/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
@@ -334,12 +334,6 @@ struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
   }
 };
 
-} // namespace
-
-namespace mlir::qco {
-
-namespace {
-
 struct QTensorAccess {
   qtensor::ExtractOp extract;
   qtensor::InsertOp insert;
@@ -438,14 +432,6 @@ static void moveScalarizedQTensorBranch(IfOp oldIf, Block* oldBlock,
   assert(carriedIndex == carriedArguments.size());
   rewriter.mergeBlocks(oldBlock, newBlock, argumentReplacements);
 
-  for (auto [indexPosition, index] : llvm::enumerate(indices)) {
-    auto access = accesses.accesses.find(index);
-    if (access != accesses.accesses.end()) {
-      rewriter.replaceAllUsesWith(access->second.extract.getResult(),
-                                  scalarArguments[indexPosition]);
-    }
-  }
-
   SmallVector<Value> scalarYields;
   scalarYields.reserve(indices.size());
   for (auto [indexPosition, index] : llvm::enumerate(indices)) {
@@ -453,6 +439,8 @@ static void moveScalarizedQTensorBranch(IfOp oldIf, Block* oldBlock,
     if (access == accesses.accesses.end()) {
       scalarYields.push_back(scalarArguments[indexPosition]);
     } else {
+      rewriter.replaceAllUsesWith(access->second.extract.getResult(),
+                                  scalarArguments[indexPosition]);
       scalarYields.push_back(access->second.insert.getScalar());
     }
   }
@@ -478,8 +466,6 @@ static void moveScalarizedQTensorBranch(IfOp oldIf, Block* oldBlock,
     rewriter.eraseOp(operation);
   }
 }
-
-} // namespace mlir::qco
 
 namespace {
 
@@ -511,18 +497,11 @@ struct ScalarizeQTensorInputs final : OpRewritePattern<IfOp> {
         continue;
       }
 
-      SmallVector<int64_t> accessedIndices;
-      accessedIndices.reserve(thenAccesses->accesses.size() +
-                              elseAccesses->accesses.size());
-      for (int64_t index : thenAccesses->accesses.keys()) {
-        accessedIndices.push_back(index);
-      }
-      for (int64_t index : elseAccesses->accesses.keys()) {
-        if (!thenAccesses->accesses.contains(index)) {
-          accessedIndices.push_back(index);
-        }
-      }
+      SmallVector<int64_t> accessedIndices(thenAccesses->accesses.keys());
+      llvm::append_range(accessedIndices, elseAccesses->accesses.keys());
       llvm::sort(accessedIndices);
+      accessedIndices.erase(llvm::unique(accessedIndices),
+                            accessedIndices.end());
       ArrayRef<int64_t> indices(accessedIndices);
 
       rewriter.setInsertionPoint(op);
@@ -542,13 +521,8 @@ struct ScalarizeQTensorInputs final : OpRewritePattern<IfOp> {
         qTensorWithoutScalars = extract.getOutTensor();
       }
 
-      SmallVector<Value> newQubits;
-      newQubits.reserve(oldQubits.size() - 1 + scalarInputs.size());
-      for (auto [oldIndex, qubit] : llvm::enumerate(oldQubits)) {
-        if (oldIndex != qTensorIndex) {
-          newQubits.push_back(qubit);
-        }
-      }
+      SmallVector<Value> newQubits(oldQubits);
+      newQubits.erase(newQubits.begin() + qTensorIndex);
       llvm::append_range(newQubits, scalarInputs);
 
       auto newIf = IfOp::create(
@@ -581,16 +555,12 @@ struct ScalarizeQTensorInputs final : OpRewritePattern<IfOp> {
                 .getResult();
       }
 
-      SmallVector<Value> replacements;
-      replacements.reserve(op.getNumResults());
-      llvm::append_range(replacements, newIf.getClassicalResults());
-      size_t newLinearIndex = 0;
-      for (size_t oldIndex : llvm::seq(oldQubits.size())) {
-        replacements.push_back(
-            oldIndex == qTensorIndex
-                ? updatedQTensor
-                : newIf.getLinearResults()[newLinearIndex++]);
-      }
+      SmallVector<Value> replacements(
+          newIf.getLinearResults().drop_back(indices.size()));
+      replacements.insert(replacements.begin() + qTensorIndex, updatedQTensor);
+      replacements.insert(replacements.begin(),
+                          newIf.getClassicalResults().begin(),
+                          newIf.getClassicalResults().end());
       rewriter.replaceOp(op, replacements);
       return success();
     }
--- a/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
@@ -360,9 +360,7 @@ analyzeQTensorBranch(Block* block, size_t qTensorArgumentIndex,
   bool reachedInsertPhase = false;
 
   while (true) {
-    if (!currentQTensor.hasOneUse()) {
-      return std::nullopt;
-    }
+    assert(currentQTensor.hasOneUse() && "expected linear typing");
     Operation* user = *currentQTensor.getUsers().begin();
     if (user->getBlock() != block) {
       return std::nullopt;
--- a/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
+++ b/mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
@@ -10,13 +10,17 @@
 
 #include "mlir/Dialect/QCO/IR/QCOOps.h"
 #include "mlir/Dialect/QCO/QCOUtils.h"
+#include "mlir/Dialect/QTensor/IR/QTensorOps.h"
 
 #include <llvm/ADT/BitVector.h>
+#include <llvm/ADT/DenseMap.h>
 #include <llvm/ADT/STLExtras.h>
 #include <llvm/ADT/STLFunctionalExtras.h>
 #include <llvm/ADT/Sequence.h>
+#include <llvm/ADT/SmallPtrSet.h>
 #include <llvm/ADT/SmallVector.h>
 #include <mlir/Dialect/Arith/IR/Arith.h>
+#include <mlir/Dialect/Utils/StaticValueUtils.h>
 #include <mlir/IR/Attributes.h>
 #include <mlir/IR/Builders.h>
 #include <mlir/IR/BuiltinTypes.h>
@@ -31,7 +35,9 @@
 #include <mlir/Support/LLVM.h>
 
 #include <cassert>
+#include <cstddef>
 #include <cstdint>
+#include <optional>
 
 using namespace mlir;
 using namespace mlir::qco;
@@ -153,19 +159,9 @@ void IfOp::getRegionInvocationBounds(
   }
 }
 
-/**
- * @brief Replace operation with the contents of a region
- *
- * @details
- * Replaces the given op with the contents of the given single-block region,
- * using the operands of the block terminator to replace operation results.
- *
- * @param rewriter The used rewriter
- * @param op The operation that is replcaed
- * @param region The region with the replacement content
- * @param blockArgs The block arguments of the region
- *
- */
+/// Replace an operation with the contents of a single-block region.
+///
+/// Use the block terminator operands to replace the operation results.
 static void replaceOpWithRegion(PatternRewriter& rewriter, Operation* op,
                                 Region& region, ValueRange blockArgs = {}) {
   assert(llvm::hasSingleElement(region) && "expected single-region block");
@@ -179,14 +175,7 @@ static void replaceOpWithRegion(PatternRewriter& rewriter, Operation* op,
 
 namespace {
 
-/**
- * @brief Remove static conditions
- *
- * @details
- * Removes a qco.if operation with a static condition and replace it with the
- * contents of the selected branch.
- *
- */
+/// Replace an if with a static condition by its selected branch.
 struct RemoveStaticCondition : public OpRewritePattern<IfOp> {
   using OpRewritePattern<IfOp>::OpRewritePattern;
 
@@ -207,22 +196,7 @@ struct RemoveStaticCondition : public OpRewritePattern<IfOp> {
   }
 };
 
-/**
- * @brief Propagate the condition into the branches
- *
- * @details
- * Allow the true region of an if to assume the condition is true
- * and vice versa. For example:
- *
- *   qco.if %cmp args(%arg0 = %q0) -> (!qco.qubit) {
- *      print(true)
- *      ...
- *   } else args(%arg = %q0) {
- *      print(false)
- *      ...
- *   }
- *
- */
+/// Let each branch use the condition value known inside that branch.
 struct ConditionPropagation : public OpRewritePattern<IfOp> {
   using OpRewritePattern<IfOp>::OpRewritePattern;
 
@@ -271,16 +245,11 @@ struct ConditionPropagation : public OpRewritePattern<IfOp> {
   }
 };
 
-/**
- * @brief Forward redundant classical results
- *
- * @details
- * Replaces a classical result with a value yielded by both branches or with an
- * earlier classical result whose pair of yielded values is identical. A
- * separate pattern removes the result and its yield operands once they become
- * unused. Linear results are intentionally excluded because their explicit
- * branch threading is part of QCO's quantum dataflow.
- */
+/// Forward redundant classical results.
+///
+/// Replace a result with a value yielded by both branches or with an earlier
+/// result whose pair of yielded values is identical. A separate pattern removes
+/// unused results. Linear results retain their explicit quantum dataflow.
 struct ForwardClassicalResults : public OpRewritePattern<IfOp> {
   using OpRewritePattern<IfOp>::OpRewritePattern;
 
@@ -320,14 +289,9 @@ struct ForwardClassicalResults : public OpRewritePattern<IfOp> {
   }
 };
 
-/**
- * @brief Remove unused classical results
- *
- * @details
- * Removes unused classical results and the corresponding operands from both
- * branch terminators. The result segment property is updated on the replacement
- * operation. The linear result suffix and all quantum dataflow remain intact.
- */
+/// Remove unused classical results and their branch yield operands.
+///
+/// Update the result segments while preserving the linear result suffix.
 struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
   using OpRewritePattern<IfOp>::OpRewritePattern;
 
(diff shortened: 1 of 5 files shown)


mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1564,8 +1813,29 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+                  dispatch<Direction, RoutingMode::Cold>(composite, bundle);
+              if (failed(preview)) {
+                return failure();
+              }
+              needsWorkspace = preview->second.nswaps != 0;
+            }
+            if (needsWorkspace) {
+              auto patch = place(composite, bundle, *rewriter);
+              bundle.applyPatch(std::move(patch));
+            }
```
MatthiasReumann: FYI: Just because I remembered seeing this when I reviewed this PR for the first time: I think this "needsWorkspace" check should be done once in the `runOnOperation` function and simply call place once with an identity mapping for all-to-all architectures. 

Once I am done with the tailored topological-sort, I wanted to optimize this "hot-path" anyway. So this could also be removed from this PR.

burgholzer: 🤖 *AI text below* 🤖

Addressed at the shared ownership boundary. `runOnOperation` now detects complete connectivity once, installs the identity placement, and skips both initial-layout search and hot routing. Sparse targets still use the existing preview path. The focused mapping, compiler, QCO IR, and QCO utility suites pass on the detached stack.

burgholzer:
🤖 *AI text below* 🤖

pre-commit.ci run

simon1hofmann:
This PR now somehow includes an old state of #2162.

burgholzer:
Hm. I wanted to disentangle the series from here on out from the lower chain. Might be that something went wrong with that. I'll try to fix that up once the main PR series gets closer to mergable.

mlir/include/mlir/Dialect/QCO/Utils/Drivers.h
```diff
@@ -44,17 +44,18 @@ using WalkProgramGraphFn =
...
+ * the function traverses the def-use chain of each qubit until a routing
+ * boundary is found. Routing boundaries include multi-qubit gates, barriers,
+ * and structured control-flow operations even when they carry only one qubit.
+ * A boundary is ready once each required input wire has visited it. This
+ * process is repeated until no more routing boundaries are found.
  *
  * The signature of the callback function is:
  *
  *     (const ReadyMap& ready, ReleasedOps& released) -> WalkResult
  *
```
denialhaag, line 55: Should this be wrapped by a see below? 🤔 

````
```c++
```
````

File: mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp [hunk 3]
```diff
@@ -369,12 +376,302 @@ struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
...
+  while (true) {
+    if (!currentTensor.hasOneUse()) {
+      return std::nullopt;
+    }
+    Operation* user = *currentTensor.getUsers().begin();
+    if (user->getBlock() != block) {
+      return std::nullopt;
+    }
+
+    if (auto extract = dyn_cast<qtensor::ExtractOp>(user)) {
```
denialhaag, line 369: This and the two if statements could be rewritten using `llvm::TypeSwitch`, but I'm not certain that that would be cleaner.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -369,12 +376,302 @@ struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
...
+        llvm::any_of(result.accesses, [](const TensorAccess& access) {
+          return !access.outputQubit;
+        })) {
+      return std::nullopt;
+    }
+    return result;
+  }
+}
+
+/** Return the access for `index`, if the branch touches that tensor element. */
```
denialhaag: If we update this PR one more time anyway, we could adapt all of the docstrings to the new style. Otherwise, we can also do that when we update all docstrings.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -369,12 +376,302 @@ struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
...
+                                PatternRewriter& rewriter) const override {
+    const auto classicalResultCount = op.getClassicalResults().size();
+    const auto oldQubits = op.getQubits();
+
+    for (const auto [tensorIndex, tensor] : llvm::enumerate(oldQubits)) {
+      const auto tensorType = dyn_cast<RankedTensorType>(tensor.getType());
+      if (!tensorType || !tensorType.hasStaticShape() ||
+          tensorType.getRank() != 1) {
+        continue;
+      }
```
denialhaag: Should this also escape early if the element type is not a `QubitType`?

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -369,12 +376,302 @@ struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
...
+ * the accessed qubits as ordinary scalar wires. This pattern extracts the
+ * union of the constant indices accessed by either branch before the if,
+ * threads only those elements as scalar qubits through both branches, and
+ * inserts the results afterwards. A tensor untouched in both branches is
+ * forwarded around the if without extracting any element. Keeping untouched
+ * elements out of the scalar operand list is important for large registers:
+ * target mapping establishes the required sequencing for the remaining
+ * physical wires when it expands the composite operation.
+ */
+struct ScalarizeTensorInputs final : OpRewritePattern<IfOp> {
```
denialhaag: A more general comment: If I'm not mistaken right now, all of this is only for (q)tensors representing qubit registers, right? Maybe it makes sense to adapt some docstrings and names to better reflect that. 🤔

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -68,6 +73,147 @@ using namespace mlir::qtensor;
...
 
+namespace detail {
+
+/** Return whether an operation or its children access classical storage. */
+static bool containsClassicalMemoryAccess(Operation* operation) {
+  bool containsAccess = false;
+  operation->walk([&](Operation* nested) {
+    const auto isClassicalStorage = [](Value value) {
+      const Type type = value.getType();
+      return isa<cbit::RegisterType>(type) || isa<BaseMemRefType>(type);
```
denialhaag: Should `BaseMemRefType` really be included here? 🤔

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -68,6 +73,147 @@ using namespace mlir::qtensor;
...
+
+    if (const auto successor = nextClassicalMemoryAccess.find(operation);
+        successor != nextClassicalMemoryAccess.end() &&
+        seen[successor->second] != sourceIndex) {
+      seen[successor->second] = sourceIndex;
+      release(successor->second);
+    }
+  }
+
+  /// A cycle can include both SSA edges and the classical-access chain.
```
denialhaag:
```suggestion
  // A cycle can include both SSA edges and the classical-access chain.
```

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -408,8 +554,16 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
     const auto stats = *routeRes;
     numSwaps += stats.nswaps;
 
-    // Fix SSA Dominance issues.
-    llvm::for_each(body.getBlocks(), [](Block& b) { sortTopologically(&b); });
+    /// Fix SSA dominance without changing classical memory order.
```
denialhaag:
```suggestion
    // Fix SSA dominance without changing classical memory order.
```

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
denialhaag, line 1: In addition to the two explicit comments, throughout this file, make sure that code comments use `//` and not `///`.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -981,6 +981,393 @@ TEST_F(QCOTest, CanonicalizesRedundantClassicalIfResults) {
...
+          %false = arith.constant false
+          qco.yield %false, %tensor3 : i1, tensor<2x!qco.qubit>
+        } {test.marker = "preserved"}
+        qtensor.dealloc %tensor1 : tensor<2x!qco.qubit>
+        return %flag : i1
+      }
+    }
+  )mlir";
+
+  auto module = parseSourceString<ModuleOp>(mlirCode, context.get());
```
denialhaag:
```suggestion
  auto moduleOp = parseSourceString<ModuleOp>(mlirCode, context.get());
```

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
```diff
@@ -981,6 +981,393 @@ TEST_F(QCOTest, CanonicalizesRedundantClassicalIfResults) {
...
+        } else args(%arg0 = %tensor0) {
+          qco.yield %arg0 : tensor<3x!qco.qubit>
+        }
+        qtensor.dealloc %tensor1 : tensor<3x!qco.qubit>
+        return
+      }
+    }
+  )mlir";
+
+  auto module = parseSourceString<ModuleOp>(mlirCode, context.get());
```
denialhaag:
```suggestion
  auto moduleOp = parseSourceString<ModuleOp>(mlirCode, context.get());
```

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
denialhaag, line 1: In addition to the two explicit comments, use `moduleOp` instead of `module`.

denialhaag (changes requested):
I admittedly didn't go through all of the logic, but this generally LGTM. I just have a few comments, none of which should be a major blocker.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer: Yeah, agreed. Let's do that and follow the new AGENTS.md guidance 👍🏼

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
burgholzer, line 1: @simon1hofmann this should likely use the fix by @MatthiasReumann instead, right?

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
burgholzer, line 1: It feels like this could be much simpler. Is this really the most efficient way to solve this?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer, line 1: Feels like these changes won't be necessary with @MatthiasReumann's solution, right?

burgholzer (changes requested):
Might have been a bit premature to declare this as ready for review. This still has quite a few dependencies on other open PRs it seems.
Might be worth getting those in first before handing this one back to Codex with all the feedback.

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
simon1hofmann: Yes, this branch still includes some code that was moved to other PRs. Already cleaned up some of them, but the mapping changes can removed here, as they're already covered by @MatthiasReumann 's PRs.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
simon1hofmann: The changes to the topological sorting can be removed, the rest should be crosschecked with the other open PRs.

mlir/include/mlir/Dialect/QCO/Utils/Drivers.h
simon1hofmann: 🤖 *AI text below* 🤖

Removed the `Drivers.h` change from this PR. #2205 replaces this traversal, so the callback documentation and its tests now remain unchanged from `main`.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Kept the explicit chain walk. The analysis carries state across the extract/insert chain and needs a clear failure path back to the outer pattern; `llvm::TypeSwitch` would add machinery without reducing that control flow.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Updated all Doxygen comments in `IfOp.cpp` to the current `///` style from the repository guidance.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Did not add a separate element-type guard. The verified `qco.if` linear-type constraints and `qtensor.extract`/`qtensor.insert` constraints already restrict this path to qubit tensors, so the guard would be redundant for valid IR.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Renamed the new analysis structures, helpers, variables, pattern, and documentation around `QTensor` and explicitly described the qubit-specific behavior.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Removed the custom topological sorter from this PR after cross-checking the open stack, so the `BaseMemRefType` handling is gone as well.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Removed this code together with the custom topological sorter. Topological reordering remains in #2240.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Removed this code together with the custom topological sorter. This PR no longer claims to restore SSA dominance while preserving classical-memory order.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Applied the distinction throughout the retained mapping changes: helper documentation uses `///`, while explanatory code comments use `//`. The superseded sorter comments were removed.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Renamed this local and every other test local introduced by this PR to `moduleOp`.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Renamed this local and every other test local introduced by this PR to `moduleOp`.

mlir/unittests/Dialect/QCO/IR/test_qco_ir.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Applied `moduleOp` consistently to all test locals introduced by this PR.

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
simon1hofmann: 🤖 *AI text below* 🤖

Removed all changes to `Mapping.h`. The custom sorter is owned by #2240, and this header now matches `main`.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Simplified the implementation substantially: branch access data now uses `DenseMap`, and branch bodies are moved with `mergeBlocks` instead of cloned through `IRMapping`. The analysis stores extract/insert operation handles so replacements also remain correct for identity threading and cross-index swaps.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Cross-checked `main` and #2184, #2185, #2205, and #2240. Removed the custom topological sorter, walker changes, CBit/memref-order behavior, and their tests. Retained only the boundary-wire selection and composite-deferral fixes that are absent from that stack, plus the target-workspace regression.

simon1hofmann:
@coderabbitai full review

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
denialhaag: Makes sense! I forgot that `qco::IfOp` only accepts linear types at the moment.

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
denialhaag: Makes sense!

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
denialhaag: Looks good!

denialhaag (approved):
Thanks a lot for addressing all of my comments, @simon1hofmann! I only cross-checked that they are all resolved now, but I didn't check the full PR again. I'm approving now so this doesn't get help up. If you want me to have another look, I'm happy to do so tomorrow morning! 🙂

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
```diff
@@ -369,12 +333,277 @@ struct RemoveUnusedClassicalResults : public OpRewritePattern<IfOp> {
...
+static std::optional<BranchQTensorAccesses>
+analyzeQTensorBranch(Block* block, size_t qTensorArgumentIndex,
+                     size_t qTensorYieldIndex) {
+  BranchQTensorAccesses result;
+  Value currentQTensor = block->getArgument(qTensorArgumentIndex);
+  bool reachedInsertPhase = false;
+
+  while (true) {
+    if (!currentQTensor.hasOneUse()) {
+      return std::nullopt;
```
burgholzer: This call is unreachable by design

burgholzer (approved):
Simplified this a bit further. But otherwise this also looks good to me now and ready to go in. We may want to eliminate that one check that is true by design.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1261,6 +1290,26 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+        if (operation == nullptr || operation == composite.op) {
+          return false;
+        }
+        assert(operation->getBlock() == composite.op->getBlock());
+        if constexpr (Direction == WireDirection::Forward) {
+          return operation->isBeforeInBlock(composite.op);
+        }
+        return composite.op->isBeforeInBlock(operation);
+      });
+    });
```
MatthiasReumann, line 1311: Could you explain to me why this is necessary? Also what is the traversal direction?

Moreover, I am not sure if the `isBeforeInBlock` check here is correct. I haven't thought it through thoroughly but make sure that this doesn't make the mistake that the order of operations in the IR (in block order) is the same as from a DAG perspective. 

```
cx q0, q1
cx q2, q3
cx q1, q4
```

describes the same graph as 

```
cx q0, q1
cx q1, q4
cx q2, q3
```

The mapping pass should process the DAG "not the IR".

mlir/lib/Dialect/QCO/IR/SCF/IfOp.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Replaced the unreachable failure branch with an assertion of the verifier-owned QTensor linearity invariant. Valid QTensor values have exactly one use, so the analysis now proceeds directly to that user.

burgholzer (approved):
Looks reasonable to me for now!
Let's get this in! :shipit:

PR #2180, opened 2026-08-20
⚡ Avoid redundant target-compilation liveness analysis

simon1hofmann:
🤖 *AI text below* 🤖

## Summary

Remove the final module-wide `remove-dead-values` pass from target compilation. The second QCO cleanup immediately before native synthesis already establishes global liveness, while native synthesis replaces one- and two-qubit unitary operations with native equivalents and reconnects their results.

Post-synthesis CSE and final target-conformance verification remain unchanged.

## Motivation

The production-shaped profile in #2134 measured the final `RemoveDeadValues` pass at 7.77 seconds of a 29.93-second target-compilation run (about 26%), including 4.09 seconds in liveness analysis, for `hwb10` targeting `FakeTorino`.

This PR removes that measured cleanup stage without changing placement, routing, native-synthesis choices, or target-conformance checks. It addresses the cleanup portion of #2134.

## Safety and testing

The added end-to-end regression compiles a program containing a private helper call, dynamic `qco.if`, and one- and two-qubit synthesis. It then reruns `RemoveDeadValues` and requires the IR to remain byte-for-byte unchanged, protecting the intended fixed point.

- `mqt-core-mlir-unittests-compiler`: 140/140 passed.
- `uvx nox -s lint`: passed.
- `uvx nox -s cpp-lint` with LLVM/Clang 22: 0 findings.
- `git diff --check`: passed.

## Stack

- Base: `main` (includes #2179)
- Follow-ups: #2181 → #2297

AI assistance: Codex assisted with diagnosis, implementation, review, testing, stack maintenance, and this description.

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [x] Any agent that created, edited, or submitted GitHub content was explicitly authorized for that scope, as required by our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/core/blob/main/docs/ai_usage.md).
- [x] Every agent-authored or agent-edited public text body begins with the visible disclosure `🤖 *AI text below* 🤖` (titles are exempt).
[...]

PR #2181, opened 2026-08-20
✨ Canonicalize classical register comparisons across frontends

simon1hofmann:
🤖 *AI text below* 🤖

## Description

- Introduce `cbit.cmp` as the canonical QC representation for comparing a complete classical register with a same-width unsigned constant.
- Parse and emit all OpenQASM 3 register comparisons (`==`, `!=`, `<`, `<=`, `>`, and `>=`) and lower them through the MemRef, QIR, Jeff, and DD consumers.
- Normalize Qiskit tuple and typed-expression conditions during import so equivalent Qiskit and OpenQASM programs produce the same QC IR when the register layout permits, and emit typed expressions during Qiskit export.
- Treat out-of-range equality constants according to ordinary unsigned semantics instead of retaining widened legacy behavior; use generic packing only for genuinely computed or remapped expressions.
- Keep wide and repeated conditions linear by sharing pure register-condition roots within an unchanged classical snapshot and folding comparisons against untouched zero-initialized registers.

## Stack

1. #2179 — classical-control-aware target mapping (merged)
2. #2180 — target-compilation liveness scaling (merged)
3. #2181 — canonical classical-register comparisons (this PR)
4. #2297 — bit-register integer casts (follow-up)

## Testing

- `uvx nox -s lint`: passed.
- `uvx nox -s cpp-lint`: passed with zero findings.
- `test/python/test_mlir_qiskit_translation.py`: 244 passed.
- `mqt-core-mlir-unittest-cbit-ir`: 9 passed.
- `release-no-mlir`: 511 passed, 1 skipped.

AI assistance: Codex assisted with implementation, review, testing, stack construction, and this description.

## Checklist


simon1hofmann:
@coderabbitai full review

docs/mlir/OpenQASM.md
```diff
@@ -145,16 +149,24 @@ QCO optimization pipeline, converts back to QC, and then exports. Calling
 {code}`mlir::QCProgram::toOpenQASM3` applies the QC cleanup pipeline but
 bypasses that QCO optimization round trip.
 
+For measurement-conditioned OpenQASM 2 programs, target compilation can expose
+the frontend's full bit-register equality as a classical SSA expression. The
+exporter recognizes that exact unchanged expression, fuses eligible direct
+measurement stores, and emits one register comparison. The constant is not
+limited to a machine integer, so this compatibility path also supports registers
+wider than 64 bits. Other expression shapes continue through the normal support
+checks below.
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Updated the documentation to use frontend/version-neutral wording. It now states that OpenQASM 3 follows the same compatibility path when every compared bit is initialized.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -2187,9 +2287,11 @@ class OpenQASMToQCEmitter {
         static_cast<int64_t>(declaration.width), declaration.name,
         program.openQASM2 ? cbit::Initialization::Zero
                           : cbit::Initialization::Undefined);
+    invalidateConditionCache();
   }
 
   void assignBit(const frontend::BitReference& target, Value value) {
+    invalidateConditionCache();
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Confirmed. The store path invalidates the classical snapshot cache at each classical write, so no code change was needed for this audit note.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
```diff
@@ -383,6 +412,531 @@ class OpenQASMEmitter {
...
+      for (const auto [index, candidate] :
+           llvm::enumerate(deadExpressionCandidates)) {
+        if (activeDead[index] &&
+            hasExternalExpressionUse(candidate.expressionIfs,
+                                     matchedOperations)) {
+          activeDead[index] = false;
+          changed = true;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Kept the current fallback without an arbitrary work budget. Result-bearing SCF is normally rejected, so aborting collection on a guessed limit could reject currently supported register equalities instead of safely falling back. A budget needs measured limits and a defined fallback policy in a separate change.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -2295,6 +2401,11 @@ class OpenQASMToQCEmitter {
     }
     scalarValues = savedScalars;
     assignState(slots, ifOp.getResults());
+    if (classicalStateGeneration == savedClassicalStateGeneration) {
+      conditionCache = savedConditionCache;
+    } else {
+      conditionCache.clear();
+    }
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Kept the explicit save/clear/restore lifecycle. Each region still needs a distinct clear point, so a guard would add an abstraction without reducing the state transitions.

docs/mlir/OpenQASM.md
```diff
@@ -199,9 +211,13 @@ arbitrary CFGs, multi-block SCF regions, dynamic qubit indices or ranges,
...
 unknown operations, and non-unitary content inside modifier regions. CBit loads,
 stores, and dynamic indices are supported. SCF results, loop-carried values,
-nonempty `scf.yield`, and `arith.select` are outside the export subset.
-Multi-operation modifier bodies must have a target qubit and cannot capture
-additional qubits from an enclosing scope.
+nonempty `scf.yield`, and `arith.select` are outside the export subset. The sole
+result-bearing SCF exception is a complete, unchanged bit-register equality over
+CBit storage produced by the OpenQASM 2 frontend. The exporter emits this
+compatibility form as one register comparison and rejects partial, mixed,
+dynamically indexed, or modified register conditions. Multi-operation modifier
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Corrected the limitation. Zero-initialized registers may omit bits that were not changed by an intervening write; uninitialized registers must constrain every bit. Mixed, dynamic, and modified forms remain rejected.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
```diff
@@ -383,6 +412,531 @@ class OpenQASMEmitter {
...
+          }
+          store = candidateStore;
+          continue;
+        }
+        auto consumer = dyn_cast<scf::IfOp>(user);
+        if (!foldedRegisterExpressionOperations.contains(user) &&
+            (!consumer ||
+             !registerEqualities.contains(consumer.getOperation()))) {
+          return;
+        }
```
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in d991c5d9c. Folded register-expression operations are now skipped during generic inline validation, so the fused measurement is not revalidated through the negating expression. Added a width-one equality-to-zero regression with strict OpenQASM analysis.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
```diff
@@ -821,6 +1379,9 @@ class OpenQASMEmitter {
   }
 
   [[nodiscard]] LogicalResult emitStore(cbit::StoreOp store) {
+    if (foldedMeasurementStores.contains(store)) {
+      return success();
+    }
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Removed the unreachable adjacent-measurement branch while preserving the foldedMeasurementStores early return, which prevents duplicate emission after measurement-store fusion.

mlir/unittests/Dialect/QC/Translation/test_openqasm3_emission.cpp
```diff
@@ -268,6 +268,358 @@ switch (selector) {
...
+}
+)mlir";
+  const DialectRegistry registry = emissionDialects();
+  MLIRContext context(registry);
+  auto moduleOp = parseSourceString<ModuleOp>(source, &context);
+  ASSERT_TRUE(moduleOp);
+
+  auto emitted = qc::translateQCToOpenQASM3(*moduleOp);
+
+  ASSERT_TRUE(succeeded(emitted));
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Added strict frontend analysis to each successful emission case in this section and changed fixtures that read or return otherwise-undefined registers to zero initialization, so validation checks fully defined emitted programs.

simon1hofmann:
@coderabbitai review

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
```diff
@@ -383,6 +412,531 @@ class OpenQASMEmitter {
...
+    return llvm::all_of(candidate.constraints, [&](const auto& constraint) {
+      const auto resource = resources.find(constraint.reg);
+      return resource != resources.end() &&
+             resource->second.kind == ResourceKind::Bit &&
+             constraint.index >= 0 && constraint.index < resource->second.width;
+    });
+  }
+
+  [[nodiscard]] static bool
+  hasExternalExpressionUse(const ArrayRef<Operation*> expressionIfs,
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Removed the prohibited top-level const qualifier from the by-value ArrayRef parameter.

mlir/unittests/Dialect/QC/Translation/test_openqasm3_emission.cpp
```diff
@@ -268,6 +268,380 @@ switch (selector) {
...
+  constexpr std::array fixtures{
+      Fixture{.width = 1, .expected = "0"},
+      Fixture{.width = 1, .expected = "1"},
+      Fixture{.width = 64, .expected = "9223372036854775808"},
+      Fixture{.width = 151,
+              .expected = "1427247692705959881058285969449495136382746624"},
+      Fixture{.width = 301,
+              .expected = "2037035976334486086268445688409378161051468393665"
+                          "936250636140449354381299763336706183397376"},
+  };
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

No additional fixture added. CompactsSharedRegisterConditionExpressionTrees already exercises two distinct constraints, and the wide-register fixtures pin nonzero bit positions. A mixed two-bit case would add overlapping coverage and would not exercise duplicate-index rejection, so I kept the test set focused.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Confirmed and fixed in 96a0568e1. Measurement fusion now walks the full folded SSA use chain and declines fusion when any result reaches an unmatched consumer. The regression covers a negated measurement condition that is compacted once, followed by a register write and an unfused reuse. Constant i1 XORs are also emitted canonically so the fallback remains valid OpenQASM.

mlir/unittests/Dialect/QC/Translation/test_openqasm3_emission.cpp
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Good point on the mixed truth values. I strengthened the existing `CompactsSharedRegisterConditionExpressionTrees` regression to require bit 0 = true and bit 1 = false, so it emits `c == 1` while preserving the nested XOR and shared-expression shape. A swapped constraint-to-index association would emit `c == 2` and now fail the test. This is in f07efbc5c.

burgholzer:
looking at this now.

burgholzer (approved):
alright. I took this for quite a spin because I had the impression that this could have been way simpler. turns out that enabled quite a few IR simplifications and broader support for comparisons across OpenQASM and Qiskit. I am happy with this now. @simon1hofmann feel free to take a look yourself and merge it afterwards. I got a few cleanups for the next pr in the stack lined up as well.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
```diff
@@ -588,6 +589,37 @@ class OpenQASMEmitter {
     if (auto load = value.getDefiningOp<cbit::LoadOp>()) {
       return emitBitReference(load.getReg(), load.getIndex());
     }
+    if (auto comparison = value.getDefiningOp<cbit::CompareOp>()) {
```
simon1hofmann, line 666: 🤖 *AI text below* 🤖

**[P1] Preserve the `cbit.cmp` snapshot before inlining it.**

`cbit.cmp` has a read effect, so its SSA result represents the register state at this operation. This code prints the comparison only when a later consumer is emitted. An intervening store therefore changes the value that controls the consumer. I reproduced `store false; %cmp = cbit.cmp eq, %c, 0; store true; scf.if %cmp` exporting as the two stores followed by `if (c == 0)`, which flips the branch result. Please either materialize a Boolean at the comparison point or reject intervening writes, as Qiskit export already does in `validateClassicalSnapshot`. Add an intervening-store regression using direct `cbit.cmp`.

simon1hofmann (commented):
🤖 *AI text below* 🤖

Five reproducible findings from the changes added after `6545b95de`. Each inline comment includes the failing case and the smallest safe direction for a fix.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
```diff
@@ -498,7 +498,8 @@ class OpenQASMEmitter {
 
   [[nodiscard]] static bool isInlineExpressionOperation(Operation& operation) {
     const auto name = operation.getName().getStringRef();
-    return isa<arith::ConstantOp, arith::CmpIOp, arith::CmpFOp>(&operation) ||
+    return isa<arith::ConstantOp, arith::CmpIOp, arith::CmpFOp,
+               cbit::CompareOp>(&operation) ||
```
simon1hofmann, line 576: 🤖 *AI text below* 🤖

**[P2] Permit `cbit.cmp` in `scf.while` condition regions.**

This makes `cbit.cmp` an inline expression, but the operation has a `MemRead` effect. `emitWhile` special-cases only `cbit::LoadOp` before requiring all other inline operations to be memory-effect-free. The frontend now lowers `while (c == 1)` to a `cbit.cmp` in the before-region, so importing and exporting that valid OpenQASM fails with `scf.while condition region must be side-effect free`. Handle `CompareOp` alongside `LoadOp` and add an import/export regression.

mlir/include/mlir/Dialect/CBit/IR/CBitOps.td
```diff
@@ -96,6 +107,28 @@ def LoadOp : CBitOp<"load"> {
...
+    The predicate must be `eq`, `ne`, `ult`, `ule`, `ugt`, or `uge`.
+
+    Example:
+    ```mlir
+    %matches = cbit.cmp eq, %c, 1 : i2 : !cbit.reg<2>
+    ```
+  }];
+
+  let arguments = (ins CBit_ComparisonPredicate:$predicate,
+      Arg<CBit_RegisterType, "register", [MemRead]>:$reg, APIntAttr:$rhs);
```
simon1hofmann, line 123: 🤖 *AI text below* 🤖

**[P2] Reject `cbit.cmp` inside unitary modifier bodies.**

This new operation reads a register, but the non-unitary deny lists in `QC/IR/Modifiers/ModifierUtils.cpp`, `QCO/IR/Modifiers/ModifierUtils.cpp`, and the QC-to-QCO preflight still list only CBit alloc/load/store. As a result, a register-reading comparison currently verifies inside `inv`, `ctrl`, and `pow`, despite the explicit `body must not ... access registers` contract. Add `cbit::CompareOp` to all three checks and their exhaustive modifier tests.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
```diff
@@ -187,33 +187,60 @@ struct ConvertCBitAllocOp final : StatefulOpConversionPattern<cbit::AllocOp> {
...
+
+  LogicalResult
+  matchAndRewrite(cbit::CompareOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter& rewriter) const override {
+    auto result = cbit::buildComparison(
+        rewriter, op.getLoc(), op.getPredicate(), op.getRhs(),
+        [&](const int64_t index) -> Value {
+          auto indexValue = LLVM::ConstantOp::create(
+              rewriter, op.getLoc(), rewriter.getI64Type(), index);
+          return loadCBit(op, adaptor.getReg(), indexValue, rewriter,
```
simon1hofmann, line 348: 🤖 *AI text below* 🤖

**[P2] Preserve the register representation across control-flow merges.**

`loadCBit` decides between `i1*` storage and returned `Result**` storage through exact SSA-value membership in `state.resultArrays`. If an `scf.if` selects a returned register in one branch and a local register in the other, `adaptor.getReg()` is a new merged pointer absent from that set. I reproduced adaptive QIR that selects the returned pointer and then performs `load i1` from it instead of calling `__quantum__rt__read_result`. Propagate the representation through merges, use one representation, or reject mixed merges before lowering the comparison.

mlir/lib/Dialect/CBit/IR/CBitOps.cpp
```diff
@@ -168,17 +169,115 @@ struct ForwardKnownLoad final : OpRewritePattern<LoadOp> {
...
+};
+
 } // namespace
 
 LogicalResult LoadOp::verify() {
   return verifyIndex(getOperation(), getReg(), getIndex());
 }
 
+LogicalResult CompareOp::verify() {
+  const auto width = static_cast<unsigned>(getReg().getType().getWidth());
```
simon1hofmann: 🤖 *AI text below* 🤖

**[P2] Compare widths without truncating the register width.**

This cast wraps a valid `!cbit.reg<4294967297>` width to `1`, so a comparison with an `i1` RHS passes verification. `convert-cbit-to-memref` then lowers the supposed whole-register comparison to one load of bit 0. Compare the values in a common non-narrowing type, or explicitly reject register widths that `APInt::getBitWidth()` cannot represent.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
burgholzer: Hm. Rejection kind of feels cleaner here. No strong opinion though.

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
burgholzer: Feels like a reasonable fix.

mlir/include/mlir/Dialect/CBit/IR/CBitOps.td
burgholzer: Good cleanup 👍🏻

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
burgholzer: Hm. Feeling like rejection is the cleanest option out of the three. At least it feels like a pretty artificial scenario without a production example.

mlir/lib/Dialect/CBit/IR/CBitOps.cpp
burgholzer: Also here, rejection seems like the best option. This feels like another corner case with no practical relevance.

burgholzer:
> 🤖 *AI text below* 🤖
> 
> Five reproducible findings from the changes added after `6545b95de`. Each inline comment includes the failing case and the smallest safe direction for a fix.

Thanks for running another pass over this. Fresh context certainly always helps. Commented on each finding. Generally, I have the feeling that simply programmatically rejecting some of these corner cases is the way to go for many of them. Should also allow us to add more simplifications throughout the code.

mlir/lib/Dialect/CBit/IR/CBitOps.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed by rejection in `ac6c5c8ec`. `cbit.cmp` now compares register and APInt widths without narrowing, so oversized registers fail verification through the existing width-mismatch diagnostic. CBit IR passes (10/10).

mlir/include/mlir/Dialect/CBit/IR/CBitOps.td
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in `0821559c3`. QC, QCO, and QC-to-QCO now classify `cbit::CompareOp` as register access inside `inv`, `ctrl`, and `pow`. The exhaustive modifier tests cover the new operation. QC IR (339/339), QCO IR (493/493), and QCToQCO (170/170) pass.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed by rejection in `f7ae48405`. Adaptive QIR now propagates returned/local CBit storage representation through CFG forwarding and `arith.select`, rejects mixed representations before LLVM lowering, and preserves returned/returned merges. QCToQIRAdaptive passes (149/149).

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed by rejection in `9214ba1fa`. OpenQASM preflight now follows transitive inline-expression users of `cbit.cmp` and rejects a snapshot that crosses a write to the same register. The regression forwards the comparison through `arith.xori` before the intervening store. The full QC Translation suite passes (181/181).

mlir/lib/Dialect/QC/Translation/TranslateQCToOpenQASM3.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in `1e6fadc81`. The exporter now permits `cbit::CompareOp` alongside `cbit::LoadOp` in supported `scf.while` condition regions. The new import/export regression validates the emitted OpenQASM in strict mode. QC Translation passes (181/181).

burgholzer:
@simon1hofmann Changes look good, thanks 🙏🏼 let's get this in once CI is green! 🚀

PR #2193, opened 2026-08-20
✨ Walk tensor chains across calls and compare untracked qubit tensors

DRovara:
## Description

🤖 *AI text below* 🤖

First of a stack replacing #1970, which combined too many independently reviewable changes. This slice contains only shared infrastructure that later slices build on; no pass and no user-facing behaviour changes.

Two independent fixes:

**`TensorIterator` across call boundaries.** A tensor chain reaching a `func.call` or `func.return` aborted the process through the iterator's `report_fatal_error` default. Both now end the chain instead — the tensor a call returns starts a life-chain of its own.

**`areModulesEquivalentWithPermutations` with untracked qubit tensors.** The comparison assumed every qubit tensor belongs to an equivalence group, which only holds for tensors descending from a `qtensor` allocation. A tensor arriving as a function argument therefore hit `DenseMap::at` on a missing key, and the tensor threaded out of a `qtensor.extract` was never mapped, tripping `IRMapping::lookup`. Untracked tensors now fall back to the regular SSA mapping, and `getOutTensor()` is mapped alongside the qubit result.

### Testing

`ComparesQubitTensorsThatDoNotDescendFromAnAllocation` in `test_qtensor_ir.cpp` covers both paths and includes a negative case so it is not vacuous. Verified by mutation: removing the `getOutTensor()` mapping reproduces the original `IRMapping::lookup` assertion.

Every suite consuming `IRVerification` passes — jeff-round-trip 143, qco-to-qc 139, qc-to-qco 168, qc-to-qir-adaptive 146, qc-to-qir-base 124, qir-ir 116, qtensor-ir 30, optimizations 121.

### AI assistance

Code and this description were produced with Claude Code (Opus 5), acting on my instructions and within the scope I authorized.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@ releases may include breaking changes.
 
 #### Passes and transformations
 
+- ✨ Add passes for quantum-specific interprocedural optimizations ([#2193])
+  ([**@DRovara**], [**@burgholzer**])
 - ✨ Add quantum loop unrolling and qubit reuse passes ([#1705], [#1718],
   [#1755], [#1756], [#1923], [#1924], [#2039], [#2118], [#2216])
   ([**@MatthiasReumann**], [**@DRovara**], [**@burgholzer**],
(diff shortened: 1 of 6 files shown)


burgholzer (approved):
This one looks good already. Let's get it in to not have it sit around for too long!

PR #2215, opened 2026-08-23
✨ Record typed compiler targets in MQT IR

burgholzer:
🤖 *AI text below* 🤖

## Description

This PR adds `#mqt.compilation_target` and its typed leaf attributes to the MQT dialect. `CompilerTarget` can materialize its immutable source facts as the attribute and reconstruct a validated C++ target from it. The round trip preserves target and site names, ordered sparse site IDs, timing metadata, connectivity, native operations, placements, durations, fidelities, and fixed or variadic operation arities. Fixed arities include zero-arity operations such as `gphase`; variadic arities record a positive inclusive minimum for arbitrarily controlled native operations.

This is the focused typed-serialization layer at the top of the stack. It builds on #2323, which builds on the compiler target introduced by #2218. It deliberately excludes payload descriptors, execution capabilities, combined target environments, module attachment, and DLTI integration. Those contracts remain separate work.

The TableGen definitions provide concrete syntax examples in the generated dialect documentation, including a complete `#mqt.compilation_target` example. Attribute verifiers reject malformed target facts before reconstruction, while the existing C++ factories remain the single source of truth for target validation and derived caches.

The changelog records this PR in the existing MQT Compiler Collection launch entry rather than adding a separate entry for unreleased functionality. Its PR-link definition remains in numeric descending order.

## Validation

- Rebased as one signed commit on #2323 at `60d5bb4db2a4672214ea793dbfe6e2d6c92c8863`.
- MQT dialect unit tests: 14 passed.
- Compiler unit tests: 146 passed.
- `cmake --build --preset release --target mlir-doc`
- `uvx nox -s lint`
- `uvx nox -s cpp-lint`
- `git diff --check`
- `git verify-commit HEAD`

Implementation and pull-request text were prepared with OpenAI Codex.

## Checklist

- [x] The changes follow the project guidelines and introduce no new warnings.


mlir/lib/Compiler/Target.cpp
```diff
@@ -862,4 +987,92 @@ CompilerTarget::synthesisBasis() const noexcept {
...
+  switch (connectivityKind()) {
+  case Connectivity::Kind::Unknown:
+    break;
+  case Connectivity::Kind::AllToAll:
+    connectivity = mqt::ConnectivityKind::AllToAll;
+    break;
+  case Connectivity::Kind::Explicit:
+    connectivity = mqt::ConnectivityKind::Explicit;
+    break;
+  }
```
burgholzer: This feels weird and like a duplication of enums that may not be necessary

mlir/lib/Compiler/Target.cpp
```diff
@@ -862,4 +987,92 @@ CompilerTarget::synthesisBasis() const noexcept {
...
+  switch (nativeOperationsKind()) {
+  case NativeOperations::Kind::Unknown:
+    break;
+  case NativeOperations::Kind::Unrestricted:
+    nativeOperations = mqt::NativeOperationsKind::Unrestricted;
+    break;
+  case NativeOperations::Kind::Explicit:
+    nativeOperations = mqt::NativeOperationsKind::Explicit;
+    break;
+  }
```
burgholzer: Same here.

burgholzer (commented):
This also looks genuinely good except for one little detail.

mlir/lib/Compiler/Target.cpp
burgholzer: 🤖 *AI text below* 🤖

Removed the duplicate C++ enum. `CompilerTarget::Connectivity::Kind` now aliases the generated `mlir::mqt::ConnectivityKind`, and `materialize` passes the value through directly. The compiler-target round-trip tests pass.

mlir/lib/Compiler/Target.cpp
burgholzer: 🤖 *AI text below* 🤖

Applied the same root fix to native-operation knowledge. `CompilerTarget::NativeOperations::Kind` now aliases the generated `mlir::mqt::NativeOperationsKind`, so the second translation switch is gone as well.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
denialhaag, line 1: Not important, but I don't think we need the `MQT_` prefixes here.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
```diff
@@ -36,4 +42,181 @@ def MQTDialect : Dialect {
...
+
+def MQT_PayloadEnvAttr : MQT_Attr<"PayloadEnv", "payload_env"> {
+  let summary = "Selected payload execution contract";
+  let description = [{
+    Records the exact payload and its effective capabilities. Producers expand
+    descriptor baselines before creating this attribute. The capability list
+    remains available when optional capability metadata is unknown;
+    `optional_capabilities_known` records whether that optional metadata is
+    complete.
+  }];
```
denialhaag, line 93: While I'm generally a fan of keeping documentation concise, I think (some of) the attributes added here could benefit from some examples in their descriptions, especially because they all end up in the rendered docs. 🤔

denialhaag (commented):
I really like the addition of these attribues! I only have a nitpick and a comment on the attribute descriptions, otherwise this LGTM! I'm already approving now, as neither of my comments is blocking. 😎

denialhaag (approved):

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: Agreed. Let's drop them!

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: Yeah, fully agree. The initial version of these was even more concise and I already urged the system to be a little more elaborate. 
I like the idea of including an explicit example of a small program using that feature as \`\`\`mlir ... \`\`\` code

burgholzer:
@simon1hofmann any feedback on this one? Given its low position in the stack, I'd like to not touch this part more than once anymore.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
```diff
@@ -36,4 +42,181 @@ def MQTDialect : Dialect {
...
+    : MQT_Attr<"TargetEnv", "target_env", [DLTIQueryInterface]> {
+  let summary = "Selected compilation and payload environment";
+  let description = [{
+    Combines typed compiler-target facts with the selected payload contract.
+    Optional DLTI extensions carry provider-specific facts. Direct access to
+    the typed fields is authoritative.
+  }];
+  let parameters = (ins "CompilationTargetAttr":$compilation_target,
+      "PayloadEnvAttr":$payload_env,
+      OptionalParameter<"::mlir::MapAttr">:$extensions);
```
simon1hofmann: 🤖 *AI text below* 🤖
[MQTDialect.td:209](https://github.com/munich-quantum-toolkit/core/blob/390e9669812cef7de7a940b0eecb137baf3f5f8d/mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td#L209) embeds #dlti.map, but MQT does not declare DLTIDialect as dependent. Standard contexts omit it in [Programs.cpp:88-92](https://github.com/munich-quantum-toolkit/core/blob/390e9669812cef7de7a940b0eecb137baf3f5f8d/mlir/lib/Compiler/Programs.cpp#L88-L92) and [mqt-cc.cpp:456-460](https://github.com/munich-quantum-toolkit/core/blob/390e9669812cef7de7a940b0eecb137baf3f5f8d/mlir/tools/mqt-cc/mqt-cc.cpp#L456-L460). Consequently, valid extension-bearing target environments cannot be parsed through QCOProgram or mqt-cc; the latter crashes after parsing fails. The test masks this by registering DLTI explicitly at [test_mqt_ir.cpp:51-54](https://github.com/munich-quantum-toolkit/core/blob/390e9669812cef7de7a940b0eecb137baf3f5f8d/mlir/unittests/Dialect/MQT/IR/test_mqt_ir.cpp#L51-L54).

simon1hofmann (approved):
Also went through this and looks pretty clean as well. Just have one finding that was flagged, probably worth double-checking in your Codex session.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: Feels like a correct finding. Let's double check and fix this where it is supposed to be fixed.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: 🤖 *AI text below* 🤖

Addressed in the rescaled implementation: the retained TableGen definitions no longer use the redundant `MQT_` record prefix.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: 🤖 *AI text below* 🤖

Addressed for every retained attribute. Each description now includes concrete syntax, and `CompilationTargetAttr` includes a complete MLIR example. The generated `mlir-doc` target passes.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: 🤖 *AI text below* 🤖

The Core 4.0 rescope removes `TargetEnvAttr` and all DLTI integration. The retained `CompilationTargetAttr` has no DLTI value or dependency, so dependent dialect registration is no longer needed in this PR.

burgholzer:
This was essentially just updated based on the changes in the stack below it. Might not hurt to give it a sanity check though.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
```diff
@@ -40,4 +44,155 @@ def MQTDialect : Dialect {
...
+}
+
+def NativeOperationAttr : MQTAttr<"NativeOperation", "native_operation"> {
+  let summary = "One native compiler-target operation";
+  let description = [{
+    The operation records its spelling, arity, parameter count, and optional
+    global or site-specific calibration data. For example,
+    `#mqt.native_operation<name = "cx",
+    arity = #mqt.operation_arity<kind = fixed, value = 2>,
+    num_parameters = 0, site_tuples = [<sites = [4, 7]>]>` records a placed
```
denialhaag: I think this would look better as a formatted `mlir` block. The rendered version doesn't look super beautiful right now.

````
```mlir
#mqt.native_operation<...>
```
````

File: mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td [hunk 6]
```diff
@@ -40,4 +44,155 @@ def MQTDialect : Dialect {
...
+
+def SiteTupleAttr : MQTAttr<"SiteTuple", "site_tuple"> {
+  let summary = "Calibration data for an ordered site tuple";
+  let description = [{
+    The tuple defines one placement of a native operation. For example,
+    `#mqt.site_tuple<sites = [4, 7], duration = 40>` records an ordered
+    two-site placement with a raw duration of 40.
+  }];
+  let parameters = (ins MQTArrayRefParameter<"int64_t">:$sites,
+      OptionalParameter<"std::optional<uint64_t>">:$duration,
```
denialhaag: Should this use `DurationUnitAttr`? 🤔

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@ releases may include breaking changes.
   [#1927], [#1935], [#1936], [#1938], [#1975], [#1976], [#2006], [#2014],
   [#2015], [#2017], [#2026], [#2028], [#2054], [#2058], [#2125], [#2136],
   [#2149], [#2150], [#2158], [#2194], [#2210], [#2211], [#2220], [#2218],
-  [#2323]) ([**@burgholzer**], [**@denialhaag**], [**@taminob**],
+  [#2323], [#2215]) ([**@burgholzer**], [**@denialhaag**], [**@taminob**],
```
denialhaag: I think we generally ordered the PR numbers by their "absolute value," right? 🤔

denialhaag (approved):
This LGTM! :shipit:

I just have two nitpicks, both of which are not blocking:

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
```diff
@@ -40,4 +44,155 @@ def MQTDialect : Dialect {
...
+
+def SiteAttr : MQTAttr<"Site", "site"> {
+  let summary = "One ordered compiler-target site";
+  let description = [{
+    Site identifiers come from the target. The array order defines dense
+    compiler vertices. For example, `#mqt.site<id = 4, name = "q4">` records
+    one named site.
+  }];
+  let parameters = (ins "int64_t":$id, OptionalParameter<"StringAttr">:$name,
+      OptionalParameter<"std::optional<uint64_t>">:$t1,
```
simon1hofmann: Using OptionalParameter<std::optional<uint64_t>> makes the generated parser call FieldParser<std::optional<uint64_t>> even after the key and equals sign were consumed. Consequently, #mqt.site<id = 0, t1 => parses successfully and prints as #mqt.site<id = 0>, silently discarding the field. The same affects t2 and the duration fields on site tuples and native operations. Require an integer whenever the key is present, retain optional storage so explicit zero remains distinct from absence, and add rejection tests.

simon1hofmann (approved):
This one also looks good, just one thing to clarify.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: Good suggestion 👍🏼 
Let's fix this now. There'll be a lot more like this, when we start polishing the documentation.

CHANGELOG.md
burgholzer: Yeah, kind of. This is sorted by eventual merge order. Let's clean this up now here.
Likely that we'll do another pass over the changelog right before the release.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: Sounds like an easy enough fix. 👍🏼

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
simon1hofmann: 🤖 *AI text below* 🤖

Done in `4fdbed100`. The native-operation example now uses a fenced `mlir` block, and the generated MLIR documentation builds successfully.

CHANGELOG.md
simon1hofmann: 🤖 *AI text below* 🤖

Done in `26697defe`. The changelog entries are now ordered numerically as #2215, #2218, #2220, and #2323.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
simon1hofmann: 🤖 *AI text below* 🤖

Done in `e0685f098`. Present timing keys now use the required `uint64_t` parser while retaining optional storage, so missing values are rejected without changing explicit duration-zero semantics. Rejection tests cover `t1`, `t2`, site-tuple duration, and native-operation duration; the MQT IR and compiler round-trip tests pass.

simon1hofmann:
@burgholzer This one is ready to be merged as well right?

burgholzer:
> @burgholzer This one is ready to be merged as well right?

Yes, let's get this is once CI is green 🚀 
That should also conclude the 4.0 split of this stack. 
We may actually want to split off the rest from that stack and form a new one after reorganizing. I'll keep that in mind for after this merges.

PR #2218, opened 2026-08-23
♻️ Generalize MLIR compiler target facts

burgholzer:
🤖 *AI text below* 🤖

## Description

Define a complete, immutable MLIR compiler-target contract before serializing it as a typed MQT dialect attribute.

- Require connectivity to be all-to-all or an explicit coupling graph.
- Require native-operation support to be unrestricted or explicitly enumerated.
- Reject missing or unrepresentable QDMI target facts during target inference.
- Represent operation arity as an exact width or a positive variadic minimum.
- Represent `gphase` as a fixed zero-arity operation and preserve its observable phase effects during native synthesis.
- Recognize structurally controlled QCO operations through variadic base-gate capabilities.
- Let the bundled DDSIM device advertise arbitrary positive controls for every canonical standard gate with one or more targets, including multi-target gates, through one exact QDMI 1.3 compatibility marker.
- Support homogeneous fixed operations above two qubits without treating format-level constructs as compiler-target operations.
- Use deterministic placement for all-to-all targets and mapping for explicit coupling graphs.
- Preserve the full nonnegative site-ID domain and cheap shared target snapshots.

The DDSIM custom property is private to MQT Core and can be removed when QDMI standardizes the same operation capability. Controlled aliases such as `mcx` and `mcp` remain available to existing adapters but are not enumerated as compiler-target capabilities.

This PR is the bottom layer of the stack. #2323 adds target-aware controlled-operation decomposition, and #2215 adds typed target serialization on top of it. This PR does not add payload capabilities, replaceable-driver work, or a dependency on the QDMI v1.4 development API.

Target synthesis relies on the canonical QCO cleanup pipeline to remove empty control modifiers. Doxygen comments in every C++ file touched by this PR consistently use `///`.

## Validation

- Compiler unit tests: 143 passed.
- Mapping unit tests: 86 passed.
- Native-synthesis unit tests: 26 passed.
- DDSIM device tests: 52 passed.
- Focused Python MLIR and Qiskit translation tests: 279 passed with the bundled superconducting-device catalog selected explicitly.
- Generated Python stubs are current.
- The release build passes.
- Full CTest reached 4,156 passes and one expected skip; its only failure was the obsolete `CUSTOM1` unsupported-property probe. The corrected isolated test passes.
- Sphinx documentation passes with warnings treated as errors.
- `uvx nox -s lint` passes.
- `uvx nox -s cpp-lint -- origin/main` reports no clang-format or clang-tidy findings.
- `git diff --check` passes.
- An independent final audit found no concrete issues.

All 13 commits were rebased onto current `main`, signed, and verified. Hosted checks for the new head are pending.

Implementation, validation, and pull-request text were prepared with OpenAI Codex.

## Checklist

### DIFF
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -27,13 +27,21 @@ compiled = compile_program(
 Target compilation accepts optimized QCO, QC, or QIR output and uses the
 canonical QCO pipeline; it cannot be combined with a custom `qco_pipeline`.
 
-The target can also be constructed directly. Omitting `couplings` selects
-all-to-all connectivity; omitting `operations` means that every operation is
-native:
+The target can also be constructed directly. Connectivity and native-operation
+metadata are unknown unless the caller states them:
 
 ```python
-target = CompilerTarget(3, couplings=[(0, 1), (1, 2)])
+target = CompilerTarget(
+    3,
+    connectivity=CompilerTarget.Connectivity([(0, 1), (1, 2)]),
+    native_operations=CompilerTarget.NativeOperations.unrestricted(),
+)
 ```
+
+Use `CompilerTarget.Connectivity.all_to_all()` for an all-to-all target. An
+empty `CompilerTarget.NativeOperations([])` means that no operation is native.
+The default-constructed metadata objects mean that the corresponding support is
+unknown; target compilation rejects an unknown property when a pass needs it.
 
 Use {py:meth}`~mqt.core.mlir.QCOProgram.compile_for_target` to apply target
 compilation to an existing QCO program. Compilation runs in place. If a pass
@@ -125,6 +133,6 @@ When exporting a program that has already been mapped to a
 {py:meth}`~mqt.core.mlir.QCProgram.to_qiskit`. The exporter maps each static
 target site ID to its index in {py:attr}`~mqt.core.mlir.CompilerTarget.sites`
 and creates a canonical physical Qiskit circuit. The circuit has one register
-named {code}`q` with {py:attr}`~mqt.core.mlir.CompilerTarget.num_qubits` qubits.
+named {code}`q` with {py:attr}`~mqt.core.mlir.CompilerTarget.num_sites` qubits.
 This option does not run target compilation or emit Qiskit layout metadata.
 Target-aware export requires static qubits whose site IDs belong to that target.
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -34,14 +34,23 @@ metadata are unknown unless the caller states them:
 target = CompilerTarget(
     3,
     connectivity=CompilerTarget.Connectivity([(0, 1), (1, 2)]),
-    native_operations=CompilerTarget.NativeOperations.unrestricted(),
+    native_operations=CompilerTarget.NativeOperations([
+        CompilerTarget.Operation("u", arity=1, num_parameters=3),
+        CompilerTarget.Operation("cx", arity=2, num_parameters=0),
+        CompilerTarget.Operation("measure", arity=1, num_parameters=0),
+        CompilerTarget.Operation("reset", arity=1, num_parameters=0),
+    ]),
 )
 ```
 
 Use `CompilerTarget.Connectivity.all_to_all()` for an all-to-all target. An
-empty `CompilerTarget.NativeOperations([])` means that no operation is native.
-The default-constructed metadata objects mean that the corresponding support is
-unknown; target compilation rejects an unknown property when a pass needs it.
+empty `CompilerTarget.NativeOperations([])` reports that no quantum operation is
+native. It can be used with passes that need only topology, but target
+compilation cannot lower quantum operations without a synthesis basis. Use
+`CompilerTarget.NativeOperations.unrestricted()` only when the target accepts
+every operation. The default-constructed metadata objects mean that the
+corresponding support is unknown; target compilation rejects an unknown property
+when a pass needs it.
 
 Use {py:meth}`~mqt.core.mlir.QCOProgram.compile_for_target` to apply target
 compilation to an existing QCO program. Compilation runs in place. If a pass
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -28,13 +28,18 @@ Target compilation accepts optimized QCO, QC, or QIR output and uses the
 canonical QCO pipeline; it cannot be combined with a custom `qco_pipeline`.
 
 The target can also be constructed directly. Connectivity and native-operation
-metadata are unknown unless the caller states them:
+support are required:
 
 ```python
 target = CompilerTarget(
     3,
     connectivity=CompilerTarget.Connectivity([(0, 1), (1, 2)]),
     native_operations=CompilerTarget.NativeOperations([
+        CompilerTarget.Operation(
+            "gphase",
+            arity=CompilerTarget.OperationArity.fixed(0),
+            num_parameters=1,
+        ),
         CompilerTarget.Operation("u", arity=1, num_parameters=3),
         CompilerTarget.Operation("cx", arity=2, num_parameters=0),
         CompilerTarget.Operation("measure", arity=1, num_parameters=0),
@@ -48,9 +53,18 @@ empty `CompilerTarget.NativeOperations([])` reports that no quantum operation is
 native. It can be used with passes that need only topology, but target
 compilation cannot lower quantum operations without a synthesis basis. Use
 `CompilerTarget.NativeOperations.unrestricted()` only when the target accepts
-every operation. The default-constructed metadata objects mean that the
-corresponding support is unknown; target compilation rejects an unknown property
-when a pass needs it.
+every operation. Creating a target from a QDMI device fails if the device does
+not provide a complete connectivity model and a representable native-operation
+set. An explicit operation arity is either fixed or variadic with a positive,
+inclusive minimum. Fixed zero represents a global-phase operation. A variadic
+capability accepts every total width from its minimum through the target's site
+count; site-specific calibration tuples are therefore available only for fixed,
+positive arities. Structural and program-format constructs are not
+compiler-target operations.
+
+Target synthesis preserves a native `gphase`. If the target does not support
+`gphase`, target synthesis preserves relative phase effects and removes only the
+unobservable global phase of the entry point.
 
 Use {py:meth}`~mqt.core.mlir.QCOProgram.compile_for_target` to apply target
 compilation to an existing QCO program. Compilation runs in place. If a pass
@@ -60,8 +74,7 @@ benchmarking, the C++ API exposes separate factories for pre-routing
 optimization, deterministic placement, topology-aware mapping, native synthesis,
 and conformance verification. Target compilation uses compact placement on
 all-to-all targets and the mapper only when the target has an explicit coupling
-graph. A target with unknown connectivity can use compact placement only when no
-non-barrier multi-site operation remains.
+graph.
 
 Target compilation preserves quantum operations even when their final qubit
 values are not measured or returned. This supports measurement-free programs,
@@ -125,8 +138,19 @@ if (!qco || !qco->compileForTarget(*target)) {
 
 The adapter accepts circuit-model devices whose operations are available
 throughout the topology in both operand orientations. Operand-symmetric gates,
-such as CZ, may report each edge once. Neutral-atom zone models require a
+such as CZ, may report each edge once. Operations with arity above two must
+report every ordered tuple of distinct sites. Neutral-atom zone models require a
 different compilation model and are rejected with a diagnostic.
+
+QDMI 1.3 cannot report an operation-arity range. The bundled DDSIM device uses
+an exact, versioned custom-operation marker to state that each canonical
+standard gate with one or more targets accepts arbitrary positive controls. The
+adapter turns such a base gate into a variadic capability whose minimum is the
+base gate's target count. For example, DDSIM reports `h` with minimum one, `rxx`
+with minimum two, and `rccx` with minimum three; each also accepts any
+additional number of controls up to the simulator's site count. Controlled
+aliases such as `mcx` and `mcp` are not enumerated as compiler capabilities.
+This private bridge can be removed when QDMI standardizes equivalent metadata.
 
 The bundled Garnet and Emerald snapshots contain available T1, T2, and fidelity
 data. Operation durations are absent because they were unavailable. See
--- a/docs/mlir/target_compilation.md
+++ b/docs/mlir/target_compilation.md
@@ -60,7 +60,8 @@ benchmarking, the C++ API exposes separate factories for pre-routing
 optimization, deterministic placement, topology-aware mapping, native synthesis,
 and conformance verification. Target compilation uses compact placement on
 all-to-all targets and the mapper only when the target has an explicit coupling
-graph.
+graph. A target with unknown connectivity can use compact placement only when no
+non-barrier multi-site operation remains.
 
 Target compilation preserves quantum operations even when their final qubit
 values are not measured or returned. This supports measurement-free programs,
--- a/docs/qdmi/ddsim_device.md
+++ b/docs/qdmi/ddsim_device.md
@@ -43,10 +43,8 @@ The device implements the full QDMI job interface (except for the
 
 ## Compile and execute QIR
 
-QDMI v1.3 cannot compactly report the DDSIM device's all-to-all topology or the
-site applicability of every operation. State the topology and reconstruct the
-fixed-arity operation list from the device metadata when compiling a program to
-QIR. Submit the resulting bitcode to the same device:
+The compiler can snapshot the DDSIM device as an all-to-all target, compile a
+program to QIR, and submit the resulting bitcode to the same device:
 
 ```python
 from mqt.core.mlir import CompilerTarget, OutputFormat, compile_program
@@ -54,24 +52,7 @@ from mqt.core.qdmi import ProgramFormat
 from mqt.core.qdmi.driver import open_device
 
 device = open_device("mqt.ddsim.default")
-operations = []
-for operation in device.operations():
-    arity = operation.qubits_num()
-    if arity is None or arity == 0:
-        continue
-    operations.append(
-        CompilerTarget.Operation(
-            name=operation.name(),
-            arity=arity,
-            num_parameters=operation.parameters_num(),
-        )
-    )
-
-target = CompilerTarget(
-    device.qubits_num(),
-    connectivity=CompilerTarget.Connectivity.all_to_all(),
-    native_operations=CompilerTarget.NativeOperations(operations),
-)
+target = CompilerTarget.from_device(device)
 program = compile_program(
     "bell.qasm",
     target=target,
--- a/docs/qdmi/ddsim_device.md
+++ b/docs/qdmi/ddsim_device.md
@@ -43,9 +43,10 @@ The device implements the full QDMI job interface (except for the
 
 ## Compile and execute QIR
 
-QDMI v1.3 cannot report unrestricted topology and operation support. State these
-known DDSIM properties when compiling a program to QIR, then submit the
-resulting bitcode to the same device:
+QDMI v1.3 cannot report the DDSIM device's all-to-all topology. State this
+temporary topology workaround and an explicit DDSIM synthesis basis when
+compiling a program to QIR. QDMI v1.4 will remove the topology workaround.
+Submit the resulting bitcode to the same device:
 
 ```python
 from mqt.core.mlir import CompilerTarget, OutputFormat, compile_program
@@ -56,7 +57,12 @@ device = open_device("mqt.ddsim.default")
 target = CompilerTarget(
     device.qubits_num(),
     connectivity=CompilerTarget.Connectivity.all_to_all(),
-    native_operations=CompilerTarget.NativeOperations.unrestricted(),
+    native_operations=CompilerTarget.NativeOperations([
+        CompilerTarget.Operation("u", 1, 3),
+        CompilerTarget.Operation("cx", 2, 0),
+        CompilerTarget.Operation("measure", 1, 0),
+        CompilerTarget.Operation("reset", 1, 0),
+    ]),
 )
 program = compile_program(
     "bell.qasm",
--- a/docs/qdmi/ddsim_device.md
+++ b/docs/qdmi/ddsim_device.md
@@ -43,8 +43,9 @@ The device implements the full QDMI job interface (except for the
 
 ## Compile and execute QIR
 
-The compiler can snapshot the DDSIM device as an all-to-all target, compile a
-program to QIR, and submit the resulting bitcode to the same device:
+QDMI v1.3 cannot report unrestricted topology and operation support. State these
+known DDSIM properties when compiling a program to QIR, then submit the
+resulting bitcode to the same device:
 
 ```python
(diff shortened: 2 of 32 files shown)


docs/qdmi/ddsim_device.md
```diff
@@ -37,16 +37,21 @@ The device implements the full QDMI job interface (except for the
 
 ## Compile and execute QIR
 
-The compiler can snapshot the DDSIM device as an all-to-all target, compile a
-program to QIR, and submit the resulting bitcode to the same device:
+QDMI v1.3 cannot report unrestricted topology and operation support. State these
+known DDSIM properties when compiling a program to QIR, then submit the
+resulting bitcode to the same device:
```
burgholzer: This is something we should fix for QDMI v1.4. This can only stay a temporary workaround.

docs/qdmi/ddsim_device.md
```diff
@@ -37,16 +37,21 @@ The device implements the full QDMI job interface (except for the
...
 from mqt.core.mlir import CompilerTarget, OutputFormat, compile_program
 from mqt.core.qdmi import ProgramFormat
 from mqt.core.qdmi.driver import open_device
 
 device = open_device("mqt.ddsim.default")
-target = CompilerTarget.from_device(device)
+target = CompilerTarget(
+    device.qubits_num(),
+    connectivity=CompilerTarget.Connectivity.all_to_all(),
+    native_operations=CompilerTarget.NativeOperations.unrestricted(),
```
burgholzer: The DDSIM device does not have unrestricted native operation support.
It does expose a full list of supported operations though. So this is only a matter of representing all-to-all connectivity explicitly.

mlir/include/mlir/Compiler/Target.h
```diff
@@ -44,6 +43,36 @@ class CompilerTarget {
...
+
+    /// Create unknown connectivity.
+    Connectivity() noexcept;
+
+    /// Create unrestricted all-to-all connectivity.
+    [[nodiscard]] static Connectivity allToAll();
+
+    /// Create explicitly enumerated connectivity.
+    [[nodiscard]] static Connectivity
+    fromCouplings(std::vector<Coupling> couplings);
```
burgholzer: Fairly sure this should be using an LLVM data type such as ArrayRef.

mlir/include/mlir/Compiler/Target.h
```diff
@@ -44,6 +43,36 @@ class CompilerTarget {
...
+    /// Return explicitly enumerated couplings, if any.
+    [[nodiscard]] llvm::ArrayRef<Coupling> couplings() const noexcept;
+
+  private:
+    friend class CompilerTarget;
+
+    Connectivity(Kind kind, std::vector<Coupling> couplings);
+
+    Kind kind_;
+    std::vector<Coupling> couplings_;
```
burgholzer: Shouldn't this use an LLVM data type?

mlir/include/mlir/Compiler/Target.h
```diff
@@ -177,19 +206,49 @@ class CompilerTarget {
...
+
+    /// Create unknown native-operation support.
+    NativeOperations() noexcept;
+
+    /// Create unrestricted native-operation support.
+    [[nodiscard]] static NativeOperations unrestricted();
+
+    /// Create explicitly enumerated native-operation support.
+    [[nodiscard]] static NativeOperations
+    fromOperations(std::vector<Operation> operations);
```
burgholzer: Same as above. LLVM data type?

burgholzer (commented):
This one feels close to ready.

mlir/include/mlir/Compiler/Target.h
burgholzer: 🤖 *AI text below* 🤖

Fixed in commit 9b961b26c. Coupling storage now uses llvm::SmallVector<Coupling>.

docs/qdmi/ddsim_device.md
burgholzer: 🤖 *AI text below* 🤖

The example now calls this out as a temporary QDMI v1.3 topology workaround and states that QDMI v1.4 will remove it.

docs/qdmi/ddsim_device.md
burgholzer: 🤖 *AI text below* 🤖

Fixed in commit 9b961b26c. The example now declares an explicit DDSIM U/CX synthesis basis plus measurement and reset instead of claiming unrestricted operation support.

mlir/include/mlir/Compiler/Target.h
burgholzer: 🤖 *AI text below* 🤖

Fixed in commit 9b961b26c. fromCouplings now takes llvm::ArrayRef<Coupling> and copies it into LLVM storage.

mlir/include/mlir/Compiler/Target.h
burgholzer: 🤖 *AI text below* 🤖

Fixed in commit 9b961b26c. fromOperations now takes llvm::ArrayRef<Operation> and stores an llvm::SmallVector<Operation>.

denialhaag (approved):
I just had a free minute to go through this, and it looks really clean to me! :shipit:

docs/mlir/target_compilation.md
```diff
@@ -27,14 +27,22 @@ compiled = compile_program(
...
-target = CompilerTarget(3, couplings=[(0, 1), (1, 2)])
+target = CompilerTarget(
+    3,
+    connectivity=CompilerTarget.Connectivity([(0, 1), (1, 2)]),
+    native_operations=CompilerTarget.NativeOperations.unrestricted(),
+)
 ```
 
+Use `CompilerTarget.Connectivity.all_to_all()` for an all-to-all target. An
+empty `CompilerTarget.NativeOperations([])` means that no operation is native.
```
simon1hofmann: Does that mean native synthesis is not called at all? Should we add an example for a target with a specific set of native operations?

docs/qdmi/ddsim_device.md
```diff
@@ -37,16 +37,27 @@ The device implements the full QDMI job interface (except for the
...
 from mqt.core.mlir import CompilerTarget, OutputFormat, compile_program
 from mqt.core.qdmi import ProgramFormat
 from mqt.core.qdmi.driver import open_device
 
 device = open_device("mqt.ddsim.default")
-target = CompilerTarget.from_device(device)
+target = CompilerTarget(
+    device.qubits_num(),
+    connectivity=CompilerTarget.Connectivity.all_to_all(),
+    native_operations=CompilerTarget.NativeOperations([
```
simon1hofmann: An example like this in `target_compilation.md`.

docs/qdmi/ddsim_device.md
```diff
@@ -37,16 +37,27 @@ The device implements the full QDMI job interface (except for the
...
+        CompilerTarget.Operation("u", 1, 3),
+        CompilerTarget.Operation("cx", 2, 0),
+        CompilerTarget.Operation("measure", 1, 0),
+        CompilerTarget.Operation("reset", 1, 0),
```
simon1hofmann: Is it clear/explained somewhere what the second and third argument in `Operation` represent?

test/python/qdmi/test_qdmi.py
```diff
@@ -552,7 +552,11 @@ def test_device_executes_qir_program(ddsim_device: Device) -> None:
 cx q[0], q[1];
 c = measure q;
 """
-    target = CompilerTarget.from_device(ddsim_device)
+    target = CompilerTarget(
+        ddsim_device.qubits_num(),
```
simon1hofmann: Unrelated to this PR, but is there any reason why it is called `qubits_num` and not `num_qubits`?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -369,16 +369,24 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
       signalPassFailure();
       return;
     }
+    if (comp->hasTwoQubitOperations &&
```
simon1hofmann: This permits unknown connectivity without two-qubit gates, but control-flow reconciliation still queries topology at [Mapping.cpp:899-908](https://github.com/munich-quantum-toolkit/core/blob/a6f7fd1cdf5e2bc3bad855c89caca1abf9a25fcf/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp#L899-L908), reaching the fatal branch at [Target.cpp:753-758](https://github.com/munich-quantum-toolkit/core/blob/a6f7fd1cdf5e2bc3bad855c89caca1abf9a25fcf/mlir/lib/Compiler/Target.cpp#L753-L758).

mlir/lib/Compiler/Target.cpp
```diff
@@ -229,7 +245,7 @@ llvm::Expected<CompilerTarget::SiteTuple>
 CompilerTarget::SiteTuple::create(std::vector<SiteId> sites,
                                   const std::optional<uint64_t> duration,
                                   const std::optional<double> fidelity) {
-  std::set<SiteId> uniqueSites;
+  llvm::SmallDenseSet<SiteId> uniqueSites;
```
simon1hofmann: This stores nonnegative int64_t IDs in SmallDenseSet, whose sentinel values overlap the advertised domain (INT64_MAX, and on LP64 also INT64_MAX-1)

simon1hofmann (changes requested):
This also looks quite good, just a few comments about the documentation and two concerns flagged by another LLM review run.

docs/mlir/target_compilation.md
burgholzer: I'd rather argue that "no operation is native" is a misconfiguration as this essentially means that you cannot compile to the target (it lacks an ISA).
Maybe this part of the documentation should make this more explicit (and the code should likely fail early on such targets)

docs/qdmi/ddsim_device.md
burgholzer: Likely no really. Maybe it would help to add keyword arguments here in the docs to make this more explicit.

test/python/qdmi/test_qdmi.py
burgholzer: None that I could think of. We can happily rename consistently.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Thanks for pointing that out. We should definitely fix this!

mlir/lib/Compiler/Target.cpp
burgholzer: Thanks. I think we could be limiting the sites to 32bits. IIRC site IDs should also be non-negative (might be worth double-checking).
Reducing to 32-bits may even increase efficiency throughout the implementation.

burgholzer:
Thanks @denialhaag and @simon1hofmann 🙏🏼 
Given how this is at the very bottom of an 11 PR stack, I'd like to only touch this once more and freeze the implementation afterwards. Are there any further comments on the above from your side?
Quick signal would be good. Then I'll kick off the implementation and the rebase.

simon1hofmann:
> Thanks @denialhaag and @simon1hofmann 🙏🏼 
> 
> Given how this is at the very bottom of an 11 PR stack, I'd like to only touch this once more and freeze the implementation afterwards. Are there any further comments on the above from your side?
> 
> Quick signal would be good. Then I'll kick off the implementation and the rebase.

Thanks for addressing my comments, I agree with all your answers.

denialhaag:
> > Thanks @denialhaag and @simon1hofmann 🙏🏼
> > Given how this is at the very bottom of an 11 PR stack, I'd like to only touch this once more and freeze the implementation afterwards. Are there any further comments on the above from your side?
> > Quick signal would be good. Then I'll kick off the implementation and the rebase.
> 
> Thanks for addressing my comments, I agree with all your answers.

Go also from my side!

docs/mlir/target_compilation.md
burgholzer: 🤖 *AI text below* 🤖

Addressed in `fe41d2a95`. The general example now declares an explicit native operation set with keyword arguments. The text also clarifies that an empty explicit set reports no native quantum operations: topology-only passes may still use such a target, while target compilation diagnoses it once lowering needs a synthesis basis.

docs/qdmi/ddsim_device.md
burgholzer: 🤖 *AI text below* 🤖

Addressed in `fe41d2a95`. The DDSIM example now derives the complete fixed-arity operation list from device metadata and combines it with the simulator's explicit all-to-all connectivity, without claiming unrestricted operation support.

docs/qdmi/ddsim_device.md
burgholzer: 🤖 *AI text below* 🤖

Addressed in `fe41d2a95`. The documentation examples now use the `arity=` and `num_parameters=` keyword arguments.

test/python/qdmi/test_qdmi.py
burgholzer: 🤖 *AI text below* 🤖

`qubits_num()` mirrors the existing QDMI Client Interface property nomenclature, including `QDMI_DEVICE_PROPERTY_QUBITSNUM` and `QDMI_OPERATION_PROPERTY_QUBITSNUM`. Renaming that public wrapper consistently is a separate QDMI-facing API change, so this focused target-contract PR keeps the established method name.

mlir/lib/Compiler/Target.cpp
burgholzer: 🤖 *AI text below* 🤖

Fixed in `fe41d2a95`. The public target contract already accepts every nonnegative `int64_t` site ID, so target and QDMI-adapter lookup sets now use sentinel-free standard containers instead of narrowing that domain. A unit test exercises `INT64_MAX` and `INT64_MAX - 1` through sites, couplings, and operation tuples.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: 🤖 *AI text below* 🤖

Fixed in `fe41d2a95`. After the pass has established that no multi-site operation remains, unknown connectivity now skips structured-control-flow topology reconciliation because child layouts cannot diverge. The regression test covers measured single-site operations in both branches of `qco.if`; explicit and all-to-all reconciliation remain unchanged.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer, line 1: I would argue that this pass should simply not run on `CompilerTarget::Connectivity::Kind::Unknown`. This would simplify quite a bit here supposedly.

docs/qdmi/ddsim_device.md
burgholzer, line 1: This feels awkward; especially the wording around QDMI v1.3
Is there no more organic solution to the underlying problem? 
This may include changes to the DDSIM device itself so that this can properly be inferred.
I want `target = CompilerTarget.from_device(device)` to just work, even for QDMI v1.3

burgholzer (commented):
Two more things to address here before this can go in.

docs/qdmi/ddsim_device.md
burgholzer: 🤖 *AI text below* 🤖

Addressed at the device boundary. DDSIM now reports one exact, namespaced QDMI 1.3 custom-property marker for its compact all-to-all topology and homogeneous fixed-arity operation applicability. The generic adapter consumes that marker without granting any facts to unmarked providers.

The documented and tested path is now simply `CompilerTarget.from_device(device)`. The Python integration test compiles and executes QIR through that exact path, and the C++ adapter test covers one-, two-, and three-site operations as well as measurement/reset and the variadic-operation exclusion.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: 🤖 *AI text below* 🤖

Agreed with the separation. I extracted deterministic topology-free placement into #2316 against `main`. Target compilation now uses placement for all-to-all targets and the existing router only for explicit coupling graphs; both share the validated allocation rewrite.

Once #2316 merges, I will rebase this PR onto it. This branch's only additional integration will be the `Unknown` case: reject remaining non-barrier multi-site operations before mutation, otherwise run placement. #2215 will be restacked after that.

burgholzer: 🤖 *AI text below* 🤖

Implemented after #2316 merged. Target compilation now sends explicit connectivity to mapping and all-to-all or unknown connectivity to placement. The shared placement pass rejects an unknown-connectivity program containing a non-barrier multi-site unitary before mutating the IR; otherwise it performs deterministic topology-free placement. The focused and full mapping/compiler suites pass.

burgholzer:
I believe I am happy with this now. Looking forward to what you think @simon1hofmann 😌

mlir/lib/Compiler/Target.cpp
```diff
@@ -507,7 +605,7 @@ llvm::Error CompilerTarget::Storage::initialize() {
   }
 
   for (const auto& specification : GATE_SPECIFICATIONS) {
```
simon1hofmann, line 607: 🤖 *AI text below* 🤖
A target containing native `u(1,3)` and variadic `x(min=1)` correctly accepts a structural CX through the new controlled-operation logic, but `supportedGates` only checks the literal name `cx`. Consequently, `synthesisBasis()` is empty and compiling a non-native two-qubit gate fails with “no usable synthesis basis,” even though U+CX is complete. Variadic Z/CZ has the same problem. Deriving CX/CZ from the corresponding variadic base capability and adding a regression test should fix it.

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
```diff
@@ -299,6 +300,28 @@ static bool requiresTargetSynthesis(Operation* operation,
   return !target.supports(operation);
 }
 
+/// Normalize relative phase effects and discard only the unobservable global
+/// phase of an entry point when the target cannot represent it.
+static LogicalResult prepareGlobalPhases(ModuleOp moduleOp,
```
simon1hofmann, line 305: 🤖 *AI text below* 🤖
`normalizeGlobalPhases` can turn `ctrl { gphase(theta) }` into `ctrl {}` followed by `p(theta)`. The latest commit removed the empty-control cleanup and deleted the test that covered this case. Running the public `createTargetNativeSynthesis` factory directly with a P-only target now encounters that empty `qco.ctrl` and fails for lack of a synthesis basis. The full compiler pipeline happens to clean it earlier, but the standalone factory has no documented canonical-input precondition and is explicitly exposed for pass-level benchmarking. Keep the cleanup/test or document and enforce the precondition.

simon1hofmann (approved):
This looks quite clean and I really like the changes here!
I just have two comments that should be checked before this can go in.

mlir/lib/Compiler/Target.cpp
burgholzer: Yeah. I think that makes sense 👍🏼 
I am still not 100% sure whether the variadic handling here is as clean as it could be yet; in the case above, it currently implies arbitrary controls.
But the semantics of an operation being variadic could also be different, e.g., a barrier can be applied to an arbitrary list of qubits.
However, I still believe this is fine for 4.0. for now and we can tweak that in 4.1+

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
burgholzer: Hm. This feels fairly hypothetical as there will (hopefully) rarely be a case where the is only a controlled global phase, but rather some meaningful operations being controlled.
Nevertheless, this may be worth fixing. Could/should we simply add the canonicalization pattern for empty control removal to the respective optimization close to where the phase normalization is currently called?

mlir/lib/Compiler/Target.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in b87d5cd1c. Variadic X and Z capabilities now classify CX and CZ as supported gates, which keeps supports(GateKind), supportedGates(), and synthesisBasis() aligned with controlled-operation support. The change preserves the current controlled-operation semantics and does not generalize other variadic operations. The regression covers both U with variadic X and U with variadic Z targets.

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in a3d8106fc. prepareGlobalPhases now removes only empty CtrlOp shells immediately after normalizeGlobalPhases. I kept the cleanup local instead of running every CtrlOp canonicalization, which could also merge or reduce non-empty controls before target planning. The restored regression uses the public synthesis factory with a P-only target and checks DD equivalence.

simon1hofmann:
@burgholzer I fixed the two open comments and this PR should we ready now.

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
burgholzer: I am wondering: should this potentially be directly integrated into the `normalizeGlobalPhases` pass? There is no value in keeping around an empty control and that pass already knows when it is pulling out a global phase as the last operation of a control modifier. Feels like the cleaner solution to me.

burgholzer:
> @burgholzer I fixed the two open comments and this PR should we ready now.

One (hopefully) last request for changes. Then this can go in 👍🏼

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
simon1hofmann: Good point, I've moved the cleanup into normalizeGlobalPhases for both QC and QCO, removed the target-specific walk.

burgholzer: I pushed one last simplification for this. Let's get this in once CI is green! 🚀

simon1hofmann: CI is failing now as llvm::make_early_inc_range(llvm::reverse(...)) is unsafe: advancing the reverse iterator leaves its base pointing at the operation that is then erased. The next iteration dereferences a dangling iterator and segfaults.

burgholzer: Damn. Interesting that this wasn't caught locally. Could you fix that? Likely just reverting my last commit. Sorry for the noise.

simon1hofmann: Fixed it, will wait for CI to be green and merge the PR 🚀

burgholzer: Given how #2323 also seems to be pretty much ready to merge and even #2215 feels ready, let's consider merging the three as one stack to safe a bit on CI resources

PR #2219, opened 2026-08-23
♻️ Define target compilation with payload specifications

burgholzer:
🤖 *AI text below* 🤖

## Description

Keep the compiler-only selected-program capability prototype independent of QDMI runtime adoption. Typed metadata and a cached target-environment analysis connect target selection to compilation without requiring unreleased QDMI APIs.

Compiler workstream: **main → #2219 → #2162**. QDMI adaptation remains the separate integration PR #2227. This compiler-only work targets **Core 4.0**, subject to Simon's compiler contract review and current CI. It does not require QDMI 1.4 or approval of the later QDMI metadata contract. #2131 tracks compiler requirements; #2365 coordinates the separate Core 4.1/QDMI 1.4 adaptation.

Retain the compiler behavior already released onto main through #2218/#2323/#2215: unknown topology or gate sets fail during inference, variadic controlled DDSIM gates and zero-arity global phase remain supported, and placement/decomposition retain their validated target. No driver replacement, batching or metadata cleanup is included.

### Local validation

- Independent release build and CTest: 3,879 passed, one existing optional-device skip.
- Targeted Python compiler/QDMI tests: 558 passed.
- Stub generation, repository lint and full changed-file C++ lint passed.
- Hosted CI must be checked separately on this final head.

AI-assisted restructuring; human design review remains required.

## Compiler contract review

Model the capabilities of the selected program/target combination rather than assuming a format's full language is executable. Keep the initial feature vocabulary bounded and informed by OpenQASM and QIR. Validate representative benchmark targets without requiring a QDMI 1.4 device. The later QDMI adapter may expose only a subset until its independent contract is ready.

## Checklist


.agent/plans/selected-payload-target-environment.md
```diff
@@ -0,0 +1,257 @@
...
+  constraints, matching the typed MQT attributes. Do not add a closed feature
+  enum or convenience support predicates. Rationale: providers and future QDMI
+  versions must round-trip new capabilities without a Core release. Date/Author:
+  2026-08-23 / Codex.
+
+- Decision: Attach `mqt.target_env` before constructing the target pass manager,
+  then make every target-dependent pass reconstruct `CompilerTarget` at
+  `runOnOperation`. Rationale: captured C++ targets create a second source of
+  truth and prevent reproducible textual pipelines. Date/Author: 2026-08-23 /
+  Codex.
```
burgholzer: I would like to challenge this point. While it is nice for the actual pipeline execution, it feels inefficient to reparse targets over and over again.
Even if passes only query the information they need.
Is this really idiomatic MLIR? Are there precedents where this is handled similarly?

.agent/plans/selected-payload-target-environment.md
```diff
@@ -0,0 +1,257 @@
...
+`mlir/include/mlir/Compiler/TargetCompilation.h`. The method now attaches the
+typed environment before it runs the canonical target pipeline. Mapping, native
+synthesis, and conformance read the environment from the module. Those passes
+live under `mlir/lib/Dialect/QCO/Transforms/`.
+
+The Python API is bound in `bindings/mlir/register_mlir.cpp`, with handwritten
+stub corrections in `bindings/patterns.txt`. The `mqt-cc` driver lives in
+`mlir/tools/mqt-cc/mqt-cc.cpp`. QDMI remains responsible only for producing a
+hardware `CompilerTarget` in this task; this task must not depend on unreleased
+QDMI payload-capability APIs.
```
burgholzer: I'd like to challenge this point here.
# already proposed how the payload-capability will be handled in QDMI. 
And while that PR is still subject to change, I would like to already see that functionality integrated here as part of the compiler work.
This also requires adapting the QDMI device implementations that are part of MQ T Core.
This could happen directly in this PR, or in another one that is stacked before  or after it.
At the very least, I would like to make sure that we offer a coherent and convenient interface for when QDMI adds that capability.
Given how we are working towards the v4 release of MQT Core, it is entirely fine that we intermittently depend on a development version of QDMI.

.agent/plans/selected-payload-target-environment.md
```diff
@@ -0,0 +1,257 @@
...
+Target compilation must know both the hardware target and the exact payload
+contract selected for execution. After this change, callers provide those two
+context-free values together. The compiler records them as one typed
+`mqt.target_env` module attribute before it runs target passes. Mapping, native
+synthesis, and conformance then reconstruct the hardware target from that
+attribute. This makes the IR the single source of target facts and makes the
+passes usable from MLIR textual pass pipelines.
+
+The behavior is visible by compiling a QCO program for an explicit target and
+payload environment, printing the result, and observing a complete typed
```
burgholzer: I am not fully sure that "payload environment" is really the best term for this. "Payload" itself is fine, but I would have rather expected something like "payload format", "payload specification", etc.

bindings/mlir/register_mlir.cpp
```diff
@@ -956,8 +1035,9 @@ LLVM bitcode.)pb");
 
   m.def("compile_program", &compileProgram, "program"_a, nb::kw_only(),
         "output"_a = mlir::ProgramFormat::QC, "inplace"_a = false,
-        "target"_a = nb::none(), "qco_pipeline"_a = "mqt-qco-default",
-        "enable_timing"_a = false, "enable_statistics"_a = false,
+        "target"_a = nb::none(), "payload_environment"_a = nb::none(),
+        "qco_pipeline"_a = "mqt-qco-default", "enable_timing"_a = false,
+        "enable_statistics"_a = false,
```
burgholzer, line 1448: It feels like the signature is a bit redundant now with `output` and `payload_environment`. At the same time, it is comfortable to simply specify the output format with a single enumerator.
This holds for the entire PR. Is this really the cleanest solution? Now's the time to still change it.

docs/mlir/target_compilation.md
```diff
@@ -14,18 +14,36 @@ stored, copied cheaply, and reused for multiple compilations.
...
 ```
 
-Target compilation accepts optimized QCO, QC, or QIR output and uses the
-canonical QCO pipeline; it cannot be combined with a custom `qco_pipeline`.
+Target compilation accepts optimized QCO, QC, OpenQASM 3, or QIR output and uses
+the canonical QCO pipeline; it cannot be combined with a custom `qco_pipeline`.
+The payload environment identifies the exact representation selected for the
+device. It is independent of the requested compiler output checkpoint. MQT Core
+does not infer payload support from an output format. Until QDMI reports this
+contract, callers must construct it from the device and payload documentation.
```
burgholzer: This is exactly the discrepancy I tried to point out in the other comment.
- There is already a WIP PR for the QDMI addition, which should already be incorporated here. MQT Core v4 will ship with QDMI v1.4
- Does it really make sense that the compiler output checkpoint is independent from the payload environment? Shouldn't the payload environment regulate the compiler output as a single source of truth?

burgholzer (commented):
This one will likely still need a bit of work to iron out the details.

bindings/mlir/register_mlir.cpp
burgholzer: 🤖 *AI text below* 🤖

Removed the redundant arguments. Targeted compilation now takes one `TargetEnvironment` and derives the terminal output from its `PayloadSpecification`. Untargeted compilation retains the convenient explicit `output` argument.

.agent/plans/selected-payload-target-environment.md
burgholzer: 🤖 *AI text below* 🤖

Replaced per-pass reconstruction with `TargetEnvironmentAnalysis`. MLIR now caches the validated `TargetEnvironment`, and the custom invalidation hook compares the current immutable `mqt.target_env` attribute with the cached attribute. Mapping, native synthesis, and conformance all request that analysis.

.agent/plans/selected-payload-target-environment.md
burgholzer: 🤖 *AI text below* 🤖

Renamed the inner contract to `PayloadSpecification` and its exact identity to `PayloadFormat`. The outer hardware-plus-payload pair remains `TargetEnvironment`.

.agent/plans/selected-payload-target-environment.md
burgholzer: 🤖 *AI text below* 🤖

Addressed through the two focused child PRs. #2226 adopts QDMI #508/#509 across the low-level Core API, bundled DDSIM and SC devices, and framework consumers. #2227 then verifies an exact advertised descriptor, groups its optional feature records, and builds the owning `TargetEnvironment`. This keeps #2219 provider-neutral while the immediate stack supplies the requested QDMI v1.4 integration.

docs/mlir/target_compilation.md
burgholzer: 🤖 *AI text below* 🤖

Addressed. The targeted compiler APIs now take one `TargetEnvironment` and derive the output from its `PayloadSpecification`; there is no independent output choice on that path. #2226 supplies the exact QDMI v1.4 payload contract and #2227 maps one advertised device descriptor and its grouped features into that environment.

burgholzer:
This looks much cleaner now. And should also be ready.

burgholzer:
@simon1hofmann @denialhaag any feedback on this one from your side?
I looked through this yesterday night already and it looked pretty clean to me.

simon1hofmann:
> @simon1hofmann @denialhaag any feedback on this one from your side? I looked through this yesterday night already and it looked pretty clean to me.

Going through this now and will provide feedback in around 15min.

simon1hofmann (approved):
Also went through the changes here and they look really clean 👍

docs/mlir/target_compilation.md
```diff
@@ -14,18 +14,35 @@ stored, copied cheaply, and reused for multiple compilations.
...
+    CompilerTarget,
+    PayloadFormat,
+    PayloadEncoding,
+    PayloadSpecification,
+    TargetEnvironment,
+    compile_program,
+)
 
 target = CompilerTarget.from_device_id("mqt.sc.iqm.garnet")
+payload = PayloadSpecification(PayloadFormat("qir", "2.1.0", "base", PayloadEncoding.BINARY))
```
denialhaag, line 27: Should we also have an enum for the `format_id` (e.g., `Format.QIR`)? 🤔

docs/mlir/target_compilation.md
```diff
@@ -100,12 +122,24 @@ if (!target) {
   return 1;
 }
 
+auto payload = mlir::PayloadSpecification::create({
+    .id = "qir",
```
denialhaag, line 172: I guess my question above percolates quite a bit. 😬

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
denialhaag, line 1: My earlier point about removing the `MQT_` prefix will probably show up during rebasing here, but just as a reminder.

docs/mlir/target_compilation.md
denialhaag, line 27: Does QIR even have a patch version? 🤔 It doesn't look like here: https://github.com/qir-alliance/qir-spec/blob/f5647346542d5a65225c3eb349847fe4df01d1b2/specification/profiles/Base_Profile.md#L179-L180

denialhaag (commented):
This also looks pretty clean to me! 😌 I just have two observations and a small reminder:

docs/mlir/target_compilation.md
burgholzer: That feels too restrictive given how the intention is that providers can provide custom formats (IQM JSON for example, or QPY, or calibration).

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
burgholzer: Good to keep it in mind!

docs/mlir/target_compilation.md
burgholzer: See above

docs/mlir/target_compilation.md
burgholzer: Not really. I was just about to argue that it is cleaner if one always specifies a full semver tripple. But one could also argue that this should actually be a version range following the python dependency specification.
That would allow to express things like "accepts OpenQASM 3" or "accepts QIR 2.1" or more. Adds a fair bit of complexity probably, but may be worth it.
Any thoughts?

denialhaag: Yeah, I missed that point a bit. Sorry about that!

denialhaag: Hmm, I do understand the intention behind requiring a full SemVer triple, but OpenQASM and QIR really seem like prime examples where people generally don't think too much about patch (and sometimes even minor) versions. I'd say relaxing the condition could be worth it. 🤔

burgholzer: Technically, both formats have a minor version that is kind of meaningful (OpenQASM 3.1 and QIR 2.1) but I agree with you that people typically do not care and just refer to them as OpenQASM 3 and QIR 2 (or even just QIR for the sake of it).
Let's what gpt-5.6 comes up with in plan mode later 😌

simon1hofmann:
@burgholzer Can we also move this to v4? @denialhaag and I already reviewed it and it's the only prerequisite for #2162.

burgholzer:
> @burgholzer Can we also move this to v4? @denialhaag and I already reviewed it and it's the only prerequisite for #2162. 

I think we can. I was just a bit hesitant because it also ties in with the QDMI changes. However, I think it is fine for us to ship this with 4.0 and tweak the details for 4.1+ including changes prompted by QDMI.

simon1hofmann:
@burgholzer What is the current state on this? are there only the open threads left to address?

burgholzer:
> @burgholzer What is the current state on this? are there only the open threads left to address?

I think these are the only open threads. I would have probably ran a ponytail review with Astra on it once more, but you may have already done that.

docs/mlir/target_compilation.md
simon1hofmann: 🤖 *AI text below* 🤖

The [QIR specification](https://github.com/qir-alliance/qir-spec/blob/f5647346542d5a65225c3eb349847fe4df01d1b2/specification/profiles/Base_Profile.md#specification-version) records major and minor version flags, not a patch flag. [5412f9fba](https://github.com/munich-quantum-toolkit/core/commit/5412f9fbaf0a9eadb439ee121f3a37c8eba7e466) now accepts one to three numeric components in C++/Python and typed MLIR input. The examples use `qir` / `2.1`; its internal normalization to `2.1.0` is a descriptor convention, not an extra QIR flag. Tests cover normalization, typed attributes, and output selection.

We are keeping exact selected versions in this PR. Colloquial “QIR 2” names a family; the exact string `2` here instead means `2.0.0` and does not silently select QIR 2.1. Family names or dependency-style ranges would need a separate selection step: choose a concrete version supported by both compiler and device, then construct the payload specification. The current [QDMI #508 proposal](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/508) likewise advertises exact descriptors without inferred compatibility.

This fixes the mandatory-patch issue. Family/range selection remains a separate design question, not a feature claimed by this change.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
simon1hofmann: 🤖 *AI text below* 🤖

Confirmed on the rebased branch: the TableGen definitions already use `MQTDialect`, `PayloadFormatAttr`, `PayloadSpecAttr`, and `TargetEnvAttr`, without the `MQT_` prefix. No additional rename was needed.

simon1hofmann:
🤖 *AI text below* 🤖

Rebased onto main at `ba3637cd1` and restacked #2162 on top. The review fixes remain in individual commits:

- [8ce67b89f](https://github.com/munich-quantum-toolkit/core/commit/8ce67b89f43e43287c94b826dacbdcc7e337d3f0): register the math dialect for runtime target-native synthesis, with a fresh-context regression test.
- [5412f9fba](https://github.com/munich-quantum-toolkit/core/commit/5412f9fbaf0a9eadb439ee121f3a37c8eba7e466): accept exact version shorthand, with C++, typed-MLIR, and Python regressions. Family/range selection remains separate, as explained in the version thread.
- [9e2339c26](https://github.com/munich-quantum-toolkit/core/commit/9e2339c26cc1c007d8a74f7ed51c57686a0a0299): use DDSIM for the executable QIR examples, show `2.1`, and document both the attached-environment and inliner prerequisites.
- [c78063ca4](https://github.com/munich-quantum-toolkit/core/commit/c78063ca45763c3308c79bf50546fb38ef9aadaf): align the plan with the compiler-only Core 4.0 scope; the separate QDMI 1.4 adaptation does not gate it.

Local validation on the restacked tree: 253 C++ tests passed (compiler, native synthesis, MQT IR), 55 Python compiler tests and six target-specific Qiskit translation tests passed, and the CLI shorthand output assembled successfully with QIR major/minor flags 2/1. Repository lint passed. Full changed-file C++ lint ran, but the local CLion clang-tidy 23 development build reports three false positives on necessary dereference parentheses. Its suggested edit fails to compile in a minimal reproducer, so those expressions were retained; the genuine trailing-comma finding was fixed. This is not a clean local C++ lint result. The build uses main's exact jeff-mlir dependency pin. Hosted CI is a separate final-head check; these are local results.

The rebase preserves main's reusable-function inlining before #2162's control-flow legalization. The existing loop/switch capability snapshot and safety limits remain intact.

mlir/include/mlir/Compiler/TargetEnvironment.h
denialhaag, line 1: Given the glossary entry, should all of the "Payload" classes be prepended with "Execution"? Or am I mixing up the two terms right now? 🤔

mlir/include/mlir/Compiler/TargetEnvironment.h
```diff
@@ -0,0 +1,165 @@
...
+struct ProgramCapability {
+  std::string id;
+  uint64_t value = 0;
+  std::vector<ProgramConstraint> constraints;
+
+  friend bool operator==(const ProgramCapability&,
+                         const ProgramCapability&) = default;
+};
+
+/// Context-free selected payload execution contract.
```
denialhaag:
```suggestion
/// Context-free selected execution payload contract.
```

mlir/include/mlir/Compiler/TargetEnvironment.h
```diff
@@ -0,0 +1,165 @@
...
+///
+/// Producers must include every effective capability, including
+/// payload-format baselines. The knowledge bit states whether the list also
+/// contains all optional device capabilities.
+class PayloadSpecification {
+public:
+  /// Create and validate a selected payload specification.
+  /// Numeric versions accept one to three components; omitted components are
+  /// zero, and the stored version always uses major.minor.patch.
```
denialhaag, line 79:
```suggestion
  /// Create and validate a selected payload specification.
  ///
  /// Numeric versions accept one to three components; omitted components are
  /// zero, and the stored version always uses major.minor.patch.
```

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
denialhaag, line 1: It would be nice if the "central" attributes included an example in their descriptions (as, e.g., `CompilationTargetAttr` does).

mlir/include/mlir/Dialect/QCO/Transforms/Passes.h
```diff
@@ -19,7 +19,7 @@
 
 namespace mlir {
 class CompilerTarget;
-} // namespace mlir
+}
```
denialhaag: This change should probably be reverted.

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
denialhaag, line 1: Not necessarily for this PR, but we should consider splitting this file into multiple files, as we do for the QC and QCO operations.

mlir/unittests/Compiler/test_compiler_pipeline.cpp
```diff
@@ -2251,9 +2289,36 @@ h q;
...
+      llvm::cantFail(CompilerTarget::create(
+          1, CompilerTarget::Connectivity::allToAll(),
+          CompilerTarget::NativeOperations::unrestricted())),
+      payload);
+
+  EXPECT_FALSE(runDefaultPipeline(std::move(program), environment));
+  /// The call binds an rvalue reference, but the unsupported output is rejected
+  /// before the function consumes the input. Inspecting the input verifies that
+  /// preservation contract.
+  /// NOLINTNEXTLINE(bugprone-use-after-move)
```
denialhaag:
```suggestion
  // The call binds an rvalue reference, but the unsupported output is rejected
  // before the function consumes the input. Inspecting the input verifies that
  // preservation contract.
  // NOLINTNEXTLINE(bugprone-use-after-move)
```

denialhaag (changes requested):
Thanks for the continued improvements, @simon1hofmann! 🙂 Overall, this looks pretty good to me. I only have a few remarks:

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Tracked separately in #2445. The follow-up proposes moving typed attribute implementations into `MQTAttributes.cpp` while retaining dialect registration and metadata hooks here, with no API or behavior changes. No file split in this PR.

mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
simon1hofmann: 🤖 *AI text below* 🤖

Added assembly examples for `PayloadFormatAttr`, `PayloadSpecAttr`, and `TargetEnvAttr` in 2a25b619b. The compound examples use a producer-defined payload to avoid implying a complete QIR capability set.

All three examples parse and round-trip through `mqt-cc`; generated MLIR docs and repository lint also pass.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.h
simon1hofmann: 🤖 *AI text below* 🤖

Restored the namespace-closing comment in 278e6a173. No other pass-header changes were reverted.

mlir/include/mlir/Compiler/TargetEnvironment.h
simon1hofmann: Not sure about that, I would have kept "Payload" 🤔 
What do you think @burgholzer?

burgholzer: Hm. The glossary also defines the general term `payload`. Haven't fully thought it through now, but sticking with `payload` for now seems fine.

denialhaag: But isn't that the problem? The general "payload" entry is (currently) referring to something rather specific (I almost read it as "compiler payload"). 🤔

denialhaag (approved):
Thanks for addressing all of my comments, @simon1hofmann! I'll leave the final comment open for now, but I'll not die on that hill; you can resolve it whenever we have come to any conclusion. I'm approving for now. Feel free to request another review if there are new changes I can have a look at! 🙂

simon1hofmann:
Also looks good to me, will leave the final check to @burgholzer before merging.

burgholzer:
I'll tackle this one next

mlir/include/mlir/Compiler/TargetEnvironment.h
burgholzer: hm. i am tempted to leave this as-is for now and look at it from a bit farther away once the whole pipeline is complete.

burgholzer (commented):
I ran this through another iteration with the goal of simplifying the implementation, which was pretty successful I believe.
I'd vote for getting this in now and looking at the fallout in subsequent audits until the release.

PR #2224, opened 2026-08-24
✨ Add Pauli twirling pass

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Add the opt-in `pauli-twirl-2q-gates` QCO pass.

The pass:

- supports CX, CZ, ECR, and iSWAP gates;
- selects uniformly from 16 Pauli twirls with a reproducible seed;
- inserts four explicit Pauli operations per supported gate;
- preserves the exact unitary with a global-phase correction when required;
- leaves unsupported gates, modifier regions, and existing global phases unchanged.

Python callers can use the existing `QCOProgram.copy()` and
`run_pass_pipeline()` methods. The change adds no dependency or dedicated
binding.

Codex assisted with repository tracing, implementation, tests, validation, and
the pull-request text. The human author remains responsible for reviewing and
accepting the contribution.

## Validation

- `mqt-core-mlir-unittest-optimizations --gtest_brief=1`: 188 tests passed.
- `mqt-core-mlir-unittests-compiler`: 131 tests passed.
- Focused Python textual-pipeline tests: 2 tests passed.
- `uvx nox -s lint`: passed.
- Independent final diff review: no findings.

## Checklist

- [x] I have added entries to the changelog for noteworthy additions.
- [x] Migration instructions are not required because this does not change a released API.
- [x] The changes pass the remote CI checks.
- [x] I have reviewed and accepted responsibility for all code changes.

**AI-assisted content:**

- [x] The agent was explicitly authorized to implement the change and create this pull request.
- [x] This public text begins with the required visible disclosure.
- [x] AI assistance is disclosed in this description.


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

burgholzer (commented):
Hey @simon1hofmann 👋🏼 In order to move this one along a little more, I kickstarted my review agent, which produced the output below that seems quite reasonable to me. Take everything with a grain of salt though.

🤖 *AI text below* 🤖

I see two main items to resolve:

1. **Define the target-compilation ownership.** The documented route applies the pass to raw QCO. The canonical target pipeline later performs two-qubit fusion, mapping, target-native synthesis, and conformance verification, while `compile_program(target=...)` rejects custom QCO pipelines. Running twirling before target compilation can miss entanglers introduced by mapping/native synthesis and can let later two-qubit fusion absorb the sampled Paulis. Running it after target compilation can introduce X/Y/Z/Id operations that are not native after conformance has already run. The smallest scope-preserving resolution for this PR is to describe the pass explicitly as a raw-IR primitive with no target-native guarantee, then track a target-aware insertion point separately. If this PR is intended to deliver hardware-ready Pauli twirling now, it needs a stage after the final native entanglers are formed and before target one-qubit lowering/conformance.

2. **Document the ensemble contract.** With the default seed, the pass produces one deterministic, logically equivalent circuit. Noise tailoring requires callers to execute and combine multiple independently seeded realizations. A short caller-owned loop over `copy()` and `run_pass_pipeline(...seed=N...)` is sufficient; I do not think this needs a new multi-output abstraction in this PR. Qiskit's corresponding API exposes both multiple twirls and target-aware one-qubit optimization, which is a useful boundary comparison: https://github.com/Qiskit/qiskit/blob/4e0f21adda3273ecc74807a51827da480220b230/qiskit/circuit/twirling.py#L36-L80

I left three non-blocking inline simplifications that together remove about 16 lines without changing behavior. The tests are not bloated; the semantic matrix checks and modifier-boundary coverage are justified.

docs/mlir/python_compiler_collection.md
```diff
@@ -255,6 +255,15 @@ custom = compile_program(
 )
 ```
 
+Pauli twirling is available as an opt-in textual pass. It supports CX, CZ, ECR,
+and iSWAP gates, keeps every inserted Pauli operation (including identities)
+explicit, and preserves the exact global phase. The seed defaults to {code}`42`:
```
burgholzer, line 300: 🤖 *AI text below* 🤖

This should state that the seed selects one realization, not the full error-suppression experiment. Because the default seed is fixed at 42, repeated optionless calls produce the same transformed program. A single logically equivalent circuit does not by itself provide the ensemble averaging that turns coherent error into a stochastic Pauli channel.

Please add a short note that callers should create copies with distinct seeds, execute those circuits, and combine their results. A small Python loop is enough; the MLIR pass does not need a multi-output API.

docs/mlir/python_compiler_collection.md
```diff
@@ -255,6 +255,15 @@ custom = compile_program(
...
+
+```{code-cell} ipython3
+twirled = compile_program(bell_qasm, output=OutputFormat.QCO)
+twirled.run_pass_pipeline("pauli-twirl-2q-gates{seed=42}")
```
burgholzer, line 304: 🤖 *AI text below* 🤖

This example proves that the textual pass is registered, but it does not yet describe a target-safe execution workflow. The current target compiler runs two-qubit fusion, mapping, target-native synthesis, and conformance in a fixed order, and `compile_program(target=...)` rejects custom QCO pipelines.

If this pass runs before target compilation, mapping and native synthesis may introduce additional physical entanglers after twirling, while two-qubit fusion may absorb the sampled Pauli gates. If it runs after target compilation, the new X/Y/Z/Id operations may not belong to the target's single-qubit basis, and conformance has already completed.

Please make the contract explicit. The smallest change for this PR is to call this a raw-QCO/IR transformation with no target-native guarantee and track target-aware placement separately. A hardware-ready workflow would instead need to run after the final native two-qubit gates are formed and before the inserted Paulis are lowered to the target's one-qubit basis.

mlir/lib/Dialect/QCO/Transforms/Optimizations/PauliTwirling.cpp
```diff
@@ -0,0 +1,244 @@
...
+  Pauli beforeSecond;
+  Pauli afterFirst;
+  Pauli afterSecond;
+  bool correctPhase;
+
+  constexpr Twirl(const Pauli preFirst, const Pauli preSecond,
+                  const Pauli postFirst, const Pauli postSecond,
+                  const bool phase)
+      : beforeFirst(preFirst), beforeSecond(preSecond), afterFirst(postFirst),
+        afterSecond(postSecond), correctPhase(phase) {}
```
burgholzer, line 51: 🤖 *AI text below* 🤖

Non-blocking simplification: `Twirl` is plain table data, and this constructor enforces no invariant or normalization. It only repeats aggregate member assignment. Removing the constructor leaves every existing table initializer valid and makes the representation more obviously data-only. This should remove six lines without changing behavior.

mlir/lib/Dialect/QCO/Transforms/Optimizations/PauliTwirling.cpp
```diff
@@ -0,0 +1,244 @@
...
+                                newFirstOut.getDefiningOp());
+  rewriter.replaceAllUsesExcept(secondOut, newSecondOut,
+                                newSecondOut.getDefiningOp());
+
+  if (twirl.correctPhase) {
+    GPhaseOp::create(rewriter, op->getLoc(), std::numbers::pi);
+  }
+}
+
+namespace {
```
burgholzer, line 209: 🤖 *AI text below* 🤖

Non-blocking simplification: the anonymous namespace opened near the top of the file is closed after the tables only to be reopened here. Keeping that namespace open through the pass class preserves the same internal linkage and removes the redundant close/reopen boundary, about four lines including the extra spacing.

mlir/lib/Dialect/QCO/Transforms/Optimizations/PauliTwirling.cpp
```diff
@@ -0,0 +1,244 @@
...
+    struct Candidate {
+      UnitaryOpInterface gate;
+      const std::array<Twirl, 16>* table;
+    };
+    SmallVector<Candidate> gates;
+    getOperation().walk([&](Operation* op) {
+      if (const auto* table = getTwirlTable(op)) {
+        gates.push_back({.gate = cast<UnitaryOpInterface>(op), .table = table});
+      }
+    });
```
burgholzer: 🤖 *AI text below* 🤖

Non-blocking simplification with a small efficiency benefit: this walk visits every operation, while `getTwirlTable` separately performs ancestor searches for `CtrlOp`, `InvOp`, and `PowOp` to rediscover that their descendants are excluded.

A pre-order walk can express the boundary once: inspect the current operation, record an eligible outer `CtrlOp` if present, then return `WalkResult::skip()` for every `CtrlOp`, `InvOp`, or `PowOp`; otherwise advance. That lets `isNestedInModifier` and its guard disappear, avoids traversing ignored modifier subtrees, and removes roughly six lines. The existing modifier tests already protect this behavior.

docs/mlir/python_compiler_collection.md
simon1hofmann: I added a short note about that.

docs/mlir/python_compiler_collection.md
simon1hofmann: Also added a note about that.

mlir/lib/Dialect/QCO/Transforms/Optimizations/PauliTwirling.cpp
simon1hofmann: Removing the `Twirl` constructor would lead to 64 designated initializer warnings.

mlir/lib/Dialect/QCO/Transforms/Optimizations/PauliTwirling.cpp
simon1hofmann: Makes sense, changed It to a pre-order walk.

mlir/lib/Dialect/QCO/Transforms/Optimizations/PauliTwirling.cpp
simon1hofmann: This would lead to four LLVM lint warnings.

burgholzer: Yeah. Let's skip this.

burgholzer (approved):
I believe this looks good now. 
It would be even nicer if this were more tightly integrated into the compiler pipeline; but that is definitely a step for the future.

simon1hofmann:
> I believe this looks good now. It would be even nicer if this were more tightly integrated into the compiler pipeline; but that is definitely a step for the future.

 Created a follow-up issue: #2265.

PR #509, opened 2023-12-25
Fix handling of output permutation and initial layout

martin-fink:
## Description

Output permutation and initial layout need to be applied after the circuit is parsed completely.

This fixes an issue where an output permutation is declared with less qubits than are used in the circuit.

```qasm
// i 0 2 1 3
// o 3 0
qubit[4] q;
```

Would result in an output permutation of
```
0: 1
1: 1
2: 2
3: 0
```

instead of

```
0: 1
3: 0
```

The issue is fixed by applying the output permutation after parsing the complete circuit.

## Checklist:

- [x] The pull request only contains commits that are related to it.
- [x] I have added appropriate tests and documentation.
- [x] I have made sure that all CI jobs on GitHub pass.
- [x] The pull request introduces no new warnings and follows the project's style guidelines.

PR #2226, opened 2026-08-24
✨ Adopt exact QDMI 1.4 payload contracts across Core

burgholzer:
🤖 *AI text below* 🤖

## Description

Keep the runtime portion of the experimental QDMI program-capability contract independently reviewable. This branch targets `v4.1`, without compiler-only #2219, replaceable drivers, metadata cleanup, or native multi-program adoption.

Native multi-program jobs and indexed results were extracted into #2373 using [QDMI #509](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/509)'s existing format enum. This prototype instead uses [QDMI #508](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/508)'s exact format descriptors, optional feature records, single-program submission, and format-defined output bytes. Mechanical Qiskit/PennyLane adaptations remain here so this runtime builds and its SDK consumers import and run coherently.

Current mainline optional shots, binary transport, DDSIM QCO simulation, session ownership, concurrent single-program execution, and fail-closed target inference are preserved. It does not replace concurrent submissions with synthetic aggregate jobs.

### Design and release gate

This is a **non-blocking Core 4.1 / QDMI 1.4 candidate**, not Core 4.0 scope or a settled public contract. [QDMI #523](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/523) and Core #2365 must record the format/capability decisions before this becomes merge-ready. In particular, provider-neutral execution guarantees, opaque payload semantics, and supported versus native operations remain design work.

Refs #2365, #2373, Munich-Quantum-Software-Stack/QDMI#508, Munich-Quantum-Software-Stack/QDMI#523.

Core #2227 is the separate adapter between this runtime and #2219's compiler model. Neither foundation depends on that adapter or on the other foundation. The development QDMI pin must be replaced with a released dependency before publishing release artifacts.

### Validation

- Independent native suite: 3,874 passed; one existing skip.
- Python QDMI, Qiskit and PennyLane suite: 399 passed.
- Generated stubs, repository lint and C++ lint passed.
- Tests retain current concurrency and optional-shot regressions, and exercise descriptors, optional feature metadata, binary/text framing, retrieval, SDK serialization and results.
- Hosted CI and design review remain separate gates.

AI assistance was used for extraction, rebasing, implementation adaptations, tests, documentation and this description.

## Interface acceptance

Describe the device-and-format-specific contract through #2365 and [QDMI #523](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/523). Do not assume full OpenQASM/QIR language support or a universal native/verbatim compilation boundary. The prototype remains a priority for Core 4.1, not a release blocker.

Before the interface change merges, link a working Core consumer and a demonstration with at least one existing provider, preferably both where applicable. Compatibility evidence suffices when a provider needs no changes. Development revisions may support these tests; published artifacts must use released dependencies.

## Checklist


include/mqt-core/qdmi/Client.hpp
```diff
@@ -69,6 +70,32 @@ concept custom_property_value =
...
+    throw std::invalid_argument(std::string(description) +
+                                " is not null-terminated");
+  }
+  if (value.find('\0') != value.size() - 1U) {
+    throw std::invalid_argument(std::string(description) +
+                                " contains an embedded null byte");
+  }
+  value.pop_back();
+  return value;
+}
```
burgholzer, line 85: I am likely the one being too dense here, but it feels like this may be introducing off-by-one errors or false requirements.
an `std::string` is _always_ null-terminated and IIRC the null terminator is not accounted in `size()`. I want to make absolutely sure that this doesn't introduce an architecture where strings are essentially `<chars>\0\0` in memory.

include/mqt-core/qdmi/Client.hpp
```diff
@@ -69,6 +70,32 @@ concept custom_property_value =
...
+
+[[nodiscard]] inline std::string
+decodeText(const std::span<const std::byte> value,
+           const std::string_view description) {
+  if (value.empty()) {
+    throw std::invalid_argument(std::string(description) +
+                                " is not null-terminated");
+  }
```
burgholzer, line 93: This error message feels wrong. are we sure this is useful, needed, and correct?

include/mqt-core/qdmi/Client.hpp
```diff
@@ -701,11 +712,11 @@ class Device {
       }
 
       qdmi::throwIfError(result, msg);
-      std::string value(size - 1, '\0');
+      std::string value(size, '\0');
       result = QDMI_device_query_device_property(device_.get(), prop, size,
                                                  value.data(), nullptr);
       qdmi::throwIfError(result, msg);
-      return value;
+      return detail::decodeText(std::move(value), msg);
```
burgholzer, line 737: This is another case where I would want to make absolutely sure this is correct and does not introduce off-by-one errors. There's more cases below.

bindings/qdmi/qdmi.cpp
```diff
@@ -105,25 +162,50 @@ NB_MODULE(MQT_CORE_MODULE_NAME, qdmiModule) {
...
+  job.def(
+      "get_results",
+      [](const qdmi::Job& self, const size_t programIndex,
+         const QDMI_Job_Result result) {
+        const auto value = self.getResults(programIndex, result);
+        return nb::bytes(reinterpret_cast<const char*>(value.data()),
+                         value.size());
+      },
+      "program_index"_a, "result"_a,
+      "Returns one indexed result as exact bytes.");
```
burgholzer: would it makes sense to flip the order of the arguments here so that the index could receive a default argument? Which would allow to simplify a few call sites.

bindings/qdmi/qdmi.cpp
```diff
@@ -185,6 +269,9 @@ when the custom slot is unsupported.)pb");
 
   job.def_prop_ro("num_shots", &qdmi::Job::getNumShots, "The number of shots.");
 
+  job.def_prop_ro("programs_num", &qdmi::Job::getProgramsNum,
+                  "The number of programs in the job.");
```
burgholzer: Related to another comment from a PR as part of this stack: should this be called `num_programs` to align better with the rest of the bindings and Python exposure. May also be worth to already rename that on the C++ side before binding to Python.

bindings/qdmi/qdmi.cpp
```diff
@@ -205,33 +292,149 @@ when the custom slot is unsupported.)pb");
...
+                   })
+      .def_prop_ro("profile",
+                   [](const QDMI_Program_Format& self) {
+                     return std::string(
+                         decodeFixedField(self.profile, "program profile"));
+                   })
+      .def_prop_ro("encoding",
+                   [](const QDMI_Program_Format& self) {
+                     return static_cast<QDMI_Program_Encoding>(self.encoding);
+                   })
```
burgholzer, line 344: Some of these feel a little bit more complicated than they need to be. Not all of these may need lambdas. Look for simplification opportunities.

bindings/qdmi/qdmi.cpp
```diff
@@ -205,33 +292,149 @@ when the custom slot is unsupported.)pb");
...
+      .def(
+          "__eq__",
+          [](const QDMI_Program_Format& self, const nb::handle other) {
+            return nb::isinstance<QDMI_Program_Format>(other) &&
+                   qdmi::equal(self, nb::cast<QDMI_Program_Format>(other));
+          },
```
burgholzer, line 350: this could likely be simplified by defining operator== on QDMI_Program_Format and using nanobind's builtin bindings for equals.

bindings/qdmi/qdmi.cpp
```diff
@@ -205,33 +292,149 @@ when the custom slot is unsupported.)pb");
...
+          nb::sig("def __eq__(self, arg: object, /) -> bool"))
+      .def("__hash__", [](const QDMI_Program_Format& self) {
+        size_t hash = std::hash<std::string_view>{}(
+            decodeFixedField(self.id, "program format ID"));
+        hash ^= static_cast<size_t>(self.version) << 1U;
+        hash ^= static_cast<size_t>(self.encoding) << 3U;
+        hash ^= std::hash<std::string_view>{}(
+                    decodeFixedField(self.profile, "program profile"))
+                << 5U;
```
burgholzer, line 359: same here. likely makes sense to directly expose the `std::hash` specialization in C++ and rely on nanobind functionality for this here.

bindings/qdmi/qdmi.cpp
```diff
@@ -205,33 +292,149 @@ when the custom slot is unsupported.)pb");
...
+          "The canonical QIR 2.1 Base Profile binary format.")
+      .def_prop_ro_static(
+          "QIR21_ADAPTIVE_TEXT",
+          [](nb::handle) { return qdmi::QIR21_ADAPTIVE_TEXT; },
+          "The canonical QIR 2.1 Adaptive Profile text format.")
+      .def_prop_ro_static(
+          "QIR21_ADAPTIVE_BINARY",
+          [](nb::handle) { return qdmi::QIR21_ADAPTIVE_BINARY; },
+          "The canonical QIR 2.1 Adaptive Profile binary "
+          "format.");
```
burgholzer, line 384: also feels like this could be simplified. I'll not repeat this for this file anymore. There are similar opportunities for simplification as pointed out already in the remainder of this file.

bindings/qdmi/qdmi.cpp
```diff
@@ -320,6 +523,11 @@ text, so the device must receive it as exact bytes. Pass ``bytes`` to
              &qdmi::Device::getSupportedProgramFormats,
              "Returns the list of program formats supported by the device.");
 
+  device.def("try_program_features", &qdmi::Device::tryGetProgramFeatures,
+             "program_format"_a,
+             "Returns the complete optional capability list for an exact "
+             "payload, or None when the metadata is unknown.");
```
burgholzer, line 523: the naming of these method is a bit odd. Can we find a better name and description and potentially even implementation for this?

bindings/qdmi/qdmi.cpp
```diff
@@ -391,39 +599,47 @@ when the custom slot is unsupported.)pb");
...
+         const std::optional<qdmi::CustomJobParameter>& custom3,
+         const std::optional<qdmi::CustomJobParameter>& custom4,
+         const std::optional<qdmi::CustomJobParameter>& custom5) {
+        std::vector<std::vector<std::byte>> bytes;
+        bytes.reserve(programs.size());
+        for (const auto& program : programs) {
+          const std::span value{static_cast<const std::byte*>(program.data()),
+                                program.size()};
+          bytes.emplace_back(value.begin(), value.end());
         }
```
burgholzer: Is there a way that we could change the implementation so that we can avoid these copies? Or is this intended so that the programs stay valid?

python/mqt/core/qdmi/__init__.pyi
```diff
@@ -138,46 +148,94 @@ class Job:
...
+    QIR21_BASE_BINARY: Final[ProgramFormat] = ...
+    """The canonical QIR 2.1 Base Profile binary format."""
 
-    CUSTOM3 = 999999997
+    QIR21_ADAPTIVE_TEXT: Final[ProgramFormat] = ...
+    """The canonical QIR 2.1 Adaptive Profile text format."""
 
-    CUSTOM4 = 999999998
+    QIR21_ADAPTIVE_BINARY: Final[ProgramFormat] = ...
+    """The canonical QIR 2.1 Adaptive Profile binary format."""
```
burgholzer, line 214: does it make sense to change these so that the version number comes at the end (similar to OpenQASM). Future additions would then start with the same prefix and only differ in the version postfix.

CHANGELOG.md
```diff
@@ -78,11 +78,17 @@ releases may include breaking changes.
 
 #### Other additions
 
+- ✨ Add backend-owned program serialization and result decoding to QDMI Qiskit
+  backends. [QDMI-on-IQM] now provides the IQM JSON integration and `MoveGate`
+  ([#2114], [#2226]) ([**@marcelwa**], [**@burgholzer**])
```
burgholzer, line 97: This feels like a leftover of the back port after the 3.9.1 release. Consider removing.

src/qdmi/devices/dd/Device.cpp
```diff
@@ -145,14 +147,33 @@ makeOperationAddresses(const std::array<OperationInfo, N>& ops) {
...
+                         .constraint_id = {},
+                         .constraint_value = 0},
+    QDMI_Program_Feature{.id = QDMI_PROGRAM_FEATURE_MEASURED_QUBIT_REUSE,
+                         .value = 0,
+                         .constraint_id = {},
+                         .constraint_value = 0},
+    QDMI_Program_Feature{.id = QDMI_PROGRAM_FEATURE_MEASUREMENT_RESULT_USE,
+                         .value = 0,
+                         .constraint_id = {},
+                         .constraint_value = 0},
```
burgholzer: Maybe I am confused here, but we definitely support these features in OpenQASM3.
I do think we also support Boolean Computation and forward branching (if-else statements) and potentially more features. Worth checking.

For QIR, we should essentially support everything currently in QIR 2.1 because our QIR runner supports it. We should indicate accordingly.

src/qdmi/devices/dd/Device.cpp
burgholzer, line 1: If I am reading this correctly, the DDSIM device only supports single program jobs. This feels like an unnecessary limitation. Jobs could either be launched in parallel in separate threads, or be run in sequence one after another.
Running in parallel requires multiple DD packages and maintaining them. This might be too much work. Running in sequence could reuse the same package, which has other issues, but should actually work smoothly.
I'll like the DDSIM device to expose as many modern QDMI features as possible so that it is a reasonable test device that comes dependency-free with MQT Core.
Try to accommodate that.

python/mqt/core/plugins/qiskit/backend.py
```diff
@@ -235,6 +235,39 @@ def is_convertible(device: QDMIDevice) -> bool:
         # Zoned operations cannot easily be represented in Qiskit's Target model
         return not any(op.is_zoned() for op in device.operations())
 
+    def _program_serializer(  # ruff:ignore[no-self-use]
```
burgholzer, line 250: Feels like a warning that should be fixed rather than ignored. Maybe I am wrong though and downstream overriders may need `self`, but it is not needed here.

python/mqt/core/plugins/qiskit/backend.py
```diff
@@ -831,12 +913,3 @@ def run(
...
         return QDMIJob(backend=self, jobs=qdmi_jobs, circuit_names=circuit_names)
-
-
-# MQT Core owns the two OpenQASM formats and registers them through the same
-# registry as everyone else, so the backend walks one ordered list of formats
-# with no format-specific branch. `mqt.core.plugins.qiskit.__init__` imports this
-# module whenever Qiskit is installed, so both formats are available as soon as
-# the adapter is.
-register_program_serializer(ProgramFormat.QASM3, _serialize_to_qasm3)
-register_program_serializer(ProgramFormat.QASM2, _serialize_to_qasm2)
```
burgholzer, line 850: This changes how program serialization for Qiskit is handled from the 3.9.1 release. Make sure all documentation is updated accordingly and there are no remaining fragments of the old integration style.

python/mqt/core/plugins/pennylane/device.py
```diff
@@ -179,10 +209,47 @@ def __init__(
...
+        supports_one_shot = (
+            reported is not None
+            and self._program_format == ProgramFormat.OPENQASM3
+            and mcm <= self._program_capabilities
+            and {"measure", "reset"} <= operations
+        )
+        self._capabilities = DeviceCapabilities(
+            qjit_compatible=False,
+            runtime_code_generation=False,
+            dynamic_qubit_management=False,
```
burgholzer, line 240: feels like the capability management could be better here. dynamic qubit management is an explicit payload capability.
Carefully check for a suitable mapping between feature of the payload and how that maps to PennyLane.

python/mqt/core/plugins/pennylane/converter.py
```diff
@@ -316,13 +350,67 @@ def _convert_qasm3(self, tape: QuantumScript) -> _ConvertedProgram:
...
+                if not self._supports_mcm():
+                    msg = "The selected payload does not support measurement-conditioned execution."
+                    raise UnsupportedOperationError(msg)
+                dependencies = operation.meas_val.measurements
+                if any(measurement.meas_uid not in seen_measurements for measurement in dependencies):
+                    msg = "A conditional uses a measurement that does not precede it."
+                    raise ValidationError(msg)
+                if len(dependencies) > 16:
+                    msg = "A conditional depends on more than 16 measurements."
+                    raise ValidationError(msg)
```
burgholzer, line 393: Is this an artificial limitation or an actual PennyLane limitation?

python/mqt/core/plugins/qiskit/backend.py
burgholzer, line 1: QDMI v1.4 also introduces multi-program job submission.
Qiskit backend `run` natively supports passing `QuantumCircuit | Sequence[QuantumCircuit]`. the latter case can now be natively be handled by QDMI instead of in the QDMI backend code itself. This should clean up the implementation quite a bit.
With the planned changes to the DDSIM QDMI backend, this can also be properly tested.

python/mqt/core/plugins/pennylane/device.py
burgholzer, line 1: I am not familiar enough with PennyLane to judge, but I would guess it also supports the submission of multiple circuits to a device as part of one job. This should now work natively with QDMI v1.4
Explore whether the device itself can be improved in that regard.

burgholzer (commented):
This one took quite a bit of time to review. and there are still several areas that I would like to see changed. However, I have the feeling this is going to be the last iteration on this before it is ready.

@denialhaag @simon1hofmann do you have any feedback on top of the above?

src/qdmi/devices/dd/Device.cpp
```diff
@@ -505,15 +613,16 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::submitQIRProgramSampling()
       // Update the measurement counts.
       ++counts_[runtime.getMeasurements()];
```
simon1hofmann: `getMeasurements()` stores slot 0 first, but QDMI requires the highest slot first. Values `[1,0,0]` therefore produce `"100"` instead of `"001"`.

src/qdmi/devices/dd/Device.cpp
simon1hofmann, line 1008: Sampled jobs always expose histograms even when bool, integer, floating-point, or container outputs are omitted. QDMI requires `QDMI_ERROR_NOTSUPPORTED` when output cannot be represented losslessly.

python/mqt/core/plugins/qiskit/backend.py
```diff
@@ -441,6 +526,16 @@ def _build_target(self) -> Target:
...
 
+        capabilities = self._program_capabilities or set()
+        for capability, instruction, name in (
+            ("forward-branching", IfElseOp, "if_else"),
+            ("counted-iteration", ForLoopOp, "for_loop"),
+            ("conditional-loop", WhileLoopOp, "while_loop"),
+            ("multiway-branching", SwitchCaseOp, "switch_case"),
+        ):
+            if capability in capabilities:
+                target.add_instruction(instruction, name=name)
```
simon1hofmann, line 537: Circuits using measurement-driven branches and measured-qubit reuse are accepted without the separately required capabilities.

python/mqt/core/plugins/qiskit/backend.py
```diff
@@ -797,8 +870,17 @@ def run(
...
+            while pending:
+                instruction = pending.pop()
                 op_name = instruction.operation.name
+                for block in getattr(instruction.operation, "blocks", ()):
+                    pending.extend(block.data)
+                if op_name in {"if_else", "for_loop", "while_loop", "switch_case"}:
+                    if op_name not in self._target.operation_names:
+                        msg = f"Unsupported control flow operation: '{op_name}'"
+                        raise UnsupportedOperationError(msg)
+                    continue
```
simon1hofmann, line 895: Circuits using measurement-driven branches and measured-qubit reuse are accepted without the separately required capabilities.

src/qdmi/devices/dd/Device.cpp
simon1hofmann, line 720: State extraction neither emits the required QIR header nor assigns `programOutput_` (observed at [:916-918](https://github.com/munich-quantum-toolkit/core/blob/b7d1f2ac128fc33b64209db1d371454dc543a775/src/qdmi/devices/dd/Device.cpp#L916-L918)).

include/mqt-core/qdmi/ProgramFormat.hpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+[[nodiscard]] constexpr bool
+isValidProgramFormat(const QDMI_Program_Format& format) noexcept {
+  return format.version != 0U && format.id[0] != '\0' &&
+         (format.encoding == QDMI_PROGRAM_ENCODING_TEXT ||
+          format.encoding == QDMI_PROGRAM_ENCODING_BINARY) &&
+         detail::isCanonicalFixedString(format.id) &&
+         detail::isCanonicalFixedString(format.profile);
+}
```
simon1hofmann, line 79: It accepts binary OpenQASM, unsupported QIR profiles and unqualified vendor IDs, causing `NOTSUPPORTED` where QDMI requires `INVALIDARGUMENT`.

src/qdmi/devices/dd/Device.cpp
```diff
@@ -764,12 +876,13 @@ auto MQT_DDSIM_QDMI_Device_Job_impl_d::getProbabilities(const size_t size,
...
 }
-auto MQT_DDSIM_QDMI_Device_Job_impl_d::getResults(const QDMI_Job_Result result,
+auto MQT_DDSIM_QDMI_Device_Job_impl_d::getResults(const size_t programIndex,
+                                                  const QDMI_Job_Result result,
                                                   const size_t size, void* data,
                                                   size_t* sizeRet)
     -> QDMI_STATUS {
   if ((data != nullptr && size == 0) ||
-      IS_INVALID_ARGUMENT(result, QDMI_JOB_RESULT)) {
+      IS_INVALID_ARGUMENT(result, QDMI_JOB_RESULT) || programIndex != 0U) {
```
simon1hofmann: Index 1 on a completed single-program job returns `INVALIDARGUMENT`; QDMI requires `OUTOFRANGE`.

src/qdmi/driver/Driver.cpp
simon1hofmann, line 642: The `>= *_MAX` checks reject QDMI’s valid custom range `999999995..INT32_MAX` instead of returning `NOTSUPPORTED`.

python/mqt/core/qdmi/__init__.pyi
simon1hofmann, line 101: The runtime binding accepts `program_index: int = 0`, but every overload omits it, so valid multi-program calls fail static type checking.

simon1hofmann (changes requested):
Went through the changes and have some comments that should be double-checked.

src/qdmi/devices/dd/Device.cpp
burgholzer: If this is true, then we may want to adjust the QIR runtime function itself so that it follows the same convention. We should generally try to stick to the same convention throughout MQT Core wherever we can; even if this means changes now.

src/qdmi/devices/dd/Device.cpp
burgholzer: Hm. This feels a bit wrong to me. Shouldn't it be up to `getHistogram` to report the error? Maybe I am misinterpreting this though?

python/mqt/core/plugins/qiskit/backend.py
burgholzer: Is that something that we can reasonably express in Qiskit? without going entirely out of our way checking the program?
This may become a little easier once we can actually wedge the compiler collection between the input program and the required output program.

src/qdmi/devices/dd/Device.cpp
burgholzer: State extraction itself should likely not do this at all. But we should definitely be exposing the output stream from QIR through the new result type, which should already be captured by the QIR runner.

include/mqt-core/qdmi/ProgramFormat.hpp
burgholzer: Fair enough. Validation could be more strict here.

src/qdmi/devices/dd/Device.cpp
burgholzer: This is true, but the actual implementation will likely change a bit based on the other review comments. It is still good to make use of `OUTOFRANGE` for out of range access to the result list.

src/qdmi/driver/Driver.cpp
burgholzer: I believe this is fixed in a later PR down the line in this PR stack. Might be worth pulling the changes in here already.

python/mqt/core/qdmi/__init__.pyi
burgholzer: This is definitely an oversight and needs to be fixed either in the bindings themselves or in the pattern file.

src/qdmi/devices/dd/Device.cpp
simon1hofmann: Yes that should be reported by `getHistogram`, but that function currently never reports that error 🤔 

🤖 AI text below 🤖
Yes, getHistogram is the better place to report this. My concern was about the current behavior, not that the check must live in this switch. getHistogram currently always succeeds, while counts_ only contains RESULT/RESULT_ARRAY bits. For AdaptiveRecordOutputs.ll, that omits the recorded integer and double values, so the three-bit histogram is only a partial representation. Under the pinned QDMI contract, the histogram queries must return QDMI_ERROR_NOTSUPPORTED in that case. Container records alone do not invalidate the histogram; that part of my wording was too broad.

python/mqt/core/plugins/qiskit/backend.py
simon1hofmann: Probably not..
Just noticed that for example in this test: [test_mock_backend.py:547-566](https://github.com/munich-quantum-toolkit/core/blob/b7d1f2ac128fc33b64209db1d371454dc543a775/test/python/plugins/qiskit/test_mock_backend.py#L547-L566) `mid-circuit-measurement` should be required due to the `circuit.measure(0, 0)`, but `boolean-computation` is not needed.

src/qdmi/devices/dd/Device.cpp
burgholzer: Hm. We may want to adjust the QDMI spec here. This feels overly restrictive. It may still make sense to export a histogram, even though there is further output in the proprietary results.
Good time to make changes to the open QDMI PRs so this becomes less of a hassle.

docs/conf.py
```diff
@@ -176,10 +177,12 @@ def format_url(self, _e: Entry) -> HRef:  # ruff:ignore[no-self-use]
 
 
 cpp_api_tagfile = ("_build/doxygen/mqt-core.tag", "cpp/", "_build/doxygen/xml")
+_qdmi_api_base = "https://munich-quantum-software-stack.github.io/QDMI/pr-preview/pr-509/"
```
denialhaag: Just noting that this likely needs to be changed before merging.

docs/Doxyfile
```diff
@@ -2101,7 +2101,7 @@ SKIP_FUNCTION_MACROS   = YES
 # the path). If a tag file is not located in the directory in which Doxygen is
 # run, you must also specify the path to the tagfile here.
 
-TAGFILES               = _build/qdmi.tag=https://munich-quantum-software-stack.github.io/QDMI/v1.3.2/
+TAGFILES               = _build/qdmi.tag=https://munich-quantum-software-stack.github.io/QDMI/pr-preview/pr-509/
```
denialhaag: Same here.

include/mqt-core/qdmi/ProgramFormat.hpp
```diff
@@ -0,0 +1,88 @@
...
+inline constexpr QDMI_Program_Format OPENQASM3{
+    .version = QDMI_MAKE_VERSION(3, 0, 0),
+    .encoding = QDMI_PROGRAM_ENCODING_TEXT,
+    .id = "openqasm",
+    .profile = ""};
+inline constexpr QDMI_Program_Format QIR21_BASE_TEXT{
+    .version = QDMI_MAKE_VERSION(2, 1, 0),
+    .encoding = QDMI_PROGRAM_ENCODING_TEXT,
+    .id = "qir",
+    .profile = "base"};
```
denialhaag, line 45: This ties in with my comments in https://github.com/munich-quantum-toolkit/core/pull/2219:

1. Does a patch version for QIR make sense?
2. Should we use enum values for `id` (and `profile`)?

denialhaag (commented):
Some additional comments from my side, none of which I'm _super_ sure about myself:

docs/conf.py
burgholzer: Yeah. It needs to at least point to a merged version on the develop branch.
Before the eventual v4 release it needs to be pinned to a released version (the v1.4.0 tag).

include/mqt-core/qdmi/ProgramFormat.hpp
burgholzer: See answers there
- debatable
- likely no because these can be customized by providers.

denialhaag: Yeah, I just noticed that this runs counter to one of the PR's ideas. Sorry about that. 😬 

> replaces the closed program-format enum with exact immutable descriptors and optional feature records

burgholzer: No worries. These PRs are quite a bit to digest.

src/qdmi/driver/Driver.cpp
burgholzer: After checking again, this might not actually be fixed yet. And it definitely should be fixed. Here would be a good place to do so.

PR #2227, opened 2026-08-24
✨ Build target environments from QDMI payloads

burgholzer:
🤖 *AI text below* 🤖

## Description

Bridge the independently developed compiler model (#2219) and QDMI runtime prototype (#2226). Snapshot an open device and one exact accepted program format into an owning `TargetEnvironment`, preserving grouped optional features, encoding and the prototype's standard-format baseline.

C++ and Python factories share the existing target inference contract: unknown topology or gate-set facts fail early, and DDSIM's arbitrary controlled-operation families and zero-arity global phase remain supported. Python device factories reuse configuration validation at the extension boundary. Examples use DDSIM when they need an executable QIR format; the model-only SC device is not presented as an execution provider.

### Integration-only base

The base branch `codex/qdmi-capability-integration` temporarily merges #2219 and #2226 **only to build and test this adapter**. It is not a new PR or serial dependency. Neither foundation depends on the other, and #2162 is not included. Once both foundations are available in `v4.1`, retarget this existing PR there and remove the temporary base. The compiler foundation targets Core 4.0 on `main`; the QDMI runtime foundation targets Core 4.1 on `v4.1`.

Native multi-program adoption (#2373), SDK batching consumers, metadata removal, replaceable drivers, and scheduler integration remain independent.

### Design and release gate

This remains a draft, non-blocking Core 4.1 / QDMI 1.4 candidate. It must not enter Core 4.0. [QDMI #523](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/523) and Core #2365 own the unresolved program-capability contract; passing prototype tests does not settle that design. Use released dependency pins before release artifacts are published.

Refs #2365, Munich-Quantum-Software-Stack/QDMI#523.

### Validation

- Independent combined-foundation native suite: 3,891 passed; one existing skip.
- Python compiler, QDMI, Qiskit and PennyLane suite: 454 passed.
- Generated stubs, repository lint and C++ lint passed.
- Tests cover malformed descriptors/features, retained target inference, snapshot and exception behavior, conflicting Python configuration, and DDSIM QIR compilation/submission.
- Fresh-build test discovery was refreshed after staging the bundled catalogues, so parameterized device tests are included.
- Hosted CI and contract review remain separate gates.

AI assistance was used for rebasing, integration adaptations, tests, documentation and this description.

## Adapter acceptance

The adapter must reflect accepted capabilities for the selected device and format, without inventing missing topology, operations, or classical features. Validate both Qiskit and PennyLane compiler-entry cases through #2365. Keep service compilation guarantees distinct from serialization and compiler target facts; the underlying QDMI design remains a non-blocking release priority.

## Checklist

### DIFF
--- a/bindings/mlir/register_mlir.cpp
+++ b/bindings/mlir/register_mlir.cpp
@@ -20,7 +20,6 @@
 #include "mlir/Dialect/QCO/Utils/DDFunctionality.h"
 #include "mlir/bench/Generate.h"
 #include "qdmi/Client.hpp"
-#include "qdmi/ProgramFormat.hpp"
 #include "qdmi/driver/SessionConfig.hpp"
 #include "qiskit/Qiskit.h"
 
--- a/bindings/mlir/register_mlir.cpp
+++ b/bindings/mlir/register_mlir.cpp
@@ -20,6 +20,7 @@
 #include "mlir/Dialect/QCO/Utils/DDFunctionality.h"
 #include "mlir/bench/Generate.h"
 #include "qdmi/Client.hpp"
+#include "qdmi/ProgramFormat.hpp"
 #include "qdmi/driver/SessionConfig.hpp"
 #include "qiskit/Qiskit.h"
 
@@ -39,6 +40,7 @@
 #include <nanobind/stl/string_view.h>
 #include <nanobind/stl/variant.h>
 #include <nanobind/stl/vector.h>
+#include <qdmi/constants.h>
 
 #include <cctype>
 #include <complex>
@@ -188,6 +190,31 @@ template <nb::exception_type Exception = nb::exception_type::value_error,
     throw nb::builtin_exception(Exception, full.c_str());
   }
   return *std::move(result);
+}
+
+[[nodiscard]] static qdmi::Device openQDMIDevice(
+    const std::string& deviceId, std::optional<std::string> baseUrl,
+    std::optional<std::string> token,
+    std::optional<std::filesystem::path> authFile,
+    std::optional<std::string> authUrl, std::optional<std::string> username,
+    std::optional<std::string> password,
+    std::optional<std::string> deviceConfig,
+    std::optional<std::filesystem::path> deviceConfigFile,
+    std::optional<std::string> custom1, std::optional<std::string> custom2,
+    std::optional<std::string> custom3, std::optional<std::string> custom4,
+    std::optional<std::string> custom5) {
+  /// Validate before crossing the extension boundary for consistent ValueError.
+  if (deviceConfig && deviceConfigFile) {
+    throw nb::value_error(
+        "device_config and device_config_file are mutually exclusive");
+  }
+  const auto overrides = qdmi::makeDeviceSessionConfig(
+      std::move(baseUrl), std::move(token), std::move(authFile),
+      std::move(authUrl), std::move(username), std::move(password),
+      std::move(deviceConfig), std::move(deviceConfigFile), std::move(custom1),
+      std::move(custom2), std::move(custom3), std::move(custom4),
+      std::move(custom5));
+  return qdmi::Session::openDevice(deviceId, overrides);
 }
 
 template <class ProgramType>
@@ -918,21 +945,13 @@ either unrestricted or explicitly enumerated native-operation support.)pb");
              std::optional<std::string> custom3,
              std::optional<std::string> custom4,
              std::optional<std::string> custom5) {
-            // Keep this preflight at the Python boundary so the public
-            // ValueError does not depend on cross-extension exception
-            // translation.
-            if (deviceConfig && deviceConfigFile) {
-              throw nb::value_error(
-                  "device_config and device_config_file are mutually "
-                  "exclusive");
-            }
-            const auto overrides = qdmi::makeDeviceSessionConfig(
-                std::move(baseUrl), std::move(token), std::move(authFile),
-                std::move(authUrl), std::move(username), std::move(password),
-                std::move(deviceConfig), std::move(deviceConfigFile),
-                std::move(custom1), std::move(custom2), std::move(custom3),
-                std::move(custom4), std::move(custom5));
-            auto device = qdmi::Session::openDevice(deviceId, overrides);
+            auto device = openQDMIDevice(
+                deviceId, std::move(baseUrl), std::move(token),
+                std::move(authFile), std::move(authUrl), std::move(username),
+                std::move(password), std::move(deviceConfig),
+                std::move(deviceConfigFile), std::move(custom1),
+                std::move(custom2), std::move(custom3), std::move(custom4),
+                std::move(custom5));
             return takeResult(mlir::compilerTargetFromDevice(device));
           },
           "device_id"_a, nb::kw_only(), "base_url"_a = std::nullopt,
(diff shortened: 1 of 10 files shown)


mlir/include/mlir/Compiler/QDMIAdapter.h
```diff
@@ -44,6 +46,28 @@ compilerTargetFromDevice(const qdmi::Device& device);
...
+/**
+ * @brief Open a registered QDMI device and snapshot one accepted payload.
+ *
+ * @details This adapter contains exceptions from the QDMI C++ API and returns
+ * them as LLVM errors. The returned environment owns all queried metadata.
+ */
+[[nodiscard]] llvm::Expected<TargetEnvironment>
+targetEnvironmentFromDeviceId(std::string_view deviceId,
+                              const QDMI_Program_Format& format);
+
```
burgholzer: This is just me trying to be pragmatic: Most devices will expose a single payload format that they support. In these cases, it would be quite convenient to define a shortcut that either just uses the device or the ID without an explicit program format in the call.
The only thing to choose would be what to do with devices that offer multiple formats. One plausible option would be to default to the first supported format, which, by definition, is the preferred format by the device.
It should likely be pointed out explicitly that there is an overload for the convenience shortcut that allows specifying an explicit format.
As a follow-up, all code that does not explicitly need it otherwise and all docs may be moved to the convenience function.

mlir/lib/Compiler/QDMIAdapter.cpp
```diff
@@ -379,6 +384,140 @@ snapshotCompilerTarget(const qdmi::Device& device) {
...
+      [](FeatureGroup& group) { return std::move(group.capability); });
+  return capabilities;
+}
+
+[[nodiscard]] static std::vector<ProgramCapability>
+payloadBaseline(const PayloadFormat& format) {
+  if (format.id != "qir" || format.version != "2.1.0" ||
+      format.profile != "adaptive") {
+    return {};
+  }
```
burgholzer, line 547: The 2.1.0 feels a bit more restrictive than this likely needs to be.

python/mqt/core/mlir.pyi
```diff
@@ -15,6 +15,7 @@ from typing import Literal, Unpack, overload
 
 import qiskit.circuit
 
+import mqt.core.qdmi
```
burgholzer, line 20: This line feels odd and I am pretty sure that we can get rid of that by better writing the bindings.

burgholzer (commented):
This largely looks good to me. I one simplification request, one lifted constraint and likely some better bindings. I have the feeling that changes in the lower layers of the PR stack will trigger a few more changes here.

I am also wondering whether there is a bit too much overlap/duplication between the MLIR and the QDMI concepts and whether some of that could be simplified while still maintaining the proper separation between the two.

@denialhaag @simon1hofmann same question here as in the other PRs: Do you also see some things that need changes?

mlir/unittests/Compiler/test_compiler_qdmi_adapter.cpp
```diff
@@ -10,21 +10,175 @@
...
+      return QDMI_ERROR_INVALIDARGUMENT;
+    }
+    auto& library = fromSession(session);
+    switch (property) {
+    case QDMI_DEVICE_PROPERTY_NAME: {
+      static constexpr std::string_view DEVICE_NAME{"adapter-test"};
+      return copyValue(DEVICE_NAME.data(), DEVICE_NAME.size() + 1U, size, value,
+                       sizeRet);
+    }
+    case QDMI_DEVICE_PROPERTY_QUBITSNUM: {
```
simon1hofmann, line 92: Was this the reason for calling it `qubits_num` in the other PR?

docs/mlir/target_compilation.md
simon1hofmann, line 44: `CompilerTarget` is not imported anymore (deleted in line 18).

docs/mlir/target_compilation.md
```diff
@@ -15,34 +15,27 @@ Open a configured QDMI device and snapshot it as a compiler target:
...
     compile_program,
 )
+from mqt.core.qdmi import ProgramFormat
 
-target = CompilerTarget.from_device_id("mqt.sc.iqm.garnet")
-payload = PayloadSpecification(PayloadFormat("qir", "2.1.0", "base", PayloadEncoding.BINARY))
-environment = TargetEnvironment(target, payload)
+environment = TargetEnvironment.from_device_id(
+    "mqt.sc.iqm.garnet",
+    ProgramFormat.QIR21_BASE_BINARY,
```
simon1hofmann, line 25: The SC provider does not advertise any [program formats](https://github.com/munich-quantum-toolkit/core/blob/6d85e61d6e40c9ab2817332d222ab628784d37e5/src/qdmi/devices/sc/Device.cpp#L254-L262) right? 
Same for the example in lines 111-112.

simon1hofmann (approved):
Only have a couple of comments about the docs, but besides that this PR also looks really good!

bindings/mlir/register_mlir.cpp
```diff
@@ -25,6 +26,7 @@
 #include <nanobind/stl/string_view.h> // NOLINT(misc-include-cleaner)
 #include <nanobind/stl/variant.h>     // NOLINT(misc-include-cleaner)
 #include <nanobind/stl/vector.h>      // NOLINT(misc-include-cleaner)
+#include <qdmi/constants.h>
```
denialhaag, line 42: We still need to streamline our includes when we have fewer PRs open, but this would comply with the current style:


```suggestion
#include "qdmi/constants.h"
```

docs/mlir/target_compilation.md
```diff
@@ -15,34 +15,27 @@ Open a configured QDMI device and snapshot it as a compiler target:
...
 the canonical QCO pipeline. The targeted overload therefore accepts one
-`TargetEnvironment` and no independent output or custom pipeline. MQT Core's
-QDMI adapter does not yet translate QDMI program-format and feature metadata, so
-callers must construct the payload specification from the device documentation.
-
-The example has no reported execution capabilities. A producer must add every
-effective capability, including the selected format's baseline. Set
-`optional_capabilities_known=True` only when the producer also knows that the
-list contains every optional device capability.
+`TargetEnvironment` and no independent output or custom pipeline.
```
denialhaag, line 38: This might just be me, but this feels overly technical and not user-facing. Again, this might just be me, but my brain switched off while I was reading. 🤠

denialhaag (approved):
This also LGTM! 🙂 I just have one nitpick and a comment on the documentation:

mlir/unittests/Compiler/test_compiler_qdmi_adapter.cpp
burgholzer: Yeah. Mainly. But we use the other wording practically everywhere else. So I would be fine with the syntactic break going from C to C++.

docs/mlir/target_compilation.md
burgholzer: Ah. We need to be careful here while the docs in CI are failing. They should still all pass locally for all of the changes here and produce correct output.

docs/mlir/target_compilation.md
burgholzer: It does not. But maybe it should.
Maybe the configuration for the device should include which formats the device claims to accept/support. Would allow a more faithful model of an IQM machine.

bindings/mlir/register_mlir.cpp
burgholzer: Yeah. You are right. Still have this wired wrongly in my brain. We did device at some point to only treat stdlib includes with angle brackets, right?
Because this is still an external header pulled in as a dependency.

docs/mlir/target_compilation.md
burgholzer: This entire page needs a revamp once the implementation is consolidated.
And some of the terminology might still need bike shedding. As pointed out in several of my own review comments, I am not quite happy yet with how some of the things are framed.

bindings/mlir/register_mlir.cpp
denialhaag: Yeah, but then we changed our minds again, following the discussion in https://github.com/FullStaQD/compiler/pull/37. I still have it on my list to apply this across the MQT. Maybe just before the v4 release is a good time, since we hopefully won't have too many open PRs at that point. 🤔

burgholzer: Yeah. Before the release makes sense.
We may even want to consider that as a point where we may use an "mqt" namespace for our code to differentiate it a tad bit more from plain mlir. Similar to how this is done in FullStaQD with `qcc`. I was initially opposed to that but I am starting to see the potential benefits.

PR #2228, opened 2026-08-24
✨ Fuse parameterized single-qubit unitary runs

### LINKED ISSUE #1764: ✨ Fuse single-qubit unitary runs with parametrized (dynamic-angle) gates
### Problem Statement

The `fuse-single-qubit-unitary-runs` (#1672) pass currently composes concrete `Eigen` matrices, so it can only fuse gates whose unitary matrix is known at compile time (see the note in `mlir/include/mlir/Dialect/QCO/Transforms/Passes.td`). Any single-qubit gate whose rotation angle is a dynamic SSA Value (e.g. a function argument or a loop-induction-derived parameter) cannot join a run and is left unfused. This means parametrized circuits — an important and not-yet-well-covered case in the community — miss out on single-qubit fusion entirely.

### Proposed Solution

Extend the pass to fuse runs containing parametrized gates by composing them symbolically instead of numerically:

- Build the composed single-qubit operation from SSA Values and emit the resulting angles as `arith` ops, analogous to how the quaternion-based `merge-single-qubit-rotation-gates` pass already merges rotation runs over symbolic values.
- Handle the extra wrinkle that full Euler resynthesis needs more than multiplication (e.g. `atan2`/`arg` for angle extraction), either by emitting those as `math`/`arith` ops or by restricting symbolic fusion to gate families with a closed-form symbolic merge.
- Keep the existing constant-matrix fast path; fall back to symbolic composition only when an operand is dynamic.

simon1hofmann:
🤖 *AI text below* 🤖

## Description

- Compose named single-qubit gates with dynamic SSA parameters through the
  existing quaternion-based symbolic path.
- Emit exact runtime sequences in all seven supported single-qubit bases:
  U, ZYZ, ZXZ, XZX, XYX, ZSXX, and R.
- Lower symbolic runs to the resolved basis before target-native synthesis.
- Cover every primitive parameterized single-qubit gate across every basis,
  both Euler singular branches, short same-axis fusion, and all target-resolved
  bases.

Dynamic `pow`, arbitrary dynamic unitary shells, controlled shells, and
symbolic two-qubit synthesis remain explicit run boundaries.

Fixes #1764

## Validation

- 236 decomposition tests passed.
- 122 optimization tests passed.
- 132 compiler tests passed.
- `uvx nox -s lint` passed.
- `git diff --check` passed.

## AI assistance

OpenAI Codex assisted with implementation, tests, review, and this pull-request
description. Three independent final reviews found no blocking correctness,
scope, or test-coverage issue. A human remains responsible for reviewing and
accepting the contribution.

## Checklist

- [x] The changes are fully tested and pass the CI checks. Local checks pass; CI is pending.


simon1hofmann:
@coderabbitai full review

mlir/unittests/Dialect/QCO/Transforms/Decomposition/test_euler_decomposition.cpp
```diff
@@ -958,6 +994,251 @@ TEST(FuseSingleQubitUnitaryRunsTest, IgnoresDynamicPowerExponent) {
...
+  rotations[1].getThetaMutable().assign(funcOp.getArgument(1));
+
+  ASSERT_TRUE(succeeded(runFuse(*owned, "zyz")));
+  rotations.clear();
+  funcOp.walk([&](RZOp op) { rotations.push_back(op); });
+  ASSERT_EQ(rotations.size(), 1U);
+  EXPECT_TRUE(
+      valueDependsOn(rotations.front().getTheta(), funcOp.getArgument(0)));
+  EXPECT_TRUE(
+      valueDependsOn(rotations.front().getTheta(), funcOp.getArgument(1)));
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Addressed in 94a99deb9. The same-axis regression now binds unequal concrete values, canonicalizes the fused expression, and checks matrix equivalence against the ordered product of both RZ operations. This distinguishes addition from subtraction or another incorrect combination.

mlir/lib/Compiler/TargetCompilation.cpp
```diff
@@ -28,6 +48,17 @@ void populateTargetCompilationPipeline(OpPassManager& pm,
...
+    if (const auto targetBasis = target.synthesisBasis()) {
+      if (const char* basis =
+              symbolicFusionBasisName(targetBasis->singleQubit)) {
+        qco::FuseSingleQubitUnitaryRunsOptions options;
+        options.basis = basis;
+        options.skipControlledBodies = true;
+        pm.addPass(qco::createFuseSingleQubitUnitaryRuns(options));
+      }
+    }
+  }
```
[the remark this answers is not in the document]
simon1hofmann: 🤖 *AI text below* 🤖

Addressed in 94a99deb9. Added an end-to-end target regression using a ZSXX target that advertises the ctrl shell as native. It verifies that the controlled region still contains exactly one dynamic qco.u and that all three parameters remain the original entry argument after target compilation.

simon1hofmann:
@coderabbitai review

simon1hofmann:
@coderabbitai review

burgholzer (commented):
Hey @simon1hofmann 👋🏼 

I also let this one through my review agent once to provide you with some feedback and I believe it identified a couple of bits and pieces that still need tweaking here.

🤖 *AI text below* 🤖

I see three actionable items:

1. Please add direct lowering fast paths for simple parameterized operations, especially canonical dynamic U to U/ZYZ/ZXZ/ZSXX and direct single-axis cases. The current singleton path constructs a quaternion and then extracts Euler angles again. This emits avoidable `sin`, `cos`, `acos`, `atan2`, and `floor` operations, plus a large arithmetic graph, where closed-form parameter rewiring suffices. Keep the general quaternion fallback for genuinely composed runs and the transformed bases that require it.
2. Please remove `skip-controlled-bodies` from the public pass and dynamic-pattern plumbing unless there is a concrete external caller. The target pipeline no longer uses it, and setting it currently suppresses RX/RY/RZ/P canonicalization globally, not only inside `qco.ctrl`. The option therefore does more than its description promises.
3. The dynamic merge can accumulate the quaternion in the conversion loop instead of buffering every quaternion and folding them in a second loop. The supported-gate predicate already controls membership, and the buffer does not provide rollback because conversion has already emitted helper IR.

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
```diff
@@ -349,7 +351,11 @@ static void lowerTargetOperation(IRRewriter& rewriter, UnitaryOpInterface op,
   rewriter.setInsertionPoint(operation);
   if (op.isSingleQubit()) {
     Matrix2x2 matrix;
-    op.getUnitaryMatrix2x2(matrix);
+    if (!op.getUnitaryMatrix2x2(matrix)) {
+      decomposition::synthesizeParameterizedUnitary1Q(rewriter, operation,
```
burgholzer, line 355: 🤖 *AI text below* 🤖

Please avoid routing every runtime-parameterized single-qubit operation through the general quaternion-to-Euler path. At this call site, even a singleton `qco.rz` or an already canonical `qco.u` is converted into a quaternion and then decomposed again. For a dynamic U, that introduces six sine/cosine evaluations plus `acos`, two `atan2` operations, two `floor` operations, and a sizable arithmetic graph. U to U/ZYZ/ZXZ/ZSXX already has direct identities and needs no inverse trigonometry.

Could `synthesizeParameterizedUnitary1Q` first handle the cheap named-operation and basis combinations by directly reusing or rewiring their SSA parameters, then fall back to `mergeDynamicChain` only for genuinely composed input or XZX/XYX/R cases that need transformed extraction? This keeps the shared general implementation without charging its runtime cost on the common singleton path.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
```diff
@@ -82,12 +83,15 @@ def FuseSingleQubitUnitaryRuns
...
+    The pass also composes supported named gates with dynamic `f64` parameters
+    and emits conservative runtime sequences in the requested basis. Dynamic
+    `pow` and arbitrary dynamic unitaries remain unchanged.
   }];
-  let options = [Option<
-      "basis", "basis", "std::string", "\"zyz\"",
-      "Target Euler basis (zyz, zxz, xzx, xyx, u, zsxx, r).">];
+  let options = [Option<"basis", "basis", "std::string", "\"zyz\"",
+                        "Target Euler basis (zyz, zxz, xzx, xyx, u, zsxx, r).">,
+                 Option<"skipControlledBodies", "skip-controlled-bodies",
```
burgholzer: 🤖 *AI text below* 🤖

Can we delete this public option? After the target-native redesign, no in-tree caller sets it. More importantly, the dynamic composer implements `true` by omitting all RX/RY/RZ/P canonicalization patterns. As a result, a top-level `rz(%a); rz(%b)` run no longer takes the cheap same-axis merge even though the option claims to affect only operations nested in `qco.ctrl`.

The minimal fix is to remove the option and its dynamic-pattern plumbing; target-native synthesis already treats modifier bodies as opaque. If a concrete standalone-pass caller needs the option, then canonicalization must be scoped so that top-level behavior remains unchanged, with a regression for that contract. That is more machinery, so I would only keep it for an actual use case.

mlir/lib/Dialect/QCO/Transforms/Optimizations/MergeSingleQubitRotationGates.cpp
```diff
@@ -752,19 +816,20 @@ struct MergeSingleQubitRotationGatesPattern final
...
    * The pass-level global-phase normalization subsequently combines and
    * normalizes the emitted correction.
    *
    * Converts every gate before rewriting so a missing conversion Case cannot
    * leave partially rewired ops.
    */
-  static LogicalResult
-  mergeDynamicChain(MutableArrayRef<UnitaryOpInterface> chain,
-                    PatternRewriter& rewriter) {
+  static LogicalResult mergeDynamicChain(
```
burgholzer: 🤖 *AI text below* 🤖

This function can be shorter and avoid the temporary quaternion allocation. Chain membership is already restricted by `isMergeable`, and the `Value` conversions for those accepted gate kinds are total. The `SmallVector<Quat<Value>>`, its reservation, and the second Hamilton-product loop therefore do not add validation or rollback safety; helper IR has already been emitted during conversion if a later conversion were to fail.

Please accumulate `qAccum` in the conversion loop, as the static path already does. That removes the buffer and second pass while preserving the exact multiplication order and behavior.

mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp
simon1hofmann: U, ZYZ, ZXZ, and ZSXX now use direct SSA identities; only XZX, XYX, and R use quaternion/Euler fallback.

mlir/include/mlir/Dialect/QCO/Transforms/Passes.td
simon1hofmann: Remove this new public option, not really needed.

mlir/lib/Dialect/QCO/Transforms/Optimizations/MergeSingleQubitRotationGates.cpp
simon1hofmann: Yes that's possible, dynamic accumulation now streams inside the conversion loop with the same multiplication order as the static path.

burgholzer (approved):
This LGTM 👍🏼 Let's get this in!

Issue #2093, opened 2026-08-13
⬆️ Adopt QDMI 1.4 in MQT Core


burgholzer:
🤖 *AI text below* 🤖

## Elevator pitch

Adopt released QDMI 1.4 in Core 4.1 through independent, testable workstreams. Stable ABI cleanup, driver replacement, and native batching must not wait for an unsettled program-capability model or scheduler policy.

## Release boundaries

- Core v4.0 uses released QDMI 1.3.x. Do not merge QDMI 1.4 ABI changes into that release.
- Core 4.1 release artifacts must use a released QDMI 1.4 tag, not a development or PR commit. Candidate revisions are permitted for development validation only.
- Program capabilities are a priority for QDMI 1.4/Core 4.1, but not a release blocker. Release the settled streams without them if the design is not ready.
- Shared static Slurm deployment and both provider migrations target Core 4.1/MQSF. They remain technically independent of QDMI 1.4. Keep the working Quantum Week setup and limit immediate changes to necessary fixes.
- Availability-based scheduler implementation is unscheduled pending its design. QDMI v2 stack changes are out of scope.

## Independent PR workstreams

| Topic | QDMI | Core |
| --- | --- | --- |
| Metadata cleanup | [#512](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/512) → [#513](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/513), rooted on `v1.4` | #2233 on `v4.1`, independently of driver replacement |
| Replaceable drivers | [#511](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/511) rooted on `v1.4` | `v4.1` → #2229 → #2230 → #2231, without compiler or cleanup ancestry |
| Multi-program jobs | [#509](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/509) with the existing program-format enum | #2373 on `v4.1` contains the low-level extraction from #2226; SDK consumers follow independently |
| Program capabilities | [#508](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/508) behind the design gate | #2226 on `v4.1` is independent of #2219; #2227 bridges both |
| Compiler-only capabilities | No QDMI 1.4 dependency | `main` → #2219 → #2162 targets Core 4.0 after compiler contract review and validation |

[...]

PR #2229, opened 2026-08-24
♻️ Load a replaceable QDMI Client driver

burgholzer:
🤖 *AI text below* 🤖

## Description

Load a replaceable QDMI Client driver through its validated C ABI while retaining session ownership, stable device IDs, and existing compiler/SDK behavior.

This is the independent foundation of the driver workstream: **v4.1 → #2229 → #2230 → #2231**. It depends only on [QDMI #511](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/511), not metadata cleanup, batching, or payload capabilities. The development pin uses that isolated ABI branch. Published artifacts must use released QDMI 1.4.

Target **Core 4.1**, never v4.0. The optional private discovery/configuration extension is in #2230; standardizing that interface remains QDMI v2 work. Coordination: #2093.

Mechanical adaptations to bindings, SDKs, compiler device opening, Slurm, and fixtures stay here so this intermediate state builds. Current optional-shot APIs, optional-device builds, variadic gates, zero-arity global phase, and fail-closed target inference are preserved.

### Local validation

- Full release build and CTest: 3,869 passed, one existing skip.
- Selected QDMI, SDK, and MLIR Python tests: 455 passed.
- Stub generation, repository lint, and full changed-file C++ lint passed.
- Hosted CI is separate and must be checked on the final pushed head.

## Deployment acceptance

Demonstrate replacing the driver in an installed C++/Python environment without rebuilding the clients or changing SDK application code. Test the packaged default driver and a standards-only replacement without Core's optional discovery symbols. Preserve the default wheel experience and record compatibility with existing provider libraries.

## Checklist


cmake/AddMQTQDMIDevice.cmake
burgholzer, line 1: This feels right, but it also feels fairly complex. Are we sure all of this is really needed to properly distribute this? Keep later PRs in the stack in mind for this.

include/mqt-core/qdmi/Client.hpp
```diff
@@ -70,6 +69,49 @@ concept custom_property_value =
     std::same_as<T, std::vector<std::byte>>;
 
 namespace detail {
+struct ClientApi {
```
burgholzer, line 71: Likely `ClientAPI` would be better suited here. There may be other places benefitting from the improved spelling.

include/mqt-core/qdmi/Client.hpp
burgholzer, line 1: This should very closely follow the implementation in the driver so that this has a common look and feel that does not diverge too much. Aligning the two should be a priority here.

include/mqt-core/qdmi/Client.hpp
```diff
@@ -406,23 +458,13 @@ class Operation;
...
-   * session is not accessible.
-   */
-  [[nodiscard]] static Device createSessionlessDevice(QDMI_Device device);
-
-  /**
-   * @brief Opens a registered QDMI device as a fresh device session.
-   * @param id Stable registered device ID.
-   * @param overrides Session values that replace registered defaults.
-   * @return An owning device wrapper for the new session.
+   * @brief Opens a Client-visible QDMI device in a fresh session.
```
burgholzer, line 506:
```suggestion
   * @brief Opens a client-visible QDMI device in a fresh session.
```
Worth establishing consistently

include/mqt-core/qdmi/Client.hpp
```diff
@@ -70,6 +69,49 @@ concept custom_property_value =
...
+};
+
+struct ClientSession {
+  ClientSession(std::shared_ptr<const ClientApi> selectedApi,
+                QDMI_Session selectedSession)
+      : api(std::move(selectedApi)), handle(selectedSession) {}
+  ~ClientSession();
+
+  ClientSession(const ClientSession&) = delete;
+  ClientSession& operator=(const ClientSession&) = delete;
```
burgholzer, line 101: should this define (defaulted) move constructors and assignment?

include/mqt-core/qdmi/Client.hpp
```diff
@@ -70,6 +69,49 @@ concept custom_property_value =
...
+
+  std::shared_ptr<const ClientApi> api;
+  QDMI_Session handle;
+};
+
+struct JobDeleter {
+  void operator()(QDMI_Job_impl_d* job) const;
+  std::shared_ptr<ClientSession> session;
+};
```
burgholzer, line 110: This feels somewhat odd. Is this really a clean, idiomatic C++20 solution?

mlir/include/mlir/Compiler/QDMIAdapter.h
```diff
@@ -69,9 +70,9 @@ targetEnvironmentFromDeviceId(std::string_view deviceId,
                               const QDMI_Program_Format& format);
 
 /**
- * @brief List the stable IDs of registered QDMI devices.
+ * @brief List the stable IDs visible to a fresh QDMI Client session.
  *
- * @details This adapter contains exceptions from QDMI registry discovery and
+ * @details This adapter contains exceptions from QDMI Client discovery and
```
burgholzer, line 53: Just annotating this here, but it holds for the entire PR:
Always mentioning "QDMI Client" or "QDMI client" does not feel quite right to be honest. It just sounds very verbose and unnatural in many of the cases.
The previous "registry" also is not quite right anymore.
Maybe "driver session" or something along those lines would be a better fit.

mlir/lib/Compiler/QDMIAdapter.cpp
```diff
@@ -581,9 +580,16 @@ targetEnvironmentFromDeviceId(const std::string_view deviceId,
...
+    ids.reserve(devices.size());
+    std::ranges::transform(
+        devices, std::back_inserter(ids),
+        [](const qdmi::Device& device) { return device.getId(); });
+    return ids;
   } catch (...) {
-    return qdmiError("Failed to discover registered QDMI devices",
+    return qdmiError("Failed to discover QDMI devices",
                      std::current_exception());
   }
```
burgholzer, line 523: It feels like this should be functionality exposed by QDMI itself as part of the client interface. Evaluate and potentially add as part of the existing PR stack in QDMI in a suitable place. Then use wherever applicable here and simplify accordingly.

bindings/qdmi/qdmi.cpp
```diff
@@ -139,11 +160,72 @@ template <typename Query>
...
 
 NB_MODULE(MQT_CORE_MODULE_NAME, qdmiModule) {
-  qdmiModule.doc() = "QDMI entities and access to MQT Core's QDMI driver.";
-  auto driver = qdmiModule.def_submodule(
-      "driver", "Register, discover, and open QDMI devices through MQT Core.");
+  qdmiModule.doc() = "QDMI Client entities.";
   bindings::registerSlurm(qdmiModule);
 
+  nb::class_<qdmi::Session>(qdmiModule, "ClientSession",
+                            "One initialized QDMI Client session.")
```
burgholzer, line 112: Can we potentially simply refer to this as `Session`? It is called `qdmi::Session` already. I believe "Session" and "Device Session" is the right distinction here.
And maybe a framing around "One initialized session with a QDMI driver" or similar. Bike shedding to be done.

docs/qdmi/driver.md
```diff
@@ -6,22 +6,35 @@ mystnb:
...
+[the DDSIM QDMI Device](ddsim_device.md).
+
+This boundary lets another QDMI 1.4 Driver implement the Client interface
+without linking to MQT Core's Driver. MQT Core checks the complete Client
+function table and the QDMI Client ABI major and minor versions before it
+allocates a session.
+
+## Driver Selection
+
+MQT Core selects the Client driver for the process after the Driver passes ABI
```
burgholzer, line 26: "Client driver" -> "QDMI driver"
sounds better IMHO and should be used everywhere.
Likely the only thing where the word "client" is really important is when one explicitly needs to distinguish between the QDMI client and device interfaces.
In any case where there is no ambiguity, we may just use "QDMI".
We should likely avoid "QDMI interface" because the "I" already stands for "Interface".

docs/qdmi/driver.md
```diff
@@ -6,22 +6,35 @@ mystnb:
...
+and function-table validation and allocates the first raw session. The selection
+order is:
+
+1. `qdmi::SessionConfig::driverPath` or Python `driver_path`;
+2. the UTF-8 `MQT_CORE_QDMI_DRIVER` environment value;
+3. the packaged `MQT::CoreQDMIDriver` library.
+
+The selection remains active until process exit. A later explicit request for a
+different Driver fails. A failed load, ABI check, symbol check, or raw-session
```
burgholzer, line 35: This feels more restrictive than it likely needs to be. Is it so complicated technically to allow switching the driver at runtime?
That should work in a pretty similar fashion as the driver loading shared device libraries "live"/at runtime.

docs/qdmi/driver.md
```diff
@@ -38,28 +51,34 @@ For example, an embedded simulator consumer can enable only the DDSIM device,
 while CUDA-Q can enable the DDSIM and superconducting devices used by its
 integration tests.
 
-The QDMI driver and QDMI libraries are available independently. Device-free
-builds can register external device libraries through
-[QDMI device configuration](configuration.md). Building MQT Core's C++ tests
-requires both bundled devices so that the complete device integration is tested.
+The Client and Driver libraries are separate shared libraries. Device-free
```
burgholzer, line 58: I totally get why the Driver library is now forced to be a shared library.
However, I do not fully understand why the C++ Client library is also forced to be shared. Couldn't this still remain open (static by default; shared in wheels)?

docs/qdmi/driver.md
```diff
@@ -38,28 +51,34 @@ For example, an embedded simulator consumer can enable only the DDSIM device,
...
+builds can use another QDMI 1.4 Driver through `driver_path` or
```
burgholzer, line 59: I do not think we need to emphasize the "1.4" as much and can just refer to this as "QDMI". It's only really important in the changelog/upgrade guide or when we talk about the explicit compatibility guarantees. This should also make the wording a little more accessible.

docs/qdmi/driver.md
```diff
@@ -38,28 +51,34 @@ For example, an embedded simulator consumer can enable only the DDSIM device,
...
 
 ## Python Bindings
 
-The QDMI interface is the low-level contract implemented by a QDMI device. The
-MQT Core QDMI driver loads device libraries and implements the QDMI client
-interface. The C++ QDMI library adds owning wrappers for QDMI devices, sites,
-operations, and jobs. The Python module exposes these QDMI entities through
-{py:mod}`mqt.core.qdmi`. Its {py:mod}`mqt.core.qdmi.driver` submodule provides
-device discovery, registration, and opening.
+The C++ QDMI library adds owning wrappers for Client sessions, devices, sites,
```
burgholzer, line 66: I like "C++ QDMI library" much more than "Client library" or "QDMI Client library". I think it is much more natural to think about this as the C++ abstraction built on top of QDMI. And it is built on top of the client interface because we want replaceable drivers and we do not want to deal with the name shifting of the device libraries. Furthermore, the driver encapsulates administration and authorization logic that the upper levels of the stack should not necessarily see. Maybe there is a good place where one could institutionalize this reasoning (after fleshing it out a little more).

docs/qdmi/driver.md
```diff
@@ -38,28 +51,34 @@ For example, an embedded simulator consumer can enable only the DDSIM device,
...
 
 ```{code-cell} ipython3
-from mqt.core.qdmi.driver import open_device, registered_device_ids
+from mqt.core.qdmi import ClientSession, open_device
 
-for device_id in registered_device_ids():
-    device = open_device(device_id)
+for discovered in ClientSession().devices:
+    device = open_device(discovered.id)
     print(device.name())
```
burgholzer, line 82: Another idea looking at this with a (more or less) fresh mind: Instead of referring to this as `ClientSession` and having to add a lot of "Client" everywhere, it would maybe be more helpful to simply refer to this as "driver" in a similar fashion as we refer to the implementers of the device interface as "devices". "client" describes the person using the (driver) library; it does not describe the library itself; as opposed to "device" which actively describes what the library implements.

src/qdmi/Slurm.cpp
```diff
@@ -111,25 +110,32 @@ namespace {
...
+  Session session;
+  auto devices = session.getDevices();
+  std::vector<std::string> deviceIds;
+  deviceIds.reserve(devices.size());
+  std::ranges::transform(devices, std::back_inserter(deviceIds),
+                         [](const Device& device) { return device.getId(); });
+  const auto deviceId = parseLicense(licenseSpec, deviceIds);
+  const auto selected =
+      std::ranges::find(deviceIds, deviceId) - deviceIds.begin();
+  auto device = devices[static_cast<size_t>(selected)];
```
burgholzer, line 138: I'd hope that this becomes simpler again based on some of the other review comments.

src/qdmi/Client.cpp
burgholzer, line 1: Given how the new code here shares quite a bit of code with the loading logic in the driver, does it make sense to introduce a shared support library (could be header-only) that centralizes a bit of the code and avoids duplication?
I'd only do this if it genuinely saves code.

src/qdmi/Client.cpp
```diff
@@ -52,8 +66,311 @@ void validateProgramFeatureResult(const QDMI_Program_Feature& feature) {
...
+}
+
+[[nodiscard]] auto packagedDriverPath() -> std::filesystem::path {
+  const auto directory = thisModuleDirectory();
+  auto filename = std::filesystem::path{MQT_CORE_QDMI_DEFAULT_DRIVER_FILENAME};
+  for (const auto& candidate :
+       {directory / filename, directory / "lib" / filename,
+        directory / "bin" / filename,
+        directory.parent_path() / "lib" / filename,
+        directory.parent_path() / "bin" / filename}) {
```
burgholzer: Should this also cover `lib64`. IIRC we had problems with that already at some point on manylinux images.

src/qdmi/Client.cpp
```diff
@@ -52,8 +66,311 @@ void validateProgramFeatureResult(const QDMI_Program_Feature& feature) {
...
+                             std::string(name));
+  }
+  return function;
+}
+
+[[nodiscard]] auto loadClient(const std::filesystem::path& path)
+    -> LoadedClient {
+  auto* const library = openLibrary(path);
+  if (library == nullptr) {
+    throw std::runtime_error("Cannot load QDMI Client driver '" +
```
burgholzer, line 241: just to annotate this once more: We shouldn't use "QDMI Client driver" but simply "QDMI driver".

src/qdmi/Client.cpp
```diff
@@ -951,89 +1259,55 @@ Job::getSparseProbabilities(const size_t programIndex) const {
...
+    if (candidateId == id) {
+      return device;
+    }
+    if (!available.empty()) {
+      available += ", ";
+    }
+    available += candidateId;
+  }
+  throw std::out_of_range("QDMI Client session has no device with ID '" +
+                          std::string(id) + "'; available IDs: " + available);
```
burgholzer, line 1208: This feels like something that should be improved in the QDMI client interface upstream so that code like this becomes way simpler (and closer to what the custom driver extension actually offered before).

src/qdmi/Client.cpp
burgholzer, line 1: I get where most of the changes are coming from here and it is quite natural that there are substantial additions given the replaceable driver. However, some of the code, especially around lifetime handling feels more complicated than it necessarily needs to be in my opinion. Maybe some opportunities for simplification arise out of the consequences of the other review comments.

burgholzer (commented):
Alright. I have quite a few ideas here for how to make this better and improve the overall wording. I really like how this is shaping up though. I did not think that replaceable drivers would feel so natural. Let's see how much of the custom driver extension will actually remain in the end.

PR #2230, opened 2026-08-24
✨ Add an optional packaged QDMI Driver extension

burgholzer:
🤖 *AI text below* 🤖

## Description

Discover installed device catalogues without importing provider code, and open a fresh session for one stable device ID through Core's optional private default-driver extension.

Driver workstream: **v4.1 → #2229 → #2230 → #2231**. This PR depends only on #2229, not metadata cleanup, batching, or compiler/payload capabilities. It targets **Core 4.1 / QDMI 1.4**, not Core 4.0. Coordination: #2093.

The private extension stages trusted manifests at lowest precedence and supports strict targeted opening. Generic Client sessions continue to work with drivers that lack these symbols. Standardizing discovery/configuration remains QDMI v2 work.

### Local validation

- Full release build and CTest: 3,873 passed, one existing skip.
- QDMI, SDK, and MLIR Python tests: 467 passed.
- Generated stubs, repository lint, and full changed-file C++ lint passed.
- Tests cover metadata-only discovery, optional symbols, selection timing, strict configuration, session lifetime, and Windows-sensitive initialization ordering.
- Hosted CI is separate and must be checked on the final pushed head.

## Extension boundary

Keep the default-driver discovery/configuration extension optional. A replacement driver exposing only the standard Client ABI must remain usable; do not require provider repositories to implement Core-private symbols. Document how the default packaged driver keeps ordinary installations seamless.

## Checklist

### DIFF
--- a/docs/qdmi/driver.md
+++ b/docs/qdmi/driver.md
@@ -36,6 +36,45 @@ different Driver fails. A failed load, ABI check, symbol check, or raw-session
 allocation does not select a Driver, so a later call can retry. MQT Core keeps
 the selected shared library loaded while its function pointers can be used.
 
+## Optional Packaged-Driver Extension
+
+MQT Core's packaged Driver adds two private symbols to the same shared library
+that exports the standard Client interface:
+
+- `MQT_CORE_QDMI_driver_add_manifest_v1` stages a trusted package manifest.
+- `MQT_CORE_QDMI_driver_session_alloc_for_device_v1` allocates a session for one
+  configured stable ID.
+
+These symbols are an MQT Core extension. They are not part of a public QDMI
+header, and another Client driver can omit them. MQT Core resolves them as
+optional symbols and calls them only through the extension API. Missing
+extension symbols do not prevent standard Client sessions. Generic
+{cpp-api:func}`qdmi::Session::openDevice` and Python
+{py:func}`mqt.core.qdmi.open_device` enumerate the standard Client device list
+and never use the private targeted-session symbol.
+
+Use {cpp-api:func}`qdmi::default_driver::addManifest` or Python
+{py:func}`mqt.core.qdmi.default_driver.add_manifest` before the packaged Driver
+freezes its registry. Staging the packaged library does not select it as the
+generic Client driver. The first successful raw standard or targeted session
+allocation selects a Client driver. A later targeted-session initialization or
+device query failure does not undo that selection.
+
+By default, the `default_driver` extension resolves MQT Core's packaged Driver
+and ignores `MQT_CORE_QDMI_DRIVER`. An explicit `driver_path` overrides that
+default for a compatible extension. The process selection rule still prevents
+switching Drivers after a successful raw allocation. Standard Client sessions
+use the selection order above, including the environment override.
+
+Use {cpp-api:func}`qdmi::default_driver::openDevice` or Python
+{py:func}`mqt.core.qdmi.default_driver.open_device` when an application
+deliberately depends on the packaged Driver. The targeted call merges manifest
+defaults with the supplied JSON or Python overrides. It rejects unsupported
+parameters and malformed configuration, propagates device-library status codes,
+and requires the session to expose exactly one device. Each call creates an
+independent session. The returned device and its derived wrappers retain that
+session until the last wrapper is destroyed.
+
 ## Building the Bundled Devices
 
 Standalone MQT Core builds include the DDSIM and superconducting QDMI device
--- a/include/mqt-core/qdmi/Client.hpp
+++ b/include/mqt-core/qdmi/Client.hpp
@@ -494,6 +494,20 @@ class Site;
 class Device;
 class Operation;
 
+namespace default_driver {
+/// Stage one package manifest in MQT Core's optional driver extension.
+void addManifest(const std::filesystem::path& path);
+
+/// Open one default-driver device with strict merged session configuration.
+/// @param id Stable device ID.
+/// @param deviceSessionJson JSON session overrides.
+/// @param driverPath Optional compatible extension path. By default, this call
+/// uses MQT Core's packaged Driver and ignores `MQT_CORE_QDMI_DRIVER`.
+[[nodiscard]] Device openDevice(
+    std::string_view id, std::string_view deviceSessionJson = {},
+    const std::optional<std::filesystem::path>& driverPath = std::nullopt);
+} // namespace default_driver
+
 /**
  * @brief Class representing the Session library.
  * @details This class provides methods to query available devices and
@@ -803,6 +817,10 @@ private:
    */
   Device(QDMI_Device device, std::shared_ptr<detail::ClientSession> session)
       : device_(device), session_(std::move(session)) {}
+
+  friend Device
+  default_driver::openDevice(std::string_view, std::string_view,
+                             const std::optional<std::filesystem::path>&);
 
   /// Wrap operation handles while retaining their owning device session.
   [[nodiscard]] std::vector<Operation>
--- a/src/qdmi/Client.cpp
+++ b/src/qdmi/Client.cpp
@@ -158,6 +158,11 @@ void closeLibrary(LibraryHandle library) { dlclose(library); }
   if (path.empty()) {
     throw std::invalid_argument("QDMI Client driver path must not be empty");
   }
+  if (path.native().find(std::filesystem::path::value_type{}) !=
+      std::filesystem::path::string_type::npos) {
+    throw std::invalid_argument(
+        "QDMI Client driver path must not contain null bytes");
+  }
   std::error_code error;
   auto normalized = std::filesystem::weakly_canonical(
       std::filesystem::absolute(path, error), error);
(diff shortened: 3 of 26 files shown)


docs/qdmi/driver.md
```diff
@@ -36,6 +36,45 @@ different Driver fails. A failed load, ABI check, symbol check, or raw-session
...
+{cpp-api:func}`qdmi::Session::openDevice` and Python
+{py:func}`mqt.core.qdmi.open_device` enumerate the standard Client device list
+and never use the private targeted-session symbol.
+
+Use {cpp-api:func}`qdmi::default_driver::addManifest` or Python
+{py:func}`mqt.core.qdmi.default_driver.add_manifest` before the packaged Driver
+freezes its registry. Staging the packaged library does not select it as the
+generic Client driver. The first successful raw standard or targeted session
+allocation selects a Client driver. A later targeted-session initialization or
+device query failure does not undo that selection.
```
burgholzer, line 61: I believe that some of this will (have to) change based on the review in the PR this is stacked on.

docs/qdmi/driver.md
burgholzer, line 1: The wording here and throughout will have to be adapted based on the feedback on nomenclature in the lower PRs.

include/mqt-core/qdmi/Client.hpp
```diff
@@ -449,6 +449,20 @@ class Site;
 class Device;
 class Operation;
 
+namespace default_driver {
```
burgholzer, line 497: Is "default_driver" really what we want to use in terms of nomenclature?
Maybe "builtin_driver" would already be better.
Outside of MQT Core, I'd generally like to refer to it as the `MQT Core QDMI Driver`. I am not sure how well this flows internally in the project.
Maybe the driver we have could be framed as _the_ QDMI Driver, while the framework also offers support for third-party drivers.
This already affects a lower PR in the stack and should correspondingly be updated there already.

src/qdmi/driver/DeviceRegistry.cpp
```diff
@@ -417,37 +433,35 @@ void appendFragments(std::vector<std::filesystem::path>& files,
...
-  if (auto xdg = environment("XDG_CONFIG_HOME")) {
-    appendIfFile(files, std::filesystem::path(*xdg) / "mqt-core" / "qdmi.json");
-  } else if (auto home = environment("HOME")) {
-    appendIfFile(files, std::filesystem::path(*home) / ".config" / "mqt-core" /
-                            "qdmi.json");
+  if (auto xdg = environmentUtf8("XDG_CONFIG_HOME")) {
+    appendIfFile(files, pathFromUtf8(*xdg) / "mqt-core" / "qdmi.json");
+  } else if (auto home = environmentUtf8("HOME")) {
+    appendIfFile(files,
+                 pathFromUtf8(*home) / ".config" / "mqt-core" / "qdmi.json");
```
burgholzer, line 476: Is the `Utf8` in these names really that important? I'd argue we can just drop it (likely already in a PR further down the stack)

src/qdmi/driver/DeviceRegistry.hpp
```diff
@@ -21,6 +23,20 @@ namespace qdmi::detail {
...
 
+/// Stages one low-precedence package manifest before the driver is frozen.
+auto stagePackageManifest(const std::filesystem::path& path) -> int;
+
+/// Freezes and returns the staged package manifests.
+[[nodiscard]] auto freezePackageManifests()
+    -> std::vector<std::filesystem::path>;
+
+/// Reopens package-manifest staging after driver construction fails.
+void rollbackPackageManifestFreeze();
```
burgholzer, line 34: Another nomenclature thing "Package Manifest" does not sound like the very best fit.
Maybe "Device Library Manifest" or something more in the direction of "Device Record(s)". I am pretty sure there is good established terminology in CS for that that we can comfortably apply to quantum. This also already applies further down the stack.

src/qdmi/driver/Driver.cpp
burgholzer, line 1: feels like a `using namespace qdmi` could reduce a bit of code here (and maybe in other `.cpp` files as well (likely already applies further down the stack).

src/qdmi/driver/Driver.cpp
```diff
@@ -105,11 +105,37 @@ namespace {
...
+  if (status == QDMI_SUCCESS) {
+    return true;
+  }
+  if (status == QDMI_WARN_GENERAL) {
+    SPDLOG_WARN("QDMI device {} returned a general warning", action);
+    return true;
+  }
+  return false;
+}
+} // namespace
```
burgholzer, line 132: Does this really help? It likely does, but I am not sure it is actually consistently applied.

src/qdmi/Client.cpp
```diff
@@ -197,9 +202,26 @@ void closeLibrary(LibraryHandle library) { dlclose(library); }
...
+  }
+  const auto packaged = packagedDriverPath();
+  return packaged.has_parent_path() ? normalizePath(packaged) : packaged;
+}
+
 struct LoadedClient {
+  /// Private ABI v1 path arguments use UTF-8.
+  using AddManifest = int (*)(const char*);
+  using SessionAllocForDevice = int (*)(const char*, size_t, const char*,
+                                        QDMI_Session*);
```
burgholzer, line 223: Another explicit mention of UTF-8 that feels a bit overkill to me. Is that really that important here?

src/qdmi/Client.cpp
```diff
@@ -281,7 +311,13 @@ template <class Function>
...
                        QDMI_device_query_operation_property);
 #undef LOAD_CLIENT_SYMBOL
-    return {library, std::move(api)};
+    auto loaded = LoadedClient{library, std::move(api)};
+    loaded.addManifest = loadOptionalSymbol<LoadedClient::AddManifest>(
+        library, "MQT_CORE_QDMI_driver_add_manifest_v1");
+    loaded.sessionAllocForDevice =
+        loadOptionalSymbol<LoadedClient::SessionAllocForDevice>(
+            library, "MQT_CORE_QDMI_driver_session_alloc_for_device_v1");
+    return loaded;
```
burgholzer, line 321: Can we somehow centralize this so that these are not just free floating strings here prone to drifting?

src/qdmi/Client.cpp
```diff
@@ -294,6 +330,10 @@ struct ClientSelection {
   LibraryHandle library{};
   std::shared_ptr<const detail::ClientApi> api;
   std::filesystem::path path;
+  LoadedClient::AddManifest addManifest{};
+  LoadedClient::SessionAllocForDevice sessionAllocForDevice{};
```
burgholzer, line 335: Does it make sense to list these as separate members here, or would it make more sense to define some kind of Extension POD to aggregate the extension functionality. Might make it easier if we want to extend it in the future.

src/qdmi/Client.cpp
burgholzer, line 1: The added code here does not only feel overly complex, it also feels like it does not really concern this PR but rather #2229 and the implementation there.
Since this adds quite a bit of complexity, it may make sense to already fold the relevant parts into #2229 and look at the big picture as to whether the added complexity really makes sense or whether there is a more minimal but equally powerful solution.

test/qdmi/driver/session_device.cpp
```diff
@@ -46,6 +48,21 @@ struct QDMI_Device_Job_impl_d {
...
+  return value == nullptr ? std::string_view{} : std::string_view{value};
+}
+
+[[nodiscard]] auto successfulStatus(const std::string_view operation) -> int {
+  const auto mode = warningMode();
+  return mode == "all" || mode == operation ||
+                 (mode == "children-null" && operation == "children")
+             ? QDMI_WARN_GENERAL
+             : QDMI_SUCCESS;
+}
```
burgholzer, line 57: This whole warning handling feels like it is adding something that we have not really used before. No one in the QDMI community that I know of has been using warnings so far, so I am not sure the ecosystem is equipped to deal with the consequences of getting `QDMI_WARN_GENERAL` instead of `QDMI_SUCCESS`.
Obviously, we could set a good precedent here, but this feels like it does not necessarily belong in this PR stack, but rather a follow-up.
This would probably also mandate something like "warnings as errors", which only further complicates the design.
For now, I think we should stick with logging warnings, but not making use of `QDMI_WARN_GENERAL`

test/qdmi/CMakeLists.txt
burgholzer, line 1: We should absolutely make sure that all of these tests added here are genuinely useful or just pin down implementation details instead of asserting useful/expected behavior. This holds for the entire PR stack!

test/qdmi/test_client_targeted_selection.cpp
burgholzer, line 1: Let's try to minimize test files that only contain a single test. If we continue like this, we'll have more test files than actual library files, which is certainly not sustainable. Mitigate that in the next round of iteration on the entire stack.
Do not add tests simply to meet the 90% coverage target. If it cannot reasonably be met without going out of ones way, it should be purposely decided that it is not necessary to cover all possible failure conditions.

burgholzer (commented):
This one also triggered quite a couple of general comments that also apply down the stack and should be considered in the next iteration.

PR #2231, opened 2026-08-24
🔧 Harden installed QDMI runtime staging

burgholzer:
🤖 *AI text below* 🤖

## Description

Make the existing QDMI runtime-staging helper work for installed CMake consumers as well as in-tree applications, including device catalogues, provider assets and Windows dependency DLLs.

Driver workstream: **v4.1 → #2229 → #2230 → #2231**. This deployment layer depends only on #2230, not metadata removal, batching or payload capabilities. It targets **Core 4.1 / QDMI 1.4**, not Core 4.0. Coordination: #2093.

Reuse the existing imported-device fixture as an installed `find_package` consumer. Preserve build-tree RPATH handling and avoid treating imported targets as build dependencies. No new Client API or provider SDK dependency is introduced.

### Local validation

- Release build and full CTest: 3,873 passed, one existing optional-device skip.
- Both installed-consumer fixture tests passed, including manifest/assets/runtime staging.
- Repository lint passed; the parent Client/extension stack also passed its C++ lint, stub and Python checks.
- Real Windows DLL loading and final-head hosted CI remain separate checks.

## Installed deployment acceptance

Verify the default and replacement driver paths in installed CMake and Python consumers, including catalogue paths and runtime dependency loading. Clients must not need a rebuild to change drivers. Record compatibility evidence with existing provider libraries; do not add a provider SDK build prerequisite.

## Checklist

### DIFF
--- a/cmake/AddMQTQDMIDevice.cmake
+++ b/cmake/AddMQTQDMIDevice.cmake
@@ -147,7 +147,33 @@ function(mqt_get_qdmi_device_targets result)
       PARENT_SCOPE)
 endfunction()
 
-# Copy in-tree QDMI runtime libraries and manifests beside a runtime consumer.
+function(_mqt_qdmi_runtime_files result target)
+  set(runtime_files "$<TARGET_FILE:${target}>")
+  if(WIN32)
+    get_target_property(concrete_target ${target} ALIASED_TARGET)
+    if(NOT concrete_target)
+      set(concrete_target ${target})
+    endif()
+    get_target_property(imported ${concrete_target} IMPORTED)
+    if(imported)
+      string(MAKE_C_IDENTIFIER "${concrete_target}-runtime-closure" runtime_closure_target)
+      if(NOT TARGET ${runtime_closure_target})
+        add_library(${runtime_closure_target} MODULE EXCLUDE_FROM_ALL
+                    "${CMAKE_CURRENT_FUNCTION_LIST_FILE}")
+        set_property(TARGET ${runtime_closure_target} PROPERTY LINKER_LANGUAGE CXX)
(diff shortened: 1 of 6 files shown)


cmake/AddMQTQDMIDevice.cmake
burgholzer, line 1: This feels overly complicated and slightly confusing, especially because of the terminology being used. "closure", "staging", "concrete", "runtime".
I think this is important work to ensure a robust installation and usage; but it needs to be prepared in a better fashion.

Our top distribution channel is the `mqt-core` Python package. It is either directly used, which needs to work with the shipped libraries. Or it is loaded in other Python-only libraries (also not critical), _or_ it is used to feed a `find_package(mqt-core)` call in another project (such as MQT QCEC, MQT DDSIM, or MQT QMAP). The latter case is probably the one that is hardest to get right across all systems when it comes to using the QDMI-related functionality.
However, I still have the feeling that most of that was already working quite alright.

The second distribution channel is likely installations in HPC centers. These would typically build MQT Core from source via tools like `spack`. I do not really foresee major problems there with the solution that we previously had in place (maybe I am missing something though). The key aspect here is that this is almost exclusively for Linux (x86 and arm64), while the Python distribution channel is much broader and needs to cover macOS and Windows.

Maybe one can simplify around that information and already include these thoughts in the iteration on the lower layers.

test/qdmi/driver/imported_device/CMakeLists.txt
burgholzer, line 1: This definitely goes out of its way to prove that something works.
I'd personally be fine if this was demonstrated locally but not directly included in the test suite.

burgholzer (commented):
Also quite a bit of meta comments here for the general direction of this series. This should complete the run of feedback through the PRs.

Issue #2250, opened 2026-08-26
Institutionalize idiomatic MLIR development in MQT Core


burgholzer:
🤖 *AI text below* 🤖

## Goal

Make idiomatic, performant, and deterministic MLIR development the default for humans and coding agents, with less review intervention. MQT Core policy is authoritative; adopted upstream guidance and repository tooling support it, while historical precedent remains evidence only.

## Authority order

1. MQT Core documented policy.
2. Scoped agent guidance and executable configuration.
3. MLIR and LLVM guidance explicitly adopted by MQT Core.
4. Selected Google practices where MQT Core deliberately differs.
5. Existing repository precedent as evidence only.

## v4 release policy

Valid-IR crashes, invalid pass output, unsafe rewrites, miscompilation, and observable nondeterminism are blockers. Performance changes require measurement or a concrete allocation or complexity argument. Public C++ or Python renames require a separate migration-impact decision. Use focused issue- and PR-sized changes; do not create a repository-wide style patch or new `lit` and FileCheck infrastructure.

## Fresh-agent smoke result

Passed on 2026-08-26. Given only a normal small MLIR coding request, a fresh agent followed the scoped guidance: it used `ModuleOp` by value without misleading `const`, added a direct GoogleTest, passed the source gate and focused validation, and proposed the conforming subject `✨ Add nested MLIR operation counter`. The disposable patch was removed.

PR #2256, opened 2026-08-26
📝 Institutionalize idiomatic MLIR development

### LINKED ISSUE #2252: Establish MQT Core MLIR policy and enforcement
🤖 *AI text below* 🤖

Parent: #2250

## Deliverable

Land the campaign foundation as one coherent change:

- MQT-owned general and MLIR development policies plus a seeded glossary.
- Scoped agent guidance for `mlir/` and `bindings/mlir/`.
- An explicit MLIR clang-tidy allowlist shared by the bindings.
- An NFC cleanup of inappropriate `const` on mutable MLIR handles.
- A dependency-free pre-commit gate with positive and negative validation.

## Acceptance

The documentation, source gate, release build, MLIR CTest label, changed-file clang-tidy, complete docs build, link check, and binding build pass. No generated documentation, new `lit` infrastructure, custom clang-tidy plugin, or subtree-wide formatting is introduced.

burgholzer:
🤖 *AI text below* 🤖

## Description

Define concise MQT-owned general and MLIR development policies, seed a repository glossary, and give coding agents a scoped high-impact rule set. Replace inherited broad MLIR style checks with an explicit allowlist, establish a semantic baseline without misleading `const` on mutable MLIR handles and views, and enforce the mechanically identifiable subset through the existing dependency-free pre-commit stack.

This establishes the foundation for the audit campaign in #2250. The remaining contract, testing, performance, determinism, and public-terminology audits stay in focused child issues. The LLVM-style documentation-comment migration is tracked separately in #2267 for completion before v4.

Local validation passed: all pre-commit and Nox lint hooks; release configure and build; the CI-equivalent `uvx nox -s cpp-lint` over every changed C++ file; the new `--all` mode over every eligible project C++ file; all 2,895 `mqt-mlir-unittests`; 279 focused Qiskit/MLIR binding tests; stub validation; MLIR and complete documentation builds; link checking; and positive and negative source-gate checks covering `OpResult`, a range view, and `TypedValue`. Both cpp-linter modes report zero findings. `bindings/mlir/.clang-tidy` is now a regular, newline-terminated, byte-identical mirror of `mlir/.clang-tidy`, and the local tooling dependencies live in a dedicated `cpp-lint` group. No new runtime dependency, `lit` infrastructure, custom clang-tidy plugin, or broad reformatting is included.

Fixes #2252

## Checklist


.pre-commit-config.yaml
```diff
@@ -81,6 +81,16 @@ repos:
...
+      - id: disallow-const-mlir-handles
+        name: Disallow const on mutable MLIR handles
+        language: pygrep
+        entry: >-
+          (?x) \b(?: const\s+(?: (?:(?:::)?mlir::)?(?:Value|Operation|Block|Region|ModuleOp) |
+          (?:::)?(?:[A-Za-z_]\w*::)*(?!(?:PlanOp|ForbiddenModifierBodyOp)\b)[A-Z]\w*Op )\b |
+          (?: (?:(?:::)?mlir::)?(?:Value|Operation|Block|Region|ModuleOp) |
+          (?:::)?(?:[A-Za-z_]\w*::)*(?!(?:PlanOp|ForbiddenModifierBodyOp)\b)[A-Z]\w*Op )\b\s+const\b )
+        files: ^(mlir|bindings/mlir)/.*\.(c|cc|cpp|cxx|h|hh|hpp|hxx)$
+        priority: 0
```
burgholzer, line 92: @denialhaag I hope this marks the last time that you had to write a review comment with a `const Value` that should drop its `const` ;o)

mlir/AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Check minimal semantic IR properties and relevant diagnostics. Cover invalid
+  input when it forms part of the contract.
+- Debug by reducing the IR, isolating the first failing pass, printing generic
+  IR and IR around the failure, disabling threading if needed, then adding the
+  reduced case as a regression test.
+
+## Terminology and maintenance
+
+- Use the preferred terms in [`docs/glossary.md`](../docs/glossary.md). Update
+  the glossary in the same change when public or ambiguous terminology changes.
```
burgholzer: I think this should be promoted to the main AGENTS.md because we will very soon also add terminology for QDMI to the glossary, so everything related to it applies globally.

docs/glossary.md
burgholzer, line 1: The glossary in https://mystmd.org/guide/glossary also has an (alphabetical) index.
It seems to me like this should be following https://mystmd.org/guide/glossaries-and-terms if it does not do that yet.

I think it also makes a lot of sense to add a list of abbreviations because our field uses quite a lot of them. Quick brainstorming: MQV, CDA, TUM, MQSC, MQT, MQSS, QDMI, DD(s), QIR, MLIR, LLVM, OpenQASM, QASM, `jeff`, `IR`

burgholzer (commented):
I already like this quite a bit. I have three inline comments here and a few more things I'd like to kick off here when addressing them. After that, it should be ready for review.

mlir/AGENTS.md
burgholzer: 🤖 *AI text below* 🤖

Promoted the same-PR glossary update rule to the root `AGENTS.md`, so it now applies repository-wide, and removed the duplicate scoped rule from `mlir/AGENTS.md`.

docs/glossary.md
burgholzer: 🤖 *AI text below* 🤖

Converted the glossary to sorted MyST glossary directives, added the requested abbreviations and names (including the distinction between OpenQASM and generic QASM), and linked the entries to Sphinx’s generated alphabetical index. The complete docs build and link check pass.

burgholzer:
@denialhaag @simon1hofmann Would you mind taking a quick look here? I would hope that this really helps to get higher quality solutions out of agentic workflows and helps us to maintain alignment with several of the upstream guides.

bindings/mlir/qiskit/QiskitImport.cpp
```diff
@@ -439,12 +441,12 @@ void emitModifiedOperation(
...
 
-void emitModifiedGate(mlir::qc::QCProgramBuilder& builder,
-                      const Instruction& instruction,
-                      const mlir::ValueRange qubits,
-                      const llvm::ArrayRef<ParameterValue> parameters,
-                      const LocalParameters& localParameters,
-                      const GlobalParameters& globalParameters) {
+static void emitModifiedGate(mlir::qc::QCProgramBuilder& builder,
+                             const Instruction& instruction,
+                             const mlir::ValueRange qubits,
```
denialhaag: I think, right?

```suggestion
                             mlir::ValueRange qubits,
```

mlir/AGENTS.md
denialhaag, line 1: Hmm, this duplicates quite a bit of information from `docs/mlir/development.md` while linking to it. I know that `docs/development.md` documents a clear priority order, but I'm still wondering if we can remove the amount of information in this file. 🤔

.pre-commit-config.yaml
denialhaag: That will reduce my review comments by half. 🥲

denialhaag (commented):
I really like this! Thanks a lot, @burgholzer! 🙂 

I just have two comments:

.pre-commit-config.yaml
burgholzer: Sorry for that. Still a good trend I believe 😜

bindings/mlir/qiskit/QiskitImport.cpp
burgholzer: Good call! Removed another bunch of offenders 🤠 

🤖 *AI text below* 🤖

Yes. `ValueRange`, `OperandRange`, and `ResultRange` are cheap non-owning views passed by value; `const` neither protects their elements nor the referenced IR. The canonical policy, scoped agent rule, and pre-commit gate now cover all three, and the existing 28 `const ValueRange` declarations were cleaned up as NFC.

mlir/AGENTS.md
burgholzer: Also good observation. Had Codex dig a bit around for best practices. The below is what it came up with.

🤖 *AI text below* 🤖

Agreed. I reduced `mlir/AGENTS.md` from 64 to 31 lines and made the layering explicit: canonical rationale and procedures live in `docs/mlir/development.md`; the scoped agent file keeps only recurring high-impact guardrails and routes agents to the relevant policy sections; mechanical rules live in tooling. This follows the common concise always-loaded instructions plus on-demand detail pattern without adding agent-specific synchronization machinery.

burgholzer:
@denialhaag should be better now! Let me know what you think.
I'll let this open for Simon's review later and pick this up first thing tomorrow morning.
This should be a fairly solid basis for getting the long chain of PRs that have piled up across the finish line.

And maybe the both of you could split the follow-up audits that have already been planned in the overall issue between you afterwards 🙌🏼

.pre-commit-config.yaml
```diff
@@ -81,6 +81,16 @@ repos:
         entry: \b(Nanobind|Numpy|Cmake|CCache|Github|PyTest|Mqt|Tum)\b
         exclude: ^(\.pre-commit-config\.yaml|vendor/)
         priority: 0
+      - id: disallow-const-mlir-handles
+        name: Disallow const on mutable MLIR handles
+        language: pygrep
+        entry: >-
+          (?x) \b(?: const\s+(?: (?:(?:::)?mlir::)?(?:Value|ValueRange|OperandRange|ResultRange|Operation|Block|Region|ModuleOp) |
```
simon1hofmann: This does not cover `const auto` on MLIR handles, but probably too difficult to check using pre-commit.

docs/mlir/development.md
```diff
@@ -0,0 +1,132 @@
...
+6. Enable dialect-conversion tracing for a conversion failure.
+7. Save a pass-pipeline crash reproducer for crashes that are not immediately
+   local.
+8. Turn the reduced case into the smallest direct regression test.
+
+`mqt-cc` registers MLIR's standard pass-manager options. Useful options include
+`--mlir-print-op-generic`, `--mlir-print-ir-before-all`,
+`--mlir-print-ir-after-failure`, `--mlir-disable-threading`,
+`--mlir-print-stacktrace-on-diagnostic`, and
+`--mlir-pass-pipeline-crash-reproducer=<path>`. `--debug-only` traces require a
```
simon1hofmann, line 122: Are `--mlir-print-op-generic`, `--mlir-disable-threading`, `--mlir-print-stacktrace-on-diagnostic` available in `mqt-cc`?

noxfile.py
```diff
@@ -58,6 +58,63 @@ def lint(session: nox.Session) -> None:
...
+    diff_base = session.posargs[0] if session.posargs else "origin/main"
+
+    if shutil.which("cmake") is None and shutil.which("cmake3") is None:
+        session.install("cmake")
+    if shutil.which("ninja") is None:
+        session.install("ninja")
+    # Match cpp-linter-action v2.21.0 and the reusable workflow inputs.
+    session.install("cpp-linter==1.13.0", "nanobind==3.0.0")
+
+    compiler_env = {"CC": "clang-22", "CXX": "clang++-22"}
```
simon1hofmann: On macOS LLVM installations the tools are normally named `clang`, `clang++` and `clang-tidy` right?

noxfile.py
```diff
@@ -58,6 +58,63 @@ def lint(session: nox.Session) -> None:
...
 
 
+@nox.session(name="cpp-lint", reuse_venv=True, venv_backend="uv")
+def cpp_lint(session: nox.Session) -> None:
+    """Reproduce the CI cpp-linter check for changed C++ files."""
+    if len(session.posargs) > 1:
+        session.error("pass at most one diff base")
+    diff_base = session.posargs[0] if session.posargs else "origin/main"
+
+    if shutil.which("cmake") is None and shutil.which("cmake3") is None:
```
simon1hofmann: `cmake3` is accepted here but the session uses `cmake`.

.pre-commit-config.yaml
```diff
@@ -81,6 +81,16 @@ repos:
...
+          (?:::)?(?:[A-Za-z_]\w*::)*(?!(?:PlanOp|ForbiddenModifierBodyOp)\b)[A-Z]\w*Op )\b |
```
simon1hofmann: `*Op` also matches ordinary C++ types such as `const CBitModifierBodyOp&`.

noxfile.py
```diff
@@ -58,6 +58,63 @@ def lint(session: nox.Session) -> None:
...
+        "-B",
+        "build/cpp-lint",
+        "--preset",
+        "lint",
+        "-DCMAKE_C_COMPILER=clang-22",
+        "-DCMAKE_CXX_COMPILER=clang++-22",
+        "-DCMAKE_BUILD_TYPE=Debug",
+        "-DBUILD_MQT_CORE_BINDINGS=ON",
+        "-DCMAKE_UNITY_BUILD=OFF",
+        "-DCMAKE_VERIFY_INTERFACE_HEADER_SETS=ON",
```
simon1hofmann: already provided by `--preset lint`.

noxfile.py
```diff
@@ -58,6 +58,63 @@ def lint(session: nox.Session) -> None:
...
+
+    compiler_env = {"CC": "clang-22", "CXX": "clang++-22"}
+    session.run(
+        "cmake",
+        "-B",
+        "build/cpp-lint",
+        "--preset",
+        "lint",
+        "-DCMAKE_C_COMPILER=clang-22",
+        "-DCMAKE_CXX_COMPILER=clang++-22",
```
simon1hofmann: already provided by `compiler_env`.

simon1hofmann (changes requested):
This looks really good already and will really help the other open PRs! I have a few comments worth checking.

noxfile.py
burgholzer: No, I believe if you pull from brew, you'll also get the postfixed versions (plus a symlink). You'll definitely get them when you use the official LLVM installer script we use in CI.
Maybe we should make this a little more robust so that it actually searches for that compiler. However, I also do not want to overcomplicate the setup here.
I believe it is fine as long as _some_ decently modern `clang` compiler is used.
So we may as well just put `clang` and `clang++` here supposedly.
We just need to avoid that Linux picks up `gcc`.

docs/mlir/development.md
burgholzer: I would think so because these are installed per default in MLIR CLI tools.
Worth double checking.

.pre-commit-config.yaml
burgholzer: Yeah. `const auto` is hard to catch. One would have to do some kind of type checking to infer the right type for auto to deduce whether `const` belongs there or not. Maybe Codex will have a better idea, but I somewhat doubt it.

noxfile.py
burgholzer: I think we can just drop the `cmake3` workaround. IIRC that has been dropped by most projects by now and we are at CMake v4 anyway already. Just looking for `cmake` should be fine.

noxfile.py
burgholzer: Good catch 👍🏼  need to deduplicte

.pre-commit-config.yaml
burgholzer: Fair enough. Maybe there's a good way to make this a little tighter.

noxfile.py
burgholzer: Good catch! Also relates to your comment from above. Should definitely be deduplicated.

.pre-commit-config.yaml
burgholzer: As expected, it is tough for the static analysis, but it works as agent guidance supposedly 😌 

🤖 *AI text below* 🤖

Agreed. The canonical policy and scoped MLIR agent guide now state that the rule includes `const auto` when it deduces an MLIR handle or view. The dependency-free text gate deliberately does not claim to infer deduced types; the ExecPlan records that semantic enforcement should wait until repeated violations justify a custom checker.

docs/mlir/development.md
burgholzer: 🤖 *AI text below* 🤖

They were not all available. `mqt-cc` registered only the pass-manager options. It now also calls MLIR's standard `registerAsmPrinterCLOptions()` and `registerMLIRContextCLOptions()`. I rebuilt the release target and confirmed that all three queried flags appear in `mqt-cc --help-hidden`.

noxfile.py
burgholzer: 🤖 *AI text below* 🤖

Removed. The dedicated lint build now receives the resolved LLVM 22 `clang` and `clang++` paths only through `CC` and `CXX`.

noxfile.py
burgholzer: I am not 100% sure whether this is flexible enough yet, but I think it is a step in the right direction.

🤖 *AI text below* 🤖

Yes. The session now accepts either `clang-tidy-22` or an unversioned `clang-tidy`, resolves that tool's LLVM binary directory, verifies that it is LLVM 22, uses the sibling `clang` and `clang++`, and passes the same directory to cpp-linter. The local reproduction selected `/usr/lib/llvm-22/bin` and reported LLVM 22.1.8.

noxfile.py
burgholzer: 🤖 *AI text below* 🤖

Removed. The `lint` preset now remains the single source for the build type, bindings, unity-build setting, and interface-header verification.

.pre-commit-config.yaml
burgholzer: 🤖 *AI text below* 🤖

Agreed. I removed the unsound suffix-wide `*Op` match and its growing exception list. The gate now checks only explicitly named core MLIR handles and range views. The policy still covers typed operation wrappers because a text check cannot distinguish them from ordinary project types reliably. A temporary test confirmed that `const ValueRange` fails while `const CBitModifierBodyOp&` passes.

noxfile.py
burgholzer: 🤖 *AI text below* 🤖

Good catch. I removed the `cmake3` fallback from the new session and the two older occurrences in `noxfile.py`. All Nox sessions now use the repository's supported `cmake` command.

burgholzer:
Alright. I believe I addressed everything.
I am done for the day. I'll pick this back up tomorrow.
If you still got feedback then, I'll incorporate it. If not, I'll proceed and merge.
For the future: do not hesitate to make tweaks to this setup if you feel like something is not quite ideal.

noxfile.py
```diff
@@ -58,6 +58,69 @@ def lint(session: nox.Session) -> None:
...
+    if len(session.posargs) > 1:
+        session.error("pass at most one diff base")
+    diff_base = session.posargs[0] if session.posargs else "origin/main"
+
+    if shutil.which("cmake") is None:
+        session.install("cmake")
+    if shutil.which("ninja") is None:
+        session.install("ninja")
+    # Match cpp-linter-action v2.21.0 and the reusable workflow inputs.
+    session.install("cpp-linter==1.13.0", "nanobind==3.0.0")
```
denialhaag: Could this be a dependency group in `pyproject.toml` so that Renovate picks it up? 🤔

denialhaag (approved):
This LGTM now!

I just stumbled across one more thing that might be worth investigating, but feel free to just resolve and merge. 😌

.pre-commit-config.yaml
simon1hofmann: 🤖 *AI text below* 🤖

Thanks for documenting the limitation. I think this is only partially addressed: the normative policy still prohibits `const auto` MLIR handles, and examples remain at `bindings/mlir/qiskit/QiskitImport.cpp:1034-1042`, `mlir/lib/Compiler/Programs.cpp:211`, and `bindings/mlir/qiskit/QiskitExport.cpp:1843-1844`, while the ExecPlan still claims a clean baseline. Could we either clean these inferred and wrapper cases or qualify the remaining clean-baseline claims?

bindings/mlir/.clang-tidy
simon1hofmann, line 1: 🤖 *AI text below* 🤖

In clones with `core.symlinks=false`, this is checked out as a regular file containing `../../mlir/.clang-tidy`. Local clang-tidy then parses that text as YAML and fails with `error: not a mapping`. This does not affect the current build jobs, but it breaks local linting in such checkouts, which are common on Windows. Could we use a real or generated configuration instead?

docs/development.md
denialhaag, line 1: I'm not sure if this is the right place for this, but we could also document a "style guide" for docstrings.

I don't know why, but I really dislike this kind of docstring:

```c++
/** A relatively short docstring. */
```

Instead, I generally prefer this:

```c++
/// A relatively short docstring.
```

At the same time, I don't like `///`-style docstrings for multi-line docstrings for some reason.

In addition, we should decide whether we want to use `@detail`, `@details`, or omit it altogether (which is fine if there's an empty line before the extended description). Also, should the extended description be right after `@detail` or `@details`, or in a new line below?

Of course, all of this does not affect functionality in any shape or form, but I would like to have something to follow (and I will even comply with `/** A relatively short docstring. */`). Maybe two or three examples in this document could already settle the discussion and ensure that agents (and we) generally adhere to a consistent style.

burgholzer: Style guide makes sense! and I believe this is the right place to do this.
It may not be your preferred way of doing it, but LLVM has a rather comprehensive guide: https://llvm.org/docs/CodingStandards.html#commenting

The example from there:
```c++
/// Does foo and bar.
///
/// Does not do foo the usual way if \p Baz is true.
///
/// Typical usage:
/// \code
///   fooBar(false, "quux", Res);
/// \endcode
///
/// \param Quux kind of foo to do.
/// \param [out] Result filled with bar sequence on foo success.
///
/// \returns true on success.
bool fooBar(bool Baz, StringRef Quux, std::vector<int> &Result);
```

Some interesting bits and pieces:
- no explicit `/brief` or `/details`
- `///` as the categoric style for doxygen comments, `/* ...*/` only for special cases.
- parameters shall only be documented if their documentation actually helps and adds value

What do you think about that? I think the main deviation from the current style is `/** ..*/ -> `/// ...`
Could you live with that? If yes, I'd try to get some of the guidelines into the development guide.

bindings/mlir/.clang-tidy
burgholzer: Hm. Windows being the painful one again 🙃 Let's change this. Even though it causes a bit of duplication.

noxfile.py
burgholzer: Yeah, good idea.
For `cpp-linter` we likely want to maintain the `==` dependency.
For `nanobind` we should follow the project `~=` style.
The comment here needs to be adjusted then.

docs/development.md
denialhaag: Yes, let's go with that! As long as it looks consistent and clean, I'm happy! 😎

.pre-commit-config.yaml
burgholzer: Thanks for pushing on this. Let's try to clean as many of them as possible. 
I'd generally be a bit careful about how much emphasis to put on the ExecPlans.
I noticed that these tend to get very long; especially if a PR is iterated quite a bit. I plan to do a cleanup of past ExecPlans in a separate PR.

docs/development.md
burgholzer: Then the big question:
Do we want to clean up the existing comments in the code base after this PR lands? then I'd create a separate follow-up tracking issue for that.

denialhaag: I think that would be nice! Maybe we can tackle that just before the v4 release. 🤔

.pre-commit-config.yaml
burgholzer: 🤖 *AI text below* 🤖

Completed the semantic baseline rather than qualifying it. A one-off `clang-query` inventory found 666 candidates across 80 files; source review removed all 642 avoidable `const` uses, including inferred handles, typed wrappers, ranges, structured bindings, and references. The 24 remaining AST matches are const-container pointers, genuinely generic code, or compiler-generated proxy declarations with no `const` in source. The expanded gate covers `TypedValue`, `BlockArgument`, `OpResult`, and the range views. The release build, all 2,895 MLIR tests, changed-file cpp-lint, and the new all-files cpp-lint mode pass.

noxfile.py
burgholzer: I think this works nicely

🤖 *AI text below* 🤖

Added a dedicated `cpp-lint` dependency group with `cpp-linter==1.13.0` and `nanobind~=3.0.0`. The Nox session installs that group, the lockfile is regenerated, and the action-alignment comment now points to the group. Both changed-file and all-files session modes pass.

bindings/mlir/.clang-tidy
burgholzer: Should be addressed now.

🤖 *AI text below* 🤖

Replaced the symlink with a regular, newline-terminated copy of `mlir/.clang-tidy`. The files are byte-identical and carry the same short mirror comment. This works with `core.symlinks=false` without adding a generator or synchronization hook.

docs/development.md
burgholzer: 🤖 *AI text below* 🤖

Added the agreed LLVM-style documentation-comment policy to `docs/development.md`: use `///`, let the first sentence summarize, separate details with a blank `///`, omit `brief` and `details`, and document parameters or returns only when useful. It includes short and multiline examples; root agent guidance only routes to the canonical policy. The 1,615 existing block comments are tracked for four mechanical pre-v4 migrations and a final source gate in #2267.

simon1hofmann (approved):
Looks good to me now, excited to see this in action in the other PRs!

AGENTS.md
```diff
@@ -26,8 +26,10 @@ MQT Core. The project-wide policy for AI-assisted contributions is
   authorization.
 - Preserve user changes and inspect the working tree before editing. Never
   discard or overwrite changes that are outside the task.
-- Follow the patterns in neighboring files and prefer the smallest change that
-  fully solves the problem.
+- Follow the repository's documented development policies and the nearest scoped
+  `AGENTS.md`. Use neighboring code as evidence of established practice, not as
+  authority when it conflicts with current policy. Prefer the smallest change
+  that fully solves the problem.
```
denialhaag, line 32: Nitpick, so feel free to ignore: Should this be two bullet points at this point?

AGENTS.md
```diff
@@ -108,16 +116,20 @@ MQT Core. The project-wide policy for AI-assisted contributions is
...
+`llvm::SmallVector` and `llvm::function_ref` where appropriate. Do not use
+C-style casts, including casts to `void`; use the appropriate C++ cast or adjust
+the code so that no cast is needed. Use C standard-library typedefs such as
+`size_t` and fixed-width integer types such as `uint64_t` without the `std::`
+namespace qualifier, while directly including the header that provides them. Do
+not use `module` as a C++ variable or parameter name because it conflicts with
+the C++20 keyword. Use `moduleOp` for `mlir::ModuleOp` values. The canonical
+general and MLIR-specific coding policies are documented in
+[`docs/development.md`](docs/development.md) and
+[`docs/mlir/development.md`](docs/mlir/development.md).
```
denialhaag, line 132: Also here: Would this be better as a list of bullet points?

docs/development.md
denialhaag: LGTM now! 😎

docs/mlir/development.md
denialhaag, line 1: Something else that just came to mind: Should we mention that we prefer including `mlir/Support/LLVM.h` and dropping `llvm::` qualifiers when possible? Or is that already covered somewhere? 🤔

denialhaag (approved):
This continues to LGTM! I found a few more things, but all of them are minor. Let's quickly merge this after this iteration before we keep it open forever. 🤠

burgholzer:
> This continues to LGTM! I found a few more things, but all of them are minor. Let's quickly merge this after this iteration before we keep it open forever. 🤠

Thanks @denialhaag 🙏🏼 
I'll actually merge this now. Let's defer these to a direct follow-up that can happen in parallel to further ongoing work.

PR #2240, opened 2026-08-25
:zap: Implement `reorderTopologically` function

MatthiasReumann:
## Description

This pull request implements the `reorderTopologically` function using the updated driver function to replace the `sortTopological` function of MLIR. 

## Benchmarks 

Grover using `scf.for` (2-118 qubits) on 10x12 grid architecture. 5 Repetitions; Average Milliseconds reported. 

| #qubits | main | branch | speedup |
|---------|------|--------|---------|
| 2       | 2    | 2      | 1       |
| 6       | 6    | 5      | 1.2     |
| 10      | 12   | 10     | 1.2     |
| 14      | 28   | 23     | 1.22    |
| 18      | 54   | 38     | 1.42    |
| 22      | 99   | 60     | 1.65    |
| 26      | 169  | 94     | 1.8     |
| 30      | 265  | 126    | 2.1     |
| 34      | 402  | 139    | 2.89    |
| 38      | 518  | 151    | 3.43    |
| 42      | 517  | 172    | 3.01    |
| 46      | 612  | 207    | 2.96    |
| 50      | 753  | 233    | 3.23    |
| 54      | 922  | 305    | 3.02    |
| 58      | 1055 | 288    | 3.66    |
| 62      | 1295 | 307    | 4.22    |
| 66      | 1528 | 332    | 4.6     |
| 70      | 1847 | 369    | 5.01    |
| 74      | 2137 | 404    | 5.29    |
| 78      | 1785 | 420    | 4.25    |
| 82      | 1972 | 457    | 4.32    |
| 86      | 2181 | 489    | 4.46    |
| 90      | 2526 | 514    | 4.91    |
| 94      | 2842 | 560    | 5.08    |
| 98      | 3177 | 601    | 5.29    |
| 102     | 3347 | 632    | 5.3     |
| 106     | 3900 | 669    | 5.83    |
| 110     | 4412 | 694    | 6.36    |
| 114     | 4850 | 707    | 6.86    |
| 118     | 5486 | 857    | 6.4     |

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [x] The changes are fully tested and pass the CI checks.
[...]

PR #2285, opened 2026-08-31
✨ Support directional compiler-target operation availability

### LINKED ISSUE #2132: ✨ Support directional compiler-target connectivity and operation availability
🤖 *AI text below* 🤖

## Problem

`CompilerTarget::Coupling` accepts a pair of site IDs, but target construction currently sorts each pair and builds adjacency in both directions. Target operations are documented as globally available; ordered `SiteTuple`s carry calibration data only. A target therefore cannot express that `cx` or `ecr` is native on `(u, v)` but not on `(v, u)`.

Current implementation:

- [`CompilerTarget` topology and operation model](https://github.com/munich-quantum-toolkit/core/blob/53325922bfebe14f17fbb8af7d07fd68bb9f9704/mlir/include/mlir/Compiler/Target.h#L30-L171)
- [Coupling canonicalization and symmetric adjacency](https://github.com/munich-quantum-toolkit/core/blob/53325922bfebe14f17fbb8af7d07fd68bb9f9704/mlir/lib/Compiler/Target.cpp#L415-L439)
- [Mapping's explicit symmetric-coupling assumption](https://github.com/munich-quantum-toolkit/core/blob/53325922bfebe14f17fbb8af7d07fd68bb9f9704/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp#L263-L279)

## Goal

Represent directional connectivity and ordered operation availability without losing the convenient symmetric and all-to-all target forms. Routing may use the underlying connected graph, but native synthesis and target conformance must respect the legal operand order of directional operations.

Adapters that receive ordered topology or operation data must preserve it rather than silently symmetrizing it.

## Minimal reproducer

Create a two-site target where `cx` is available only on `(0, 1)`. Compiling `cx` on `(1, 0)` must either synthesize an equivalent circuit using legal target operations or fail with a precise diagnostic. It must not accept the reversed gate as native.

## Acceptance criteria

- The target API distinguishes `(u, v)` from `(v, u)` for directional connectivity and operation availability.
- Mapping, native synthesis, and conformance consistently use that direction information.
- Unsupported directions are legally rewritten or rejected; no reversed native operation survives verification.
- Symmetric and all-to-all targets retain straightforward construction and existing behavior.
- C++ and Python tests cover asymmetric `cx`/`ecr`, symmetric topology, and an unreachable direction.

Related work: [#1709](https://github.com/munich-quantum-toolkit/core/pull/1709) introduced coupling-map injection. [#2119](https://github.com/munich-quantum-toolkit/core/pull/2119) fixes sparse target-site indexing and is intentionally separate from directionality.

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Adds exact ordered operation applicability to `CompilerTarget` and uses it to support targets with one-way CX, ECR, and RZX gates.

- Separates operation applicability from optional calibration data, including unrestricted, explicitly empty, and explicit ordered site tuples.
- Preserves ordered applicability through QDMI, typed MQT attributes, Python bindings, serialization, mapping, synthesis, and target conformance.
- Wraps the immutable compiler target in a cached `MappingTarget`: topology traversal stays bidirectional, while mapping assigns the lower cost to the physical direction supported by the selected entangler.
- Precomputes directional two-qubit costs in O(E), giving average O(1) lookups in the mapping hot path instead of repeated support scans.
- Keeps the mapping pass synthesis-free; native synthesis performs any required direction repair and checks the exact emitted direction.
- Handles symmetric operations by reordering operands only when the reversed physical tuple is the supported one.

Fixes #2132

## Verification

- Core release suites: Compiler 156/156, MQT IR 15/15, Mapping 96/96, and NativeSynthesis 37/37 passed.
- Python MLIR tests: 49/49 passed.
- Repository lint, C++ lint, strict documentation build, generated Python stubs, and `git diff --check` passed.
- Regression tests cover ambiguous one- and two-qubit sites and maximum valid target site IDs.
- Benchpress MQT integration: 53/53 passed, including focused one-way CX/ECR applicability and timeout round-trip tests.

## AI assistance

Codex implemented, reviewed, and tested the change under explicit user authorization. Human review and acceptance remain required before merge.

## Checklist


simon1hofmann:
@MatthiasReumann Would you mind taking at first look here to see if this goes in the right direction?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -191,6 +198,84 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+    std::array sites{target.siteForVertex(hw0), target.siteForVertex(hw1)};
+    if (supportsOnSites(entry, target, sites)) {
+      return 0.F;
+    }
+    std::ranges::reverse(sites);
+    if (supportsOnSites(entry, target, sites)) {
+      return 1.F;
+    }
+    return std::numeric_limits<float>::infinity();
+  }
```
MatthiasReumann: After a first glance, I think this function is the only relevant part of this change (at least for the mapping pass implementation). 

Maybe the information of how expensive a path in the coupling graph is could be integrated into the `CompilerTarget`: There could be something like a member variable (`phantomCouplings`?), which don't exist physically but can be achieved using synthesis logic. 

Compute and cache these edges once because recomputing these every time querying the path costs will be very, very, expensive in the mapping pass. 

To query the path cost would then be a lookup into the "real" coupling map, followed by a lookup into the "phantom edges" - analogously to this function here. If this is integrated into the target, the changes in the mapping pass should reduce to a handful of lines. 

---

I would also argue to keep the mapping pass as "synthesis-free" as possible. It shouldn't have to care about directionality of gates; or native gate sets. This is the task of the subsequent synthesis pass.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1009,49 +1098,27 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
-
-      if (block[0].operation() != block[1].operation()) {
-        return;
-      }
+  /// Return a two-qubit operation's program indices in semantic operand order.
+  /// Wire traversal records ready indices in wire order. The iterator's qubit
+  /// is the operation result on that wire in both traversal directions, so use
+  /// the ordered results to recover the operation's operand order.
+  [[nodiscard]] static IndexPairType
+  orderedPrograms(Operation* operation, const ArrayRef<size_t> indices,
```
MatthiasReumann: Not sure what happened with this function here, but keep the `Window` as is and this logic doesn't need to change.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: > It shouldn't have to care about directionality of gates; or native gate sets. This is the task of the subsequent synthesis pass.

I mostly agree, but to some degree the mapping pass has to care about the native ISA of the device in order to make good decisions for mapping/routing. So it is not perfectly easy to separate.

MatthiasReumann: I was hoping that we could "abstract that away" by simply updating path costs / path classification in the `CompilerTarget` (or some mapping-specific wrapper).

burgholzer: In my mind, the `CompilerTarget` is an immutable snapshot of a device used for compilation. I am not 100% sure that it is the "right" place to have this kind of abstraction.
However, I also think that this should not necessarily be embedded in some mapping logic. The notion of a `cost` for a certain gate is very likely useful beyond mapping/routing and should generally be accessible. Maybe this is some kind of cached data structure built on top of the CompilerTarget.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -213,19 +215,24 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
     /// Construct a non-root node from its parent node. Apply the given swap to
     /// the layout of the parent node.
     Node(Node* parent, const IndexPairType& swap, const Window& window,
-         const CompilerTarget& target, const Parameters& params)
+         const CompilerTarget& target, const TargetGateCosts* gateCosts,
```
MatthiasReumann: Could you try to merge the target and the gateCosts into one `MappingTarget` class (maybe wait on #2316) which wraps all the path-cost logic? I still think there are way too many changes in the Mapping pass  😅

simon1hofmann: #2316 was just merged, so will rebase and try that out now 👍

burgholzer: Maybe it makes sense to wait on the other Mapping pass changes from your PR stack as well before revisiting this here.

simon1hofmann: Makes sense, will wait for the stack then.

burgholzer: https://github.com/munich-quantum-toolkit/core/pull/2240 just merged. So any pre-requisites for this should be in.

burgholzer:
I started an iteration on this already now. Will look at it in more detail tomorrow.

burgholzer:
Still working through this one. Feels too large still, even after the simplifications.

burgholzer:
@simon1hofmann Alright. I think I simplified this as far as I could. Would you mind checking the current implementation and whether it fulfills all goals you had in mind?

simon1hofmann:
> @simon1hofmann Alright. I think I simplified this as far as I could. Would you mind checking the current implementation and whether it fulfills all goals you had in mind?

Thanks a lot for the changes, the skipped benchpress test were successful based on this PR 🚀

burgholzer:
I'll merge this now, given how it seems to resolve the underlying issues 🙌🏼

PR #2297, opened 2026-08-31
✨ Support fixed-width classical registers across formats

### LINKED ISSUE #2289: ✨ Support bit-register casts in the OpenQASM frontend
🤖 *AI text below* 🤖

### Problem Statement

The OpenQASM frontend cannot import a program that casts a multi-bit measurement register to an integer for use in a condition.

A minimal reproducer is:

```python
from mqt.core.mlir import QCProgram

source = """OPENQASM 3.0;
include "stdgates.inc";

bit[2] syndrome;
qubit[2] q;

syndrome[0] = measure q[0];
syndrome[1] = measure q[1];

if (uint[2](syndrome) == 3) {
  x q[0];
}
"""

QCProgram.from_qasm_str(source)
```

This fails with:

```text
OpenQASM frontend error: expected expression
```

Using a direct comparison instead,

```openqasm
if (syndrome == 3) {
```

also fails:

```text
OpenQASM frontend error: identifier 'syndrome' is not a scalar value
```

OpenQASM 3 permits casts from `bit[n]` to `uint[n]` or `int[n]` when the widths match. The conversion uses the register's little-endian representation. See the [OpenQASM type conversion rules](https://openqasm.com/versions/3.1/language/types.html#casting-specifics).

This prevents the MQT Compiler Collection from importing valid dynamic circuits that branch on a multi-bit measurement result. It is also the frontend failure exposed by the `mqt-cc` benchmark in `ucc-bench`.

Reproduced on `main` at `35d3dc2cb87dc9ed4904e9db7eb43257ad3d4527`.

### Proposed Solution

Add support for classical bit-register casts in the OpenQASM frontend.

At minimum, the frontend should:

- Parse sized casts such as `uint[2](syndrome)` and `int[2](syndrome)`.
- Convert a `bit[n]` register to an integer of the same width using the OpenQASM bit order.
- Permit the cast result in expressions, comparisons, and conditions.
- Report a clear diagnostic for unsupported or incompatible widths.
- Add frontend tests for casts of measured registers and their use in `if` conditions.

Supporting direct comparisons such as `syndrome == 3` would also be useful, but support for the explicit standards-compliant cast is sufficient to express the condition.

Related issues:

- #1733
- #1732
- #32

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Preserve fixed-width classical-register semantics across OpenQASM and Qiskit without reconstructing bit-load graphs in each exporter.

- Add `cbit.read` and `cbit.write` for whole-register snapshots and atomic updates. Keep `cbit.cmp` as the compact register-versus-constant operation for signed and unsigned predicates.
- Represent runtime bitwise not, and, or, xor, shifts, rotations, and popcount with MLIR's fixed-width integer operations.
- Support all six OpenQASM comparisons. Exact-width `int[N]` casts use signed two's-complement ordering; direct bit-register and `uint[N]` comparisons use unsigned ordering.
- Import and export Qiskit `Store` for Clbits, indexed ClassicalRegisters, and whole ClassicalRegisters, including nested control flow.
- Encode signed comparisons in Qiskit's unsigned expression model by XOR-biasing the sign bit. Recognize only that exact encoding on import instead of canonicalizing arbitrary Boolean graphs.
- Validate register snapshots and reject stale or cross-region uses that a target would otherwise re-evaluate after an intervening write.
- Share one CBit decomposition for MemRef and Adaptive QIR consumers.

Runtime shift distances must be unsigned and less than the register width. Constant overshifts fold to zero. Qiskit expressions remain limited to 64 bits. Standalone Qiskit variables remain outside the CBit register model.

jeff supports arbitrary-width `cbit.cmp` through per-bit array reads and Boolean logic. General `cbit.read` and `cbit.write` remain rejected because jeff has neither arbitrary-width integer casts nor logical right shift; promoting values would require incomplete logical-width emulation. QIR Base rejects the same general whole-register forms, while Adaptive QIR lowers internal values.

Fixes #2289

## Validation

- Fresh full C++ lint build after rebasing onto `main`.
- 1,646 tests passed across ten affected C++ unit-test binaries.
- `uv run --no-sync pytest -q test/python/test_mlir_qiskit_translation.py`: 253 tests passed.
- `uvx nox -s stubs`
- `uvx nox -s lint`
- `uvx nox -s cpp-lint`
- `git diff --check`
- Fresh hosted CI is pending for `7bc22a446`.

Codex assisted with implementation, simplification, independent review, validation, and this description under human direction.

## Checklist

- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals. (Not applicable for unreleased v4 functionality.)
- [x] I have added migration instructions to the upgrade guide (if needed). (Not applicable.)
- [x] The changes are fully tested and pass the CI checks. (Local validation passes; fresh hosted CI is pending.)


simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai full review

simon1hofmann:
@coderabbitai full review

burgholzer:
Also made this one a fair bit more general and hopefully usable.
I haven't checked all the details here yet, but it may still be interesting to look at already.

burgholzer (approved):
Alright, this has become a bit of a monster. But I still think this is quite useful and expands the capabilities of the OpenQASM and Qiskit support.
Maybe you could run it through a sanity check on your side (maybe that finds a few simplifications)?
This is good to go from my side.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
```diff
@@ -140,24 +141,35 @@ static LogicalResult prepareCBitRegisterReads(Operation* moduleOp,
...
     } else if (representation == RETURNED_CBIT_REGISTER) {
       state.returnedCBitReads.insert(operation);
     }
   };
   moduleOp->walk(
       [&](cbit::LoadOp loadOp) { prepareRead(loadOp, loadOp.getReg()); });
-  moduleOp->walk([&](cbit::CompareOp compareOp) {
-    prepareRead(compareOp, compareOp.getReg());
+  moduleOp->walk([&](cbit::StoreOp storeOp) {
+    const auto representation = representations.lookup(storeOp.getReg());
```
simon1hofmann, line 159: 🤖 *AI text below* 🤖

`prepareCBitRegisterAccesses` propagates representations through CFG arguments and `arith.select`, but not through `func.call` arguments or results. Passing a returned CBit register to a helper therefore leaves the callee argument unclassified (`0`), so loads interpret QIR `Result**` storage as local `i1*` storage and stores bypass this returned-register guard, potentially corrupting result pointers. Could we either propagate representation across call edges or explicitly reject CBit-typed calls? A helper-call regression test would cover both paths.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -1074,88 +1093,76 @@ class OpenQASMToQCEmitter {
...
+            distance.type == frontend::ScalarType::Uint
+                ? std::get<uint64_t>(distance.constant)
+                : static_cast<uint64_t>(std::get<int64_t>(distance.constant));
+        if (amount >= expression.width) {
+          return arith::ConstantIntOp::create(opBuilder, loc, type, 0);
+        }
+        shift = arith::ConstantIntOp::create(opBuilder, loc, type,
+                                             static_cast<int64_t>(amount));
+      } else {
+        shift = emitExpression(opBuilder, expression.distance, {});
```
simon1hofmann: 🤖 *AI text below* 🤖

A dynamic distance is truncated and passed directly to `arith.shli`/`arith.shrui`. For a `bit[3]` value, a runtime distance of 3 already produces poison; larger distances can also wrap during truncation and produce an unrelated result. This differs from the constant path above, which correctly returns zero for an overshift, and from OpenQASM's bit-shift semantics. Could we guard the dynamic shift and return zero when the distance is at least the register width?

mlir/lib/Target/OpenQASM/OpenQASMSemantics.cpp
```diff
@@ -2243,6 +2486,30 @@ class SemanticAnalyzer {
                             .type = ScalarType::Uint,
                             .bitVector = bitVector});
     }
+    if (expression.kind == Expr::Kind::IntCast ||
```
simon1hofmann, line 2682: 🤖 *AI text below* 🤖

This new cast path is not gated by `program.openQASM2`. As a result, an OpenQASM 2 file containing `if (uint[2](c) < 1)` is accepted and canonicalized into a relational register comparison, bypassing the format's equality-only condition restriction. Could we reject sized integer casts in OpenQASM 2 before canonicalization and add a regression test?

simon1hofmann:
> Alright, this has become a bit of a monster. But I still think this is quite useful and expands the capabilities of the OpenQASM and Qiskit support. Maybe you could run it through a sanity check on your side (maybe that finds a few simplifications)? This is good to go from my side.

I went through the changes and they look quite good to me. Also ran it through ponytail again, but that did not flag any over engineering or possible simplifications. Just three findings regarding correctness, see comments.

mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
burgholzer: Hm. I do not have a strong preference here. Slight tendency to simply reject CBit types in calls. CBit types can be read and written to, so this can always be done before a call.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
burgholzer: Adding such a guard seems reasonable.
For a majority of the circuits we are currently considering, the dynamic part is fairly hypothetical. So we should not overengineer the solution and make sure that the common path is efficient and fast.

mlir/lib/Target/OpenQASM/OpenQASMSemantics.cpp
burgholzer: We could reject that. At the same time, our IR can handle this.
I would actually lean towards being way more lenient in the OpenQASM 2 import path and allow almost arbitrary OpenQASM 3 concepts as long as that allows to simplify the code.
As long as it produces IR that we can work with, I do not think we need to be overly restrictive on the import side.

burgholzer:
> I went through the changes and they look quite good to me. Also ran it through ponytail again, but that did not flag any over engineering or possible simplifications. Just three findings regarding correctness, see comments.

Thanks! I added a few replies. I'll hand that back to Codex, iterate once, then likely merge this. It feels ready 😌

burgholzer (approved):
Alright, I took this for another spin after your comments. I repeatedly tried to cut the complexity from this PR. However, this is as far as I got. To be fair, I think this works quite will now.
If you are happy with it as well, then let's get it in.

simon1hofmann:
> Alright, I took this for another spin after your comments. I repeatedly tried to cut the complexity from this PR. However, this is as far as I got. To be fair, I think this works quite will now.
> If you are happy with it as well, then let's get it in.

LGTM, thanks a lot for your work in this!

PR #2316, opened 2026-09-01
♻️ Separate target placement from routing

burgholzer:
🤖 *AI text below* 🤖

## Description

Separate deterministic qubit placement from topology-aware routing.

Target compilation now uses a compact placement pass for all-to-all targets and reserves the existing mapping pass for explicit coupling graphs. Both paths share the validated allocation discovery, capacity check, and dynamic-to-static rewrite, so the split does not duplicate the fragile part of the transformation.

This is the focused prerequisite for #2218. After this PR merges, #2218 can add its three-state connectivity policy without making the router handle an unknown topology: explicit connectivity uses mapping, all-to-all uses placement, and unknown connectivity rejects residual multi-site operations before placement.

Validated locally with:

- 86 mapping tests
- 139 compiler tests
- the generated MLIR documentation build
- the full repository lint suite
- the canonical changed-file C++ lint suite with no clang-format or clang-tidy findings

No changelog or upgrade-guide entry is needed because this only restructures the unreleased MQT Core v4 target-compilation work.

## Checklist

### DIFF
--- a/mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
+++ b/mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
@@ -25,10 +25,7 @@ namespace qco {
 /// Create a deterministic placement pass for a compiler target.
 std::unique_ptr<Pass> createPlacementPass(const CompilerTarget& target);
 
-/**
- * @brief Create a mapping pass instance for a compiler target.
- * @returns a pass object.
- */
+/// Create a mapping pass for a compiler target with explicit topology.
 std::unique_ptr<Pass> createMappingPass(const CompilerTarget& target,
                                         MappingPassOptions options);
 
--- a/mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
+++ b/mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
@@ -22,6 +22,9 @@ class CompilerTarget;
 
 namespace qco {
 
+/// Create a deterministic placement pass for a compiler target.
+std::unique_ptr<Pass> createPlacementPass(const CompilerTarget& target);
+
 /**
  * @brief Create a mapping pass instance for a compiler target.
  * @returns a pass object.
--- a/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
@@ -66,7 +66,6 @@ namespace mlir::qco {
 using namespace mlir::qtensor;
 
 #define GEN_PASS_DEF_MAPPINGPASS
-#define GEN_PASS_DEF_PLACEMENTPASS
 #include "mlir/Dialect/QCO/Transforms/Passes.h.inc"
 
 namespace {
@@ -319,17 +318,19 @@ applyPlacement(Region& body, const CompilerTarget& target, const Layout& layout,
 
 namespace {
 
-struct PlacementPass : impl::PlacementPassBase<PlacementPass> {
-  PlacementPass() = default;
+struct PlacementPass final
+    : PassWrapper<PlacementPass, OperationPass<ModuleOp>> {
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(PlacementPass)
+
   explicit PlacementPass(const CompilerTarget& compilerTarget)
       : target(compilerTarget) {}
 
+  void getDependentDialects(DialectRegistry& registry) const override {
+    registry.insert<QCODialect>();
+  }
+
 protected:
   void runOnOperation() override {
-    if (!target) {
-      llvm::reportFatalUsageError("No compiler target specified!");
-    }
-
     auto moduleOp = getOperation();
     auto func = mqt::getEntryPoint(moduleOp);
     if (!func) {
@@ -340,7 +341,7 @@ protected:
 
     auto computation = discoverComputation(func);
     if (failed(computation) ||
-        failed(checkCapacity(func, *target, *computation))) {
+        failed(checkCapacity(func, target, *computation))) {
       signalPassFailure();
       return;
     }
@@ -348,12 +349,12 @@ protected:
     const auto layout = Layout::fromMapping(
         llvm::to_vector(llvm::seq(computation->wires.size())));
     IRRewriter rewriter(&getContext());
-    applyPlacement(func.getFunctionBody(), *target, layout, *computation,
+    applyPlacement(func.getFunctionBody(), target, layout, *computation,
                    rewriter);
   }
 
 private:
-  std::optional<CompilerTarget> target;
+  CompilerTarget target;
 };
 
 struct MappingPass : impl::MappingPassBase<MappingPass> {
--- a/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
@@ -66,15 +66,300 @@ namespace mlir::qco {
 using namespace mlir::qtensor;
 
 #define GEN_PASS_DEF_MAPPINGPASS
+#define GEN_PASS_DEF_PLACEMENTPASS
 #include "mlir/Dialect/QCO/Transforms/Passes.h.inc"
 
 namespace {
+
+using Wires = SmallVector<WireIterator>;
+
+struct WireInfos {
+  /// Return the mapped wire index of a program index.
+  [[nodiscard]] size_t lookupIndex(const size_t prog) const {
+    assert(containsProgram(prog) && "program index is not mapped");
+    return programToIndex_[prog];
+  }
+
+  /// Return the mapped program index of a wire index.
+  [[nodiscard]] size_t lookupProgram(const size_t index) const {
+    return indexToProgram_[index];
+  }
+
+  /// Bidirectionally map a wire index to a program index.
+  /// Overwrites existing mappings.
+  void insertOrUpdate(const size_t index, const size_t prog) {
+    if (index >= indexToProgram_.size()) {
+      indexToProgram_.resize(index + 1);
+    }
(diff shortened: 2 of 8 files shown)


mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
simon1hofmann, line 1: Should we document here that `createMappingPass` now requires an explicit topology?

simon1hofmann (approved):
Went through the changes and they look good to me 👍 

I have one comment found by ponytail, where I'm not entirely sure if we should fix it that way:

[P2] The TableGen declaration creates an unsafe targetless public pass.

[Passes.td](https://github.com/munich-quantum-toolkit/core/blob/5f52947580123f622513bc7e577a4e1d7ae9243e/mlir/include/mlir/Dialect/QCO/Transforms/Passes.td#L117-L131) generates `createPlacementPass()` and `registerPlacementPass()` with no `CompilerTarget`. That default-constructs [PlacementPass](https://github.com/munich-quantum-toolkit/core/blob/5f52947580123f622513bc7e577a4e1d7ae9243e/mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp#L322-L356), whose execution exits the process through `reportFatalUsageError`.

The canonical pipeline is safe because it calls the target-taking overload. The problem is the valid-looking public factory and generated `place-qubits` registration.

Ponytail fix: remove `PlacementPass` from TableGen and implement it as an explicitly target-bound `PassWrapper`, matching [TargetNativeSynthesisPass](https://github.com/munich-quantum-toolkit/core/blob/5f52947580123f622513bc7e577a4e1d7ae9243e/mlir/lib/Dialect/QCO/Transforms/NativeSynthesis/TargetSynthesis.cpp#L441-L449). 

Expose only:
`createPlacementPass(const CompilerTarget&)`

This removes code and an unusable API.

mlir/include/mlir/Dialect/QCO/Transforms/Mapping/Mapping.h
burgholzer: It's kind of documented in the pass documentation itself already, but may not hurt to make it explicit here as well. 👍🏼

burgholzer: 🤖 *AI text below* 🤖

Yes. The factory documentation now states that mapping requires an explicit topology.

I also adopted the review-body suggestion: placement is now an internal target-bound `PassWrapper`, and the generated targetless factory plus `place-qubits` registration have been removed. The full mapping and compiler suites, generated docs, repository lint, and changed-file clang-tidy pass. An independent Ponytail audit of the resulting complete PR diff found nothing further to remove.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -354,43 +576,46 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
     if (!func) {
-      mod.emitError() << "does not contain an entry point function";
+      moduleOp.emitError() << "does not contain an entry point function";
       signalPassFailure();
       return;
     }
 
-    auto comp = discoverComputation(func);
-    if (failed(comp)) {
+    if (failed(validateRoutingOperations(func))) {
```
MatthiasReumann, line 594: Do we really need such a check or is this one of those _assumptions_ we've talked a few hours ago?

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -70,83 +70,305 @@ using namespace mlir::qtensor;
...
+struct PlacementPass final
+    : PassWrapper<PlacementPass, OperationPass<ModuleOp>> {
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(PlacementPass)
+
+  explicit PlacementPass(const CompilerTarget& compilerTarget)
+      : target(compilerTarget) {}
+
+  void getDependentDialects(DialectRegistry& registry) const override {
+    registry.insert<QCODialect>();
+  }
```
MatthiasReumann, line 330: I think TableGen can generate these automatically.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -70,83 +70,305 @@ using namespace mlir::qtensor;
...
-  private:
-    /// Maps the i-th wire index to a program index.
-    SmallVector<size_t> indexToProgram_;
-    /// Maps a program index to the i-th wire index.
-    SmallVector<size_t> programToIndex_;
-    /// Program indices that have corresponding wires.
-    DenseSet<size_t> programs_;
-  };
+    const auto layout = Layout::fromMapping(
+        llvm::to_vector(llvm::seq(computation->wires.size())));
```
MatthiasReumann, line 350: I think a `Layout::identity` utility would be great here for ease of understanding.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -354,43 +576,46 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
       llvm::reportFatalUsageError("No compiler target specified!");
     }
```
MatthiasReumann, line 577: Given that the `PlacementPass` uses a non-optional `CompilerTarget` I think it could make sense to apply the same logic here.

MatthiasReumann (changes requested):
Looks great! Pretty much the changes that I would expect. 

Left some minor comments; otherwise let's get this merged!

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Agreed! Let me add that real quick!

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Hm. I am not too sure. Passes may only assume that the incoming IR is valid. There is valid IR that the routing pass cannot be applied to. So I would be tempted to argue that this verification here makes sense. I'll run this through Codex to make sure.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Here this makes much more sense in my opinion. 👍🏼 A Compiler target is a pre-condition for running the pass.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Let's see. I think this may have to do with this just being a pass wrapper.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Codex seems to agree 👇🏼 

🤖 *AI text below* 🤖

Yes, this check is necessary. A higher-arity unitary is valid QCO IR, but this pass only routes one- and two-qubit operations. The routing code interprets ready unitaries as operand pairs, so treating decomposition as an unchecked assumption could silently produce incorrect output. Keeping the check before placement also ensures failure without mutating the IR. The same validation already existed in `discoverComputation`; this PR only separates it so the placement-only path can accept higher-arity operations.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: I figured that this was due to something along the lines below. We may be able to adjust this if we make changes to the textual MLIR pipeline. Not in this PR though.

🤖 *AI text below* 🤖

TableGen can generate this boilerplate only by also generating a no-argument factory and pass-registry constructor. There is no `CompilerTarget` available to textual pass registration, so that would recreate the unusable targetless placement API removed after the previous review. The manual target-bound wrapper is intentional and matches the existing target-native synthesis and conformance passes.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Codex was a bit picky here. But thinking about it a little more deeply, I think I agree. As long as there are no other uses, the utility does not really provide that much value.

🤖 *AI text below* 🤖

I checked the complete repository before adding this. This is the only identity-layout construction, so a public helper would add a header API, implementation, and contract test for one explicit two-line expression. Under the focused scope of this PR, keeping `Layout::fromMapping(llvm::to_vector(llvm::seq(n)))` is the smaller design, so I am leaving it local after all.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: I am afraid codex also has a point here. We may want to take this feedback into a refactor of the textual IR in a follow up (after your stack of PRs has been merged).

🤖 *AI text below* 🤖

I agree with the invariant, but the clean change is not a one-line member replacement. The generated base owns six options, the statistic, cloning, documentation, and the targetless factory and registration. Making the target mandatory therefore requires converting the complete mapper to a manual target-bound pass and relocating that generated surface. All supported callers already use `createMappingPass(target, options)`. That legacy cleanup is broader than this focused prerequisite, so I am leaving it unchanged here instead of adding a partial workaround.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
MatthiasReumann: Good to know, thanks!

MatthiasReumann (approved):
LGTM now 👍🏻

Issue #2255, opened 2026-08-26
Audit MLIR pass, verifier, and rewrite contracts


burgholzer:
🤖 *AI text below* 🤖

Parent: #2250

## Scope

Audit passes, verifiers, rewrite patterns, conversions, and dialect registration against the MQT-owned MLIR policy and adopted upstream guidance.

Prioritize valid-input crashes, invalid pass output, non-local verifier assumptions, undeclared dependent dialects, unbounded recursion, untruthful rewrite APIs, missed constant matching, and duplicated traits, interfaces, conversions, or utilities.

## Delivery rules

Fix internal objective violations in focused PRs with regression tests. Treat crashes, unsafe rewrites, miscompilation, and invalid output as v4 blockers. Record lower-risk findings separately rather than expanding a blocker fix into a style patch.

PR #2319, opened 2026-09-01
🐛 Verify modifier bodies at the dialect boundary

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Reject invalid operations at the QC and QCO modifier boundary instead of adding pass-local guardrails to measurement lifting.

Modifier bodies may contain operations implementing `UnitaryOpInterface` and pure, speculatable, qubit-free classical support computations. Effectful or unknown operations, register access, and non-unitary operations carrying qubits are rejected by the dialect verifiers. Measurement lifting can consequently assume valid IR and no longer needs its redundant memory-effect checks.

The QIR conversion also rejects controlled global phases before mutating or inlining the control region. This preserves the existing diagnostic for valid multi-block classical computations that produce a non-hoistable phase angle.

Part of #2255.

## Validation

- QC IR: 339/339 tests passed
- QCO IR: 494/494 tests passed
- QC-to-QIR adaptive: 149/149 tests passed
- QCO optimizations: 194/194 tests passed
- Total affected suites: 1,176/1,176 tests passed
- `nox -s lint`

## Checklist

- [x] The QC and QCO modifier documentation describes the verifier contract.


DRovara (commented):
That's an edge case I didn't think about, but I also have some concerns about it.

In particular, I'm not very convinced by the example test case. I assume the `@observe` function should represent a function that contains a measurement. But it's not really reasonable to invert a measurement.

In fact, I would say the contents of an `qco.inv` modifier should always be unitary. That means the only way to have actual memory effects in there is to have a `gphase` operation, because that adds a global phase with no targeted qubits (but actually, in that case, I don't even know if the `inv` modifier allows 0 target qubits.

Similarly, `qco.ctrl` also only makes sense for unitaries.


What are your thoughts on that?

simon1hofmann:
> That's an edge case I didn't think about, but I also have some concerns about it.
> 
> In particular, I'm not very convinced by the example test case. I assume the `@observe` function should represent a function that contains a measurement. But it's not really reasonable to invert a measurement.
> 
> In fact, I would say the contents of an `qco.inv` modifier should always be unitary. That means the only way to have actual memory effects in there is to have a `gphase` operation, because that adds a global phase with no targeted qubits (but actually, in that case, I don't even know if the `inv` modifier allows 0 target qubits.
> 
> Similarly, `qco.ctrl` also only makes sense for unitaries.
> 
> What are your thoughts on that?

Thanks a lot for the review, you're right, the func.call example is invalid modifier content. I will retarget this PR to reject unknown/effectful non-unitary operations in both modifier verifiers while permitting pure classical support and UnitaryOpInterface operations.

The audit produced a lot of edge cases that are quite hypothetical.

mlir/lib/Conversion/QCToQIR/QIRCommon/QIRCommon.cpp
```diff
@@ -336,6 +336,12 @@ struct ConvertQCCtrlOp final : StatefulOpConversionPattern<CtrlOp> {
                                          "Nested CtrlOps are not supported");
     }
 
+    bool hasGlobalPhase = false;
+    op.getRegion().walk([&](GPhaseOp) { hasGlobalPhase = true; });
+    if (hasGlobalPhase) {
+      return op.emitError("Controlled GPhaseOps cannot be converted to QIR");
+    }
+
```
DRovara: I guess a (multi-)controlled `GPhaseOp` is also equivalent to a (controlled) targeted `P` gate, so rather than rejecting, one could also translate it to that?

mlir/lib/Dialect/QC/IR/Modifiers/ModifierUtils.cpp
```diff
@@ -56,6 +48,30 @@ LogicalResult verifyModifierBody(Operation* modifierOp, Block& body) {
...
+        return !isPure(&operation) ||
+               operation
+                   .walk([](Operation* nested) {
+                     const auto carriesQubit =
+                         llvm::any_of(nested->getOperandTypes(),
+                                      containsQubit) ||
+                         llvm::any_of(nested->getResultTypes(), containsQubit);
+                     return isa<UnitaryOpInterface>(nested) || carriesQubit
+                                ? WalkResult::interrupt()
+                                : WalkResult::advance();
```
DRovara: Is there any situation where the `isa<UnitaryOpInterface>(nested)` really matters here? If nested was a unitary, then `carriesQubit` would be true anyways, right?

mlir/lib/Dialect/QCO/IR/Modifiers/ModifierUtils.cpp
```diff
@@ -55,6 +48,30 @@ LogicalResult verifyModifierBody(Operation* modifierOp, Block& body) {
...
+               operation
+                   .walk([](Operation* nested) {
+                     const auto carriesQubit =
+                         llvm::any_of(nested->getOperandTypes(),
+                                      containsQubit) ||
+                         llvm::any_of(nested->getResultTypes(), containsQubit);
+                     return isa<UnitaryOpInterface>(nested) || carriesQubit
+                                ? WalkResult::interrupt()
+                                : WalkResult::advance();
+                   })
```
DRovara: Same as above.

mlir/lib/Dialect/QC/IR/Modifiers/ModifierUtils.cpp
```diff
@@ -56,6 +48,30 @@ LogicalResult verifyModifierBody(Operation* modifierOp, Block& body) {
         "arguments");
   }
 
+  const auto hasNonUnitaryOperation =
+      llvm::any_of(body.without_terminator(), [](Operation& operation) {
+        if (isa<UnitaryOpInterface>(operation)) {
+          return false;
+        }
+        return !isPure(&operation) ||
```
DRovara: I'm not 100% sure but I think this allows measurements.

`MeasureOp` is marked as `Pure` and has no nested operation to walk.

DRovara (changes requested):
Hi @simon1hofmann, thanks for the update! Looks quite reasonable, I think it's good to make these verifications in advance.

I do have some minor concerns on the correctness/completeness of the implementation though, but I'm not 100% sure on any of them.

mlir/lib/Conversion/QCToQIR/QIRCommon/QIRCommon.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Agreed. A controlled global phase can be expressed as a phase gate on one control, controlled by the remaining controls. The existing `normalizeGlobalPhases` pass already performs that rewrite when the angle can be hoisted, and both QIR pipelines run it before conversion. The extra rejection added here duplicated the downstream check, so I removed it. Region-local, non-hoistable angles remain unsupported because they require CFG-aware modifier lowering; that is separate from this measurement-lifting fix.

mlir/lib/Dialect/QC/IR/Modifiers/ModifierUtils.cpp
simon1hofmann: 🤖 *AI text below* 🤖

The generic verifier was the wrong abstraction and has been removed because it rejected valid structured quantum flow. For completeness, `qc.gphase` implements `UnitaryOpInterface` but has no qubit operands or results, so the `isa` check was not redundant within that predicate. The branch now leaves the existing explicit QC verifier unchanged.

mlir/lib/Dialect/QCO/IR/Modifiers/ModifierUtils.cpp
simon1hofmann: 🤖 *AI text below* 🤖

The generic verifier was the wrong abstraction and has been removed because it rejected valid structured quantum flow. For completeness, `qco.gphase` implements `UnitaryOpInterface` but has no qubit operands or results, so the `isa` check was not redundant within that predicate. The branch now leaves the existing explicit QCO verifier unchanged.

mlir/lib/Dialect/QC/IR/Modifiers/ModifierUtils.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Good catch. The branch no longer relies on purity or type inference for this. I removed the generic verifier and restored the explicit recursive check that rejects `qc.measure`, including measurements nested in modifier bodies. The existing direct and nested verifier tests cover this behavior.

mlir/lib/Conversion/QCToQIR/QIRCommon/QIRCommon.cpp
DRovara: I've resolved these comments now insofar that they no longer appear in the code, but by now the implementation is back to how it was before my first comment, so I guess it's not fully done yet.

simon1hofmann: 🤖 *AI text below* 🤖

Your concern is now addressed by #2435, which has merged into main. The QC and QCO modifier verifiers reject effectful non-unitary support operations, including the `func.call @observe()` used by this PR's remaining regression. That fixture is therefore invalid under the current contract. Reviewing the remaining measurement-lifting guards against current core did not reveal a correctness case that still requires them.

Controlled global phases are handled by the existing phase normalization, and #2435 also preserves eager parameter computations during modifier rewrites.

Closing #2319 as superseded by #2435.

PR #2320, opened 2026-09-01
🐛 Preserve yield permutations when unrolling loops

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Preserve result permutations when fully unrolling yield-only quantum loops. The shortcut that erases an empty loop now applies only when every yielded value is the corresponding iteration argument; non-identity yields are handled by the regular unrolling rewrite so their value mapping is retained.

A regression covers a one-iteration loop that swaps two yielded qubits.

Part of #2255.

## Validation

- `QuantumLoopUnrollTest.PreservesYieldOnlyPermutation`
- Full quantum-loop-unroll suite: 6/6 tests passed
- `nox -s lint`

## Checklist

- [x] Documentation remains accurate; no documentation update is required for this internal correctness fix.

### DIFF
--- a/mlir/lib/Dialect/QCO/Transforms/Optimizations/QuantumLoopUnroll.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Optimizations/QuantumLoopUnroll.cpp
@@ -60,6 +60,15 @@ static SmallVector<scf::ForOp> collectQuantumLoops(FunctionOpInterface func) {
   return loops;
 }
 
+static bool hasIdentityYieldOnlyBody(scf::ForOp loop) {
+  if (!llvm::hasSingleElement(loop.getBody()->getOperations())) {
+    return false;
+  }
+  return llvm::equal(
+      loop.getRegionIterArgs(),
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
This LGTM 👍🏼 
Just wondering: Does this also work with unrolling multiple loop rounds when the loop body is just a permutation?
Asked differently, if the test were to run two iterations, would it result in the true "empty" circuit again with 0->0 and 1->1 mapping?
It's not really necessary to add an explicit test for this as part of the PR; I just want to make sure.

simon1hofmann:
> This LGTM 👍🏼 
> Just wondering: Does this also work with unrolling multiple loop rounds when the loop body is just a permutation?
> Asked differently, if the test were to run two iterations, would it result in the true "empty" circuit again with 0->0 and 1->1 mapping?
> It's not really necessary to add an explicit test for this as part of the PR; I just want to make sure.

Yes: the swap body is not treated as identity, so loopUnrollFull carries each iteration’s yielded values into the next. One iteration maps (q0, q1) → (q1, q0); two map back to (q0, q1), leaving the empty identity circuit.

PR #2321, opened 2026-09-01
🐛 Avoid releasing static QIR qubits

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Avoid lowering source qubit deallocations to `__quantum__rt__qubit_release` when Adaptive QIR uses static allocation. Static qubits are runtime-owned handles rather than dynamically allocated resources, so their source deallocations must be discarded instead of passed to the dynamic release API.

A regression verifies that valid static-allocation input emits no release call.

Part of #2255.

## Validation

- `QCToQIRAdaptiveNativeTest.DoesNotReleaseStaticQubits`
- Full Adaptive QIR suite: 147/147 tests passed
- `nox -s lint`

## Checklist

- [x] Documentation remains accurate; no documentation update is required for this internal correctness fix.

### DIFF
--- a/mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
+++ b/mlir/lib/Conversion/QCToQIR/QIRAdaptive/QCToQIRAdaptive.cpp
@@ -412,6 +412,10 @@ struct ConvertQCDeallocOp final : StatefulOpConversionPattern<DeallocOp> {
   matchAndRewrite(DeallocOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter& rewriter) const override {
     auto& state = getState();
+    if (state.allocationMode == AllocationMode::Static) {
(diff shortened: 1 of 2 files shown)


burgholzer (changes requested):
In general, I would almost argue that it is invalid IR to deallocate a static qubit. @simon1hofmann might be worth evaluating whether that would be the cleaner assumption here.
The test added here feels like it is promoting an anti-pattern and should rather be ruled out by design.

simon1hofmann:
🤖 *AI text below* 🤖

Thanks, that makes sense. Deallocating a static qubit should be treated as
invalid IR rather than handled defensively in the QIR conversion pass. The
added test therefore encodes an anti-pattern, and the pass-local guard is the
wrong ownership boundary.

Closing this PR without merge. If we want to enforce the invariant explicitly,
that should be separate verifier-level work.

PR #2323, opened 2026-09-01
✨ Preserve target-native controlled operations

burgholzer:
🤖 *AI text below* 🤖

## Description

This PR makes the canonical target-compilation pipeline preserve controlled operations that an all-to-all `CompilerTarget` reports as native. It reuses the existing multi-controlled decomposition pass with one target-aware factory; the standalone and command-line pass remain target-independent, and explicit-connectivity targets retain the established decomposition before mapping.

This lets the DDSIM target capability introduced by #2218 apply broadly rather than special-casing `mcx` and `mcp`. The compiler regression covers arbitrarily controlled H, RX, RXX, SWAP, RCCX, X, and phase operations plus bare RCCX and global phase. An end-to-end QDMI regression compiles a Qiskit circuit to QIR, executes it on DDSIM, and compares the exact complex statevector—including nonzero global phase—with Qiskit. Minimum dependency sessions skip this regression because they install Qiskit 1.1, while the registered translator supports Qiskit 2.5.x and explicit candidate builds.

This PR is one signed commit stacked on #2218 at `c7e94438069dfa6635da7e45bf9b3642b8922af8`. It adds no new abstraction or dependency: target ownership stays in the existing pass, `CompilerTarget::supports` remains the single capability check, and unsupported operations continue through the current decomposition path.

Capability checks are intentionally not cached. `CompilerTarget` already indexes operations in a `StringMap`, and the pass performs one cheap structural lookup for each candidate `CtrlOp` or `RCCXOp`. An operation-pointer cache would need invalidation without reducing demonstrated cost.

The changelog records #2323 in the existing MQT Compiler Collection launch entry rather than adding a separate entry for unreleased functionality. Its PR-link definition block remains numeric descending. New Doxygen comments use `///`.

## Validation

- Compiler unit tests: 144 passed.
- Current Qiskit 2.5 QDMI/DDSIM statevector test: 1 passed.
- Minimums Qiskit 1.1 session: 1 expected skip.
- `uvx nox -s lint`
- `uvx nox -s cpp-lint -- origin/codex/generalize-compiler-target`
- `git diff --check`
- `git verify-commit HEAD`
- An independent final audit found no concrete issue.

Implementation and pull-request text were prepared with OpenAI Codex.

## Checklist

- [x] The changes follow the project guidelines and introduce no new warnings.

### DIFF
--- a/mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp
@@ -23,6 +23,7 @@
 #include <mlir/IR/MLIRContext.h>
 #include <mlir/IR/PatternMatch.h>
 #include <mlir/IR/Value.h>
+#include <mlir/IR/Visitors.h>
 #include <mlir/Support/LLVM.h>
 #include <mlir/Support/LogicalResult.h>
 #include <mlir/Transforms/GreedyPatternRewriteDriver.h>
@@ -1447,11 +1448,30 @@ protected:
                        CompilerTarget::Connectivity::Kind::AllToAll
             ? &*target_
             : nullptr;
+
+    SmallVector<Operation*> rewriteOps;
+    getOperation().walk<WalkOrder::PreOrder>([&](UnitaryOpInterface op) {
+      if (nativeTarget != nullptr &&
+          nativeTarget->supports(op.getOperation())) {
+        return WalkResult::skip();
+      }
+      if (isa<CtrlOp, RCCXOp>(op)) {
+        rewriteOps.push_back(op.getOperation());
+      }
+      return WalkResult::advance();
+    });
+    if (rewriteOps.empty()) {
+      return;
+    }
+
     RewritePatternSet patterns(&getContext());
     patterns.add<DecomposeControlledGatePattern, DecomposeRCCXPattern>(
         &getContext(), minQubits, nativeTarget);
 
-    if (failed(applyPatternsGreedily(getOperation(), std::move(patterns)))) {
+    GreedyRewriteConfig config;
+    config.setStrictness(GreedyRewriteStrictness::ExistingAndNewOps);
+    if (failed(
+            applyOpPatternsGreedily(rewriteOps, std::move(patterns), config))) {
       signalPassFailure();
     }
   }
--- a/mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp
+++ b/mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp
@@ -8,6 +8,7 @@
  * Licensed under the MIT License
  */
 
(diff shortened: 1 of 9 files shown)


burgholzer:
This one also feels ready and was iterated quite heavily today.

mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp
```diff
@@ -1318,14 +1319,19 @@ namespace {
...
+                                          const CompilerTarget* target)
+      : OpRewritePattern<CtrlOp>(context), minQubits_(minQubits),
+        target_(target) {}
 
   LogicalResult matchAndRewrite(CtrlOp op,
                                 PatternRewriter& rewriter) const override {
     if (op.getNumQubits() < minQubits_) {
       return failure();
     }
+    if (target_ != nullptr && target_->supports(op.getOperation())) {
```
simon1hofmann: Returning failure here preserves only the outer `CtrlOp`; `applyPatternsGreedily` still visits its body.

For example:
```
OPENQASM 3.0;
include "stdgates.inc";
qubit[4] q;
ctrl @ rccx q[0], q[1], q[2], q[3];
```

Consider a valid four-site all-to-all target with native U, CX, and rccx with variadic minimum four. It supports a four-qubit ctrl @ rccx, but not the bare three-qubit RCCX inside that shell. The [RCCX pattern](https://github.com/munich-quantum-toolkit/core/blob/60d5bb4db2a4672214ea793dbfe6e2d6c92c8863/mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp#L1407-L1418) therefore expands the body. The outer control then contains multiple unitaries, is no longer considered native, and final conformance fails.

simon1hofmann (approved):
This one also looks good, just found one corner case to check.

mlir/lib/Dialect/QCO/Transforms/Decomposition/DecomposeMultiControlled.cpp
burgholzer: Hm. I am not sure I fully follow. Or at least I am confused.
A valid target configuration would likely also contain the fixed `rccx` gate (and not only the variadic one). So it would likely not be decomposed.

More generally though, I believe the decomposition should not recurse into control modifiers, should it?
More generally, the pattern driver should not recurse into operations that are already deemed supported, right?
Can we fix that?

simon1hofmann: This can occur for a valid target: variadic rccx with minimum arity 4 supports the four-qubit ctrl @ rccx shell but not its nested three-qubit rccx. I updated the greedy driver to skip regions of target-native unitary operations.

burgholzer: Pushed one small improvement for this. I think this should also be ready to go in now :shipit:

PR #2336, opened 2026-09-02
✨ Add reusable quantum functions to QC and QCO

burgholzer:
🤖 *AI text below* 🤖

## Description

Add a frontend-neutral function model for reusable quantum helpers:

- mark unitary `func.func` definitions with `mqt.unitary` and represent their calls with `qc.call` or `qco.call`;
- expose complete callback-based function and call APIs in both program builders;
- use one positional QCO ABI: ordinary results first, then the updated value of each borrowed scalar qubit argument;
- preserve supported functions and calls across QC/QCO conversion while rejecting representations that would lose information; and
- treat generic `func.call` operations as local wire-analysis boundaries, avoiding speculative interprocedural mapping and caches.

The deliberately narrow unitary contract covers gate definitions needed by future OpenQASM, Qiskit, and jeff integrations without result annotations, recursive matrix synthesis, or a new function abstraction. This branch is based directly on `main` and is independent of #2196 and its stack.

### Validation

- Full Debug and Release C++ suites: 3,840 passed and one expected skip each.
- `uvx nox -s cpp-lint`, `uvx nox -s lint`, and `git diff --check` passed.
- Doxygen output retains the full documentation for all 90 generated gate methods in each builder, with no new warnings relative to `main`. The `@file` summary is retained.
- The latest documentation correction changes no non-comment C++ tokens. Existing trailing Doxygen syntax and inline unused-parameter comments are preserved.

Build and C++ lint validation used #2352's dependency and test-RPATH fixes as a temporary local overlay. That overlay was removed afterward and is not part of this PR; both PRs remain based on `main`.

Codex assisted with implementation, documentation restoration, and local validation.

## Checklist


burgholzer:
@MatthiasReumann @DRovara I haven't looked through all the details here yet, but this PR may be interesting for the both of you.
Matthias, it addresses some of the concerns that you may have had on the WireIterator call handling.
Damian, this was mostly inspired by the lack of any production user/consumers of the IPO passes and multiple function handling. This triggered me, which eventually led to the implementation here which feels quite nice already and actually rather close to some of the things we also had in the original RFC for the dialect rewrite but never got to so far. I plan to add the respect new capabilities to all the supported formats in follow-up PRs.
This is more so an FYI than a call for action.

mlir/lib/Dialect/QCO/Utils/FunctionUtils.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "mlir/Dialect/QCO/Utils/WireIterator.h"
+
+#include <llvm/ADT/STLExtras.h>
+#include <mlir/Dialect/Func/IR/FuncOps.h>
+#include <mlir/IR/SymbolTable.h>
+
+using namespace mlir;
+using namespace mlir::qco;
+
+FailureOr<unsigned> mlir::qco::traceQubitArgument(func::FuncOp function,
```
MatthiasReumann, line 23: Yes! This is exactly what I had planned for the function calling! Love it 👍🏻 

Maybe I'm missing something: But shouldn't this function return an unsigned at the end?

burgholzer: It does in the following block that is part of the `while(true) {...}` loop:
```
if (auto argument = dyn_cast<BlockArgument>(value)) {
  if (argument.getOwner() == &function.getBody().front() &&
      isa<QubitType>(argument.getType())) {
    return argument.getArgNumber();
  }
  return failure();
}
```

MatthiasReumann: I see, thanks!

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
```diff
@@ -304,6 +309,180 @@ verifyEntryPoint(Operation* operation, const NamedAttribute attribute) {
   return success();
 }
 
+[[nodiscard]] static bool hasQCQubit(Type type) {
+  return isa<qc::QubitType>(type);
+}
+
+[[nodiscard]] static bool hasQCOQubit(Type type) {
+  return isa<qco::QubitType>(type);
+}
```
DRovara: Just a minor thing, but why are these called `hasQCQubit` and not `isQCQubit`?

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
```diff
@@ -304,6 +309,180 @@ verifyEntryPoint(Operation* operation, const NamedAttribute attribute) {
...
+
+template <typename CallOp>
+[[nodiscard]] static LogicalResult
+verifyNoUnitaryRecursion(func::FuncOp function) {
```
DRovara, line 317: I'm not 100% sure but I believe the `unitary` marker could never be valid for any function that can be considered recursive to begin with, because a recursive function will always need a control flow operation to identify the base case / stop the recursion, which would fail verification in `verifyQCUnitaryBody`.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
DRovara, line 1: The changes here actually clash quite a bit with the IPO stack.

The IPO passes are specifically made to work *not only* with unitary functions, but arbitrary functions. And we needed the wire/tensor iterators to thread corresponding variables of qubit types and qtensor types through functions.

Accordingly, the IPO pass will need *some* functionality like this either way, so maybe removing them is not the best choice?

(I imagine the AI model decided to remove the code because no active consumer of the WireIterator/TensorIterator changes has been merged in yet, so it didn't notice that it was actually necessary)

DRovara (commented):
Awesome, this looks pretty cool. I like the idea of the unitary marker for unitary functions.

I have to admit I am a bit behind and still haven't fully grasped the idea behind the MQT dialect here (unitary functions aren't necessarily MQT-exclusive), but I also understand just having that as a joint source to avoid duplication.

I have a few minor comments and one major concern regarding the IPO passes, where we will have to think of *some* solution.

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
burgholzer: Good point. I'd even go a step further and question the justification for the presence of these functions entirely. They are just one-liners. I'll drop them.

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
burgholzer: Hm. You might have a point here. Let me try to verify that assumption and potentially remove the duplicate verification.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
burgholzer: The AI model actually had the context of the entire IPO PR stack.
However, it heavily argued that it would be preferable to first streamline the most practical cases and make the most use of existing IPO functionality in MLIR.
@MatthiasReumann also already voiced concerns about the previous implementation here.
So my stance would be that I'd rather put this on a clean slate right now, build the infrastructure for practically relevant quantum programs and the connection to/from OpenQASM, Qiskit, and jeff; and then add the necessary capabilities based on that.
I hope this makes sense.

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
DRovara: At the very least I can't think of a way to write a recursive function that is not an infinite loop and only uses unitary operations.

We might eventually say that certain applications of classically controlled gates still count as unitaries (after all you can view them as the sum of two unitary matrices scaled by a constant factor 1 or 0, depending on the condition outcome), but the way it is right now, we reject any `qco.if` as non-unitary.

mlir/include/mlir/Dialect/QCO/Utils/WireIterator.h
MatthiasReumann: My concerns: I think the user should be responsible for constructing new `WireIterator`s when reaching a call-op (using the newly added function). This keeps the `WireIterator` lean and non-recursive.

DRovara: Sure, that's a fair point, if the goal was to set up a clean baseline and then build further requirements ontop, if necessary.

I interpreted the replacement as "from now on, let's use only this", which wouldn't have worked, but I'm fine with giving it another try on the new infrastructure once this is merged.

burgholzer: Great. Then let's role with this 👍🏼 
Trying to get a few more other PRs in, then I'll rebase, clean up, and proceed with a first merge here.

mlir/lib/Dialect/MQT/IR/MQTDialect.cpp
burgholzer: Hm. `qco.if` is non-unitary kind of by definition. There is no way to write down a single unitary matrix for an if-else expression.

DRovara: Yes but I would argue it is "unitary enough" that I wouldn't rule out some way to mathematically prove a recursive function that uses only a single if-else block can still be unitary. At the end of the day

```
if (c) {
  x q;
}
else {
  z q;
}
```

can be expressed as

$$c \begin{bmatrix} 0 & 1 \\\ 1 & 0 \end{bmatrix} + (1 - c)\begin{bmatrix} 1 & 0 \\\ 0 & -1 \end{bmatrix}$$

Generally, recursions can be translated to loops and loops can be unrolled if the bounds are constant, leaving us without any non-unitary in such a case. 
That's what I meant, while `qco-if` is non-unitary, there might be a future where we extend our definition of unitary functions to still allow single `qco.if`s in such cases.

burgholzer: I am not sure I fully agree with the reasoning here.
If I look at this through my classical simulation lense, then I see two scenarios:
- `c` depends on a runtime value, in which case I cannot describe the whole program with one unitary and would also have to resort to individual sampling instead of batch sampling -> not unitary
- `c` can be resolved at compile-time, at which point the `if` becomes irrelevant and would be eliminated by constant propagation and/or canonicalization.

That being said, this does not concern the PR here and we can happily look into extending the scope in future changes.

DRovara: Fair points, I get what you mean. But yeah it doesn't matter at all for this PR.

Maybe it's an interesting side thought at some point that could lead into a nice little student thesis, but I'd argue it's not even that impactful to begin with.

burgholzer:
Alright, I believe this is as clean as it gets for now. Tagging this for auto-merge so that the other depending PRs can be updated over night.

PR #2341, opened 2026-09-03
✨ Support register-wide Qiskit classical initialization

### LINKED ISSUE #2332: ✨ Support register-wide Qiskit classical initialization
🤖 This response was generated with AI assistance.

### Problem Statement

Qiskit classical bits start at zero, whereas an uninitialized OpenQASM 3 `bit[n]` declaration has undefined contents. The Qiskit serializer therefore currently prepends one explicit `store(false)` operation per classical bit to preserve Qiskit semantics.

Qiskit can instead store zero into a complete `ClassicalRegister`, which its OpenQASM 3 exporter writes as:

```openqasm
bit[n] c;
c = 0;
```

The MQT OpenQASM frontend currently rejects that RHS because whole-register assignments require a bit-register value. This prevents replacing the per-bit initialization with one register-wide operation.

### Proposed Solution

Support a well-defined whole-register initialization path for the OpenQASM emitted by Qiskit, then simplify the serializer to initialize each classical register once.

Before implementation, determine whether `c = 0` is handled as a standard OpenQASM assignment conversion or as an explicitly documented Qiskit-compatibility form. In either case:

- define width and nonzero-value behavior precisely;
- preserve undefined initialization for OpenQASM declarations that have no assignment;
- cover single- and multi-bit registers, multiple registers, and partially measured registers;
- replace the per-bit serializer loop only after the frontend accepts the exported program;
- coordinate with #2289, which tracks bit-register-to-integer casts in the opposite direction.

Follow-up to #2288 and discussion https://github.com/munich-quantum-toolkit/core/pull/2288#discussion_r3908878547.

simon1hofmann:
🤖 *AI text below* 🤖

## Summary

- accept Qiskit's literal-zero whole-register initialization as a narrow OpenQASM compatibility form
- initialize each Qiskit classical register once while retaining scalar initialization for loose classical bits
- preserve undefined native OpenQASM registers and reject every other scalar whole-register assignment
- document the compatibility behavior and cover semantic, lowering, and serialization paths

Closes #2332.

## Validation

- complete OpenQASM target suite: 177 passed
- complete Qiskit mock-backend module: 21 passed
- serializer checked with Qiskit 1.1 and 2.5
- C++ lint: 0 format and 0 tidy findings
- complete repository lint suite


simon1hofmann:
🤖 *AI text below* 🤖

The register-wide form has been removed. Qiskit 1.1 and 2.5 model `ClassicalRegister` values as integers, and their public OpenQASM 3 export path cannot emit a standards-compliant bit-string literal. The existing per-bit `false` stores remain the smallest portable implementation, so this PR now has no effective diff and is being closed. Issue #2332 remains available if Qiskit later exposes a suitable export representation.

PR #2351, opened 2026-09-03
🐛 Preserve mapped feed-forward control

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Prevent routing from creating a cyclic SSA dependency through later classical
control. An independent wire can advance past a conditional whose measurement
depends on an unresolved two-qubit gate. Borrowing the conditional's output for
a SWAP can then make the unresolved gate depend on itself.

Defer structured control while earlier wire operations still need routing, but
let it pass terminal sinks and output-only measurements. Preserve the wire value
at a composite's block-order boundary, including when another wire has advanced
through a gate using the composite's classical result. Exhausted idle wires in
nested backward traversal do not block other composites.

Classify consecutive measurements once per unchanged traversal, sharing slice
and register work across the run. Ignore register accesses before the run so
that earlier control does not force unnecessary measured-qubit reuse. Prefer
terminal measurements independently of the selected target profile.

Keep main's memory-effect-aware sorter unchanged. Mapping regressions cover
crossed control, terminal-wire progress, consecutive measurements, multiple
result users, and register effects. Sorter regressions cover repeated stores and
whole-register writes followed by indexed loads. No public API, new dependency,
changelog entry, or migration change is needed for this unreleased v4 correction.

## Validation

- At `aa1b13cf8`, the release build passes 100 mapping, 192 QCO utility, and
  181 compiler tests with GCC 13 and LLVM/MLIR 23.1.0 on ARM64 Linux.
- Added regressions cover idle nested wires, conditional gate parameters,
  terminal measurements after earlier register control, and first/last result
  control across consecutive measurements with shared stores.
- Repository lint and full-file C++ lint against fixed base `ce608b082` pass
  with zero findings.
- With the rebased #2452 exporter, all 398 Qiskit translation/loop tests pass,
  and the previously crashing nested Qiskit circuit compiles through the target
  pipeline. These are combined-stack checks.
- The synthetic 4,000-measurement mapping probe improved from 327.85 ms to
  2.398 ms before the boundary and earlier-access fixes; this is not a corpus
  or whole-program speed claim.
- Benchpress was not rerun. Hosted checks must validate the new head; earlier
  results do not validate this update.

### Previously recorded integration validation

At `8e7ad2934`, a fresh wheel passed 306 Qiskit translation tests, two one-qubit
synthesis regressions, and all 68 Benchpress integration tests. The following
Benchpress results were recorded at that revision and were not rerun for this
update.

### Remaining Benchpress gaps

Six of the 31 guarded feed-forward profiles pass the existing integration checks;
25 stop at the strict textual event-order guard. That guard can reject valid
reordering and these failures alone do not prove changed semantics. The old
three-qubit deterministic counterexample now preserves `11` before and after
compilation.

Direct Core diagnostics on all 25 rejected profiles produce valid mapped IR and
preserve their 3,416 conditional counts and comparison/read-epoch multisets.
These structural checks are not equivalence proofs. All 25 then fail native
Qiskit export when a delayed store crosses another measurement or control/modifier
operations.

BV100 still fails native export when measurements are grouped before their
stores. A minimal valid `measure; measure; store; store` program reproduces this
exporter restriction. #2439 supports intervening quantum operations, not another
measurement. Keep the integration guards and export fallback pending separate
fixes; this PR does not claim to enable the full guarded matrix or pass the full
Benchpress corpus.

Codex assisted with the merge, scope reduction, regressions, validation, and this
description. Hosted CI for this update remains pending.

## Checklist


mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1251,9 +1251,6 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
                     [](auto&) { return Direction == WireDirection::Backward; })
                 .template Case<IfOp, IndexSwitchOp, scf::ForOp, scf::WhileOp>(
                     [&](auto&) {
-                      if (indices.size() == 1) {
-                        return true;
-                      }
```
MatthiasReumann, line 1251: This is the most trivial fix. The problem is that the SCF op (the `IfOp` in the added unit-test) depends on the classical outcome of the measurement indirectly via a side-effect chain (thus it is not actually executable). 

In the future we should include classical dependencies in the "advancement" logic. I am currently not sure how to best model this.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1232,11 +1232,36 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+                                .Case<cbit::LoadOp, cbit::StoreOp>(
+                                    [&](auto ls) {
+                                      return ls.getIndex() == store.getIndex();
+                                    })
+                                .template Case<cbit::ReadOp, cbit::WriteOp>(
+                                    [](auto) { return true; })
+                                .Default([](Operation*) { return false; });
+                          });
+                    }
+                  }
```
MatthiasReumann: Another example why a linearly-typed `cbit` dialect would be nice. If we modeled classical registers this way, the `getForwardSlice` call below could handle all of the above as well!

MatthiasReumann:
@coderabbitai full review

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1251,9 +1276,6 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
                     [](auto&) { return Direction == WireDirection::Backward; })
                 .template Case<IfOp, IndexSwitchOp, scf::ForOp, scf::WhileOp>(
                     [&](auto&) {
-                      if (indices.size() == 1) {
-                        return true;
-                      }
                       if (visited.insert(op).second) {
                         composites.emplace_back(op, indices);
```
simon1hofmann, line 1349: 🤖 *AI text below* 🤖

**[P1] Prevent terminal wires from blocking routing progress**

Removing the one-wire fast path can cause an infinite loop. A terminal measurement or sink on q0 blocks an independent one-qubit conditional on q1; lookahead still sees a later CX(q1,q2), but routing repeatedly returns an empty SWAP sequence without advancing. Ensure composite deferral cannot leave routing without progress.

Reproduced with an explicit three-site triangle, `ntrials=1`, `seed=42`: measure q0, sink q0, apply X to q1 inside `qco.if true`, then CX(q1,q2) and final sinks. The variant with only the early sink also hangs. Both inputs verify and finish on main `ec799daa0`; both exceed a 10-second timeout on `74c6f7435`.

simon1hofmann (commented):
🤖 *AI text below* 🤖

Two reproduced regressions in the current implementation, detailed inline. The existing 96 mapping and 187 QCO utility tests pass; the focused comparisons against main expose the failures.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1232,11 +1232,36 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
+                                !store->isBeforeInBlock(op)) {
+                              return false;
+                            }
+                            return TypeSwitch<Operation*, bool>(op)
+                                .Case<cbit::LoadOp, cbit::StoreOp>(
+                                    [&](auto ls) {
+                                      return ls.getIndex() == store.getIndex();
+                                    })
+                                .template Case<cbit::ReadOp, cbit::WriteOp>(
+                                    [](auto) { return true; })
```
simon1hofmann: 🤖 *AI text below* 🤖

**[P2] Do not treat register overwrites as feed-forward**

A later `StoreOp` or `WriteOp` does not necessarily consume the measured value. Counting an overwrite as adaptive lets routing place SWAPs after otherwise terminal measurements. Require an actual dependency on subsequent quantum work.

Reproduced on a three-site line with `ntrials=1`, `seed=0`: CX(q0,q1), CX(q0,q2), measure/store q0, CX(q1,q2), then measure/store q1 into the same register element, using the same SSA index and returning the register. Main keeps the measurements terminal and the mapped output exports as Base QIR. This PR inserts a SWAP consuming the first measurement's output qubit, and export fails with `QIR Base Profile forbids using a qubit after measurement`.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1179,18 +1154,132 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
...
-  /// Advance past all executable gates and return operations with nested
-  /// regions and the respective wire indices. Stops when no more executable
-  /// gates are found. The function positions each wire on a non-executable
-  /// two-qubit gate or a composite unitary, if possible. The function never
-  /// advances past sink-like operation and thus, each wire will never reach the
-  /// sentinel state.
+  /// Return whether quantum work follows a measurement through its wire, SSA
+  /// results, or any register effects. This function assumes the direction
+  /// WireDirection::Forward.
+  static bool measurementNeedsRouting(MeasureOp measurement) {
```
MatthiasReumann: @burgholzer This is the logic required to verify if a measured qubit is allowed to be reused for mapping (essentially: is-adaptive?). I know that there is an attribute in the QIR translation which specifies the profile. Any chance we could specify such an attribute for each function (for now just the `entry_point`) earlier in the compilation pipeline? If so, we could simplify this function dramatically (... in a future PR for `v4.1`).

MatthiasReumann (approved):

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
burgholzer: Let me explore this 👍🏼

burgholzer: I pushed a couple of fixes to the branch here and also ran an analysis for your idea above. Below's the outcome of that. As a consequence, I would leave this as-is for now.

🤖 AI text below 🤖

The selected profile already reaches mapping through TargetEnvironmentAnalysis, so we would not need another per-function attribute.
However, I think we need to distinguish “the target permits measured-qubit reuse” from “this measurement must advance so dependent quantum work can proceed.” Even for an adaptive target, we should keep measurements terminal whenever possible. A profile flag therefore would not replace the dependency analysis here.
For now, we have cached the analysis for consecutive measurements and excluded preceding register accesses that unnecessarily forced reuse. I would keep that logic and use the existing target environment if we later need to enforce backend-specific reuse or reset restrictions.

burgholzer (approved):
Pushed a couple of fixes. Should be good now. Let's get this merged together with the stacked PR 🚀

PR #2369, opened 2026-09-04
✨ Support break, continue, and general while loops

### LINKED ISSUE #2355: 🐛 OpenQASM Translation Fails for `scf.while`
`qc::translateQCToOpenQASM3` fails for the following QC program 
```mlir
module {
  func.func @main() -> i64 attributes {mqt.entry_point} {
    %alloc = memref.alloc() : memref<2x!qc.qubit>
    %0 = memref.load %alloc[%c0] : memref<2x!qc.qubit>
    %1 = memref.load %alloc[%c1] : memref<2x!qc.qubit>
    %cst = arith.constant 1.5707963267948966 : f64
    scf.while : () -> () {
      qc.h %0 : !qc.qubit
      qc.h %1 : !qc.qubit
      %true = arith.constant true
      scf.condition(%true)
    } do {
      scf.yield
    }
    qc.barrier %0, %1 : !qc.qubit, !qc.qubit
    %c0_0 = arith.constant 0 : index
    memref.store %0, %alloc[%c0_0] : memref<2x!qc.qubit>
    %c1_1 = arith.constant 1 : index
    memref.store %1, %alloc[%c1_1] : memref<2x!qc.qubit>
    memref.dealloc %alloc : memref<2x!qc.qubit>
    return %c0_i64 : i64
  }
}
```

with 

```
loc("<qco-program-builder>":1:1): error: OpenQASM emission error: scf.while condition region must be side-effect free
```

although it should be valid OpenQASM.

burgholzer:
🤖 *AI text below* 🤖

## Description

OpenQASM export rejects valid `scf.while` loops with executable operations before the condition. Support these loops, scalar loop state, and nested `break` and `continue` statements across OpenQASM and Qiskit, with round trips through jeff.

- Lower early exits through `QCProgramBuilder` CFG construction and upstream CFG-to-SCF lifting. Both jumps share one decision block carrying complete scalar state. Canonical do-while programs round-trip as one `scf.while` without an exit-only `scf.if`.
- Export both SCF regions with separate input/result tuples, parallel assignments, and preserved measurement snapshots. Qiskit uses native local variables, captures, stores, and breaks without adding classical bits for compiler temporaries.
- Merge initialization facts across reachable exits and retain detailed, contextual exporter diagnostics in Python exceptions.

The jeff-mlir dependency pins the decoder correction in [upstream PR #52](https://github.com/unitaryfoundation/jeff-mlir/pull/52), which handles independent input/output tuples for while and switch. No source patch is applied.

Termination diagnostics, external runtime input variables, and runtime gate parameters remain unsupported. Constant-true loops remain valid IR. This change affects unreleased v4 functionality, so standalone changelog and upgrade-guide entries are not applicable.

Fixes #2355

Local validation:

- 501 MLIR Python tests pass, including 31 focused loop tests. The loop module skips under Qiskit 1.1.0, matching the supported translation versions.
- 3,117 MLIR C++ tests pass.
- `uvx nox -s lint`, `uvx nox -s cpp-lint`, and `uvx nox -s stubs` pass. Generated stubs have no public API diff.
- Integration coverage combines reset, measurement, wide register comparisons, and stale snapshots in a do-while loop. Wide register expressions remain supported without widening scalar locals; invalid register captures raise Python exceptions.
- Initialization checks distinguish assignments reached before continue from skipped assignments. QCO-to-QC preserves distinct results of index-producing operations so real qubit permutations retain their stores.
- Round trips check scalar values, quantum operation order, initialization, captures, canonical do-while shape, deterministic exports, unchanged source IR, and buffered output on failure.

Codex assisted with implementation, tests, documentation, and simplification review. Human review and hosted CI remain pending.

## Checklist


burgholzer:
@simon1hofmann tagging you for a review here. I've done multiple iterations here and checked large parts of the code. But it wouldn't hurt to have another pair of eyes (or model) check this.

@MatthiasReumann this should address your use case, hopefully.

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
```diff
@@ -2483,8 +2585,104 @@ class OpenQASMToQCEmitter {
...
+  template <typename Loop>
+  void emitLoopWithJumps(const Loop& loop, ValueRange gateParameters,
+                         ValueRange gateQubits) {
+    const auto slots = mutatedState(loop.body);
+    const auto savedScalars = scalarValues;
+    SmallVector<Value> initial;
+    Value step, stop;
+    bool indexRange = false;
+    if constexpr (std::is_same_v<Loop, frontend::ForStatement>) {
+      indexRange = loop.provenPositiveRange;
```
simon1hofmann, line 2597: 🤖 *AI text below* 🤖

**Critical:** This `i64` fast path does not prove that `induction + step` is representable. For example, `for int i in [9223372036854775806:2:9223372036854775806] { continue; }` is a singleton range and must execute once. Here, the subsequent increment wraps to `INT64_MIN`, while the signed comparison against the exclusive bound (`INT64_MAX`) remains true forever, so the program never reaches the statement after the loop. Please either use the existing widened range representation for this path or additionally prove the progression cannot overflow.

bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -2048,48 +2186,111 @@ collectFor(mlir::scf::ForOp loop, ExportState& state,
...
+  auto breakOp = std::make_unique<ExportedControlFlow>();
+  breakOp->kind = ControlFlowKind::Break;
+  breakBody.instructions.push_back(
+      {.kind = ExportedInstruction::Kind::ControlFlow,
+       .controlFlow = std::move(breakOp)});
+  exit->blocks.push_back(std::move(breakBody));
+  body.instructions.push_back({.kind = ExportedInstruction::Kind::ControlFlow,
+                               .controlFlow = std::move(exit)});
+  auto exitWrites = state.unconditionalWrites;
+  auto tail = collectBlock(after, state, controlFlowDepth + 1);
```
simon1hofmann: 🤖 *AI text below* 🤖

**Critical:** `collectBlock(after, ...)` records direct `qc.gphase` operations in `tail.globalPhase`, but the splice below moves only `tail.variables` and `tail.instructions`. Consequently, a general/state-carrying loop such as `while (i < 2) { gphase(pi/4); i += 1; }` exports successfully with `while_loop.blocks[0].global_phase == 0.0` instead of `pi/4` per iteration, silently losing circuit semantics on round-trip. Please merge `tail.globalPhase` into `body.globalPhase` as well.

simon1hofmann (commented):
🤖 *AI text below* 🤖

Two critical correctness findings:

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
burgholzer: Seems like a fairly unrealistic corner case. Let's find the minimal solution for this.

bindings/mlir/qiskit/QiskitExport.cpp
burgholzer: Feels reasonable. Let's fix 👍🏻

simon1hofmann (approved):
Mainly looked at the tests and skimmed over the rest of the code, looks quite good 👍🏻 
Also let Codex review it again in detail and it reported two issues worth double-checking.

burgholzer:
> Mainly looked at the tests and skimmed over the rest of the code, looks quite good 👍🏻 
> Also let Codex review it again in detail and it reported two issues worth double-checking. 

Nice. Replied to both. Would you mind letting your Codex session address these?

simon1hofmann:
> > Mainly looked at the tests and skimmed over the rest of the code, looks quite good 👍🏻 
> > Also let Codex review it again in detail and it reported two issues worth double-checking. 
> 
> Nice. Replied to both. Would you mind letting your Codex session address these?

Sure 👍🏻

mlir/lib/Dialect/QC/Translation/OpenQASMToQCEmitter.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in [`1f0a07451`](https://github.com/munich-quantum-toolkit/core/commit/1f0a074512ad282d578639794e946680efb09169). The fast-path proof now also bounds `stop + step`; ranges that could overflow use the existing widened lowering. Added a regression where `[1:9223372036854775807:1]` with `continue` executes once.

bindings/mlir/qiskit/QiskitExport.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Fixed in [`1f0a07451`](https://github.com/munich-quantum-toolkit/core/commit/1f0a074512ad282d578639794e946680efb09169). The after-region now remains the continuation branch of the synthesized exit conditional, so its global phase runs only for true-condition iterations and not for the final condition-only visit. Added a Qiskit round-trip amplitude regression.

burgholzer:
Thanks for the changes @simon1hofmann 🙏
Looks good to me! 
Let's get this in once CI is green! 🚀

PR #2435, opened 2026-09-07
🐛 Enforce unitary effect contracts

### LINKED ISSUE #2434: 🐛`Pure` and `RecursiveMemoryEffects` Traits on Modifier Operations
The modifier operations of the QCO dialect have the `Pure` and `RecursiveMemoryEffects` trait, which to my understanding should be invalid.

```TableGen
def CtrlOp : QCOOp<"ctrl",
                   traits = [UnitaryOpInterface, AttrSizedOperandSegments,
                             AttrSizedResultSegments, SameOperandsAndResultType,
                             SameOperandsAndResultShape,
                             SingleBlockImplicitTerminator<"YieldOp">, Pure, // <---
                             RecursiveMemoryEffects]> { // <---
```

Pure is an alias of `NoMemoryEffect` and `AlwaysSpeculatable`. Hence, adding `RecursiveMemoryEffects` is probably wrong 🧐

burgholzer:
🤖 *AI text below* 🤖

## Description

Model `qco.ctrl`, `qco.inv`, `qco.pow`, `qco.if`, and `qco.index_switch` with both `RecursivelySpeculatable` and `RecursiveMemoryEffects`. Their own region-container behavior has no direct effect, but both properties depend on the nested operations. In particular, a nested `qco.gphase` carries `MemWrite`, so its enclosing modifier is neither memory-effect-free nor speculatable and cannot be erased or speculated as pure.

Keep generic `qco.call` effects conservatively at `MemWrite`, with calls non-speculatable. Generic effect queries never inspect sibling function bodies, so they respect MLIR's function-pass isolation.

Compute more precise effects once inside the module-level qubit-reuse pass. Bottom-up summaries share work across transitive callees and call sites; unknown effects and cyclic dependencies never become proven effect-free. The pass keeps unitary helper bodies unchanged while rewriting callers and discards all summaries afterward. This preserves reuse across phase-free calls without persistent cache invalidation.

Strengthen the QC and QCO modifier and `mqt.unitary` function verifiers to admit only operations implementing the dialect `UnitaryOpInterface` plus memory-effect-free classical operations that have no regions and carry no qubits. Parameter computation is eager when the invocation is reached and need not be speculatable. Modifiers transform the quantum action on the computation's defined input domain, not the parameter computation. Classical SSA values may still be captured from above. Memory access, classical-bit register access, and non-unitary region operations such as SCF are rejected.

Reuse the QC modifier verifiers in the QC-to-QCO preflight and remove modifier-specific structured-control-flow remapping that is no longer valid. General structured control flow outside modifiers remains supported.

Controlled global phases retain their quantum semantics: global-phase normalization converts a phase inside `qco.ctrl` into the corresponding phase operation on its controls instead of extracting it as an unconditional global phase.

Preserve dependent parameter chains when unrolling modifiers and extracting global phases. These moves cross only eager modifier boundaries, never surrounding classical conditions. Nested modifier canonicalizers retain parameter producers before erasing their former scopes.

Regression tests cover non-speculatable classical computation in unitary helpers and all QC/QCO modifiers, classical conditional boundaries, nested canonicalization, transitive callee effects, callee mutations between pass invocations, unresolved and recursive calls, helper preservation in nested symbol tables, and qubit reuse across phase-free calls. A callee with a global phase remains an effect barrier.

No new dependencies.

Fixes #2434

## Checklist

### DIFF
--- a/mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
+++ b/mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
@@ -35,9 +35,15 @@ def MQTDialect : Dialect {
     `mqt.entry_point` marks the single public, defined `func.func` program entry
     in a module.
     `mqt.unitary` marks a private function that defines a unitary operation.
-    Its body admits unitary operations and pure, region-free classical
-    computation, matching the quantum modifier-body contract. Unitary
-    operations may retain effects such as a scoped global-phase contribution.
+    Its body admits unitary operations and memory-effect-free, region-free
+    classical computation, matching the quantum modifier-body contract. Parameter
+    computation is evaluated eagerly when the invocation is reached; it need not
+    be speculatable. The quantum action is unitary on the computation's defined
+    input domain. Modifiers transform that action, not the parameter computation:
+    inverse does not reverse it, control does not conditionally evaluate it, and
+    power does not repeat it, including at exponent zero. Undefined inputs do not
+    guarantee traps. Unitary operations may retain effects such as a scoped
+    global-phase contribution.
     `#mqt.compilation_target` records compiler-target facts as typed IR.
   }];
 
--- a/mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
+++ b/mlir/include/mlir/Dialect/MQT/IR/MQTDialect.td
@@ -35,6 +35,9 @@ def MQTDialect : Dialect {
     `mqt.entry_point` marks the single public, defined `func.func` program entry
     in a module.
     `mqt.unitary` marks a private function that defines a unitary operation.
+    Its body admits unitary operations and pure, region-free classical
+    computation, matching the quantum modifier-body contract. Unitary
+    operations may retain effects such as a scoped global-phase contribution.
     `#mqt.compilation_target` records compiler-target facts as typed IR.
   }];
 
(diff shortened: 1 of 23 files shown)


burgholzer:
@MatthiasReumann This seems to make way more sense to me. Let me know what you think 😌

MatthiasReumann (commented):
LGTM 👍🏻

So I fully understand this: If I mark an operation as `Pure` and with `RecursiveMemoryEffects` in MLIR, all operations within the pure operation have to be pure also?

burgholzer:
> So I fully understand this: If I mark an operation as `Pure` and with `RecursiveMemoryEffects` in MLIR, all operations within the pure operation have to be pure also?

My understanding is: `Pure == NoMemoryEffects + AlwaysSpeculatable`. 
An operation marked with `NoMemoryEffects` and `RecursiveMemoryEffects` means that the operation itself has no memory effects, but overall memory effects shall be determined by the memory effects of the recursively contained operations. This is, e.g., important when considering a controlled-gphase operation. The control modifier itself has no memory effects, but the global phase operation does. So it cannot simply be pulled out.

Where one needs to be a little careful is the `AlwaysSpeculatable` trait. AFAIU, it does not check the body of an operation and unconditionally marks the operation as speculatable. This was not correct in general.

I am still tracing a few things here though that do not quite make sense to me (yet).

burgholzer:
Alright. This is much cleaner and stricter now. Modifiers are really meant to be a light wrapper and represent unitaries. More complicated code can be deferred to proper functions.

burgholzer (commented):
Okay, I believe this is finally clear and ready to be merged.
Let's see over the week if this stands up as a contract for how we treat modifiers and functions, while we work out more details.

PR #2439, opened 2026-09-07
🐛 Export measurements across quantum operations

simon1hofmann:
🤖 *AI text below* 🤖

## Description

The native Qiskit exporter rejects valid programs when a SWAP or another quantum
operation separates a measurement from the store of its result. Allow constants,
unitary operations (including barriers), and resets between them. Keep the
measurement at its original position and use its existing classical destination,
without adding classical bits.

Continue rejecting intervening classical accesses and control flow: fusing the
store into the measurement writes the destination earlier and may otherwise
change the program's meaning. Update the documented export contract.

This addresses the exporter restriction discussed in #2351 independently of the
mapping fixes and the sorter changes in #2436. Neither PR is a dependency.

No standalone changelog entry or migration instructions are needed for this
correction to unreleased v4 functionality.

## Validation

- Merged main at `6b9385280`, including #2435 and #2444, and rebuilt the MLIR
  bindings at `db22ece60`.
- All 306 Qiskit translation tests passed with Python 3.13 and Qiskit 2.5.2.
- All 13 focused delayed-measurement tests passed. Ten positive cases cover X,
  SWAP, reset, barrier, and inverse modifiers through QC and QCO conversion.
  They check operation order, wire and destination identity, classical-bit
  count, and measured value.
- Intervening bit and whole-register writes and the deferred classical-snapshot
  case remain rejected. The existing control-flow checks still pass.
- `uvx nox -s lint` and `git diff --check` passed.
- `uvx nox -s stubs` passed and produced no tracked stub changes.
- `uvx nox -s cpp-lint -- upstream/main` passed with zero clang-format or
  clang-tidy findings, checking the full changed C++ file with clang-tidy 23.

Implementation, tests, documentation, and this description were prepared with
AI assistance via Codex. CI and human-review confirmations remain open for the
updated revision.

## Checklist

### DIFF
--- a/bindings/mlir/qiskit/QiskitExport.cpp
+++ b/bindings/mlir/qiskit/QiskitExport.cpp
@@ -2025,8 +2025,11 @@ static void validateControlFlowDepth(const size_t controlFlowDepth) {
   }
   for (auto* operation = measure->getNextNode(); operation != store;
        operation = operation->getNextNode()) {
+    // Fusion writes the destination at the measurement. Quantum gates and
+    // resets cannot observe that earlier classical write and stay in place.
     if (operation == nullptr ||
-        !llvm::isa<mlir::arith::ConstantOp>(operation)) {
+        !llvm::isa<mlir::arith::ConstantOp, mlir::qc::UnitaryOpInterface,
+                   mlir::qc::ResetOp>(operation)) {
       return false;
     }
   }
--- a/docs/mlir/python_compiler_collection.md
+++ b/docs/mlir/python_compiler_collection.md
@@ -287,13 +287,15 @@ satisfy the existing snapshot checks.
 
 Each exported measurement must write to one static public CBit in the same
 block. Destinations may be reused; later measurements overwrite earlier values
-in program order. A measurement's destination store must follow it directly,
-apart from constant operations. A conditional or otherwise delayed destination
-store is rejected because Qiskit cannot preserve it as one measurement
-instruction. The measurement result may feed supported classical expressions
-after that store. General scalar control flow saves live measurement results in
-native variables before later writes. Deferred measurement expressions require
-an unchanged destination CBit.
+in program order. Constants, unitary quantum operations (including barriers),
+and resets may separate a measurement from its destination store. The exporter
+keeps the measurement at its original position and writes the destination there.
+Other intervening operations, including classical accesses and control flow, are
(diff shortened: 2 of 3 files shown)


bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -2025,8 +2025,11 @@ static void validateControlFlowDepth(const size_t controlFlowDepth) {
   }
   for (auto* operation = measure->getNextNode(); operation != store;
        operation = operation->getNextNode()) {
+    // Fusion writes the destination at the measurement. Quantum gates and
+    // resets cannot observe that earlier classical write and stay in place.
     if (operation == nullptr ||
-        !llvm::isa<mlir::arith::ConstantOp>(operation)) {
+        !llvm::isa<mlir::arith::ConstantOp, mlir::qc::UnitaryOpInterface,
+                   mlir::qc::ResetOp>(operation)) {
```
MatthiasReumann, line 2032: Under our current assumptions, this probably works. However, I wonder if it were possible to also use the `getRecursiveEffects` function here to figure out if an operation writes to the same "location" as the `StoreOp` 🤔 This may be the more future proof solution.

simon1hofmann: 🤖 *AI text below* 🤖

Thanks! `mlir::getEffectsRecursively` could help here, but a same-location write check alone would miss exporter-specific dependencies. Fusion writes the classical destination at the measurement, so intervening reads also matter.

There is a deferred-read case already covered by `test_delayed_measurement_store_is_rejected`:

```text
old = load c
measured = measure q
if old { quantum gate }
store measured, c
```

The recursive effects of the `if` only describe the quantum operations; the load is outside it. However, the exporter evaluates the condition from `c` at the consumer, so writing the measurement result into `c` early would change the condition. An intervening measurement also needs care: its QC effects describe the qubit, while export introduces a write to its classical destination.

I would keep the conservative check in this PR. A broader effect-based check should come with an update to the classical-snapshot analysis that accounts for these fused writes.

burgholzer: Keep in mind that I am currently working on the Memory Effects in #2435.
I am still working out the exact details, but this may change a few things here.

burgholzer (approved):
Yeah, this works 👍🏻

PR #2287, opened 2026-08-31
🔍 Audit MLIR pass, verifier, and rewrite contracts

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Records five reproduced correctness failures in MLIR lowering and JeFF import,
with minimal CLI examples, source evidence, ownership, and limits for each
proposed fix. The audit is based on `main` at `6cc98f059`.

The diff contains only one 271-line [audit report](https://github.com/munich-quantum-toolkit/core/blob/b2f4a6296b169270849860eb8f221cb330877891/.agent/audits/mlir-contracts.md).
There are no production or test changes; the findings remain open for focused
fixes. The report also reconciles the merged and closed split PRs.

Passes may assume verified IR. The retained findings concern valid source
programs or an external import boundary. Generic pass-entry validation,
invalid-QCO regressions, blanket failure-atomicity requirements, and speculative
resource caps are excluded.

## Confirmed findings

1. **High:** QIR Base moves a later gate before a measurement of the same qubit,
   changing the recorded result.
2. **High:** QC-to-QCO aborts on valid branches that use the same dominating
   qubit reference.
3. **High:** a malformed JeFF file aborts the importing process instead of
   returning an import error.
4. **Medium:** sparse static qubit IDs produce insufficient QIR resource
   metadata; ID 7 is emitted with `required_num_qubits="1"`.
5. **Medium:** Adaptive QIR moves a branch-local qubit release into the epilogue,
   producing a dominance error and failing compilation.

## Validation

- Built the compiler and compiler tests with LLVM/MLIR 23.1.0 and the pinned
  jeff-mlir dependency.
- All 163 compiler tests passed.
- Executed all five CLI reproducers after rebuilding the merged branch. They
  confirm the existing failures; this PR does not implement their fixes.
- `uvx nox -s lint` and `git diff --check` passed.
- Current GitHub CI has not been verified as part of this update.

Tracks #2255.

## Checklist

- [x] The pull request's diff is focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality. — N/A: report only; executed reproducers are included in the report.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals. — N/A: internal audit report only.
- [x] I have added migration instructions to the upgrade guide (if needed). — N/A: no API or behavior changes.
- [x] The changes follow the project's style guidelines and introduce no new warnings.
[...]

PR #2446, opened 2026-09-07
🐛 Enforce quantum allocation and QIR output contracts

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Dynamic QC/QCO qubit allocations must occur directly in the entry block of the `mqt.entry_point` function. The shared MQT verifier enforces this program contract, builders reject invalid insertion points, and standalone QIR and mapping passes invoke the same allocation check before transforming input.

Adaptive QIR preserves the control-flow position of quantum releases. Scalar results now use dynamic allocation alongside result registers, and only owned results are released. The QIR builder applies the same result policy independently of qubit allocation and releases owned qubits at finalization.

Shared QIR output preparation requires one entry-function return and validates measurement/store fusion before rewriting returns or deleting stores. Fusion requires the same block, an available index, and no intervening classical interference; stores to distinct constant indices remain supported. Unknown effects, overlapping writes, and changed execution conditions produce diagnostics. `qc.yield` is effect-free so recursive analysis can recognize supported quantum modifier bodies.

Addresses the conditional-allocation finding in [#2287](https://github.com/munich-quantum-toolkit/core/pull/2287#discussion_r3950384043) and the allocation, lifetime, result-ownership, and output-store audit findings. No new dependencies.

Validation: 1,952 native tests passed across the mapping, compiler, MQT/QC/QCO/QIR IR, QC-to-QCO, Base/Adaptive QIR, and OpenQASM suites after integrating main (`784519cf9`). The release build used AppleClang 21 with LLVM/MLIR 23.1.0 and `ENABLE_IPO=ON`. The compiler and MQT suites were rebuilt and passed again after fixing two lint warnings in the merged tests. `uvx nox -s lint` and `uvx nox -s cpp-lint` passed; C++ lint checked every line of the changed C++ files with clang-tidy 23 and reported zero findings. Hosted CI for the updated head is pending; the full Python and documentation suites were not run locally.

The allocation-verifier tests use SCF without an added CF dependency, and the conditional-release regression uses `scf.if`. CF remains an internal QIR lowering dependency and is used to test rejection of unsupported multiple returns. Changelog and upgrade-guide updates for unreleased v4 functionality are deferred to the release cleanup.

GPT-6 via Codex assisted the implementation, regression tests, validation, and PR text.

## Checklist


burgholzer (changes requested):
I am wondering whether this is the right scope to reject this. 
Shouldn't we reject this at the QC/QCO dialect verification level already? I do not think our current infrastructure can handle programs with allocations placed in SCF regions or modifiers (which already forbid them).
Even functions, at least how they are currently implemented, do not really support this.

simon1hofmann:
🤖 *AI text below* 🤖

@burgholzer Yes, if this is a general allocation invariant, checking it once during QC/QCO program verification makes sense.

There is existing behavior to align: [this QC-to-QCO test](https://github.com/munich-quantum-toolkit/core/blob/e6b2518ee12c7603f3bf150f638c018ba921efb7/mlir/unittests/Conversion/QCToQCO/test_qc_to_qco.cpp#L1717-L1745) explicitly verifies and converts an allocation inside `scf.if`, and the ordinary QC/QCO function builders track helper-local resources. Modifiers already reject scalar qubit allocations, while `mqt.unitary` functions reject non-unitary operations. A general ban would therefore tighten the current contract, not just move an existing verifier check.

Is the intended rule that dynamic quantum allocations may occur only in the entry block of the `mqt.entry_point` function, with ordinary helper functions receiving their quantum resources as arguments?

If so, I propose reworking this PR to enforce that rule consistently for `qc.alloc`, `qco.alloc`, qubit `memref.alloc`, and quantum `qtensor.alloc`; align the builders, documentation, and verification tests; then remove the Adaptive-specific allocation scan. Checking only the scalar allocation operations would leave the register cases uncovered. Classical allocations and static qubit references would stay out of scope.

The code is unchanged pending agreement on that scope.

burgholzer:
> Is the intended rule that dynamic quantum allocations may occur only in the entry block of the mqt.entry_point function, with ordinary helper functions receiving their quantum resources as arguments?

Yeah, I would be tempted to say so.
This feels like an assumption to go with and simplify around.
We may loosen that in the future, but for now I actually think it may help to be explicit.

simon1hofmann:
🤖 *AI text below* 🤖

@burgholzer Implemented the agreed rule in 6fc5241a5 and rebased onto `main`
at ec799daa0.

Dynamic `qc.alloc`, `qco.alloc`, qubit `memref.alloc`, and `qtensor.alloc`
must now appear directly in the entry block of the `mqt.entry_point` function.
The shared MQT entry-point verifier owns this check; QC/QCO program imports
also enforce it for modules without an entry marker and caller-supplied
contexts. Builders, documentation, and tests follow the same rule. Helpers
receive quantum resources as arguments; classical allocations and static qubit
references remain unchanged.

Removed the Adaptive-specific allocation scan and the redundant Mapping
placement check. Mapping now discovers allocations directly in the entry block.

Local validation passed: the full lint-preset build, all 2,358 configured MLIR
CTest entries with LLVM/MLIR 23.1.0, `uvx nox -s lint`, and whole-changed-file
`uvx nox -s cpp-lint` against the new base. Local clang-tidy was 23.0.0git;
these results do not attest to CI for this push.

burgholzer (approved):
I iterated on this quite a bit and added another QIR audit on top of it. This does a bit more now, but I believe it does it well.
If you are fine with the changes as well, let's get this in.

PR #2452, opened 2026-09-07
🐛 Export independently scheduled measurements

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Stacked on #2351, updated through parent commit `aa1b13cf8`. This PR fixes the native Qiskit exporter boundary exposed by the
remaining 25 Benchpress feed-forward profiles, separately from the routing
correction in the parent PR. Merge the parent first, then retarget this PR to
`main`.

Mapping may legally separate a measurement from its classical destination by
other measurements or independent control flow. Permit destination fusion when
recursive effects cannot access that bit, using CBit's non-aliasing register
contract and static-index disambiguation. Keep measurements at their original
quantum positions and add no scratch classical bits.

Move each safely fusible store immediately after its measurement on the
exporter's existing IR clone before indexing writes. Snapshot analysis then uses
the actual emitted order, without synthetic measurement writes or special
handling of other measurements' future stores. Preserve late constant indices,
the parent's indexed lookup, and scalar snapshot support. The input program is
unchanged.

Overlapping accesses, unknown effects, and unsupported stale snapshots remain
diagnosed. Stale register snapshots wider than 64 bits remain unsupported.
The unique measurement-destination requirement remains an exporter limitation,
not a QC dialect invariant; multiple classical destinations are outside this PR.

### Current validation

- At `5ea693d50`, rebuilt Python 3.13 bindings with Qiskit 2.5.2 pass all 398
  tests in `test_mlir_qiskit_translation.py` and `test_mlir_loops.py`.
- Added cases cover measured-bit control before a delayed store, disjoint-bit
  reads without redundant scalar snapshots, late destination indices, input
  preservation, and shared destinations with matching write order. Reversed
  writes and conflicting effects remain rejected.
- The parent's formerly crashing nested Qiskit circuit now compiles through
  the target pipeline.
- Stub generation, repository lint, and full-file C++ lint against fixed parent
  `aa1b13cf8` pass. Generated stubs introduce no API changes.
- Hosted checks must validate this new head; previous-head results are not
  validation of this update.

### Historical Benchpress validation

At source commit `7dad9e19e`, a fresh wheel with Qiskit 2.5.0 passed all 42
targeted Benchpress native-export checks: 31 previously guarded profiles, ten
existing profiles, and BV100. No OpenQASM fallback was used. All 4,621
conditionals remained, nested gates passed Qiskit basis/connectivity analysis,
and BV100 retained exactly 99 classical bits and measurements. All 80
Benchpress integration tests and the then-current 330 Qiskit translation tests
passed; the preceding build failed 12 of the new positive cases.

These Benchpress checks were not rerun for this update. The old integration
snapshot-rejection test predates the parent's scalar snapshot support and has
not been revalidated.

The historical Benchpress checks used a corresponding local integration update:
pin that PR snapshot, retire the temporary textual event-order checker, retain
input restrictions, and enable only the validated profiles. Those adapter
changes are not part of this Core diff. The full benchmark suite was not
restarted; structural counts are not a general dynamic-program equivalence
proof.

No changelog or migration entry is needed for this unreleased-v4 fix with no
public API changes. Implementation, tests, and this description were prepared
with AI assistance through Codex. Human review remains required.

## Checklist

### DIFF
--- a/bindings/mlir/qiskit/QiskitExport.cpp
+++ b/bindings/mlir/qiskit/QiskitExport.cpp
@@ -49,6 +49,7 @@
 #include <mlir/IR/Region.h>
 #include <mlir/IR/Value.h>
 #include <mlir/IR/ValueRange.h>
+#include <mlir/Interfaces/SideEffectInterfaces.h>
 #include <mlir/Support/WalkResult.h>
 #include <mlir/Transforms/GreedyPatternRewriteDriver.h>
 #include <nanobind/nanobind.h>
@@ -1687,12 +1688,34 @@ exportExpression(mlir::Value value, ExportState& state,
   return result;
 }
 
+[[nodiscard]] static mlir::cbit::StoreOp
+measurementDestination(mlir::qc::MeasureOp measure) {
+  mlir::cbit::StoreOp destination;
+  for (auto* user : measure.getResult().getUsers()) {
+    if (auto store = llvm::dyn_cast<mlir::cbit::StoreOp>(user)) {
+      if (destination) {
+        throw std::runtime_error(
+            "QC measurement has more than one classical destination");
+      }
+      destination = store;
+    }
+  }
+  if (!destination) {
+    throw std::runtime_error(
+        "QC measurement is missing a static classical destination");
+  }
+  return destination;
+}
+
 /// Index writes in block order, including effects of nested operations.
 static void indexWrites(mlir::Block& block, ExportState::WriteIndex& index) {
   index.try_emplace(&block);
   for (auto& operation : block) {
     llvm::DenseSet<mlir::Value> modified;
-    if (auto store = llvm::dyn_cast<mlir::cbit::StoreOp>(operation)) {
+    if (auto measure = llvm::dyn_cast<mlir::qc::MeasureOp>(operation)) {
+      /// Fusion writes the destination at the measurement's position.
+      modified.insert(measurementDestination(measure).getReg());
+    } else if (auto store = llvm::dyn_cast<mlir::cbit::StoreOp>(operation)) {
       modified.insert(store.getReg());
     } else if (auto write = llvm::dyn_cast<mlir::cbit::WriteOp>(operation)) {
       modified.insert(write.getReg());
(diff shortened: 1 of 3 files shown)


bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -1694,10 +1695,36 @@ exportExpression(mlir::Value value, ExportState& state,
...
 
+[[nodiscard]] static mlir::cbit::StoreOp
+measurementDestination(mlir::qc::MeasureOp measure) {
+  mlir::cbit::StoreOp destination;
+  for (auto* user : measure.getResult().getUsers()) {
+    if (auto store = llvm::dyn_cast<mlir::cbit::StoreOp>(user)) {
+      if (destination) {
+        throw std::runtime_error(
+            "QC measurement has more than one classical destination");
+      }
```
MatthiasReumann, line 1700: We can keep this assumption for now. However, generally, this could be valid right? @burgholzer

bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -2017,6 +2031,71 @@ static void validateControlFlowDepth(const size_t controlFlowDepth) {
...
+                             ? mlir::WalkResult::advance()
+                             : mlir::WalkResult::interrupt();
+                }
+                if (auto load = llvm::dyn_cast<mlir::cbit::LoadOp>(candidate)) {
+                  return disjointClassicalBit(load.getReg(), load.getIndex(),
+                                              destination)
+                             ? mlir::WalkResult::advance()
+                             : mlir::WalkResult::interrupt();
+                }
+                if (auto interface =
```
MatthiasReumann:
```suggestion
                if (auto mem =
```

bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -2017,6 +2031,71 @@ static void validateControlFlowDepth(const size_t controlFlowDepth) {
...
+  const auto bit = mlir::getConstantIntValue(index);
+  const auto destinationBit = mlir::getConstantIntValue(destination.getIndex());
+  return bit && destinationBit && *bit != *destinationBit;
+}
+
+[[nodiscard]] static bool
+canFuseMeasurementAcross(mlir::Operation& operation,
+                         mlir::cbit::StoreOp destination) {
+  return !operation
+              .walk<mlir::WalkOrder::PreOrder>([&](mlir::Operation* candidate) {
```
MatthiasReumann: A `TypeSwitch` that returns a `WalkResult` could work nicely here instead of the `if`s.

bindings/mlir/qiskit/QiskitExport.cpp
```diff
@@ -2017,6 +2031,71 @@ static void validateControlFlowDepth(const size_t controlFlowDepth) {
...
+                                              destination)
+                             ? mlir::WalkResult::advance()
+                             : mlir::WalkResult::interrupt();
+                }
+                if (auto load = llvm::dyn_cast<mlir::cbit::LoadOp>(candidate)) {
+                  return disjointClassicalBit(load.getReg(), load.getIndex(),
+                                              destination)
+                             ? mlir::WalkResult::advance()
+                             : mlir::WalkResult::interrupt();
+                }
```
MatthiasReumann: Not totally sure about this, but: I think this case should also be handleable via the `MemoryEffectOpInterface` case below. 

```
if (!effect.getValue()) return;
if (effect.getValue() == destination.getReg()) {
  if (isa<Store> || isa<Load>() { return true; /* check -index */ }
  return false;
}
```

You may need to check some edge cases such as `cbit.alloc` and `cbit.sink` 🧐

MatthiasReumann (commented):
Solid work 🗿 Left some nitpicks. Nothing too dramatic! I'll leave it to @burgholzer to judge how these changes fit into the bigger picture of the QASM export.

bindings/mlir/qiskit/QiskitExport.cpp
simon1hofmann: 🤖 *AI text below* 🤖

Yes, this is valid QC IR: the measurement returns an `i1` that can be stored in
several destinations. This is an existing native-Qiskit exporter limitation,
not a dialect invariant; this PR moves the check into a shared helper.

Keeping the restriction here as suggested. Supporting multiple destinations
would require one measurement followed by classical copies that preserve write
order and captured values. That extension is outside this PR's scope.

burgholzer (approved):
Also pushed a couple of fixes here. Let's get this in! 🚀

PR #2537, opened 2026-09-11
✨ Synthesize two-qubit blocks in the native basis

burgholzer:
🤖 *AI text below* 🤖

## Description

Target compilation and synthesis without routing now resynthesize constant two-qubit blocks in the selected native basis while preserving cheaper native operations. For example, non-native `RZZ(0.3); RXX(0.4)` needs two CZ gates together instead of four individually; a native SWAP/RZZ pair stays at two gates when block synthesis would need three. Unrestricted targets also optimize their blocks.

Both pipelines reuse the existing fusion implementation at two stages:

- Before placement, replace a block only when both its two-qubit IR operation count and native cost decrease. This removes cancelled interactions before routing and keeps compact gate pairs intact. Native support is checked without assigning physical sites.
- After placement/routing, resynthesize only when native two-qubit cost decreases, using the assigned sites and operand direction. Individual lowering preserves supported native gates and supplies the comparison cost.

Run scanning, Weyl decomposition, phase handling, and emission remain shared. Numerical block failures fall back to individual lowering; the standalone target-independent fusion factory remains available. The selected synthesis basis uses one entangler family and does not search arbitrary native mixtures or calibration costs. No new dependencies or block IR are added.

Cleanup runs at the boundaries that require it, removing five redundant passes from target compilation and three from synthesis without routing. Full cleanup remains before unrolling because dead work affects its resource budget. Tensor shrinking remains after unrolling, and CSE remains before native synthesis to reuse unchanged measurement reads. Both pipelines share the post-placement tail; the generic cleanup pipeline remains available to its other callers.

### Validation

At `1565801f9`, the release build and 3,539 CTest cases passed, with one existing QDMI skip. All 90 Python MLIR tests, whole-file C++ lint, repository lint, and complete executable documentation/internal link checks passed. The native-synthesis suite contains 61 passing tests. Coverage includes full-unitary and phase preservation, pre-routing cancellation, native preservation, reversed connectivity, unrestricted optimization, numerical fallback, and routed measurement results. The two cleanup regressions cover tensor shrinking after unrolling on two-site targets and reuse of unchanged classical reads across an intervening measurement.

### Measurements

Both Benchpress Small studies use 42 circuits across four abstract topologies, five alternating samples per version and case, and 1,680 successful samples per study. The environment is ARM64, GCC 13.3, LLVM/MLIR 23.1.0, the release preset with LTO disabled, CPU 0 affinity, default mapping options, and seed 42. Targets use `id/sx/x/rz/cz` with measurement, reset, and global phase. Timing includes QC/QCO conversion, the target pipeline, cleanup, and OpenQASM export; file import, target construction, and output writes are excluded.

The block-fusion study compared `07374719b` with MQT Core 4.0.0 (`8d9bcfb03`). Native two-qubit counts improved in 41 of 168 cases and never regressed. Depth improved in 38 of 152 comparable cases and never regressed. Geometric mean count/depth reductions were 2.0%/2.2% on square, 5.4%/5.9% on heavy-hex, and 4.8%/5.1% on linear targets; all-to-all quality was unchanged. Runtime was broadly unchanged before the cleanup refinement.

The cleanup study compares `1565801f9` against `a904d3308`, isolating the scheduling change with shared dependency libraries and identical compiler options. Baseline source copies compile as separate translation units; the candidate uses the normal project libraries and unity-build configuration.

| Topology | Compilation time change from cleanup |
| --- | ---: |
| All-to-all | -6.58% |
| Square | -6.48% |
| Heavy-hex | -6.10% |
| Linear | -6.42% |
| All cases | **-6.40%** |

These are geometric means of paired ratios using five-sample medians. Every case median improved, from 1.74% to 11.18%; median within-case sample ranges were about 3.2% for both variants. All 168 outputs were byte-identical across variants and deterministic across repeats. Native counts were unchanged in all 168 cases; depth was unchanged in all 152 cases where the straight-line metric applies. The 16 structured-control-flow cases have no depth metric. The two studies use different baselines; their timing results are not a single measured speedup against 4.0.0.

Separate probes preserve full matrices, including phase, and show 1,024-gate RZZ/RXX synthesis falling from 2,048 native entanglers to two in the release-baseline comparison. The final cleanup probes have maximum matrix error `3.59906e-12`, unchanged from their baseline. These results cover the stated workloads and settings, not a general speedup or global synthesis optimum. Benchmark drivers, exact build metadata, raw measurements, and PNG/PDF plots remain in the local experiment archive.

Codex assisted with implementation, review, validation, and this description.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ releases may include breaking changes.
 
 ## [Unreleased]
 
+- ✨ Optimize constant two-qubit blocks before placement and during native
+  synthesis. Remove cancelled interactions before routing and preserve cheaper
+  native operations in both target compilation and synthesis. ([#2537])
+  ([**@burgholzer**])
+
 ## [4.0.0] - 2026-09-11
 
 ### MQT Core 4: a compiler foundation built on MLIR and LLVM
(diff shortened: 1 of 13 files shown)


burgholzer (commented):
Took a bit of iteration, but this looks much cleaner now. 
It should be faster _and_ better across the board.

PR #2545, opened 2026-09-12
♻️ Unify native results and isolate DDSIM execution

burgholzer:
🤖 *AI text below* 🤖

## Description

Replace recoverable exceptions in native QDMI, benchmark, DD, and QIR setup APIs with `mqt::LogicalResult` and `mqt::FailureOr<T>`. The small attributed LLVM port keeps standalone Core independent of LLVM. MLIR signatures use upstream result types. Scoped thread-local diagnostics carry messages, severity, error categories, and original QDMI statuses; shared boundary adapters preserve Python exception categories and C status codes.

Keep dependency exceptions inside private JSON translation units, using nlohmann JSON's normal exception-enabled parser once per input. Retain duplicate-key, schema, range, UTF-8, and source-diagnostic checks. GCC and Clang compile the migrated algorithms without exception handling; MSVC retains its normal STL configuration.

QIR allocation failures recover through explicit error-output pointers. Other unhandled runtime failures diagnose and terminate execution. DDSIM now parses, compiles, and executes QASM and QIR in reusable private worker processes. Abnormal exits fail the assigned QDMI job without publishing partial results or replaying it; concurrent and later jobs remain usable. Cancellation and shutdown terminate and reap workers. Workers install beside the provider and participate in runtime copying and wheel relocation. This contains DDSIM crashes; it is not a hostile-code sandbox or isolation for arbitrary providers.

C++ migration: use fallible factories, check `failed(result)` before dereferencing, and install `ScopedDiagnosticHandler` before invoking an operation to capture its diagnostic. Infallible operations retain ordinary values or `void`; optional successful absence remains `std::optional<T>`. Python signatures, QDMI statuses, sampling, and lazy dense/sparse result materialization remain supported. The native API, QIR, and DDSIM guides document the contracts.

Local validation covers 3,594 native cases and 597 standalone cases (one existing optional SC query skip in each), 1,453 Python tests, stub generation, executable documentation, repository lint, and whole-file C++ lint. Linux wheel and installed-consumer relocation checks pass. After the complexity review, the full native suite and all 304 affected Python tests pass again. macOS/Windows packaging and execution, including MSVC mixed-exception validation, remain pending on the revised implementation. Local builds disable IPO because the installed LLVM static libraries conflict with GCC LTO.

Matched performance measurements against upstream main and the previous branch show QIR sampling back at upstream latency; small valid JSON and tiny Bell execution retain about 6% and 8–10% overhead. Worker isolation adds cold-start and transport costs: reused QASM jobs measured about 2.45 ms versus 2.22 ms upstream, and reused QIR jobs about 5.59 ms versus 3.70 ms with substantial variability. These measurements precede the final complexity-only simplifications; no speedup is claimed for that review.

AI assistance: Codex implemented the changes, performed one requested complexity review, applied its three simplifications, and ran local validation. The simplifications share worker setup and remove redundant QIR checks and supervisor state.

## Checklist


burgholzer:
@simon1hofmann @denialhaag This PR contains a proposal for an exception-free MQT Core (using exceptions only where absolutely necessary). I iterated this PR quite intensely and tried to condense it. However, it is still a monster, simply because so many existing APIs work with exceptions.

I am not really asking for a detailed review from you here. Rather an assessment whether you feel we should be doing this. Most of the changes in the PR are of a rather mechanical nature (replacing throwing APIs by ones that forward errors). I am not sure myself that I like this, but maybe mostly because it looks slightly unusual.

If we decide to go with this, then this PR should be mostly good already.
If we decide against it, I'd at least try to extract a couple of useful changes from the PR into a separate one that we can more easily get merged.

Let me know what you think!

.github/workflows/ci.yml
```diff
@@ -133,7 +126,9 @@ jobs:
       build-project: true
       build-target: mqt-core-lint-headers
       use-sccache-gha: false
-      files-changed-only: true
+      # GitHub omits patches from large diffs. Restore changed-file selection once
+      # the action includes https://github.com/cpp-linter/cpp-linter/pull/216.
+      files-changed-only: false
```
denialhaag: Just so that we don't forget:

```suggestion
      files-changed-only: true
```

bindings/Result.hpp
denialhaag: Even though they are very much internal, some of the functions could use brief docstrings.

mlir/tools/mqt-cc/mqt-cc.cpp
denialhaag: This is a bit off-topic, but we could use this opportunity to streamline our usage of `failed(blah)` vs. `blah.failed()`, preferring the former. This file, for example, still has quite a few instances of `blah.failed()`.

denialhaag (commented):
After scrolling through this PR a few times now, this is starting to grow on me. I agree that it feels a bit unusual, but MQT Core has become an MLIR/LLVM-centric repository with v4, and if that's how things are done upstream, why not do it here? 🙂

burgholzer:
> if that's how things are done upstream, why not do it here? 🙂

I can get behind that sentiment! I just wanted to make sure that people are on board with this, as it is quite a substantial change throughout the library. If we are pulling through with this, I'd also through a couple more tokens at it to make sure that we get it "right" the first time.

@simon1hofmann do you have any strong feelings on this?

mlir/tools/mqt-cc/mqt-cc.cpp
burgholzer: Yeah, agreed. Let's put that on the "throwing more tokens at this" list!

bindings/Result.hpp
burgholzer: Agreed 👍🏼 even if just brief ones.

.github/workflows/ci.yml
burgholzer: The respective PR has merged already and has been released. This merely requires an update in the workflows repo and a new patch release. Could you get that going?

simon1hofmann:
> > if that's how things are done upstream, why not do it here? 🙂
> 
> I can get behind that sentiment! I just wanted to make sure that people are on board with this, as it is quite a substantial change throughout the library. If we are pulling through with this, I'd also through a couple more tokens at it to make sure that we get it "right" the first time.
> 
> @simon1hofmann do you have any strong feelings on this?

I skimmed through the changes a little bit, but it's quite a lot and probably have to check with other PRs to not create too many conflicts.
Generally I also think we should do it like upstream projects :)

burgholzer:
> > > if that's how things are done upstream, why not do it here? 🙂
> > 
> > 
> > I can get behind that sentiment! I just wanted to make sure that people are on board with this, as it is quite a substantial change throughout the library. If we are pulling through with this, I'd also through a couple more tokens at it to make sure that we get it "right" the first time.
> > @simon1hofmann do you have any strong feelings on this?
> 
> I skimmed through the changes a little bit, but it's quite a lot and probably have to check with other PRs to not create too many conflicts. Generally I also think we should do it like upstream projects :)

yeah, we should not keep this open for too long if we want to push for it. otherwise, we will constantly accumulate conflicts (like at this moment). @simon1hofmann maybe you can take some more time over the course of the coming days to form a little bit more grounded opinion. then we can decide early next week.

.github/workflows/ci.yml
denialhaag: Do we want to temporarily pin a development version, or do I create a minor release now that also releases the functionality tested in #2603.

burgholzer: Feel free to release a new minor version once the discussion around the auto-fixer are settled 😌

PR #2546, opened 2026-09-12
✨ Preserve structured compilation and execution

burgholzer:
🤖 *AI text below* 🤖

## Description

Structured quantum programs with borrowed registers, indexed loops, runtime powers, or computed outputs could fail during conversion/export or exhaust the DD execution budget. This change supports them through existing compiler and runtime representations:

- Preserve fixed-size borrowed registers through private QC/QCO helpers. Slots retain their original qubit identities; quantum arguments must be disjoint, whole-register boundaries must be complete, and quantum results preserve argument order. QC references remain usable across helper calls and identity stores. Physical reference buffers round-trip through QCO, and structured builders reject resource permutations for both standalone and register-backed qubits.
- Export logical qubit indices and constant angle tables with existing OpenQASM constructs. Mapped OpenQASM uses bounded unrolling to expose static physical qubits. Adaptive QIR retains indexed loops on all-to-all targets with unrestricted operations.
- Limit DD conditional loops to 100,000 iterations; counted loops use widened trip counts, and recursive calls remain rejected. Reuse matrix powering and share fixed-gate lowering between constant and runtime exponents, preserving global phase.
- Record computed Adaptive QIR outputs using Boolean storage and records. Boolean records join DDSIM shot/count strings through per-shot execution, including constant Boolean outputs. Eligible measurement-only programs retain batch sampling.

OpenQASM import omits runtime assertions and their supporting arithmetic. Static diagnostics and negative-index wrapping remain; runtime bounds, nonzero steps, and exponent requirements are documented preconditions. Register accesses also require in-bounds, disjoint slots and identity-preserving stores when these cannot be established statically. Mapped OpenQASM and constrained-topology routing require specialization within the existing 65,536-operation unrolling budget; runtime indices that cannot be specialized remain unsupported. Changelog and upgrade-guide entries are deferred to release preparation.

Validation: optimized Clang/MLIR 23 build; 3,581 native test entries (3,580 passed and one expected Slurm skip); 52 Python compilation/submission tests against the rebuilt package, including standard and iterative QPE through mapped OpenQASM and QIR; full-matrix angle-table round trips; constant-Boolean per-shot fallback checks; executable docs and generated-link checks; repository lint and whole-file C++ lint. Hosted CI for the updated commit is pending.

AI assistance was used for implementation, tests, review, and this description.

## Checklist


burgholzer (commented):
@simon1hofmann @denialhaag Can I hand this off to one of you? I currently do not have the capacity to fully think this through. I think these are genuinely very useful improvements.

simon1hofmann:
> @simon1hofmann @denialhaag Can I hand this off to one of you? I currently do not have the capacity to fully think this through. I think these are genuinely very useful improvements.

I can have a look at this tomorrow 👍🏻

simon1hofmann:
@burgholzer I went through the changes and I think they all make sense. I pushed some more fixes and additional tests.

burgholzer:
> @burgholzer I went through the changes and I think they all make sense. I pushed some more fixes and additional tests.

Many thanks! The only thing I am slightly concerned about is that one of the new commits mentions that some programs cannot be round tripped between QC and QCO (anymore). I find that a bit unfortunate if it's true. We should always be able to go back and forth between our dialects. Is this something that we can fix here?

simon1hofmann:
🤖 *AI text below* 🤖

@burgholzer Yes. The rejection test exposed a limitation of the earlier lowering; it did not introduce the limitation. Ordinary reference stores did not preserve the ownership information needed to reconstruct linear QCO values.

Pushed the fix in commit [730599de9](https://github.com/munich-quantum-toolkit/core/commit/730599de9a0b96564a05dea6cabb08b8baa93684). It adds explicit `qc.take`/`qc.put` operations and converts them back to `qtensor.extract`/`qtensor.insert`. This covers register-slot permutations, cross-register transfers, and replacement with allocated qubits, without inserting quantum SWAP gates. The existing borrowed-reference path remains for functions that do not change slot contents. Adaptive QIR lowers the ownership operations to reference loads/stores at the terminal conversion boundary.

Tests cover repeated semantic round trips, dynamic indices, nested helpers/loops, invalid reference lifetimes, and execution through Adaptive QIR on DDSIM. All 711 tests in the two conversion suites, Adaptive QIR suite, and compiler suite pass; repository lint passes. Whole-PR C++ lint still reports existing diagnostics, with none on the newly added code in focused whole-file checks.

This fixes the borrowed-register round-trip case discussed here. Arbitrary reference stores still need explicit ownership operations, and the existing restrictions on scalar region-state correspondence and register lifetimes remain.

burgholzer (commented):
Alright. I went through this another time. @simon1hofmann I simplified quite a bit from your additions because I did not feel that the additional operations were justified.
I think this is about as clean as it gets at the moment. In the interest of moving along, I'll now merge this PR so that follow-ups can build on it.
If you see anything that doesn't make sense, then please file a follow-up PR! 🙏🏼

PR #2551, opened 2026-09-12
✨ Expose native target mapping controls

simon1hofmann:
🤖 *AI text below* 🤖

## Description

Add shared `CompilationOptions` for timing, statistics, a compilation-wide seed,
and native mapping trials, refinement iterations, routing lookahead, and search
memory limits. C++, Python, and `mqt-cc` use the same controls through target
compilation, synthesis,
custom pass pipelines, and QDMI source submission.
Compiler settings are accepted only through `CompilationOptions`; separate
timing/statistics arguments are removed.

An explicit seed controls Core's mapper, Pauli twirling, and numerical synthesis
retries. It overrides pass-local seeds, including nested pipelines. The scoped
seed is captured in MLIR crash reproducers and restored after success or failure.
The new CLI tests use GoogleTest and check both exit status and diagnostics.

### Usage

```python
from mqt.core.mlir import CompilationOptions, MappingOptions, compile_program

options = CompilationOptions(
    seed=7,
    enable_timing=True,
    enable_statistics=True,
    mapping=MappingOptions(
        trials=4, iterations=2, lookahead=10, search_memory_limit=8 * 1024 * 1024
    ),
)
compiled = compile_program(source, target=device, options=options)
# The same options work with submit_program(source, target=device, options=options).
```

Typed programs accept `options` on `compile_for_target`, `synthesize_for_target`,
and `run_pass_pipeline`. C++ compiler entry points accept a final
`const CompilationOptions&`. Low-level target pipeline builders accept
`MappingOptions`; execute their pass managers with `runWithCompilationOptions`
to apply the seed and instrumentation with the same precedence as the compiler API.

```console
mqt-cc input.qasm --qdmi-device mqt.sc.iqm.garnet \
  '--payload-spec=#mqt.payload_spec<format = <id = "qir", version = "2.1.0", profile = "base", encoding = text>, capabilities = [], optional_capabilities_known = false>' \
  --seed 7 --mapping-trials 4 --mapping-iterations 2 --mapping-lookahead 10 \
  --mapping-search-memory-limit 8388608
```

`--seed` also works without a device and with custom or isolated pass pipelines.

### Limitations

- `seed=None` preserves each pass's default or explicit seed. Execution sampling has a separate seed.
- Omitted trials use the available logical CPU count. Set seed and trials for repeatable mapping with a fixed build, input, target, and mapping controls; layouts may change between releases.
- Trials and iterations must be positive. Iterations default to one forward/backward refinement round. Lookahead defaults to 20 additional two-qubit gates; zero considers only the current gate. Lookahead storage follows actual gates, including when the ceiling is `SIZE_MAX`. All-to-all placement ignores valid mapping controls. The mapper may also consider deterministic candidates.
- `search_memory_limit` (C++ `searchMemoryLimit`) estimates retained node/layout bytes per search and concurrent trial, defaulting to 64 MiB. Zero disables node expansion. Container overhead, caches, and IR are extra; this does not cap total process memory. Changing the budget can change layouts and gate counts; more memory does not guarantee fewer gates.
- Target compilation still requires structured QCO/SCF input and the existing native-synthesis input subset. These controls do not prescribe a layout; #2553 adds layout selection and reporting.
- Compilation options do not apply to an already compiled payload. Source submission accepts `options=None` to mean no supplied compiler settings; other compiler entry points take `CompilationOptions` directly.

### Breaking API change

Replace Python `enable_timing=True, enable_statistics=True` keywords with
`options=CompilationOptions(enable_timing=True, enable_statistics=True)`.
In C++, replace trailing timing/statistics booleans with
`CompilationOptions{.enableTiming = true, .enableStatistics = true}`.
For source `submitProgram`, the options object follows the custom job parameters.
Omit `options` when defaults are sufficient; for compilation and typed methods,
replace explicit `options=None` with omission or `CompilationOptions()`.

### Validation

- 162 Python compiler/QDMI tests pass, including exact mapping diagnostics and zero/small search budgets through typed and payload compilation.
- 228 native compiler, 115 mapping, and 61 target-synthesis tests pass, including stored-seed precedence and restoration and maximum-lookahead compilation.
- Three CLI CTests pass; the documented target command succeeds with normal and `SIZE_MAX` lookahead, and with zero, 1 KiB, and 8 MiB search budgets.
- Stub generation, repository lint, and full-file C++ lint pass.

CI for the latest push is pending.

Release documentation follows the current release-preparation policy. This PR adds no changelog or upgrade-guide entry.

## Checklist


bindings/mlir/register_mlir.cpp
```diff
@@ -360,7 +361,8 @@ qdmiFormat(mlir::ProgramFormat output) {
 compileProgramForTarget(const nb::object& program, const nb::object& target,
                         std::optional<QDMI_Program_Format> programFormat,
                         std::optional<mlir::ProgramFormat> output, bool inplace,
-                        bool enableTiming, bool enableStatistics) {
+                        bool enableTiming, bool enableStatistics,
+                        mlir::MappingOptions mapping) {
```
burgholzer: Should this be called `mappingOptions`, `mapping_options`, etc. throughout the PR?
Also, is this a good place to add this in the function signature?
Should the `enableTiming` and `enableStatistics` options also be grouped somehow? 
Maybe under one umbrella `options`?
I am pretty sure there are some established best practices in the classical community for how to best handle this and make it convenient for people to use.

bindings/mlir/register_mlir.cpp
```diff
@@ -1185,6 +1190,18 @@ Programs own their MLIR module. Conversions can consume a program; use
...
           "Return the textual MLIR representation of this program.");
 
+  nb::class_<mlir::MappingOptions>(
+      m, "MappingOptions",
+      "Native mapping controls. Set seed and trials for reproducible mapping "
+      "with a fixed Core build, input, and target. All-to-all placement "
+      "ignores valid mapping options.")
+      .def(nb::init<size_t, std::optional<size_t>>(), nb::kw_only(),
+           "seed"_a = 42, "trials"_a = nb::none())
+      .def_rw("seed", &mlir::MappingOptions::seed, "Native mapper seed.")
```
burgholzer: The seed should likely be something that is more general. 
I would hope that one can set one seed that controls the entire compilation pipeline and any randomization.

docs/mlir/target_compilation.md
burgholzer, line 1: A bit verbose for my taste.

mlir/unittests/Compiler/mqt-cc/verify_mapping.cmake
burgholzer, line 1: do we really want these kind of CMake testing? feels a bit cumbersome.

CHANGELOG.md
burgholzer, line 1: same question on the changelog here as in the other PR.

burgholzer (changes requested):
I have a couple of broader questions on the design here.
I believe we should spend a bit of effort to come up with a good system for these options.
The general addition here makes sense to me.

bindings/mlir/register_mlir.cpp
simon1hofmann: Makes sense, wasn't sure if we already want to change existing API contracts, but probably makes sense for the next minor releases and should we pretty easy to upgrade using AI 🤔 

🤖 AI text below 🤖
Reworked this into a single `CompilationOptions` object containing seed, timing, statistics, and nested `MappingOptions`. Compilation entry points now take an options argument, which is keyword-only in Python; the separate timing/statistics arguments are removed.
The nested field is named mapping, so callers use options.mapping.trials without repeating “options”. For example: `CompilationOptions(enable_timing=True, mapping=MappingOptions(trials=4))`. This also lets us extend compiler settings without adding more function parameters.

bindings/mlir/register_mlir.cpp
simon1hofmann: The seed now controls all current randomized compiler passes: placement/routing, native synthesis and two-qubit fusion, and Pauli twirling.

bindings/mlir/register_mlir.cpp
burgholzer: I think it is fine for us to break the compiler API in the next 1-2 minor releases as we are using it more and more and learn more how we'd want to use it.
I'll give this another read, but it already sounds pretty good!

burgholzer (approved):
I pushed a couple of changes and simplifications here. These look pretty decent to me 🙌🏼
I'll quickly explore one last thing, then I'll likely merge this!

burgholzer (approved):
Let's get this in :shipit:

PR #2562, opened 2026-09-15
✨ Rank routing candidates by native gate cost

burgholzer:
🤖 *AI text below* 🤖

## Description

Selecting routes by SWAP count can miss a smaller circuit after native synthesis. Rank completed candidates by final native two-qubit gate count, then dependency depth, preserving candidate order on exact ties. One regression now chooses 4 SWAPs and 13 CZ gates over 3 SWAPs and 14 CZ gates; another keeps 11 CZ gates while reducing depth from 10 to 7.

Reuse the existing post-placement pipeline through a shared QCO helper. Candidate generation, forward/backward refinement, bounded A*, and public options stay unchanged. Topology-only targets and speculative synthesis failures retain SWAP scoring; errors from final compilation remain visible. Include the prerequisite sink canonicalization fix for SCF loop-result forwarding, which can temporarily leave a qubit producer in use after its sink is removed.

Native synthesis adds work per candidate. Counts are static, and maximum block depth is only a tie-breaker for structured control flow. No new dependencies or API migrations are required.

### Validation

- Release build and 979 tests passed: 118 mapping, 61 target synthesis, 572 QCO IR, and 228 compiler tests. Regressions cover ranking, serial/parallel determinism, 256 native-basis semantic configurations across CX/CZ/ECR/sqrt-iSWAP, quiet fallback, and SCF canonicalization.
- `uvx nox -s cpp-lint` passed for every changed C++ implementation/test file; `uvx nox -s lint` passed.
- MLIR reference generation and `uvx nox --non-interactive -s docs` passed, including executable examples and internal links.

### Benchmark evidence and cost

On DGX Spark ARM64 with GCC 13.3, LLVM 23.1, 20 trials, and a 256 MiB search budget per trial, the production integration reproduced all 40 prior QASMBench CZ outputs byte for byte. Those 20 circuits on square/heavy-hex targets improve gate counts in 18/40 cases against the recorded `e74143005` baseline, with no count regressions: 10,706 → 10,352 total native gates. The production base is `b897f04be`; its intervening change exposes mapping controls, which the comparison sets explicitly. The large `factor247_n15` baseline did not complete the earlier screen and is excluded.

The refreshed small-case median was 49.7 ms, with 25.8 s aggregate process CPU time versus 3.94 s in the recorded baseline. Host load differed, so these are resource observations, not a speedup claim.

The 1,000-qubit random grid case (4 rounds, seed 42, 20 trials, 512 MiB per-search budget, 16 GiB address-space limit) completed in 64.36 s at 10.75 GiB peak RSS and 1037 s process CPU time. Its 71,453 native gates and depth 5,379 match the explored candidate. Benchmark harnesses and data remain outside the repository.

GPT-6 via Codex implemented the change, applied the Ponytail cleanup/review, ran validation, and prepared this PR under explicit authorization. Human review remains pending.

## Checklist


burgholzer (commented):
@MatthiasReumann @simon1hofmann This is a first step towards more reasonable scoring heuristics in the routing algorithm. I am pretty sure this can be extended in the future to also include fidelity (if available on the device) or even estimated success probability (if duration and coherence times are available).

@MatthiasReumann could you re-run your current experiments with this branch and compare results for the current evaluation? I am leaning towards merging this, but would want to make sure it also improves what you are seeing.

mlir/lib/Dialect/QCO/Transforms/Mapping/Mapping.cpp
```diff
@@ -1130,6 +1155,70 @@ struct MappingPass : impl::MappingPassBase<MappingPass> {
     return best->bundle.layout;
   }
 
+  FailureOr<std::pair<size_t, size_t>> scoreNative(const Layout& layout,
+                                                   std::deque<Node>& nodes) {
```
MatthiasReumann, line 1159: I really like the direction; especially thinking towards fidelity based routing! However, I have a few problems with the current implementation:

### Cloning

Because we've spent quite some time in the past couple of months to avoid any unnecessary copies of the program, I am not a big fan of the `clone` here. Especially since (as far as I can tell) the only reason for that to be necessary is to call the `PassManager`. Seems like a dirty quick-fix to me. I'd suggest to run a few benchmarks on decomposed, unrolled grover programs, which are easy to generate and reasonably large (in terms of #operations but also in byte size). 

### Usage of `PassManager`

Executing a `PassManager` inside a pass is LLVM anti-pattern. 

 > More specifically, a PassManager should generally never need to be constructed within a Pass.

However, there seems to be ["dynamic" pass pipelines](https://mlir.llvm.org/docs/PassManagement/#dynamic-pass-pipelines) that can be used for exactly this use-case. I still would want to avoid the `clone` though. Something like a "cold" synthesis would be nice. 

### Cost-Model? 
Couldn't we not also integrate this information into the cost model of the A*? This might be slightly tricky to get right; but could pay off in the long run.

burgholzer: Very good points! Let me investigate this a little further.
One upfront note: the copy is mainly so that the routing trials are non-destructive on the original program. However, maybe there is a way to perform some kind of cold synthesis as well (similar to the cold routing).

burgholzer:
Closing this in favor of #2607

